PR #6, opened 2016-10-26
Add UDP Datagram as specified in rfc768

adventureloop:
https://www.ietf.org/rfc/rfc768.txt


GreyCat:
UDP contain "length", so shouldn't it be like that?

```
  - id: body
    size: length - 8
```

PR #8, opened 2016-12-11
Add CramFS format

athre0z:
For now, with compressed blocks as `size-eof: true` field until iterator access or another mechanism that allows parsing this data becomes available.


GreyCat:
Thanks!

PR #9, opened 2016-12-19
Added initial DNS implementation

Glitch0011:
Isn't a complete implementation but should support the average DNS record. 

Notable features this sample highlights are:

- How to currently work with bitfields and bitflags
- How to use repeat-expr
- Self-offsetted instances using types

The following improvements to Kaitai would really help:
- Fix to the fact you can't really use any non-alpha char in enum values (even in quotes)
- Meta should really have a doc tag (already in the works I see)
- Bitflags to help avoid hacks (also in the works)
- The ability to have an OR in an if statement (to avoid hack-fixes)
- Hierarchical parsing of data with multiple ksy files (to allow for a full packet rather than just the DNS payload)
- Better doc (I plan to cover this)

### DIFF
--- a/network/dns_frame.ksy
+++ b/network/dns_frame.ksy
@@ -1,0 +1,121 @@
+meta:
+  id: dns_packet
+  title: DNS (No support for Auth-Name + Add-Name for simplicity)
+  application: https://www.ietf.org/rfc/rfc1035.txt
+  endian: be
+seq:
+  - id: transaction_id
+    doc: "ID to keep track of request/responces"
+    type: u2
+  - id: flags
+    type: u2
+    doc: "Please add support for bit-flags!"
+    enum: flags_type
+  - id: questions
+    doc: "How many questions are there"
+    type: u2
+  - id: answer_rrs
+    doc: "How many answers are there"
+    type: u2
+  - id: authority_rrs
+    doc: "How many authority answers are there"
+    type: u2
+  - id: additional_rrs
+    doc: "How many additional answers are there"
+    type: u2
+  - id: queries
+    type: query
+    repeat: expr
+    repeat-expr: questions
+  - id: answers
+    type: answer
+    repeat: expr
+    repeat-expr: answer_rrs
+types:
+  query:
+    seq:
+      - id: name
+        type: domain_name
+      - id: type
+        type: u2
+        enum: type_type
+      - id: class
+        type: u2
+        enum: class_type
+  answer:
+    seq:
+      - id: name
+        type: domain_name
+      - id: type
+        type: u2
+        enum: type_type
+      - id: class
+        type: u2
+        enum: class_type
+      - id: ttl
+        doc: "Time to live"
+        type: u4
+      - id: data_length
+        doc: "This suggests that multiple domain-names can be returned, unsupported currently due to lack of example"
+        type: u2
+      - id: domain_name
+        type: domain_name
+  domain_name:
+    seq:
+      - id: name
+        type: label
+        repeat: until
+        doc: "Repeat until the length is 0 or it is a pointer (bit-hack to get around lack of OR operator)"
+        repeat-until: "(_.length & 0b00111111) == 0"
+  label:
+    seq:
+      - id: length
+        type: u1
+      - id: pointer
+        doc: "RFC1035 4.1.4: The OFFSET field specifies an offset from the start of the message"
+        doc: "If the first two bits are raised it's a pointer-offset to a previously defined name"
+        type: pointer
+        if: "((length & 0b11000000)) == 192"
+      - id: name
+        doc: "Otherwise its a string the length of the length value"
+        type: str
+        encoding: "ASCII"
+        size: length
+        if: "((length & 0b11000000)) != 192"
+  pointer:
+    seq:
+      - id: pointer
+        doc: "Read one byte, then offset to that position, read one domain-name and return"
+        type: u1
+    instances:
+      contents:
+        io: _root._io
+        pos: pointer
+        type: domain_name
+enums:
+  flags_type:
+    0x0100: "Query_Standard_Query_Not_Auth_No_Trunc_Recursive_Non_Auth_Unacceptable"
+    0x8180: "Response_Standard_Query_Not_Auth_No_Trunc_Recursive_Can_Recurse_No_Auth_Non_Auth_Unacceptable_No_Error"
+    0x8183: "Response_Standard_Query_Not_Auth_No_Trunc_Recursive_Can_Recurse_No_Auth_Non_Auth_Unacceptable_No_Such_Name"
+  class_type:
+    1: "IN"
+    2: "CS"
+    3: "CH"
+    4: "HS"
+  type_type:
+    1: "A"
+    2: "NS"
+    3: "MD"
+    4: "MF"
+    5: "CNAME"
+    6: "SOA"
+    7: "MB"
+    8: "MG"
+    9: "MR"
+    10: "NULL"
+    11: "WKS"
+    12: "PTR"
+    13: "HINFO"
+    14: "MINFO"
+    15: "MX"
+    16: "TXT"


network/dns_frame.ksy
```diff
@@ -0,0 +1,121 @@
+meta:
+  id: dns_packet
```
GreyCat: File should be named "dns_packet.ksy" as well. As far as I see, RFC never refers to DNS packet as "DNS frame", as file name suggests.

network/dns_frame.ksy
```diff
@@ -0,0 +1,121 @@
...
+  title: DNS (No support for Auth-Name + Add-Name for simplicity)
+  application: https://www.ietf.org/rfc/rfc1035.txt
```
GreyCat: We were discussing in https://github.com/kaitai-io/kaitai_struct/issues/59 to use `doc-ref` for that (or just use a comment right now). "Application" is meant to identify a particular piece of software that this format is used for. Given that network packages are highly standartized and used by tens of thousands (if not hundreds) of different software / firmware / hardware implementations everywhere, probably it's better to keep clean of "application".

network/dns_frame.ksy
```diff
@@ -0,0 +1,121 @@
...
+        repeat: until
+        doc: "Repeat until the length is 0 or it is a pointer (bit-hack to get around lack of OR operator)"
+        repeat-until: "(_.length & 0b00111111) == 0"
+  label:
+    seq:
+      - id: length
+        type: u1
+      - id: pointer
+        doc: "RFC1035 4.1.4: The OFFSET field specifies an offset from the start of the message"
+        doc: "If the first two bits are raised it's a pointer-offset to a previously defined name"
```
GreyCat: It's YAML, so proper way to format that one would be something like:

```yaml
doc: >
  RFC1035 4.1.4: The OFFSET field specifies an offset from the start of the message.
  If the first two bits are raised it's a pointer-offset to a previously defined name.
```

Doing duplicate "doc" is useless, only one of them will remain.

network/dns_frame.ksy
```diff
@@ -0,0 +1,121 @@
...
+        type: u2
+      - id: domain_name
+        type: domain_name
+  domain_name:
+    seq:
+      - id: name
+        type: label
+        repeat: until
+        doc: "Repeat until the length is 0 or it is a pointer (bit-hack to get around lack of OR operator)"
+        repeat-until: "(_.length & 0b00111111) == 0"
```
GreyCat: There is an OR operator in the expression language. You can use something like:

```yaml
repeat-until: _.length == 0 or _.is_pointer
```

network/dns_frame.ksy
```diff
@@ -0,0 +1,121 @@
...
+    7: "MB"
+    8: "MG"
+    9: "MR"
+    10: "NULL"
+    11: "WKS"
+    12: "PTR"
+    13: "HINFO"
+    14: "MINFO"
+    15: "MX"
+    16: "TXT"
```
GreyCat: Having no end of file newline is evil :)

network/dns_frame.ksy
```diff
@@ -0,0 +1,121 @@
...
+        doc: "RFC1035 4.1.4: The OFFSET field specifies an offset from the start of the message"
+        doc: "If the first two bits are raised it's a pointer-offset to a previously defined name"
+        type: pointer
+        if: "((length & 0b11000000)) == 192"
+      - id: name
+        doc: "Otherwise its a string the length of the length value"
+        type: str
+        encoding: "ASCII"
+        size: length
+        if: "((length & 0b11000000)) != 192"
```
GreyCat: Why not use a calculate value here and refer to it everywhere? Why use binary on one side and decimal on the other (192)? Why use so many parenthesis?

```yaml
    instances:
      is_pointer:
        value: "(length & 0b1100_0000) != 0b1100_0000"
```

network/dns_frame.ksy
```diff
@@ -0,0 +1,121 @@
...
+    seq:
+      - id: pointer
+        doc: "Read one byte, then offset to that position, read one domain-name and return"
+        type: u1
+    instances:
+      contents:
+        io: _root._io
+        pos: pointer
+        type: domain_name
+enums:
```
GreyCat: There are really no point in using quotes here. Enum labels are not strings or anything like that anyway. They should be treated as identifiers (and kept in `lower_underscore_case`). Even if there's no check right now, one would be added in future.

network/dns_frame.ksy
```diff
@@ -0,0 +1,121 @@
...
+        type: domain_name
+      - id: type
+        type: u2
+        enum: type_type
+      - id: class
+        type: u2
+        enum: class_type
+      - id: ttl
+        doc: "Time to live"
+        type: u4
```
GreyCat: RFC suggests that it's signed ("3.2.1. Format").

network/dns_frame.ksy
```diff
@@ -0,0 +1,121 @@
...
+      - id: name
+        type: domain_name
+      - id: type
+        type: u2
+        enum: type_type
+      - id: class
+        type: u2
+        enum: class_type
+      - id: ttl
+        doc: "Time to live"
```
GreyCat: Um, adding unit of measurement (i.e. that it's specified in seconds) would be good for documentation ;) Even RFC 1035 specifies that.

network/dns_frame.ksy
```diff
@@ -0,0 +1,121 @@
...
+    doc: "How many additional answers are there"
+    type: u2
+  - id: queries
+    type: query
+    repeat: expr
+    repeat-expr: questions
+  - id: answers
+    type: answer
+    repeat: expr
+    repeat-expr: answer_rrs
```
GreyCat: Something suggests me that there should be "authority answers" and "additional answers" here?

network/dns_frame.ksy
```diff
@@ -0,0 +1,121 @@
...
+  endian: be
+seq:
+  - id: transaction_id
+    doc: "ID to keep track of request/responces"
+    type: u2
+  - id: flags
+    type: u2
+    doc: "Please add support for bit-flags!"
+    enum: flags_type
+  - id: questions
```
GreyCat: RFC refers to these fields as QDCOUNT, ANCOUNT, NSCOUNT, ARCOUNT, or "question", "answer", "authority records", "additional records" sections. I'm not sure about reusing these names, but at the very least we should use stable naming scheme everywhere and refer to the same stuff repeatedly using the same identifiers.

network/dns_frame.ksy
```diff
@@ -0,0 +1,121 @@
...
+    doc: "How many questions are there"
+    type: u2
+  - id: answer_rrs
+    doc: "How many answers are there"
+    type: u2
+  - id: authority_rrs
+    doc: "How many authority answers are there"
```
GreyCat: They are definitely *not* referred to as "authority answers" in RFC. They are "name server resource records" in "the  authority records section". Referring to them as "answers" gives a false impression that they might be somewhat connected to ANCOUNT and, as far as I see, they are not.

network/dns_frame.ksy
```diff
@@ -0,0 +1,121 @@
...
+    instances:
+      contents:
+        io: _root._io
+        pos: pointer
+        type: domain_name
+enums:
+  flags_type:
+    0x0100: "Query_Standard_Query_Not_Auth_No_Trunc_Recursive_Non_Auth_Unacceptable"
+    0x8180: "Response_Standard_Query_Not_Auth_No_Trunc_Recursive_Can_Recurse_No_Auth_Non_Auth_Unacceptable_No_Error"
+    0x8183: "Response_Standard_Query_Not_Auth_No_Trunc_Recursive_Can_Recurse_No_Auth_Non_Auth_Unacceptable_No_Such_Name"
```
GreyCat: Probably it's better to reimplement these as named calculated instances for the time being.

GreyCat:
Thanks for your contribution! I've added some points in the review that could be improved. Answering your concerns:

> Fix to the fact you can't really use any non-alpha char in enum values (even in quotes)

I don't quite understand that, but generally, enum labels are supposed to be IDs (they are usually formatted as normal identifiers in the target languages), so it's natural to make them adhere to the same rules.

> Meta should really have a doc tag (already in the works I see)

Already added in current git, just use a `doc` top-level tag.

> Bitflags to help avoid hacks (also in the works)

Using calculated values for bitfields isn't really a "hack". Bitfield support we're discussing now would be more like a syntactic sugar for the same (i.e. calculated instances).

> The ability to have an OR in an if statement (to avoid hack-fixes)

"OR" is available, see review comments for details.

> Hierarchical parsing of data with multiple ksy files (to allow for a full packet rather than just the DNS payload)

It's there. Just load a whole bunch of *.ksy files from `network/` at once, and you can parse everything starting from the pcap capture file, for example.

> Better doc (I plan to cover this)

Thanks :)

network/dns_frame.ksy
Glitch0011: You're totally right, but wtf? The largest field in the record is signed and yet never uses a negative value?

Glitch0011: It's something to do with relative vs absolute timing. Eww.

PR #10, opened 2016-12-20
Improvements based upon feedback (see description)

Glitch0011:
- Comment used rather than application
- Flags instance used rather than enum
- Updated docs and ids to align with RFC rather than guesswork
- Added unit to ttl
- Added support for A-class records now I've found more samples
- Updated expressions to use newly discovered "or" function and instanced fields for clarity
- Removed dead code
- Dequoted enum-values based upon feedback (needed to leave null given it causes problems otherwise as a literal)


GreyCat:
Great, thanks!

PR #11, opened 2016-12-21
Extend Mach-O file descriptor (basic structure + code signature)

koczkatamas:


GreyCat:
Thanks! A few concerns that I have right now:

1. File misses trailing newline
2. Enums use uppercase characters (they are identifiers, they should be converted according to target language rules) → probably we would enforce that as a compile error as some point in time

PR #14, opened 2017-01-15
Added parser for optionroms.

KOLANICH:


KOLANICH:
@GreyCat

GreyCat:
Sorry for the review taking so much time, I'm currently having some health problems :(

First of all, I'd like to request this file to be reformatted to traditional YAML format, in particular:

* all indents using 2 spaces
* `key: value` formatted like that
* Arrays inside the map key being indented, i.e. like that:

```yaml
seq:
  - id: foo
    type: u1
  - id: bar
    type: u1
```

* Longer `doc` strings should be probably spread among several lines and written using `>` syntax, i.e.:

```yaml
  doc: >
    This is an ordinal value that indicates the revision number of this structure only
    and does not imply a level of compliance with the Plug and Play BIOS version.
```

* .ksy file must have a trailing newline.

Probably it's easier to apply some kind of YAML pretty printer.

KOLANICH:
@GreyCat

executable/optionrom.ksy
```diff
@@ -0,0 +1,474 @@
+meta:
+  id: optrom
```
KOLANICH: >it describes a part of firmware

Firmware is executable code, right? And optionroms are actually legacy bios modules, they are discovered and loaded by loader code in runtime, so they are more like executable file formats.

More, I guess we don't need a separate firmware folder at all. Everything is a computer, why do we discriminate them against x86?

GreyCat: > Firmware is executable code, right?

Not always. Firmware is actually closer to a file system, in respect that it stores both executable and non-executable parts, structured in some sort of way that bootloader would be able to locate them. To be frank, when it *is* just executable code, it usually means that Kaitai Struct has nothing to do with it.

In case of option ROMs, as I see, .ksy describes tons of tables which are *not* executable code, but used by executable code to configure hardware stuff.

> Everything is a computer, why do we discriminate them against x86?

Absolutely no discrimination. I propose the following rule of thumb:

* Some complex binary data (or its parts) that's normally flashed into some ROM chip and there's no user-space tools to access it as a file system → `firmware` (note that "firmware" might it turn contain references to filesystems, and it's totally ok)
* Small, non-container binary data, usually flashed on some sort of dedicated hardware part as EEPROM that are read by other components of the system for configuration purposes (i.e. EDID, SPD, NFC tags, PCI & SCSI configuration pages, etc) → `hardware`
* Container for (arbitrary) data structures which is implicity named "file system" and/or there are tools for some OSes which allow us to mount it / use it as file system → `filesystem`
* Container files on regular "file systems", which are mostly used to store executable code + some data to aid loading it in regular operating systems → `executable`

It doesn't matter if it's x86, ARM, Xtensa, or something exotic.

Any more opinions? Cc @koczkatamas @LogicAndTrick @tschoening?

tschoening: I agree with @GreyCat, `firmware` sounds more reasonable to me than `executable`.

koczkatamas: GreyCat +1

executable/optionrom.ksy
```diff
@@ -0,0 +1,474 @@
...
+      - id: len
+        type: u1
+        doc: "The length of the BIOS32 Service Directory Header.  This is measured in units of paragraphs (16 bytes)."
+      - id: checksum
+        type: u1
+        doc: "The BIOS32 Service Directory Header checksum. This is a value which makes the cumulative ADD value of all bytes in the Header equal to 0h."
+      - id: reserved
+        size: 5
+  pirt:
+    doc: "https://web.archive.org/web/20101127024139/http://www.microsoft.com/whdc/archive/pciirq.mspx"
```
KOLANICH: ```/types/pirt/doc-ref: unknown key found, expected: doc, enums, instances, meta, seq, types```

executable/optionrom.ksy
KOLANICH: >To be frank, when it is just executable code, it usually means that Kaitai Struct has nothing to do with it.

It's true, "executable formats"  having no data tables for loaders are not executable formats.

>In case of option ROMs, as I see, .ksy describes tons of tables which are not executable code, but used by executable code to configure hardware stuff.

This is because the modules have not landed yet. In fact, it's better to move most of tables out of this file and load them as modules. In this case pci_data_struct,  pci_expansion_rom_header,  legacy_option_rom_header and pnp_hdr make ```executable``` file format and the most of other headers go to ```hardware```.  In fact some of the structs should go to ```firmware``` as they were meant to be used as containers for other headers (according to the spec), but ... I have not found any case they were really used in this way, so let them be a part of ```executable``` format.

KOLANICH:
@GreyCat

KOLANICH:
@arekbulski What was that?

arekbulski:
Ah sorry, I was just looking at this thread and seen a post as clutter (obsolete), so I just removed it. At construct forum I just feel free to drop posts if they are either no longer relevant or just chit that, or for whatever reason not useful. Sometimes I forget which repo I am browsing, Construct or Kaitai. Sorry, I wont do that again.

PR #20, opened 2017-02-02
Add iTXt and zTXt chunks

s4chin:
In the `compressed_text_chunk`, decompression yields a `iso8859-1` text(https://www.w3.org/TR/PNG/#11zTXt). But I didn't know how to decompress and decode, so I left it at that.
In the `international_text_chunk`, the `language_tag` should be `iso646`(https://www.w3.org/TR/PNG/#11iTXt), but it gave me an error. Using `iso8859-1` worked fine for me.


GreyCat:
> But I didn't know how to decompress and decode, so I left it at that.

Actually I see that you've used `process: zlib`, and it seems to work?

> the `language_tag` should be `iso646` (https://www.w3.org/TR/PNG/#11iTXt), but it gave me an error. Using `iso8859-1` worked fine for me.

ISO 646 is actually a family of national encodings (i.e. ISO 646-HU, ISO 646-GB, etc). PNG standard cites "ISO 646.IRV:1991", which also known as [CP1009 and CP20105](https://en.wikipedia.org/wiki/Code_page_1009). Wikipedia says that it's effectively ASCII with one major twist — code 36 corresponds to ¤, not $ — but looks like Wikipedia is wrong here. According to [IBM documentation](https://www-01.ibm.com/software/globalization/cp/cp01009.html), ¤ was used in 1983 version of standard, but since 1991, this standard is full equivalent of ASCII.

Many iconv-based conversion engines (i.e. in Ruby, Python, C++) know this encoding as `ISO_646.IRV:1991`. And they do depict 36 as $ as well.

So, I believe, it boils down to just "ASCII" here. Definitely not "iso8859-1". Could you fix that?

s4chin:
Yes, `process:zlib` worked indeed. Output is like this `b'Example output'`

GreyCat:
Thanks! You can always use something like:

```yaml
      - id: text_datastream
        process: zlib
        size-eos: true
        type: string_latin1
# ...
types:
  string_latin1:
    seq:
      - id: str
        type: str
        size-eos: true
        encoding: iso8859-1
```

This is a bit clunky, but you'll get your text decoded as `chunk.body.text_datastream.str`.

PR #23, opened 2017-02-10
Update jpeg and exif files

### LINKED ISSUE #22: jpeg.ksy doesn't have all segments declared
So I'm using this version which solves a lot of problems - https://github.com/koczkatamas/kaitai_struct_formats/tree/fix/compile-workarounds.
As all the `appn` segments are not declared there, it gives an error when it encounters one or more undeclared segments. Should I add them and submit a PR?

s4chin:
Fix #22


GreyCat:
Thanks!

PR #24, opened 2017-02-18
Add binary image lay for andes firmware found in mediatek 76x0U devices

adventureloop:
Add a `firmware` directory and a firmware image format for an obscure platform.

I can see there being many `firmware` image formats and I think these are distinct from `executable` formats which already have a directory. 

The format specified here is enough to parse out a firmware image for use with driver development, but it misses the bit fields that encode the firmware version. The correct way to do this is [documented in a blog post here][1], it is just a simple bit mask:

    print("fw version: {}.{}.{}"
        .format(
            (fw_ver & 0xf000) >> 8,
            (fw_ver & 0x0f00) >> 8,
            fw_ver & 0x00ff))

I am not sure how to implement this and I am not entirely sure it is required.

There are also firmware images with the documented header used as a format trailer. The code I have currently tries to parse the header, checks the lengths and if that fails tries to parse the trailer. I don't think there is a way to model this in kaitai, or that it is desirable. 

[1]: http://adventurist.me/posts/0225/

### DIFF
--- a/firmware/andes_firmware.ksy
+++ b/firmware/andes_firmware.ksy
@@ -1,10 +1,14 @@
 meta:
   id: andes_firmware
   endian: le
+  title: Andes Firmware Image layout as seen in MT76 Wifi Chipsets
+  license: CC0-1.0
+  application: Firmware Image wifi chipset
+  doc: Firmware image found with MediaTek MT76xx wifi chipsets. 
 seq:
   - id: image_header
     type: image_header
-	size: 32
+    size: 32
   - id: ilm
     size: image_header.ilm_len
   - id: dlm
--- a/firmware/andes_firmware.ksy
+++ b/firmware/andes_firmware.ksy
@@ -1,0 +1,28 @@
+meta:
+  id: andes_firmware
+  endian: le
+seq:
+  - id: image_header
+    type: image_header
+	size: 32
+  - id: ilm
+    size: image_header.ilm_len
+  - id: dlm
+    size: image_header.dlm_len
+types:
+  image_header:
+    seq:
+      - id: ilm_len
+        type: u4
+      - id: dlm_len
+        type: u4
+      - id: fw_ver
+        type: u2
+      - id: build_ver
+        type: u2
+      - id: extra
+        type: u4
+      - id: build_time
+        type: str
+        size: 16
+        encoding: UTF-8


firmware/andes_firmware.ksy
```diff
@@ -0,0 +1,28 @@
+meta:
+  id: andes_firmware
+  endian: le
+seq:
+  - id: image_header
+    type: image_header
+	size: 32
```
GreyCat: Please fix this indent for `size`?

firmware/andes_firmware.ksy
```diff
@@ -0,0 +1,28 @@
+meta:
+  id: andes_firmware
+  endian: le
```
GreyCat: Generally, there is minimal requirement for files in this repository to have `/meta/license` and it's highly preferred to have some metadata, like `/meta/title`, `/meta/application`, and `/doc`.

Could you please at least add a `license: XXX`, where `XXX` is a [valid SPDX license identifier](https://spdx.org/licenses/) for this work?

GreyCat:
Thanks for your contribution! I've made several minor comments, could you take a look at them, so I could merge this right away? :)

GreyCat:
> I am not sure how to implement this and I am not entirely sure it is required.

The most straightforward way is to use value instances, i.e. something like that:

```yaml
instances:
  fw_ver_1:
    value: '(fw_ver & 0xf000) >> 8'
  fw_ver_2:
    value: '(fw_ver & 0x0f00) >> 8'
  fw_ver_3:
    value: 'fw_ver & 0x00ff'
```

A slightly more modern way would be to replace this:

```yaml
      - id: fw_ver
        type: u2
```

with bit-sized integer reading. Given that it's little-endian, probably this would work:

```yaml
      - id: fw_ver_3
        type: u1
      - id: fw_ver_1
        type: b4
      - id: fw_ver_2
        type: b4
```

But, yeah, I'm not quite sure that it's worth doing so as well.

GreyCat:
> There are also firmware images with the documented header used as a format trailer. The code I have currently tries to parse the header, checks the lengths and if that fails tries to parse the trailer. I don't think there is a way to model this in kaitai, or that it is desirable.

I'm not quite sure that I understand what you mean here. Do I get it correctly that there are two distinct structures, one is basically what you've specified as `image_header` in this file, and there's some other structure, which seem to have the same length data, but it's located at the end of the file, after `ilm` and `dlm`? And you want to make sure that everything's all right while parsing the file and compare lengths from the first structure and the second structure?

adventureloop:
From looking at vendor code there is a format
    
    hdr
    ilm
    dlm

and

    ilm
    dlm
    hdr

From the image there isn't anyway to figure out which is which,the drivers look at the chipset to decide. The scripts I have for parsing these are just looking for valid images, so I have to try the first format and if that doesn't parse try the second format. 

I suspect the trailer format is a separate file.

GreyCat:
Sorry for somewhat late reply and thanks for your contribution!

PR #25, opened 2017-03-18
Create iNES.ksy

DarkLinkXXXX:
Will add another PR for the mapper enum, because there's 256 relevant mapper values and I'm not adding all of those in one day.


GreyCat:
* `game/` category is mostly for file formats related to particular games. iNES format is not a game by itself, but rather a container for arbitrary executable dump for 6502 CPU. It's the same logic that, like, almost every Windows game has an PE executable (.exe), we don't put that format in `game/`, we put it into `executable/`. Given that it's a very specific type of executable that is itself a full firmware, I'd recommend to put it into `firmware/`.

* File must be named as `meta/id`, i.e. `ines.ksy`, otherwise there would be problems with importing it from other ksys, etc.

* `instances/mapper` has broken formatting (4 spaces instead of 2)

* `instances/mapper` is a value instance, so its type is inferred automatically. `type` has no meaning there; previous versions of compiler just ignored it, modern compiler should give a clear error message.

* `instances/mapper` expression is kinda overkill. `upper_nibble` is `b4`, so it's guaranteed to be at most 4 bits long. You can do just '_root.header.f6.lower_nibble | (_root.header.f7.upper_nibble << 4)'

DarkLinkXXXX:
I believe I have addressed all your concerns, thank you for the feedback.

GreyCat:
One more important detail I've overlooked: please add `meta/license`? ;)

DarkLinkXXXX:
@GreyCat Done.

GreyCat:
Thanks!

PR #26, opened 2017-03-20
Created stdlib folder with some building bloccks in it

KOLANICH:


stdlib/uuid.ksy
```diff
@@ -0,0 +1,49 @@
+meta:
+  id: uuid
+  endian: le
+  license: Unlicense
+doc: "UUID according to https://www.ietf.org/rfc/rfc4122.txt"
```
GreyCat: I think I would object to adding `uuid` type automatically to every UUID one might encounter. It is always unnecessary for parsing purposes (I doubt something in any format will ever depend on contents of internal UUID fields), and, 99.9% of the time, it is completely unnecessary for practical purposes: they're intended to be used as 128-bit unique IDs, and that's it.

Knowing internal structure of UUIDs is needed mostly to debug generation of UUIDs to ensure their uniqueness, or for forensic purposes. For both of that, you need to know which family of UUIDs you're debugging. Unfortunately, contrary to what RFC 4122 states, it's not always possible to determine it automatically, i.e. by analyzing bits of 8th byte. There are quite a few implementations that use completely random 128 bits as UUIDs, thus everything can happen.

So, I'd vote to do good quality implementation of UUIDs, but (1) keep variants and layout separate, in different ksy files, application of particular format to be chosen manually, (2) not encourage to import them in formats unless necessary.

stdlib/uuid.ksy
```diff
@@ -0,0 +1,49 @@
...
+    doc: The high field of the clock sequence multiplexed with the variant
+  - id: clock_seq_low
+    type: u1
+    doc: The low field of the clock sequence
+  - id: node
+    type: u4be
+    doc: The spatially unique node identifier
+  - id: node1
+    type: u2be
+    doc: The spatially unique node identifier
```
GreyCat: This particular sequence describes only one possible variant, which is the one described in RFC. [Microsoft's version](https://msdn.microsoft.com/en-us/library/windows/desktop/aa373931(v=vs.85).aspx), for example, is pretty different even in layout.

stdlib/uuid.ksy
```diff
@@ -0,0 +1,49 @@
...
+
+instances:
+  variant:
+    doc: >
+      The variant field determines the layout of the UUID.  That is, the interpretation of all other bits in the UUID depends on the setting of the bits in the variant field.  As such, it could more accurately be called a type field; we retain the original term for compatibility.  The variant field consists of a variable number of the most significant bits of octet 8 of the UUID.
+    pos: 0
+    type: b1
+    enum: variants
+    repeat: until
+    repeat-until: _ == 0
```
GreyCat: This is wrong in so many ways and it just won't compile:

1. Variant-determing bits are in 8-th octet, i.e. it's `pos: 7`, not `pos: 0`.
2. You don't really need to read "until 0 is encountered", you need to just need to read at most 3 bits. In theory, 0 might never come, and this will crash with EOS exception.
3. If you use `type: b1`, it gives you a boolean value. It can't be compared to 0 in `repeat_until`, it should be `repeat-until: _ == false` or `repeat-until: not _`.
4. After using `repeat`, you're essentially getting an array of booleans, not an integer. Thus you can't apply `enum` on top of array of booleans.

But most of all, the fundamental flaw is that you're parsing a certain format first (using `seq`), and then start to analyze whether it was that format or not.

stdlib/uuid.ksy
```diff
@@ -0,0 +1,49 @@
...
+enums:
+  variants:
+    0: ncs_backward_compatibility
+    2: current
+    6: microsoft_backward_compatibility
+    7: future
+  versions:
```
GreyCat: This enum is unused.

tschoening:
> I'm not completely sure about stdlib/ name, though — it looks like we're describing some structures in
> some language's or platform's stdlib, like libc, libstdc++ or something like that? May be there are better
> names? How about common/?

Comparing to the other names, we have some wide topic, like `archive`, with concrete implementations, like `zip`, so your argument holds. We would need some more specific topic for an UUID, something like use-case/file-format transparent `structs` or such? Until that is found, I vote for `common` or `other` or `stuff` or such as well over `stdlib`.

stdlib/c_str_ptr.ksy
```diff
@@ -0,0 +1,13 @@
+meta:
+  id: c_str_ptr
+  endian: le
+  license: Unlicense
+seq:
+  - id: ptr
+    type: u2
```
KOLANICH: It seems we need templates. #135

KOLANICH: https://github.com/kaitai-io/kaitai_struct/issues/135

LogicAndTrick:
I use/abuse the `common` name in all my projects so it sounds good to me. Other options could be `global` or `shared`.

koczkatamas:
Common sounds good. I also use "helper" or "util" namespace (mostly for codes and not structures), I presume "common" is better.

GreyCat:
So, let it be `common` then. I'll start it by transferring `vlq_*.ksy` there.

PR #27, opened 2017-03-23
Added specpr format.

KOLANICH:


GreyCat:
Thanks, that looks really cool! Can you recommend some data samples so I could check how it works?

GreyCat:
Found the files, ksy seems to work after some adjustments. I'll add some comments in the review, merge it, and fix some issues. Please check out the comments.

scientific/spectroscopy/specpr.ksy
```diff
@@ -0,0 +1,254 @@
+meta:
+  id: specpr
+  file-extension: spec
+  endian: be
+  license: unlicense
```
GreyCat: This is not a valid [SPDX expression](https://spdx.org/licenses/). `Unlicense` is.

scientific/spectroscopy/specpr.ksy
```diff
@@ -0,0 +1,254 @@
+meta:
+  id: specpr
+  file-extension: spec
```
GreyCat: As far as I see, none of the files at ftpext.cr.usgs.gov has this extension. Moreover, looks like this format came from classic UNIX background, which frequently used no extensions at all. Are you sure about `.spec`?

scientific/spectroscopy/specpr.ksy
```diff
@@ -0,0 +1,254 @@
...
+  id: specpr
+  file-extension: spec
+  endian: be
+  license: unlicense
+  #encoding: "utf-8"
+  encoding: "ascii"
+  title: "SPECtrum Processing Routines Data Format 3/4/88"
+doc: >
+  Specpr records are fixed format, 1536 bytes/record. Record number counting starts at 0. Binary data are in IEEE format real numbers and non-byte swapped integers (compatiible with all Sun Microsystems, and Hewlett Packard workstations (Intel and some DEC machines are byte swapped relative to Suns and HPs). Each record may contain different information according to the following scheme.
+  You can get some library of spectra from ftp://ftpext.cr.usgs.gov/pub/cr/co/denver/speclab/pub/spectral.library/splib06.library/
```
GreyCat: Generally, that's not what supposed to be `doc` element for top-level type. Top-level docstring is supposed to explain *what* is this format is all about, i.e. what are the contents, where / how / what for it is used, who designed it, who are the providers of the software, etc. Implementation details are supposed to be put into individual types documentation.

Last, but not least: I suppose in this case it's ok to just copy text descriptions verbatim, as this is probably public domain work (being US government official's work), but in general case you can't just copy text as is, especially not under different license — it would be copyright violation.

scientific/spectroscopy/specpr.ksy
```diff
@@ -0,0 +1,254 @@
...
+seq:
+  - id: records
+    type: record
+    repeat: eos
+types:
+  icflag:
+    doc: "it is big endian"
```
GreyCat: That's totally helpful docstring.

scientific/spectroscopy/specpr.ksy
```diff
@@ -0,0 +1,254 @@
...
+      size: 1536 - 4
+      type:
+          switch-on: icflag.type
+          cases:
+            'record_type::data_initial': data_initial
+            'record_type::data_continuation': data_continuation
+            'record_type::text_initial': text_initial
+            'record_type::text_continuation': text_continuation
+enums:
+  record_type: # if I use 0b notation it doesn't work
```
GreyCat: We don't support expression language for enum IDs, so it's processed by YAML, and YAML [only supports decimal, hex and octal](http://yaml.org/spec/1.2/spec.html#id2761509) integer literals. In theory, we can add conversion from strings like `0b1010111` manually...

scientific/spectroscopy/specpr.ksy
```diff
@@ -0,0 +1,254 @@
...
...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
```
GreyCat: You've beaten your previous record here ;)

Your file has 50K of binary zeroes appended to it instead of a trailing newline. Plus broken line endings (somehow your git haven't converted CRLF -> LF on checkin).

PR #30, opened 2017-04-05
Add cpio archive, old binary variant, little-endian

davidhicks:
Source: https://people.freebsd.org/~kientzle/libarchive/man/cpio.5.txt

### DIFF
--- a/archive/cpio_old_le.ksy
+++ b/archive/cpio_old_le.ksy
@@ -1,0 +1,64 @@
+meta:
+  id: cpio_old_le
+  title: cpio archive, old binary variant, little-endian
+  file-extension: cpio
+  license: CC0-1.0
+  endian: le
+  encoding: ASCII
+seq:
+  - id: files
+    type: file
+    repeat: eos
+types:
+  file:
+    seq:
+      - id: header
+        type: file_header
+      - id: path_name
+        type: str
+        size: header.path_name_size - 1
+      - id: string_terminator
+        contents: [0x00]
+      - id: path_name_padding
+        contents: [0x00]
+        if: header.path_name_size % 2 == 1
+      - id: file_data
+        size: header.file_size.least_significant_bits + header.file_size.most_significant_bits * 65536
+      - id: file_data_padding
+        contents: [0x00]
+        if: (header.file_size.least_significant_bits + header.file_size.most_significant_bits * 65536) % 2 == 1
+      - id: end_of_file_padding
+        size: 1
+        contents: [0x00]
+        repeat: eos
+        if: path_name == 'TRAILER!!!' and (header.file_size.least_significant_bits + header.file_size.most_significant_bits * 65536) == 0
+  file_header:
+    seq:
+      - id: magic
+        contents: [0xC7, 0x71]
+      - id: device_number
+        type: u2
+      - id: inode_number
+        type: u2
+      - id: mode
+        type: u2
+      - id: user_id
+        type: u2
+      - id: group_id
+        type: u2
+      - id: number_of_links
+        type: u2
+      - id: r_device_number
+        type: u2
+      - id: modification_time
+        type: four_byte_unsigned_integer
+      - id: path_name_size
+        type: u2
+      - id: file_size
+        type: four_byte_unsigned_integer
+  four_byte_unsigned_integer:
+    seq:
+      - id: most_significant_bits
+        type: u2
+      - id: least_significant_bits
+        type: u2


archive/cpio_old_le.ksy
```diff
@@ -0,0 +1,64 @@
+meta:
+  id: cpio_old_le
+  title: cpio archive, old binary variant, little-endian
+  file-extension: cpio
+  license: CC0-1.0
+  endian: le
+  encoding: ASCII
```
GreyCat: Probably not really ASCII. I heavily suspec that they're not "strings" per se, but rather just byte arrays without any specific encoding.

archive/cpio_old_le.ksy
```diff
@@ -0,0 +1,64 @@
...
+      - id: path_name
+        type: str
+        size: header.path_name_size - 1
+      - id: string_terminator
+        contents: [0x00]
+      - id: path_name_padding
+        contents: [0x00]
+        if: header.path_name_size % 2 == 1
+      - id: file_data
+        size: header.file_size.least_significant_bits + header.file_size.most_significant_bits * 65536
```
GreyCat: That would be better incapsulated in `four_byte_unsigned_integer` type as value instance. And probably it's better to replace `* 65536` with `<< 16`.

GreyCat:
Thanks!

GreyCat:
I've applied two changes, please take a look:

1. I've fixed end-of-lines to use regular *nix `\n` style (which is git standard, and which is correctly transformed on the fly into `\r\n` style for Windows systems)
2. Added value instance named `value` in `four_byte_unsigned_integer` to simplify the code and reduce repetitions.

archive/cpio_old_le.ksy
davidhicks: Yep, that's true, it has no specified encoding. I just put ASCII in as a "default encoding" to keep KS happy

archive/cpio_old_le.ksy
davidhicks: Thanks, agreed that's a good idea. I'll look into using value instances more in the future.

davidhicks:
Thanks @GreyCat : changes look good. Sorry about the line endings, I had copied the .ksy file from a Windows environment without realising. Will look into using value instances in the future--thanks for the tip!

archive/cpio_old_le.ksy
GreyCat: Ok, I'll remove encoding and `type: str` then? It would be up to the application to decide how to treat these byte arrays (i.e. using current locale encoding, some in-app setting or something like that)

davidhicks: It was really only required to make the "path_name" field more accessible/usable (as opposed to displaying a byte array). Happy to change to a byte array if that is better, as there is never going to be a valid "encoding" for the path name (given the character set is file-system specific)

GreyCat: This is kind of "fake usability", because it breaks whole format parsing when one encounters a non-ASCII character in an archive. Doing `pathName.toString(someEncoding)` is not totally hard, but at least it puts the liability on the user, not .ksy author(s).

koczkatamas: Maybe worth adding pathNameAscii / pathNameUTF8 instances which makes the convertation as a property? This way maybe it will be even more easier for the user the understand what happens here.

GreyCat: `pathNameAscii` is kind of redundant if you already have `pathNameUTF8`. The only stuff that `pathNameAscii` might provide is throwing exceptions on non-ASCII character codes (and that's usually not what you want).

Generally, I'm not sure that it would be super useful. There are tons of non-UTF8 encodings even  if we're talking only about Western countries and only about UNIX world: iso8859-1, iso8859-2, iso8859-7, iso8859-15, cp437, etc - see https://www.freebsd.org/doc/handbook/using-localization.html - and for end-user, doing something like `.toString("whatever")`, or, in worst case, `KaitaiStream.bytesToString(..., "whatever")`, is probably not that hard.

koczkatamas: We could take a look how cpio libraries work and find out which encodings are commonly used, but on the other hand that would be more effort we can spend on this format.

What we can do is support a WebIDE (/ksv) extension which gives a hint about that this field should be visualized as a string with and maybe give a best effort encoding (eg. UTF8) for visualizing (for example: `-webide-display-as: string`, `-webide-encoding: utf8`). What do you think, is it a good idea or not?

GreyCat: Yeah, that's probably a good idea. I mean, we should have multiple ways to display a byte array to begin with. The obvious ones are:

* `"foo\0bar\n\xC7"` - C-style, ASCII shown directly, everything else is C-like encoded
* `"foo\x00bar\x0A\xC7"` - C-style, ASCII shown directly, everything else is hex-encoded
* `66 6F 6F 00 62 61 72 0A C7` - hex dump
* `[102, 111, 111, 0, 98, 97, 114, 10, 199]` - array of decimal integers (probably least useful, but still might prove be of some use in uncommon cases)

Bonus points in C-like strings would have color highlighting for special characters ;)

And, yeah, probably we'll need a way to:

1. Specify "default" visualization in the .ksy
2. Experiment with switching representations on the fly using some sort of menu / buttons for one particular field

PR #31, opened 2017-04-07
Add ICC profile, version 4

davidhicks:
Specification for profile version 4.3.0.0:
http://www.color.org/specification/ICC1v43_2010-12.pdf

Tested against sample version 4.* profiles available at:
https://bugzilla.mozilla.org/show_bug.cgi?id=488800

This script adds support for most ICC-defined tag types, however in some
cases the support provided is basic. Some further improvements which
need to be made include:
- using enums for tags of type signature_type (split signature_type into
multiple independent types so they can each use a different enum)
- finish defining some of the more complex numerical/tabular tag types

This script will only parse version 4.* profiles even though previous
version 2.* profiles are quite similar in structure and can be partially
parsed using this script as well. Another improvement would be to create
an icc_2.ksy file, and split common components into a separate .ksy
file.

### DIFF
--- a/image/icc_4.ksy
+++ b/image/icc_4.ksy
@@ -1,0 +1,1701 @@
+meta:
+  id: icc_4
+  title: ICC profile, version 4
+  license: CC0-1.0
+  encoding: ASCII
+  endian: be
+seq:
+  - id: header
+    type: profile_header
+  - id: tag_table
+    type: tag_table
+types:
+  profile_header:
+    seq:
+      - id: size
+        type: u4
+      - id: preferred_cmm_type
+        type: u4
+        enum: cmm_signatures
+      - id: version
+        type: version_field
+      - id: device_class
+        type: u4
+        enum: profile_classes
+      - id: color_space
+        type: u4
+        enum: data_colour_spaces
+      - id: pcs
+        type: str
+        size: 4
+      - id: creation_date_time
+        type: date_time_number
+      - id: file_signature
+        contents: [0x61, 0x63, 0x73, 0x70]
+      - id: primary_platform
+        type: u4
+        enum: primary_platforms
+      - id: profile_flags
+        type: profile_flags
+      - id: device_manufacturer
+        type: device_manufacturer
+      - id: device_model
+        type: str
+        size: 4
+      - id: device_attributes
+        type: device_attributes
+      - id: rendering_intent
+        type: u4
+        enum: rendering_intents
+      - id: nciexyz_values_of_illuminant_of_pcs
+        type: xyz_number
+      - id: creator
+        type: device_manufacturer
+      - id: identifier
+        size: 16
+      - id: reserved_data
+        size: 28
+    types:
+      version_field:
+        seq:
+          - id: major
+            contents: [0x04]
+          - id: minor
+            type: b4
+          - id: bug_fix_level
+            type: b4
+          - id: reserved
+            contents: [0x00, 0x00]
+      profile_flags:
+        seq:
+          - id: embedded_profile
+            type: b1
+          - id: profile_can_be_used_independently_of_embedded_colour_data
+            type: b1
+          - id: other_flags
+            type: b30
+    enums:
+      cmm_signatures:
+        0x41444245: adobe_cmm #ADBE
+        0x41434D53: agfa_cmm #ACMS
+        0x6170706C: apple_cmm #appl
+        0x43434D53: color_gear_cmm #CCMS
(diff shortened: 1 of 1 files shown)


GreyCat:
Super! Looks like that's huge amount of work :)

A few little notes:

* That's a file with Windows EOLs again %)
* Please keep trailing end-of-file newline, git (and lots of other software) likes it that way
* These will currently fail to compile on C# due to name clash:

```yaml
      - id: profile_flags
        type: profile_flags
      - id: device_manufacturer
        type: device_manufacturer
```

* While declaring a enum inside a type, like that:

```yaml
  device_attributes:
    seq:
      - id: reflective_or_transparency
        type: b1
        enum: device_attributes_relfective_or_transparency
# ...
    enums:
      device_attributes_relfective_or_transparency:
        0: reflective
        1: transparency
```

You don't really need to repeat that `device_attributes_` in enum's name, as types effectively form a namespace, and that's guaranteed by KS.
* There is a typo in `relfective`, I'll fix that.

davidhicks:
Thanks for the review @GreyCat 

I'll have to check why I'm getting CRLF line endings--sorry about that!

Is there some more information on the scope of variable names for id, type, enum, instances, etc? I had a look at some other translators where a prefix/suffix is potentially added to id, type, enum and instances names which makes the names unique per type (id|type|enum|instances). Is this not the case for C#? Or perhaps the namespace for id and type names are common, but enum and instances names are in their own namespace? Also, what happens if a name is defined globally or in a parent, and then redefined in a narrower scope?

I'll tidy up the unnecessary device_attributes_* prefix.

GreyCat:
> Is there some more information on the scope of variable names for id, type, enum, instances, etc?
> Also, what happens if a name is defined globally or in a parent, and then redefined in a narrower scope?

You're right, this is not very well defined in the docs :(

Generally, attributes — both from `seq` and `instances` — are treated equally and they're available to be called only in the type where they're defined. That makes sense, as they refer to some particular object instance.

On the other hand, referring to types and enums is done in a manner common to C, C++, C#, Java and Ruby. That means, when you ask for `foo`, first we search for that exact name inside current type. If it's not found, we go up a level in nesting hierarchy and ask there, and so on, up to the topmost type (or *types*, if we have something imported). This way if something's defined inside the topmost `/types` or `/enums`, if will be potentially available for everyone to use by name, unless it is "shadowed" by the type or enum with the same name closer by hierarchy. For example:

```yaml
seq:
  - id: one
    type: foo # refers to /foo
types:
  foo: {}
  bar:
    seq:
      - id: two
        type: foo # still refers to /foo
    types:
      baz:
        seq:
          - id: three
            type: foo # but this one refers to /foo/bar/baz/foo
          - id: four
            size: three.super_inner_thing # legit usage
        types:
          foo:
            seq:
              - id: super_inner_thing
                type: u1
```

> I had a look at some other translators where a prefix/suffix is potentially added to id, type, enum and instances names which makes the names unique per type (id|type|enum|instances). Is this not the case for C#? Or perhaps the namespace for id and type names are common, but enum and instances names are in their own namespace?

Namespacing of target language entities, such as classes, enums, etc, is obviously beyond control of KS. Java, for example, has distinct namespaces for class names, (local variables + instance variables), method names. You can do crazy stuff like that if you want, and it compiles:

```java
public class A {
  public A() {
    int A = 1;
    System.out.println(A);
    System.out.println(A());
  }
  public static int A() {
    return 2;
  }
}
```

In Ruby, some stuff is kept separate by convention (i.e. class names = constants => start with upper case letter, local variables and methods start with lower case letter), and some are separated using special syntax (i.e. `@object_vars`, `@@class_vars` and `$global_vars`).

However, things quickly get messy with C++ and C#:

* For C++, "default" STL-style class names clash with member names (i.e. `foo_bar`) 
* For C#, the same goes on using upper camelcase style (i.e. `FooBar`)

I've added a very simple and crude workaround for C++ for that, i.e. adding `_t` suffix to all class names. KS-to-C# compiler's author opted to not commit such an atrocity ;), and thus having both subtype and member named `header` gets us two `Header` ids which clash :( To add an insult to injury, KS currently have no means to track that and can't even give a warning or anything.

Probably things will get *slightly* better if we'll implement custom naming configuration (i.e. something like `ksc -t csharp --type-prefix=T ...` to get all types automatically prefixed with `T`), but, unfortunately, it doesn't solve the core problem, rather shifts the blame to the user ;)

davidhicks:
Thanks @GreyCat , that's a great explanation! When I get some free time I'll see about proposing some changes to the user guide to include that information

PR #32, opened 2017-04-07
.gitattributes: require LF ending for .ksy files

davidhicks:
For .ksy files, ensure that line endings are LF and not CRLF on
check-in.


davidhicks:
Further documentation on .gitattributes at: https://git-scm.com/docs/gitattributes

.gitattributes
```diff
@@ -1 +1 @@
-*.ksy linguist-language=yaml
+*.ksy linguist-language=yaml eol=lf
```
generalmimon, line 1: Looking at various examples of the `eol` attribute on the internet (e.g. [docs.github.com](https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings#example) or [git-scm.com](https://git-scm.com/docs/gitattributes#_end_of_line_conversion)), it seems to me that the `eol` attribute is used only with `text`, not only by itself. Probably makes sense to add it?

```suggestion
*.ksy linguist-language=yaml text eol=lf
```

generalmimon (approved):
@davidhicks Thanks!

PR #33, opened 2017-04-28
Added format image/dicom_le.ksy

sorgloomer:
Hi there,

I am not sure this grammar does not cover all cases, but was working for all my datasets. At elast in he WebIDE.

### DIFF
--- a/image/dicom_le.ksy
+++ b/image/dicom_le.ksy
@@ -1,0 +1,147 @@
+meta:
+  id: dicom
+  file-extension: dcm
+  endian: le
+seq:
+  - id: file_header
+    type: t_file_header
+  - id: dataset
+    type: t_dataset
+types:
+  t_file_header:
+    seq:
+      - id: preamble
+        size: 128
+      - id: magic
+        contents: 'DICM'
+  t_dataset:
+    seq:
+      - id: elements1
+        type: t_dataentry_state_implicit
+        if: not _io.eof
+        repeat: until
+        repeat-until: _io.eof or _.is_transfer_syntax_change_explicit
+      - id: elements2
+        type: t_dataentry_state_explicit
+        repeat: eos
+  t_dataentry_state_implicit:
+    seq:
+      - id: tag_group
+        type: u2
+      - id: tag_elem
+        type: u2
+      - id: header
+        type: t_entry_header_implicit
+        if: entry_implicit
+      - id: header
+        type: t_entry_header_explicit
+        if: not entry_implicit
+      - id: content
+        size: header.content_length
+        if: has_content
+      - id: elements
+        type: t_dataentry_state_implicit
+        if: has_elements and not is_definite
+        repeat: until
+        repeat-until: _.is_closing_tag
+      - id: elements
+        type: t_elements
+        if: has_elements and is_definite
+        size: header.content_length
+    types:
+      t_elements:
+        seq:
+          - id: elements
+            type: t_dataentry_state_implicit
+            repeat: eos
+    instances:
+      entry_implicit:
+        value: tag_group != 2 or tag_group == 0xfffe
+      is_closing_tag:
+        value: tag_group == 0xfffe and tag_elem & 0xff00 == 0xe000 and tag_elem != 0xe000
+      has_elements:
+        value: (tag_group == 0xfffe and tag_elem == 0xe000) or header.is_seq
+      has_content:
+        value: not has_elements
+      is_definite:
+        value: header.content_length != 0xffffffff
+      p_is_transfer_syntax_change_explicit_1:
+        # '1.2.840.10008.1.2.1\0'
+        value: content == [49, 46, 50, 46, 56, 52, 48, 46, 49, 48, 48, 48, 56, 46, 49, 46, 50, 46, 49]
+      p_is_transfer_syntax_change_explicit_2:
+        value: content == [49, 46, 50, 46, 56, 52, 48, 46, 49, 48, 48, 48, 56, 46, 49, 46, 50, 46, 49, 0]
+      is_transfer_syntax_change_explicit:
+        value: tag_group == 2 and tag_elem == 16 and (p_is_transfer_syntax_change_explicit_1 or p_is_transfer_syntax_change_explicit_2)
+
+  t_dataentry_state_explicit:
+    seq:
+      - id: tag_group
+        type: u2
+      - id: tag_elem
+        type: u2
+      - id: header
+        if: header_is_implicit
+        type: t_entry_header_implicit
+      - id: header
+        if: not header_is_implicit
+        type: t_entry_header_explicit
+      - id: content
+        size: header.content_length
+        if: has_content
+      - id: elements
+        type: t_dataentry_state_explicit
+        if: has_elements and not is_definite
+        repeat: until
+        repeat-until: _.is_closing_tag
+      - id: elements
+        type: t_elements
+        if: has_elements and is_definite
+        size: header.content_length
+    types:
+      t_elements:
+        seq:
+          - id: elements
+            type: t_dataentry_state_explicit
+            repeat: eos
+    instances:
+      header_is_implicit:
+        value: tag_group == 0xfffe
+      is_closing_tag:
+        value: tag_group == 0xfffe and tag_elem & 0xff00 == 0xe000 and tag_elem != 0xe000
+      has_elements:
+        value: (tag_group == 0xfffe and tag_elem == 0xe000) or header.is_seq
+      has_content:
+        value: not has_elements
+      is_definite:
+        value: header.content_length != 0xffffffff
+
+  t_entry_header_explicit:
+    seq:
+      - id: vr
+        type: str
+        encoding: 'ascii'
+        size: 2
+      - id: p_reserved
+        type: u2
+        if: length_is_long
+      - id: p_content_length_u4
+        type: u4
+        if: length_is_long
+      - id: p_content_length_u2
+        type: u2
+        if: not length_is_long
+    instances:
+      content_length:
+        value: length_is_long ? p_content_length_u4 : p_content_length_u2
+      length_is_long:
+        value: not (vr == 'AE' or vr == 'AS' or vr == 'AT' or vr == 'CS' or vr == 'DA' or vr == 'DS' or vr == 'DT' or vr == 'FL' or vr == 'FD' or vr == 'IS' or vr == 'LO' or vr == 'PN' or vr == 'SH' or vr == 'SL' or vr == 'SS' or vr == 'ST' or vr == 'TM' or vr == 'UI' or vr == 'UL' or vr == 'US' or vr == 'LT')
+      is_seq:
+        value: vr == 'SQ'
+
+  t_entry_header_implicit:
+    seq:
+      - id: content_length
+        type: u4
+    instances:
+      is_seq:
+        value: content_length == 0xffffffff
--- a/image/dicom/dicom_common.ksy
+++ b/image/dicom/dicom_common.ksy
@@ -1,12 +1,13 @@
 meta:
-  id: dicom
+  id: dicom_common
+  title: Digital Imaging and Communications in Medicine (DICOM) file format
   file-extension: dcm
+  license: MIT
   endian: le
-seq:
-  - id: file_header
-    type: t_file_header
-  - id: dataset
-    type: t_dataset
+doc-ref: http://dicom.nema.org/medical/dicom/current/output/html/part10.html
+doc: |
+  Common types for little endian DICOM files
+seq: []
 types:
   t_file_header:
     seq:
@@ -14,17 +15,7 @@ types:
         size: 128
       - id: magic
         contents: 'DICM'
-  t_dataset:
-    seq:
-      - id: elements1
-        type: t_dataentry_state_implicit
-        if: not _io.eof
-        repeat: until
-        repeat-until: _io.eof or _.is_transfer_syntax_change_explicit
-      - id: elements2
-        type: t_dataentry_state_explicit
-        repeat: eos
-  t_dataentry_state_implicit:
+  t_dataentry_implicit:
     seq:
       - id: tag_group
         type: u2
@@ -40,7 +31,7 @@ types:
         size: header.content_length
         if: has_content
       - id: elements
-        type: t_dataentry_state_implicit
+        type: t_dataentry_implicit
         if: has_elements and not is_definite
         repeat: until
         repeat-until: _.is_closing_tag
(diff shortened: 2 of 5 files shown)


GreyCat:
Cool! Thanks for your contribution. I'll give a more detailed review a bit later, right now I'd like to ask one minor, but important detail: please add `license: ...` tag to your work (with [SPDX license expression](http://spdx.org/licenses/)), so it would be valid open source that everyone can collaborate to.

Also, I've just remembered that @KOLANICH announced that he was working on DICOM as well earlier in issue https://github.com/kaitai-io/kaitai_struct/issues/134 — but probably yours is more complete anyway. May be he'll want to consider contributing to this one.

image/dicom_le.ksy
```diff
@@ -0,0 +1,147 @@
...
+      - id: tag_group
+        type: u2
+      - id: tag_elem
+        type: u2
+      - id: header
+        type: t_entry_header_implicit
+        if: entry_implicit
+      - id: header
+        type: t_entry_header_explicit
+        if: not entry_implicit
```
GreyCat: Actually, this will result in duplicate declaration of a variable in most statically typed languages. It's better to use switching type here:

```yaml
      - id: header
        type:
          switch-on: entry_implicit
          cases:
            false: t_entry_header_explicit
            true: t_entry_header_implicit
```

It will subsequently reveal lack of static typing on some of the other clauses in this file, see below.

image/dicom_le.ksy
```diff
@@ -0,0 +1,147 @@
...
+        if: has_content
+      - id: elements
+        type: t_dataentry_state_implicit
+        if: has_elements and not is_definite
+        repeat: until
+        repeat-until: _.is_closing_tag
+      - id: elements
+        type: t_elements
+        if: has_elements and is_definite
+        size: header.content_length
```
GreyCat: Same stuff with `elements` here, but this one is slighlty more complex. Right now you can't switch between `repeat` (i.e. array) type and simple type (without `repeat`). So I'd propose to just have something like `elements_definite` or `elements_indefinite` here.

image/dicom_le.ksy
```diff
@@ -0,0 +1,147 @@
+meta:
+  id: dicom
+  file-extension: dcm
+  endian: le
```
GreyCat: As I've mentioned, we totally need `/meta/license` here. Also it would be very helpful to add `/meta/title` with official full format name, `/doc-ref` and, if possible, a few more detailed paragraphs of `/doc` — answering general questions about what this format is for, how is it related to other formats, where it is used, where can one obtain samples of the data in this format.

image/dicom_le.ksy
```diff
@@ -0,0 +1,147 @@
...
+        if: length_is_long
+      - id: p_content_length_u4
+        type: u4
+        if: length_is_long
+      - id: p_content_length_u2
+        type: u2
+        if: not length_is_long
+    instances:
+      content_length:
+        value: length_is_long ? p_content_length_u4 : p_content_length_u2
```
GreyCat: Unfortunately, YAML parser in IDE seems to be more forgiving than most other parsers, so string one actually must be quoted to be compilable with ksc.

image/dicom_le.ksy
```diff
@@ -0,0 +1,147 @@
...
+          - id: elements
+            type: t_dataentry_state_implicit
+            repeat: eos
+    instances:
+      entry_implicit:
+        value: tag_group != 2 or tag_group == 0xfffe
+      is_closing_tag:
+        value: tag_group == 0xfffe and tag_elem & 0xff00 == 0xe000 and tag_elem != 0xe000
+      has_elements:
+        value: (tag_group == 0xfffe and tag_elem == 0xe000) or header.is_seq
```
GreyCat: Using `header_is.seq` is a bad idea even in dynamic languages, as there is no guarantee that `is_seq` attribute will exist. If it won't, it will result in runtime error. With our switching type, it's better to use check and a cast:

```yaml
      has_elements:
        value: (tag_group == 0xfffe and tag_elem == 0xe000) or (not entry_implicit and (header.as<t_entry_header_explicit>.is_seq))
```

image/dicom_le.ksy
```diff
@@ -0,0 +1,147 @@
...
+      has_content:
+        value: not has_elements
+      is_definite:
+        value: header.content_length != 0xffffffff
```
GreyCat: `header.content_length` is also inaccessible when using a switching type, as compiler has no idea that all types that might appear here have `content_length` attribute (although implemented differently). Given that `content_length` is used in so many places, I'd advise to use something like:

```yaml
      content_length:
        value: >-
          entry_implicit ?
          header.as<t_entry_header_explicit>.content_length :
          header.as<t_entry_header_implicit>.content_length
      is_definite:
        value: content_length != 0xffffffff
```

image/dicom_le.ksy
```diff
@@ -0,0 +1,147 @@
...
+      - id: tag_elem
+        type: u2
+      - id: header
+        type: t_entry_header_implicit
+        if: entry_implicit
+      - id: header
+        type: t_entry_header_explicit
+        if: not entry_implicit
+      - id: content
+        size: header.content_length
```
GreyCat: Also just `content_length`, see below.

image/dicom_le.ksy
GreyCat: And `size: header.content_length` → `size: content_length` as well.

image/dicom_le.ksy
```diff
@@ -0,0 +1,147 @@
...
+        if: has_content
+      - id: elements
+        type: t_dataentry_state_explicit
+        if: has_elements and not is_definite
+        repeat: until
+        repeat-until: _.is_closing_tag
+      - id: elements
+        type: t_elements
+        if: has_elements and is_definite
+        size: header.content_length
```
GreyCat: Everything related to duplicate `header` and `elements` applies to this type as well.

image/dicom_le.ksy
```diff
@@ -0,0 +1,147 @@
...
+        size: 2
+      - id: p_reserved
+        type: u2
+        if: length_is_long
+      - id: p_content_length_u4
+        type: u4
+        if: length_is_long
+      - id: p_content_length_u2
+        type: u2
+        if: not length_is_long
```
GreyCat: In theory, this should be also replaced with a switched type:

```yaml
- id: content_length
  type:
    switch-on: length_is_long
    cases:
      false: u2
      true: u4
```

In practice, though, there is a known and yet unfixed bug that won't allow to use such attribute in ternary operator, so don't fix that just yet ;)

image/dicom_le.ksy
```diff
@@ -0,0 +1,147 @@
...
+      is_seq:
+        value: vr == 'SQ'
+
+  t_entry_header_implicit:
+    seq:
+      - id: content_length
+        type: u4
+    instances:
+      is_seq:
+        value: content_length == 0xffffffff
```
GreyCat: Maybe it's actually signed, and it would be `content_length == -1`?

image/dicom_le.ksy
sorgloomer: Yeah, I tried to use switching types, but then I could not read even the common fields of those types. I had a feeling that this is a hack, but I was going to use it from the webide anyway... Now that I know there is typecasting in the system I will look for ways to improve it so it works in other languages too

image/dicom_le.ksy
sorgloomer: I don't know what happens in ksc when content-length is negative. Also, I dont know if it is legal in a dicom fileto have length > 0x7fffffff. I will just leave it this way for now, maybe we will find somebody who knows dicom better

image/dicom_le.ksy
sorgloomer: I am not sure I understand, shall I use `"xx"` for strings?

GreyCat: No, I just mean:

```yaml
value: 'length_is_long ? p_content_length_u4 : p_content_length_u2'
```

KOLANICH:
>I've just remembered that @KOLANICH announced that he was working on DICOM 

Not quite that.
>Here is my todo list (this means I haven't started working on them) for now

That was a wishlist. The formats I have started working on are in my repo. So I glad that someone have implemented the format I haven't started working on.

GreyCat:
@sorgloomer Just wanted to remind if you'd consider adding a license, so I can merge this thing? ;)

sorgloomer:
@GreyCat Sry for the delay, it's not forgotten, there will be some progress this week.

GreyCat:
Thanks for your contribution!

sorgloomer:
@GreyCat Hmm, so the PR was automatically updated. The problem is that this does not work currently :D The docs got updated, the grammars have been restructured, but I couldn't make the imported types to work :( So, just keep this in mind

GreyCat:
I'll fix it, don't worry. Generally, as you're importing modules from one directory (one into another), relative imports would work just well here. I wonder if they work properly in WebIDE.

sorgloomer:
Alright, thanks! Could you please let me know if you manage to make it compile? There is still work left with it. For example I did not fix those type confusion things, and I remember having parsing errors with one of my test datasets too.

GreyCat:
Right now the main problem is with `dicom_common`. You seem to use it as a container for collection of classes, i.e. the main type `dicom_common` is essentially empty, but it has subtypes like `t_file_header`, `t_dataentry_implicit`, etc. It means that it would be translated into stuff like `DicomCommon::TFileHeader`, `DicomCommon::TDataentryImplicit` in most languages, thus one can't address then as just `type: t_file_header` from external files.

Unfortunately, current implementation does not allow addressing non top-level types from imported files, i.e. something like `type: dicom_common.t_file_header` would raise an error. This is partially intentional, is generally files were supposed to represent a type, not to be used as just a container for types.

Let me think a while on how to solve it.

GreyCat:
Could you elaborate, what was the original idea of splitting stuff into many different files? If the idea was just to provide a choice for API end-user, whether one wants to use "auto detect", or strictly "implicit" or "explicit" versions, then technically nothing prevents from having a hierarchy like this:

* dicom_le (`seq` corresponds to `dicom_le_generic`, i.e. use "auto detect" with `t_dataset_generic`)
  * file_implicit (`seq` corresponds to `dicom_le_implicit`)
    * t_dataset_implicit
  * file_explicit (`seq` corresponds to `dicom_le_explicit`)
    * t_dataset_explicit
  * t_dataset_generic
  * t_dataentry_implicit
  * t_dataentry_explicit
  * t_entry_header_implicit
  * t_entry_header_explicit
  * ...

This way most end users would just use `DicomLe` class. Those who want specific kind of parsing, would use `DicomLe::FileImplicit` and `DicomLe::FileExplicit`.

What do you think of it?

GreyCat:
Also, last, but not least: I believe this is a pretty complex format, so could we create a repository in spirit of ["big formats" model](https://github.com/kaitai-io/kaitai_struct/issues/165)? Basically, it means that there would be dedicated repository for DICOM .ksy file(s) + some test data there to ensure that it works + some test / dump code for simple test suite.

sorgloomer:
Sidenote: It is particularly hard to find dicom files with appropriate licensing. If someone has some, that would be great. I have a collection of dicoms I hunted from the internet, but I don't remember the origins and the licenses. Let me check for a few online.

As for the particular splitting of files: I only used the webIDE so far, where one file is one format (I believe). Having implicit/explicit/generic is just for convenience: if the enduser doesn't like entries1 and entries2 collections, he can choose a different grammar given that the format is known beforehands. So that's all, a way would have been to copy the full grammar 3 times and do proper modifications to them but my guts are against duplication.

GreyCat:
There are quite a few libraries / applications that work with DICOM. They surely must have some test suite files or something. For example, [Invesalius](https://www.cti.gov.br/invesalius) or [Medcon](http://xmedcon.sourceforge.net/) might have some? OTOH, actually I happen have some DICOM files from my own X-rays and CT. If all else fails, I guess we can use these, but they're probably not very comprehensive.

> he can choose a different grammar given that the format is known beforehands

Then the scheme I've proposed covers it. The user basically should choose between `DicomLe` vs `DicomLe::FileImplicit` and `DicomLe::FileExplicit`. We just need to mention these features in the `doc`.

GreyCat:
I've started [dicom.ksy](https://github.com/kaitai-io/dicom.ksy) repository, in which I've imported existing files, then I've merged them all in `dicom_le.ksy` (hopefully, one day it will become `dicom.ksy`, embracing both little and big-endian). Please tell me if that's ok with you. I've added you as admin to that repo.

GreyCat:
Ok, I've made some progress with dicom.ksy, which resulted in lots of restructuring. The current dicom.ksy result seems to work properly with all my real-life test files + all test files in explicit LE + implicit LE transfer syntaxes I've found. I've imported tag dictionary as enum, so now one can access `tag`.

The trick I've ultimately employed is having stuff like that:

```yaml
  t_data_element_explicit:
    seq:
      # ...
      - id: elements_implicit
        type: t_data_element_implicit
        repeat: eos
        if: is_transfer_syntax_change_implicit
  t_data_element_implicit:
    seq:
      # ...
      - id: elements
        type: t_data_element_explicit
        repeat: eos
        if: is_transfer_syntax_change_explicit
```

That is, every data element that changes transfer syntax now tries to "consume" the rest of the stream, as one of the seq attributes called `elements`.

I've also made a [simple dumper app](https://github.com/kaitai-io/dicom.ksy/blob/master/dump_dicom.rb), which produces more or less clean JSON dumps of DICOM files. Something [like that](https://gist.github.com/GreyCat/9408d30255b0943d970a572d19f7256e) — that is rtplan.dcm from pydicom test suite.

As for test data sources, I've found:

* ftp://medical.nema.org/MEDICAL/Dicom/DataSets/ — canonical sample data sets from standard originators
* https://github.com/darcymason/pydicom/tree/master/tests/test_files — pydicom also has a nice test suite, which borrows a lot from nema files

GreyCat:
Steps that I'd like to improve further:

* Convert VR to a enum based on `u2`; having VR as string might look handy, but it is ultimately much slower than having a proper enum. Enum is faster and also has other potential advantages (like an ability to have much more readable "age_string" instead of "AS", "date" instead of "DA", etc).
* Convert tag contents properly according to VR, not just leave it as a stream of bytes. I.e. "DA" should parse YYYYMMDD construct into a proper object that will provide separate `year`, `month` and `day` attributes.
* Supply "virtual" VR for implicit syntax, based on a dictionary or switch. Use this "virtual" VR to decode contents properly as well.
* Implement "explicit BE" transfer syntax. Shouldn't be that hard, that would be third `t_data_element_explicit_be` type and switching to it from other types.
* Import tons of test and do a proper test suite to ensure that it stays stable.

PR #39, opened 2017-05-22
Added ID3v2.3.0 and ID3v2.4.0 specifications

AlexAltea:
This does not include the specifications for each of the frames because that would be overkill (just too many of them, each with many quirks), but their data can be accessible as raw bytes.

More information about frames:
* ID3v2.3.0 frames: http://id3.org/id3v2.3.0#Declared_ID3v2_frames
* ID3v2.4.0 frames: http://id3.org/id3v2.4.0-frames

Everything else works exactly as expected. One minor remark is that since the standard does not require specifying how many frames are there (or how long the padding is), the only way of recognizing the last frame is reading the 4-bytes afterwards and comparing against `\x00\x00\x00\x00`. This creates a *NULL*-frame with some MP3 files. However, this behavior occurs in all audio players as well.

### DIFF
--- a/media/id3v2_3.ksy
+++ b/media/id3v2_3.ksy
@@ -1,0 +1,136 @@
+# ID3v2.3.0 Specifications
+# http://id3.org/id3v2.3.0
+
+meta:
+  id: id3v2_3
+  endian: be
+  file-extension:
+    - mp3
+
+seq:
+  - id: tag
+    type: tag
+
+types:
+  # Section 3. ID3v2 overview
+  tag:
+    seq:
+      - id: header
+        type: header
+      - id: header_ex
+        type: header_ex
+        if: header.flags.flag_headerex
+      - id: frames  
+        type: frame
+        repeat: until
+        repeat-until: _io.pos + _.size > header.size.value or _.is_invalid
+      - id: padding
+        if: header.flags.flag_headerex
+        size: header_ex.padding_size - _io.pos
+
+  # Section 3.1. ID3v2 header
+  header:
+    seq:
+      - id: magic
+        contents: 'ID3'
+      - id: version_major
+        type: u1
+      - id: version_revision
+        type: u1
+      - id: flags
+        type: flags
+      - id: size
+        type: u4be_synchsafe
+    types:
+      flags:
+        seq:
+          - id: flag_unsynchronization
+            type: b1
+          - id: flag_headerex
+            type: b1
+          - id: flag_experimental
+            type: b1
+          - id: reserved
+            type: b5
+
+  # Section 3.2. ID3v2 extended header
+  header_ex:
+    seq:
+      - id: size
+        type: u4
+      - id: flags_ex
+        type: flags_ex
+      - id: padding_size
+        type: u4
+      - id: crc
+        type: u4
+        if: flags_ex.flag_crc
+    types:
+      flags_ex:
+        seq:
+          - id: flag_crc
+            type: b1
+          - id: reserved
+            type: b15
+
+  # Section 3.3. ID3v2 frame overview
+  frame:
+    seq:
+      - id: id
+        type: str
+        size: 4
+        encoding: ASCII
+      - id: size
+        type: u4
+      - id: flags
+        type: flags
+      - id: data
+        size: size
+    types:
+      flags:
+        seq:
+          - id: flag_discard_alter_tag
+            type: b1
+          - id: flag_discard_alter_file
+            type: b1
+          - id: flag_read_only
+            type: b1
+          - id: reserved
+            type: b5
+          - id: flag_compressed
+            type: b1
+          - id: flag_encrypted
+            type: b1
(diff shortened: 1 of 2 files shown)


GreyCat:
Could you add `meta/license` tag with a valid open source license ([SPDX name](https://spdx.org/licenses/))?

AlexAltea:
Sorry for ignoring the `meta/license` tag. I realized you have to remind people about it rather often. :-)
It's now fixed. I've added GPLv2.0 since that's what appears in the footer of http://id3.org/.

GreyCat:
Merged it, thanks!

As for licensing — GPLv2.0 in the footer of id3.org refers to licensing of MoinMoin wiki, an engine that runs their web site. It totally has nothing to do with site contents per se (i.e. with ID3 technology), and even if it had anything, your work is your work, it is an implementation of that of specification, you're free to license it as you see fit.

`GPL-2.0` (which means "GNU General Public License v2.0 only") is generally a so-so choice for licensing stuff nowadays. On one hand, it's not "copyleft enough", i.e. it's not as protective as GPLv3. On the other hand, it has tons of compatibility issues, even with other free/open source license (for example, with Apache v2 or GPLv3).

Generally, stuff that we keep in "formats" repo is more-or-less trivial transcription of existing specifications, so it's kind of pointless to "disallow" proprietary products to borrow them — especially in ID3 parsing world, when there are so many permissive solutions already available. I mean, TagLib is LGPL+Mozilla, ID3Lib is LGPL (earlier versions are even public domain), etc, etc. Thus, I'd advice to use a more permissive license (if you're ok with that) — i.e. `MIT` or may be even `CC0-1.0`.

If you'd really want to keep GPL — that's totally your right, of course :) May be at least think of going with `GPL-2.0+` then, to make it compatible with more free/open source software?

AlexAltea:
Oh, my bad, I wasn't sure *who* had to pick the license. Let's stick with `CC0-1.0` then. If you want to update the `meta/license` fields yourself now, feel free to do so. Otherwise, I'll update it with my next PR, as I'm preparing more specifications. :-)

GreyCat:
Ok, got it updated. I wonder if it would be a good idea to merge both id3v2_3 and id3v2_4 into some universal v2-style tag format, that would automatically choose what to parse based on `version_major` / `version_revision`?

Also, some code review. I believe that this:

```yaml
      - id: frames  
        type: frame
        repeat: until
        repeat-until: _io.pos + _.size.value > header.size.value or _.is_invalid
```

and all this

```yaml
  padding:
    seq:
      - id: padding
        size: _root.tag.header.size.value - _io.pos
```

is done for the sole purpose of setting tag body size to `_root.tag.header.size`? Why don't just use a substream here, i.e. something like

```yaml
  tag:
    seq:
      - id: header
        type: header
      - id: header_ex
        type: header_ex
        if: header.flags.flag_headerex
      - id: body
        size: header.size.value # minus something, if applicable
        type: tag_body
  tag_body:
    seq:
      - id: frames
        repeat: until
        repeat-until: _io.eof or _.is_invalid
      - id: footer
        type: footer
        if: _parent.header.flags.flag_footer
      # no need to handle padding manually, substream skips
      # all these bytes automatically for you
```

AlexAltea:
Regarding merging formats, I find them too wildly different (nothing in common other than header and synchsafe integers). The poor adoption of ID3v2.4.0 only proves this fact. Also, there was an `id3v1_1.ksy` file already so I thought having a per-subversion specification would be the most logical thing to do.

That said, feel free to reorganize the specifications as you see adequate. :-)

> repeat-until: _io.eof or _.is_invalid

Although unlikely, it could happen that there's no padding following the list of frames, but there's a footer afterwards. In such a case neither `_io.eof` nor `_.is_invalid` would be triggered. I'm really not sure how to optimize the specifications without breaking anything.

GreyCat:
> I find them too wildly different (nothing in common other than header and synchsafe integers).

Understood, makes sense. "Synchsafe integers" are may be candidates to move to /common, though, as they might be used in many other formats.

I also wanted to note that we're currently discussing a model of repositories for relatively complex formats in https://github.com/kaitai-io/kaitai_struct/issues/165 — that idea implies that there is one repo per such format, and this repo contains not only the .ksy file itself, but also some test data and test suites to make sure that .ksy works as intended. By any chance you'd be interested in joining this initiative and making a separate id3 ksy repository?

AlexAltea:
> "Synchsafe integers" [...] might be used in many other formats.

Unfortunately, they seem to be specific to ID3 tags, but I guess it would still make sense to move them to `/common` nonetheless. See: https://en.wikipedia.org/wiki/Synchsafe.

> By any chance you'd be interested in joining this initiative and making a separate id3 ksy repository?

Sure, although at a later point in time. These ID3v2 specifications are part of a larger project, which will eventually result in different specifications and tests files (hopefully serialization for a subset of Kaitai too, though that's TBD). I'll try to release as much as possible.

PR #41, opened 2017-05-30
Added Gran Turismo GTFS .vol files support

mnakamura1337:
So, this is clearly not VNs, I hope you'll accept that into main repo.

Tested on GT2, probably should work on GT1 as well. And, yeah, I've even managed to not forget to add the `meta/license` tag ;)

All reverse engineering is my own work. I might do KaitaiFS module for this later.


GreyCat:
Thanks! Can't really test it, but it seems to compile just fine. Looking forward for a KaitaiFS module ;)

PR #43, opened 2017-06-01
Stop reading PNGs after the IEND chunk

Natronium:
If there are extra bytes at the end of the PNG (e.g. if the PNG is embedded in some other sort of file), the parser will try to read in new chunks even if they're nonsensical, and will often fail because of
invalid data in the garbage chunk (like a "chunk size" larger than the file) By stopping at the IEND chunk, the parser won't keep trying to read more chunks past the end of the file.

### DIFF
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -19,7 +19,8 @@ seq:
   # The rest of the chunks
   - id: chunks
     type: chunk
-    repeat: eos
+    repeat: until
+    repeat-until: _.type == "IEND" or _io.eof
 types:
   chunk:
     seq:


GreyCat:
Could you provide some pointers on why this is desired behavior (i.e. some place in PNG standard mandates that, or libpng does that, etc)?

We should as well think of `repeat-until: _.type == "IEND" or _io.eof`, i.e. allowing parsing of corrupt files without an IEND chunk? From practical point of view, it looks like at least libpng allows that.

Natronium:
#### The standard:
According to [5.6 Chunk Ordering](https://www.w3.org/TR/PNG/#5ChunkOrdering) in the PNG spec, the (only) IEND chunk is the last chunk in the file. and the [IEND chunk specification](https://www.w3.org/TR/PNG/#11IEND) says that the IEND chunk "marks the end of the PNG
datastream", so it looks like any data after the IEND chunk isn't part of the PNG by definition.

#### libpng:
From [the comment in pngread.c](https://github.com/glennrp/libpng/blob/0c440812db5f3f8929db87ff840bc51fd09a8af3/pngread.c#L756) that describes `png_read_end`: 

> `[png_read_end] will not read past the end of the file`

and the code agrees: `png_read_end` reads [until the `PNG_HAVE_IEND` mode flag is set](https://github.com/glennrp/libpng/blob/0c440812db5f3f8929db87ff840bc51fd09a8af3/pngread.c#L923), and, when it encounters the IEND chunk, it calls [`png_handle_iend`](https://github.com/glennrp/libpng/blob/0808d7585198158fb3abdbb83fa29ed82bd48fb1/pngrutil.c#L1096), which sets the `PNG_HAVE_IEND` mode flag.

Reading a truncated PNG without an IEND chunk definitely makes sense, good catch.

I've updated the PR with the `or _io.eof` case, and added the details from the spec, and libpng, to the commit message.

GreyCat:
Top notch work, thanks a lot!

PR #45, opened 2017-06-22
Removed redundant type

ujjwal96:
Removes a redundant type in tls client hello parser
Earlier -
```
for cs in cipher_suites.cipher_suites:
    print(cs.cipher_suite)
```
screenshot from 2017-06-22 06-15-37
Now -
```
for cs in cipher_suites.cipher_suites:
    print(cs)
```
screenshot from 2017-06-22 06-16-38

### DIFF
--- a/network/tls_client_hello.ksy
+++ b/network/tls_client_hello.ksy
@@ -54,14 +54,9 @@ types:
         type: u2
 
       - id: cipher_suites
-        type: cipher_suite
+        type: u2
         repeat: expr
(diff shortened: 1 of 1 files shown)


GreyCat:
Thanks!

GreyCat:
May be we shall import cipher suite list as enums here?
https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4

PR #46, opened 2017-07-12
Initial support for Microsoft WAV audio files

johnwbyrd:
This ksy file supports PCMWAVEFORMAT, WAVEFORMATEX and WAVEFORMATEXTENSIBLE subtypes, which means it should work for multiple channel and other esoteric formats.  It gives special consideration to fmt, data, bext, and cue chunks, which have specific meaning to the Microsoft format.

### DIFF
--- a/media/wav.ksy
+++ b/media/wav.ksy
@@ -1,0 +1,272 @@
+meta:
+  id: wav
+  file-extension: wav
+  title: Microsoft WAVE audio file
+  endian: le
+  license: BSD-3-Clause-Attribution
+  encoding: ASCII
+doc: |
+  The WAVE file format is a subset of Microsoft's RIFF specification for the
+  storage of multimedia files. A RIFF file starts out with a file header
+  followed by a sequence of data chunks. A WAVE file is often just a RIFF
+  file with a single "WAVE" chunk which consists of two sub-chunks --
+  a "fmt " chunk specifying the data format and a "data" chunk containing
+  the actual sample data.
+  This Kaitai implementation was written by John Byrd of Gigantic Software
+  (jbyrd@giganticsoftware.com), and it is likely to contain bugs.
+doc-ref: http://soundfile.sapp.org/doc/WaveFormat/
+doc-ref: https://www.loc.gov/preservation/digital/formats/fdd/fdd000001.shtml
+seq:
+  - id: riff_id
+    contents: RIFF
+  - id: file_size
+    type: u4
+  - id: wave_id
+    contents: WAVE
+  - id: chunks
+    type: chunks_type
+    size: file_size - 5
+  - id: unused
+    size-eos: true
+    
+instances: 
+  format_chunk:
+    value: _root.chunks.chunk[0].data 
+
+types:  
+  chunks_type:
+    seq:   
+      - id: chunk
+        type: chunk_type
+        repeat: eos
+ 
+  chunk_type:
+    seq: 
+      - id: chunk_id
+        type: u4be
+      - id: len
+        type: u4 
+      - id: data
+        size: len
+        type: 
+          switch-on: chunk_id 
+          cases:
+            0x666d7420: format_chunk_type
+            0x62657874: bext_chunk_type
+            0x63756520: cue_chunk_type
+            0x64617461: data_chunk_type
+            
+  bext_chunk_type:
+    seq:
+    - id: description
+      size: 256
+      type: str
+    - id: originator 
+      size: 32
+      type: str
+    - id: originator_reference
+      size: 32
+      type: str
+    - id: origination_date
+      size: 10
+      type: str
+    - id: origination_time
+      size: 8
+      type: str
+    - id: time_reference_low
+      type: u4
+    - id: time_reference_high
+      type: u4
+    - id: version
+      type: u2
+    - id: umid
+      size: 64
+    - id: loudness_value
+      type: u2
+    - id: loudness_range
+      type: u2
+    - id: max_true_peak_level
+      type: u2
+    - id: max_momentary_loudness
+      type: u2
+    - id: max_short_term_loudness
+      type: u2
+      
+  cue_chunk_type:
+    seq:
+      - id: dw_cue_points
+        type: u4
+      - id: cue_points
+        type: cue_point_type
+        repeat: expr
+        repeat-expr: dw_cue_points
+        if: dw_cue_points != 0
+        
+  cue_point_type:
+    seq:
+      - id: dw_name
+        type: u4
+      - id: dw_position
+        type: u4
+      - id: fcc_chunk
+        type: u4
+      - id: dw_chunk_start
+        type: u4
+      - id: dw_block_start
+        type: u4
+      - id: dw_sample_offset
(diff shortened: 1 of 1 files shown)


GreyCat:
Thanks! I'll add some comment along the commit, please take a look? ;)

media/wav.ksy
```diff
@@ -0,0 +1,272 @@
+meta:
+  id: wav
+  file-extension: wav
+  title: Microsoft WAVE audio file
+  endian: le
+  license: BSD-3-Clause-Attribution
```
GreyCat: This license probably needs adding a license text to the file, i.e. that "Redistribution and use in source and binary forms, with or without modification, are permitted... EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE."?

GreyCat: Probably it's worth to add it as a comment at the beginning of the file?

johnwbyrd: You can change the license to any reasonable license, so long as it's not "you can sue John Byrd for screwing this up".

johnwbyrd: It's just a single ksy, which may be useful to some audio programmer types.

media/wav.ksy
```diff
@@ -0,0 +1,272 @@
...
+  The WAVE file format is a subset of Microsoft's RIFF specification for the
+  storage of multimedia files. A RIFF file starts out with a file header
+  followed by a sequence of data chunks. A WAVE file is often just a RIFF
+  file with a single "WAVE" chunk which consists of two sub-chunks --
+  a "fmt " chunk specifying the data format and a "data" chunk containing
+  the actual sample data.
+  This Kaitai implementation was written by John Byrd of Gigantic Software
+  (jbyrd@giganticsoftware.com), and it is likely to contain bugs.
+doc-ref: http://soundfile.sapp.org/doc/WaveFormat/
+doc-ref: https://www.loc.gov/preservation/digital/formats/fdd/fdd000001.shtml
```
GreyCat: This actually won't work as intended. All YAML parsers that we use would provide just one `doc-ref` element out of these. May be it's worth to redesign `doc-ref` element to allow multiple references...

johnwbyrd: ok, I say go for the second one if we have to choose

media/wav.ksy
```diff
@@ -0,0 +1,272 @@
...
+    contents: RIFF
+  - id: file_size
+    type: u4
+  - id: wave_id
+    contents: WAVE
+  - id: chunks
+    type: chunks_type
+    size: file_size - 5
+  - id: unused
+    size-eos: true
```
GreyCat: This is not really necessary to grab the end of file as a distinct attribute. Actually, if anyone wants it, it will have the file pointer positioned right here after finishing the parsing, so it's easy to read it (or invoke another parser, etc), I guess?

media/wav.ksy
```diff
@@ -0,0 +1,272 @@
...
+    
+instances: 
+  format_chunk:
+    value: _root.chunks.chunk[0].data
```
GreyCat: `chunks.chunk[0].data` should be enough here, you're technically already in `_root`.

media/wav.ksy
johnwbyrd: Most of the WAVs I looked at seemed to have a single byte containing zero after the end of the RIFF chunk.  That zero can go away or stay, as you prefer.

media/wav.ksy
johnwbyrd: True that :)

media/wav.ksy
```diff
@@ -0,0 +1,272 @@
...
+    value: _root.chunks.chunk[0].data 
+
+types:  
+  chunks_type:
+    seq:   
+      - id: chunk
+        type: chunk_type
+        repeat: eos
+ 
+  chunk_type:
```
GreyCat: I wonder if having type named `chunk_type` and enum named `chunk_type` would clash in some languages. Need to check.

johnwbyrd: Thanks, I didn't notice that one.  Best we just do an end-run around the question and rename one of them, until it's guaranteed to work.

media/wav.ksy
```diff
@@ -0,0 +1,272 @@
...
+    seq: 
+      - id: chunk_id
+        type: u4be
```
GreyCat: I believe you've created `chunk_type` enum to be actually used here as `enum: chunk_type`? May be you've encountered any problems and abandoned the idea?

media/wav.ksy
```diff
@@ -0,0 +1,272 @@
...
+        type: u4 
+      - id: data
+        size: len
+        type: 
+          switch-on: chunk_id 
+          cases:
+            0x666d7420: format_chunk_type
+            0x62657874: bext_chunk_type
+            0x63756520: cue_chunk_type
+            0x64617461: data_chunk_type
```
GreyCat: This would look better with enums, like

```yaml
cases:
  'chunk_type::fmt': format_chunk_type
  'chunk_type::bext': bext_chunk_type
```

etc.

johnwbyrd: Agreed, but when I tried to write that out with enums, the ksy compiler refused to touch it -- it wasn't entirely clear to me why that was, so I left them as constants.  If you can get your idea to compile in the web IDE, please do :)

media/wav.ksy
johnwbyrd: Yep, you got it.  I got it to "proof of concept" and decided I'd hand it over to people who actually knew the dsl.

media/wav.ksy
GreyCat: I'd actually go with a link to "official" (i.e. Microsoft's documentation) in this case, i.e. http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/Docs/RIFFNEW.pdf

`loc.gov/.../fdd/...` link should be probably in `meta/xref` in abbreviated form.

media/wav.ksy
```diff
@@ -0,0 +1,272 @@
+meta:
+  id: wav
+  file-extension: wav
+  title: Microsoft WAVE audio file
+  endian: le
+  license: BSD-3-Clause-Attribution
+  encoding: ASCII
```
GreyCat: Actually, this is incredibly messy ;) Microsoft's documentation uses terms like "STR", "BSTR", "ZSTR", etc, which never specify encoding. In practice, most software in Windows uses "ANSI" encoding there, i.e. some arbitrary single-bit encoding, specific to locale of the OS that was used to create the file...

johnwbyrd: True that, but I know something Microsoft doesn't -- Microsoft stole this format from Electronic Arts, and EA designed the format at a time when it was ASCII or nothing.  ANSI hadn't been invented yet.  Ergo, ASCII.

media/wav.ksy
johnwbyrd: Can we go with http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/WAVE.html instead please?  There's a lot of ancillary information I used from that web site as well.  Sorry to be overly choosy :/

PR #49, opened 2017-08-19
mach_o.ksy: Provide bitmask-like access to vm_prot by using a bX-based declaration

FilippoBiga:
As per discussion on issue 222 on the main repository (https://github.com/kaitai-io/kaitai_struct/issues/222).


GreyCat:
Thanks!

PR #50, opened 2017-08-21
Extend and improve support for the Mach-O file format

FilippoBiga:
This pull request aims at providing the following changes:

- Add support for arm64 cpu type
- Fix handling of `LC_MAIN` load command
- Fix `load_upward_dylib` load command type
- Add support and handle most of the missing commands, adding the declarations of the struct they represent where necessary:
	
	- `LC_ROUTINES`
	- `LC_ENCRYPTION_INFO` and `LC_ENCRYPTION_INFO_64`
	- `LC_SEGMENT_SPLIT_INFO`
	- `LC_ROUTINES_64`
	- `LC_SUB_FRAMEWORK`, `LC_SUB_CLIENT`, `LC_SUB_UMBRELLA` and `LC_SUB_LIBRARY`
	- `LC_LINKER_OPTION`
	- `LC_TWOLEVEL_HINTS`
	- `LC_ID_DYLINKER`
	- `LC_DYLD_ENVIRONMENT`
	- `LC_DYLIB_CODE_SIGN_DRS`
	- `LC_VERSION_MIN_{IPHONEOS,TVOS,WATCHOS}`

Please, let me know if the declaration needs to be improved with regards to style conventions or other factors. I would be happy to make any change, if required.

### DIFF
--- a/executable/mach_o.ksy
+++ b/executable/mach_o.ksy
@@ -196,6 +196,7 @@ types:
             'load_command_type::dylib_code_sign_drs' : linkedit_data_command
             'load_command_type::code_signature'      : code_signature_command
             'load_command_type::encryption_info_64'  : encryption_info_command_64
+            'load_command_type::twolevel_hints'      : twolevel_hints_command
     -webide-representation: '{type}: {body}'
   vm_prot:
     seq:
@@ -751,6 +752,12 @@ types:
       - id: cryptid
         type: u4
       - id: pad
+        type: u4
+  twolevel_hints_command:
+    seq:
+      - id: offset
+        type: u4
+      - id: nhints
         type: u4
   version_min_command:
     seq:
--- a/executable/mach_o.ksy
+++ b/executable/mach_o.ksy
@@ -175,6 +175,8 @@ types:
             'load_command_type::symtab'              : symtab_command
             'load_command_type::dysymtab'            : dysymtab_command
             'load_command_type::load_dylinker'       : dylinker_command
+            'load_command_type::id_dylinker'         : dylinker_command
+            'load_command_type::dyld_environment'    : dylinker_command
             'load_command_type::uuid'                : uuid_command
             'load_command_type::version_min_macosx'  : version_min_command
             'load_command_type::version_min_iphoneos': version_min_command
--- a/executable/mach_o.ksy
+++ b/executable/mach_o.ksy
@@ -171,6 +171,7 @@ types:
           switch-on: type
           cases:
             'load_command_type::segment_64'              : segment_command_64
+            'load_command_type::dyld_info'               : dyld_info_command
             'load_command_type::dyld_info_only'          : dyld_info_command
             'load_command_type::symtab'                  : symtab_command
             'load_command_type::dysymtab'                : dysymtab_command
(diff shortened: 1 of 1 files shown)


executable/mach_o.ksy
```diff
@@ -730,6 +754,72 @@ types:
...
+        - id: reserved2
+          type: u4
+        - id: reserved3
+          type: u4
+        - id: reserved4
+          type: u4
+        - id: reserved5
+          type: u4
+        - id: reserved6
+          type: u4
```
GreyCat: May be it's better to do single `reserved` field with `size: 24` here?

executable/mach_o.ksy
```diff
@@ -730,6 +754,72 @@ types:
...
+  twolevel_hints_command:
+    seq:
+      - id: offset
+        type: u4
+      - id: nhints
+        type: u4
+  linker_option_command:
+    seq:
+      - id: count
+        type: u4
```
GreyCat: BTW, according to the [style guide](http://doc.kaitai.io/ksy_style_guide.html) we're trying to introduce, it's generally suggested that it's better to call this `num_strings`, to maintain the same consistent naming style everywhere.

executable/mach_o.ksy
```diff
@@ -730,6 +754,72 @@ types:
...
+      - id: pad
+        type: u4
+        if: _root.magic == magic_type::macho_be_x64 or _root.magic == magic_type::macho_le_x64
+  twolevel_hints_command:
+    seq:
+      - id: offset
+        type: u4
+      - id: nhints
+        type: u4
```
GreyCat: Again, if that's "number of hints", it should be probably `num_hints`.

executable/mach_o.ksy
```diff
@@ -730,6 +754,72 @@ types:
...
+      - id: strings
+        type: strz
+        encoding: utf-8
+        repeat: expr
+        repeat-expr: count
+  sub_command:
+    seq:
+        - id: name
+          type: lc_str
```
GreyCat: Too much space in YAML alignment here, I guess.

executable/mach_o.ksy
FilippoBiga: Thank you for pointing me to the style guide. Question: do the naming conventions apply to fields that already have a specific name (i.e. I used the name of the fields defined by Apple)?

GreyCat: First of all, note that this "style guide" is currently in development (see https://github.com/kaitai-io/kaitai_struct/issues/140 for details), so it's not exactly mandatory to apply every principle.

For naming, it's more like a common sense question. If that's a very common name that everyone knows and it is used in many different implementations, etc. — it probably a good idea to leave it as is (although note that you can't have IDs with upper case letters and various special characters anyway).

In most other cases, usually it's a good idea to match suggested style with "num_", "len_", "ofs_" prefixes, and keep original names using `-orig-id` for search purposes. For example, even if there is some particular naming convention used in some original implementation, usually it's better to cater for KS users (who would be familiar with KS convention), not users of that original implementation.

FilippoBiga: Ok, thank you for the clarification. I will be performing the suggested changes tomorrow, and hopefully I will have time to also add `-orig-id` where needed throughout the whole specification.

GreyCat: Meanwhile, I'll merge this PR. Hope @koczkatamas won't mind, everything seems to be clear and legit here (except for space formatting, I guess, but that's easy to fix).

koczkatamas: lgtm

PR #51, opened 2017-08-22
mach_o.ksy: Fix styling issues and add -orig-id

FilippoBiga:
As discussed in https://github.com/kaitai-io/kaitai_struct_formats/pull/50, this pull request applies the required style changes and adds `-orig-id` to some members of the declaration.

### DIFF
--- a/executable/mach_o.ksy
+++ b/executable/mach_o.ksy
@@ -769,11 +769,11 @@ types:
     seq:
       - id: offset
         type: u4
-      - id: nhints
+      - id: num_hints
         type: u4
   linker_option_command:
     seq:
-      - id: count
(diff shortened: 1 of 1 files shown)


executable/mach_o.ksy
```diff
@@ -782,44 +782,24 @@ types:
         repeat-expr: count
```
GreyCat: It won't compile: `count` is now `num_strings`.

FilippoBiga: f95d590 fixes the issue, thanks.

GreyCat:
Thanks! I'll fix the whitespace formatting problems, if you don't mind :)

PR #52, opened 2017-08-25
Added mdt format.

KOLANICH:


KOLANICH:
P.S. 133 files changed looks like some kind of glitch. Feel free to create a commit yourself.

koczkatamas:
Permissions were changed from `644` to `755` (`a+x`).

GreyCat:
Thanks for the contribution! I've recomitted them both as https://github.com/kaitai-io/kaitai_struct_formats/commit/de6a0ab4847fe6e43e78275b2fe12b9df2b9e49b

I've moved it around the directories slightly and fixed file names / meta IDs for consistency. Will probably review them now and try to fix some trivial things as well.

GreyCat:
Well, it compiles, it kind of works, but not exactly. From all the files I've tried, only one (test_structure.mdt) actually gets parsed completely without errors.

Also, reading bitmap image as true array of `s2` is just such a huge overkill %) Shall we at least do that a byte array?

KOLANICH:
1 some frame types may be not implemented for now, for example raman ones
2 the spec is not quite finished, I'm going to improve it somewhen.

PR #54, opened 2017-09-05
Updated iNES format

MetaDark:
### Overview
- Removed `meta/endian`: The specification doesn't define an endianness and doesn't require one
- Removed `ines` file extension: The specification only defines `.nes` as a valid file extension
  - This is debatable, even though the spec doesn't define it, it may be useful to define in the KSY file anyway
- Renamed a few fields for clarity
- Added enums for b1 types that don't fit true/false
- Added missing fields for the body of the ROM file

A collection of test ROMs can be found at https://wiki.nesdev.com/w/index.php/Emulator_tests. These tests are designed for emulator implementations, however they can also be used to test just the file format.

### DIFF
--- a/firmware/ines.ksy
+++ b/firmware/ines.ksy
@@ -1,10 +1,10 @@
 # TODO: Support NES 2.0
-doc-ref: https://wiki.nesdev.com/w/index.php/INES
 meta:
   id: ines
   file-extension: nes
   encoding: ASCII
   license: WTFPL
+doc-ref: https://wiki.nesdev.com/w/index.php/INES
 seq:
   - id: header
     type: header
--- a/firmware/ines.ksy
+++ b/firmware/ines.ksy
@@ -1,3 +1,4 @@
+# TODO: Support NES 2.0
 doc-ref: https://wiki.nesdev.com/w/index.php/INES
 meta:
   id: ines
@@ -32,8 +33,7 @@ types:
       - id: f10
         type: f10
         doc: this one is unofficial
-      - id: zero_fill
-        size: 4
+      - contents: [0, 0, 0, 0, 0]
     instances:
       # TODO: Add an enum for mapper. https://wiki.nesdev.com/w/index.php/List_of_mappers
       mapper:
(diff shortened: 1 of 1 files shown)


GreyCat:
Thanks!

Cc @DarkLinkXXXX, the original author, for review (in case if GitHub hasn't sent an update notification already). There are quite a few changes.

@MetaDark, while we're here, could I ask you to see if you could apply our [KSY style guide](http://doc.kaitai.io/ksy_style_guide.html) to this ksy? For example, may be `prg_rom_size` could be renamed to `len_prg_rom`?

firmware/ines.ksy
```diff
@@ -24,73 +36,113 @@ types:
...
+          - id: tv_system
+            type: b2
+            enum: tv_system
+            doc: if 0, NTSC. If 2, PAL. If 1 or 3, dual compatible.
+        enums:
+          tv_system:
+            0: ntsc
+            1: dual
+            2: pal
+            3: dual
```
GreyCat: This won't compile for majority of static language targets (like C++, Java, C#), generating something like that:

```java
    public enum TvSystem {
        NTSC(0),
        DUAL(1),
        PAL(2),
        DUAL(3);
// ...
```

which yields something like "error: variable DUAL is already defined in enum TvSystem".

MetaDark: Hmm, is there a way to model this better in KSY? Should it be preprocessed before being mapped to an enum?

GreyCat: I'd go with `dual1` and `dual2` for the time being. When we'll add writing/generation capabilities, this way you could specify which value you want exactly?

GreyCat:
Thanks! Let's wait for the comment from @DarkLinkXXXX then — it looks awesome from my perspective :)

GreyCat:
Well, a week passed, so it's probably ok :) Hopefully @DarkLinkXXXX won't oppose it :) Thanks!

DarkLinkXXXX:
Wew, sorry about that.  I wasn't familiar with the spec before I wrote ines.ksy so I'm pretty indifferent.

PR #55, opened 2017-09-24
Added lvm2

KOLANICH:
Comments?

### DIFF
--- a/filesystem/lvm2.ksy
+++ b/filesystem/lvm2.ksy
@@ -1,0 +1,164 @@
+meta:
+  title: Logical Volume Manager version 2
+  id: lvm2
+  endian: le
+  encoding: ascii
+  application:
+    - linux
+    - grub2
+    - lvm tools
+    - libvslvm
+  license: GFDL-1.3+
+doc: |
+  === Building a test file
+  ....
+  dd if=/dev/zero of=image.img bs=512 count=$(( 4 * 1024 * 2 ))
+  sudo losetup /dev/loop1 image.img
+  sudo pvcreate /dev/loop1
+  sudo vgcreate vg_test /dev/loop1
+  sudo lvcreate --name lv_test1 vg_test
+  sudo losetup -d /dev/loop1
+  ```
+doc-ref: https://github.com/libyal/libvslvm/blob/master/documentation/Logical%20Volume%20Manager%20(LVM)%20format.asciidoc
+instances:
+  sector_size:
+    value: 512 # TODO: how about 4k sectors?
+seq:
+  - id: pv
+    type: physical_volume
+types:
+  physical_volume:
+    seq:
+      - id: empty_sector
+        size: _root.sector_size 
+      - id: label
+        type: label
+    types:
+      label:
+        seq:
+          - id: label_header
+            type: label_header
+          - id: volume_header
+            type: volume_header
+        types:
+          label_header:
+            seq:
+              - id: signature
+                contents: "LABELONE"
+              - id: sector_number
+                type: u8
+                doc: "The sector number of the physical volume label header"
+              - id: checksum
+                type: u4
+                doc: "CRC-32 for offset 20 to end of the physical volume label sector"
+              - id: label_header_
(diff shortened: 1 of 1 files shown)


filesystem/lvm.ksy
```diff
@@ -0,0 +1,163 @@
...
+  id: lvm
+  endian: le
+  encoding: ascii
+  application:
+    - linux
+    - grub2
+    - lvm tools
+    - libvslvm
+  license: GNU-FDL-1.3+
+doc: >
```
GreyCat: This must be `|`, not `>`, or it will be parsed as one long line, not separate lines (as your formatting implies here).

filesystem/lvm.ksy
```diff
@@ -0,0 +1,163 @@
...
+  title: Logical Volume Manager
+  id: lvm
+  endian: le
+  encoding: ascii
+  application:
+    - linux
+    - grub2
+    - lvm tools
+    - libvslvm
+  license: GNU-FDL-1.3+
```
GreyCat: 1. That expression in SPDX would be [GFDL-1.3+](https://spdx.org/licenses/GFDL-1.3.html), not "GNU-FDL-1.3+".
2. I see that you're borrowing lots of stuff from https://github.com/libyal/libvslvm/blob/master/documentation/Logical%20Volume%20Manager%20(LVM)%20format.asciidoc — which is probably licensed on the same terms as whole libvslvm — i.e. [LGPLv3+](https://github.com/libyal/libvslvm/blob/master/COPYING). One can't just slap GFDL on that — you'll need author's permission. Have you contacted Joachim Metz? May be he'll be intersted in this ksy as well ;)

KOLANICH: That document is licensed under GNU Free Documentation License.

filesystem/lvm.ksy
KOLANICH: It's by design: I think that manual wrapping is evil, every modern text editor has automatical wrapping, which takes into account window width and text size.

KOLANICH:
@GreyCat: I have updated the PR

filesystem/lvm.ksy
GreyCat: Ah, I've missed info inside the document itself. Thanks!

filesystem/lvm2.ksy
```diff
@@ -0,0 +1,164 @@
...
+    - libvslvm
+  license: GFDL-1.3+
+doc: |
+  === Building a test file
+  ....
+  dd if=/dev/zero of=image.img bs=512 count=$(( 4 * 1024 * 2 ))
+  sudo losetup /dev/loop1 image.img
+  sudo pvcreate /dev/loop1
+  sudo vgcreate vg_test /dev/loop1
+  sudo lvcreate --name lv_test1 vg_test
```
GreyCat: BTW, this command returns on several systems where I've tried that:

```
  No command with matching syntax recognised.  Run 'lvcreate --help' for more information.
```

Probably the correct one is something like:

```
sudo lvcreate --size=4M --name lv_test1 vg_test
```

Also, at least on the systems I've tested, minimal extent size is 4MiB, and it's impossible to create LV on a PV that totals 4MiB (as you use `dd` which creates 4MiB image.img). Using a larger file fixes that problem as well.

filesystem/lvm2.ksy
```diff
@@ -0,0 +1,164 @@
...
+              - id: signature
+                contents: "LABELONE"
+              - id: sector_number
+                type: u8
+                doc: "The sector number of the physical volume label header"
+              - id: checksum
+                type: u4
+                doc: "CRC-32 for offset 20 to end of the physical volume label sector"
+              - id: label_header_
+                type: label_header_
```
GreyCat: Um, probably a bad idea to use names like that — both `label_header` and `label_header_` would be transformed into `LabelHeader` in camel case style, and will clash.

GreyCat:
It seems to work and pass borderline requirements ;) Merged in, thanks!

PR #57, opened 2017-09-27
Added .editorconfig

KOLANICH:
.editorconfig is a file setting up source code formatting  (tabs/spaces, how many spaces in a tab, presence new lin in the end of file)  and eol style. There are plugins for some editors.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+root = true
+
+[*.ksy]
+charset=utf-8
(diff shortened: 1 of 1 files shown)


KOLANICH:
@GreyCat

GreyCat:
While I have actually failed to find any editors that support this out of the box, probably it's still a good idea to support this standard. Just hope it would work for somebody.

PR #59, opened 2017-09-27
Virtualbox Virtual Disk Image

KOLANICH:


GreyCat:
Looks really cool! A few minor suggestions...

filesystem/vdi.ksy
```diff
@@ -0,0 +1,228 @@
+meta:
+  id: vdi
+  endian: le
+  encoding: utf-8
+  application:
+    - VirtualBox
+    - QEMU
+    - VMWare Workstation
+  license: GPL3.0+
```
GreyCat: 1. Probably something like `title: VDI (native VirtualBox Disk Image)` would be helpful.
2. Suggested order of `meta` entries is somewhat different in [style guide](http://doc.kaitai.io/ksy_style_guide.html).

filesystem/vdi.ksy
```diff
@@ -0,0 +1,228 @@
...
+doc-ref: https://github.com/qemu/qemu/blob/master/block/vdi.c
+#  - https://github.com/qemu/qemu/blob/master/block/vdi.c
+#  - https://www.virtualbox.org/browser/vbox/trunk/src/VBox/Storage/VDI.cpp
+#  - https://forums.virtualbox.org/viewtopic.php?t=8046
```
GreyCat: I'd rather put forums.virtualbox.org link here as a main one, as it is the essential source. Everything else (especially non-VirtualBox code) is reinterpretation — I'd mention them only if they contain any non-trivial clarifications / better info.

filesystem/vdi.ksy
```diff
@@ -0,0 +1,228 @@
...
+    instances:
+      subheader_size_is_dynamic:
+        value: version.major>=1
+      header_size:
+        value: (subheader_size_is_dynamic?header_size_optional:336)
+      
+      bmap_size:
+        value: ((header_main.blocks_in_image * 4+header_main.geometry.sector_size-1) / header_main.geometry.sector_size)*header_main.geometry.sector_size
+      bmap_offset:
+        value: header_main.offset_bmap
```
GreyCat: Probably it's better to use `ofs_` and `len_` prefixes consitently for stuff like that, as suggested in the style guide. Also, if you've decided to rename "bmap" into "blocks_map" (which I totally support), it should be consistent everywhere as well.

filesystem/vdi.ksy
```diff
@@ -0,0 +1,228 @@
...
+
+      blocks_offset:
+        -orig-id: data_offset
+        value: header_main.offset_data
```
GreyCat: Again, it's really harsh. If it's "blocks", then it should be "blocks" everywhere.

filesystem/vdi.ksy
```diff
@@ -0,0 +1,228 @@
...
+        if: subheader_size_is_dynamic
+        
+      - id: header_main
+        size: header_size
+        type: header_main
+    instances:
+      subheader_size_is_dynamic:
+        value: version.major>=1
+      header_size:
+        value: (subheader_size_is_dynamic?header_size_optional:336)
```
GreyCat: Spaces are welcome between operators and operands.

filesystem/vdi.ksy
```diff
@@ -0,0 +1,228 @@
...
+              - id: heads
+                type: u4
+
+              - id: sectors
+                type: u4
+
+              - id: sector_size
+                type: u4
+          flags:
+            seq: #little endian assummed
```
GreyCat: I believe VirtualBox is strictly little-endian anyway?

filesystem/vdi.ksy
```diff
@@ -0,0 +1,228 @@
...
+              - id: fixed
+                type: b1
+              - id: reserved2
+                type: b8
+  blocks_map:
+    seq:
+      - id: index
+        type: block_index
+        repeat: expr
+        repeat-expr: _root.header.header_main.blocks_in_image
```
GreyCat: `num_blocks` or something like that.

GreyCat:
Thanks for this contribution!

filesystem/vdi.ksy
KOLANICH: I used QEMU source code as the main one: it is a bit simpler than vbox one. The names and docs are taken from there.  VBox source was used to look up some minor details like enum members and an additional geometry field.

filesystem/vdi.ksy
KOLANICH: The format is little endian, in qemu comments it is written that some fields are little endian even in memory.

filesystem/vdi.ksy
KOLANICH: It is speculated in multiple sources that VDI is an abbreviation for VirtualBox Disk Image.

PR #62, opened 2017-09-29
Added hashcat hccap

KOLANICH:
also added some additional formats

### DIFF
--- a/network/802.11/hccap.ksy
+++ b/network/802.11/hccap.ksy
@@ -1,0 +1,61 @@
+meta:
+  id: hccap
+  title: Hashcat capture old format
+  license: Unlicense
+  file-extension: hccap
+  application:
+    - Hashcat
+    - aircrack-ng
+  endian: le
+  encoding: utf-8
+  imports:
+    - /network/mac_address
+    - /network/802.11/wpa2_nonce
+doc: |
+  Native format of Hashcat password "recovery" utility
+  A sample of file for testing can be downloaded from https://web.archive.org/web/20150220013635if_/http://hashcat.net:80/misc/example_hashes/hashcat.hccap
+doc-ref: https://hashcat.net/wiki/doku.php?id=hccap
+seq:
+  - id: records
+    type: hccap_record
+    repeat: eos
+types:
+  hccap_record:
+    seq:
+      - id: essid
+        type: strz
+        size: 36
+        
+      - id: ap_mac
+        type: mac_address
+        doc: the bssid(MAC) of the access point 
+      
+      - id: station_mac 
+        type: mac_address
+        doc: the MAC address of a client connecting to the access point 
+        
+      - id: station_nonce
+        type: wpa2_nonce
+      
+      - id: ap_nonce
+        type: wpa2_nonce
+        
+      - id: eapol_buffer
+        size: 256
+        
+      - id: eapol_size
+        type: u4
+        doc: size of eapol
+
+      - id: keyver
+        type: u4
+        doc: the flag used to distinguish WPA from WPA2 ciphers. Value of 1 means WPA, other - WPA2
+      
+      - id: keymic
+        size: 16
+        doc: the final hash value. MD5 for WPA and SHA-1 for WPA2 (truncated to 128 bit)
+    instances:
+      eapol:
+        io: eapol_buffer._io
+        pos: 0
+        size: eapol_size
--- a/network/802.11/hccapx.ksy
+++ b/network/802.11/hccapx.ksy
@@ -1,0 +1,88 @@
+meta:
+  id: hccapx
+  title: Hashcat capture format
+  license: Unlicense
+  application:
+    - Hashcat
+    - aircrack-ng
+  endian: le
+  encoding: utf-8
+  file-extension: hccapx
+  imports:
+    - /network/mac_address
+    - /network/802.11/wpa2_nonce
+doc: |
+  Native format of Hashcat password "recovery" utility
+doc-ref: https://hashcat.net/wiki/doku.php?id=hccapx
+seq:
+  - id: records
+    type: hccapx_record
+    repeat: eos
+types:
+  hccapx_record:
+    seq:
+      - id: signature
+        contents: "HCPX"
+      - id: version
+        type: u4
+        doc: the version number of the .hccapx file format
+
+      - id: ignore_replay_counter
+        type: b1
+        doc: |
+          Indicates if the message pair matching was done based on replay counter or not.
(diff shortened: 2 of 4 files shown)


KOLANICH:
@GreyCat

network/802.11/eapol_key.ksy
```diff
@@ -0,0 +1,8 @@
+meta:
+  id: eapol_key
+  title: IEEE 802.11i EAPOL
+  license: Unlicense
+doc-ref: IEEE 802.11
+seq:
+  - id: key
+    size: 256
```
GreyCat: Generally, I'm not too keen on adding such "placeholder" types for types. For example, in many applications, keys and nonces are literally just arrays of (random) bytes of a certain length. They'll never be more than that, and I heavily doubt that there would be any better replacements for them in any languages (expect for may be trivial containers). Right now all they do is add *lots* of unnecessary complication for both ksy writers and end-users.

Besides, in many cases, such as "nonce", "mac_address", or even "eapol_key" is a way too ambiguous name. For example, there are several items that can be referred to as "EAPOL keys":

* EAPOL-Key Confirmation Key (KCK) - 16 bytes
* EAPOL-Key Encryption Key (KEK) - 16 bytes
* whole EAPOL-Key frame

And, yeah, I still have to remind you of EOF newlines.

network/802.11/hccap.ksy
```diff
@@ -0,0 +1,57 @@
...
+    - /network/mac_address
+    - /network/802.11/eapol_key
+    - /network/802.11/wpa2_nonce
+doc: |
+  Native format of Hashcat password "recovery" utility
+  A sample of file for testing can be downloaded from https://web.archive.org/web/20150220013635if_/http://hashcat.net:80/misc/example_hashes/hashcat.hccap
+doc-ref: https://hashcat.net/wiki/doku.php?id=hccap
+seq:
+  - id: records
+    type: hccap
```
GreyCat: This is literally a very bad idea to be that ambiguous. Why not use `hccap_record` or something for the individual record type, why make a clash here?

network/802.11/hccap.ksy
```diff
@@ -0,0 +1,57 @@
...
+    repeat: eos
+types:
+  hccap:
+    seq:
+      - id: essid
+        type: strz
+        size: 36
```
GreyCat: I *heavily* doubt that ESSID is an UTF-8 string. AFAIR, lots of devices use a very arbitrary choice of encoding, so I guess this one should be treated as byte array.

network/802.11/hccap.ksy
```diff
@@ -0,0 +1,57 @@
...
+        
+      - id: ap_mac
+        type: mac_address
+        doc: the bssid(MAC) of the access point 
+      
+      - id: stantion_mac
```
GreyCat: "Stantion"?

network/802.11/hccap.ksy
```diff
@@ -0,0 +1,57 @@
...
+      
+      - id: ap_nonce
+        type: wpa2_nonce
+        
+      - id: eapol
+        type: eapol_key
+        
+      - id: eapol_size
+        type: u4
+        doc: size of eapol
```
GreyCat: If it's size of something called "eapol", it should be `len_eapol`.

GreyCat:
My apologies for late review and thank you for your contribution. I've added some in-line comments. To summarize, my proposals:

* Replace (at least for a time being) custom types with simpler `size: XXX` notation.
* Fix EOF problems
* Fix type name clashing (it's legal, but it's definitely very confusing to end user to use Hccap vs Hccap.Hccap types).

network/802.11/eapol_key.ksy
KOLANICH: I don't want to remember eapol key length everywhere where it can encounter, so I moved it into a separate type. It's easier to maintain. Also it gives self-documentation. Looking at the type name we understand what is it.

About eol in the eof. Why do we need these eol's in the end of files? They are redundant, since the file is ended. They may pay somewhere, for example if you copy the file content into a memory buffer anf replace the ```\n```s with nullbytes, but a good program cannot rely on that. So I vote for removing empty lines from the end of all the ksy files in the repo. See the proposed .editorconfig PR.

network/802.11/hccap.ksy
KOLANICH: OK.

network/802.11/hccap.ksy
KOLANICH: https://en.wikipedia.org/wiki/Station_(networking)

GreyCat: ... And? staNtion is some kind of pun on "station" for "Networking"?

KOLANICH: OH SHI...

network/802.11/eapol_key.ksy
GreyCat: As I've noted above, there's basically no such thing as "eapol key length". Even in the context of "eapol" buffer in hccap, it's variable length field that takes up to 256 bytes, padded to 256 bytes to make up a fixed-size C struct — as described [in their wiki](https://hashcat.net/wiki/doku.php?id=hccapx#detailed_structure). So, this "eapol_key.ksy" is actually wrong even in respect to hccap, and I highly doubt that anyone else would use exactly same 256-byte buffer.

> It's easier to maintain.

I really doubt that. Writing `size: XXX` vs adding separate dozen-line file.

> Also it gives self-documentation. Looking at the type name we understand what is it.

In majority of the cases, you'll have an attribute named in a way that it gives out its meaning. And, on the other hand, it's very easy to fall into trap of extracting a certain type as some sort of "type of global interest", while it turns out that it's a minor internal thing. You've already felt into it with "eapol_key.ksy". "mac_address.ksy" is probably also a bad idea, but due to another reason: it probably should be platform-specific external type, and having `imports` of some KS type makes that more complicated.

> So I vote for removing empty lines from the end of all the ksy files in the repo.

This is definitely not a place to discuss it, and it's definitely not a good idea to have some files in one way and some in the other. Please state your arguments in https://github.com/kaitai-io/kaitai_struct/issues/140, but please do some research and come with the hard facts, not just opinions.

network/802.11/hccap.ksy
KOLANICH: It is not used for parsing the file format.

GreyCat: I think it should be — as far as I understand, it is the size of meaningful portion of 256-byte array. Everything else there would be just random garbage, that you don't really want to use as a "key".

GreyCat:
Better late than never, I've imported and reworked these formats. Thanks!

PR #66, opened 2017-10-16
executable/elf: add missing PAX_FLAGS segment type

rossburton:
This is used by the PaX Linux security patch.  Verification can be seen at
https://pax.grsecurity.net/binutils-2.16-pt-pax-flags-200506102235.patch:

+#define PT_PAX_FLAGS	(PT_LOOS + 0x5041580) /* PaX flags */

### DIFF
--- a/executable/elf.ksy
+++ b/executable/elf.ksy
@@ -318,6 +318,7 @@ enums:
     6: phdr
     7: tls
 #    0x60000000: loos
+    0x65041580: pax_flags
     0x6fffffff: hios
 #    0x70000000: loproc
 #    0x7fffffff: hiproc


rossburton:
Adding the segment types is good but is there a way in kaitai to have an enum that doesn't cause parse failures if an unexpected value appears? I'm trying to use the ELF parser in a project but I may have to swap the segment type enum for a bare int because it throws an error for every unknown segment.

GreyCat:
Thanks for the addition!

Enum behavior is language-dependent. For example, in Java it will just turn out to be `null`. I recall that there were some discussion on the channel about making it language-independent, but there was no consensus, and nobody raised an issue about it. Feel free to propose that in https://github.com/kaitai-io/kaitai_struct/issues/new

PR #67, opened 2017-10-16
Create openpgp.ksy

cugu:


security/openpgp.ksy
```diff
@@ -0,0 +1,508 @@
...
+            #true: new_packet
+            false: old_packet
+            
+  old_packet:
+    seq:
+      - id: length
+        type: 
+          switch-on: _parent.length_type
+          cases:
+            0: u1
```
KOLANICH: Consider creating a enum.

security/openpgp.ksy
```diff
@@ -0,0 +1,508 @@
...
+            subpacket_types::policy_uri: policy_uri
+            subpacket_types::key_flags: key_flags
+            subpacket_types::signers_user_id: signers_user_id
+            subpacket_types::reason_for_revocation: reason_for_revocation
+            subpacket_types::features: features
+            subpacket_types::signature_target: signature_target
+            subpacket_types::embedded_signature: embedded_signature
+      
+  subpacket_length:
+    -webide-representation: '{length}'
```
KOLANICH: Is it a variable length integer? Please check ```common``` directory.

security/openpgp.ksy
```diff
@@ -0,0 +1,508 @@
...
+        type: u2
+      - id: rsa_e
+        size: 3
+      
+  user_id_packet:
+    seq:
+      - id: user_id
+        size-eos: true
+        type: str
+        encoding: utf-8
```
KOLANICH: consider koving to ```meta```

security/openpgp.ksy
cugu: Nope, it's different.

security/openpgp.ksy
cugu: What would be the advantage in this case?

security/openpgp.ksy
cugu: Good idea, utf-8 is default for the complete packet.

security/openpgp.ksy
KOLANICH: No real advantage, just a more fancier look.

security/openpgp.ksy
```diff
@@ -0,0 +1,503 @@
...
+        type: u1
+        # enum: TODO 5.2.1
+      - id: public_key_algorithm
+        type: u1
+        enum: public_key_algorithms
+      - id: hash_algorithm
+        type: u1
+        enum: hash_algorithms
+      - id: hashed_subpacket_length
+        type: u2
```
GreyCat: If we're to follow the [style guide](http://doc.kaitai.io/ksy_style_guide.html), then it should be probably called something like `len_hashed_subpacket`. Are you ok with that?

security/openpgp.ksy
```diff
@@ -0,0 +1,503 @@
...
+      - id: public_key
+        type: public_key_packet
+      - id: string_to_key
+        type: u1
+      - id: symmetric_encryption_algorithm
+        type: u1
+        enum: symmetric_key_algorithm
+        if: string_to_key >= 254
+      - id: more
+        size-eos: true
```
GreyCat: I believe this is just padding, isn't it? If that's it, then it's not really needed, I guess.

cugu: Actually `more` is the secret key. I fixed it.

security/openpgp.ksy
cugu: I made a couple of changes according to the style guide.

GreyCat:
@cugu I've tried to use your format spec with several different files generated by gnupg using `--encrypt` and `--sign` commands with default, `--openpgp`, `--rfc4880`, `--rfc4880bis`, `--rfc2440`, `--pgp6`, etc. with no luck with any one of them. How can I test it and what files is it supposed to parse?

GreyCat:
Thanks for the explanations, merging in!

security/openpgp.ksy
```diff
@@ -0,0 +1,504 @@
+meta:
+  id: pgp
+  title: OpenPGP Message Format
```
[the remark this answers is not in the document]
GreyCat: I'll fix this one.

security/openpgp.ksy
```diff
@@ -0,0 +1,504 @@
+meta:
+  id: pgp
```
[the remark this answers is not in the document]
cugu: `openpgp_message` is fine. 
I'm totally okay with any other group. `security` could be for keys and certificates.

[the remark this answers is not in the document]
GreyCat: I'm slightly hesitating about "security", as so far it looks like this is the only security theme-specific format that would fit this category. Virtually all of other security-related formats I know (like certificates, public/private keys, etc) use generic formats widely used for other purposes as well, i.e. ASN.1 BER/CER/DER, PEM, simple base64 encoding, text-based formats, XML, etc.

Can you provide several more examples of formats that would fit this category?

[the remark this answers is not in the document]
cugu: While for example ssh keys and ssl certificates might be encoded in base64 the underling data is a binary format AFAIK. Anyway I don't mind any other category ;-)

PR #68, opened 2017-10-16
added xm

KOLANICH:
It is not very ready (I'm stuck with problems with packed data, https://github.com/KOLANICH/kaitai_struct_formats/commit/2e90f42ac2ed48d662b397798a2d9b327d53886f, it surprisingly doesn't work even though it is consistent with the spec, it may be byte alignment issue (even though all the ```packed_note``` structs take integer count of bytes), I tried to remove ```align_to_``` call from note and the parsing finished without an error (I have not verified the result)), but instrument part is quite ready.

### DIFF
--- a/media/tracker_modules/xm.ksy
+++ b/media/tracker_modules/xm.ksy
@@ -1,0 +1,303 @@
+meta:
+  id: xm
+  title: Extended Module
+  application:
+    - FastTracker 2
+    - Protracker
+    - MilkyTracker
+    - libmodplug
+    - Mikmod
+  file-extension: xm
+  license: Unlicense
+  endian: le
+  encoding: utf-8
+doc-ref: |
+  http://sid.ethz.ch/debian/milkytracker/milkytracker-0.90.85%2Bdfsg/resources/reference/xm-form.txt
+  ftp://ftp.modland.com/pub/documents/format_documentation/FastTracker%202%20v2.04%20(.xm).html
+seq:
+  - id: preheader
+    type: preheader
+  - id: header
+    size: preheader.header_size - 4
+    type: header
+  - id: patterns
+    type: pattern
+    repeat: expr
+    repeat-expr: header.number_of_patterns
+  - id: instruments
+    type: instrument
+    repeat: expr
+    repeat-expr: header.number_of_instruments
+types:
+  preheader:
+    seq:
+      - id: signature0
+        contents: 'Extended Module: '
+      - id: module_name
+        size: 20
+        type: strz
+        doc: Module name, padded with zeroes
+      - id: signature1
+        contents: [0x1a]
+      - id: tracker_name
+        size: 20
+        type: strz
+        doc: Tracker name
+      - id: version_number
+        type: version
+        doc: "Format versions below [0x01, 0x04] have a LOT of differences. Check this field!"
+      - id: header_size
+        type: u4
+        doc: Header size << Calculated FROM THIS OFFSET, not from the beginning of the file! >>
+    types:
+      version:
+        seq:
+          - id: minor
+            type: u1
+            doc: currently 0x04
+          - id: major
+            type: u1
+            doc: currently 0x01
+        instances:
+          value:
+            value: (major<<8) | minor
+  header:
+    seq:
+      - id: song_length
+        type: u2
+        doc: Song length (in pattern order table)
+      - id: restart_position
+        type: u2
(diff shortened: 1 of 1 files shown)


media/tracker_modules/xm.ksy
```diff
@@ -0,0 +1,280 @@
+meta:
+  id: xm
+  title: Extended Module
+  application:
+    - FastTracker 2
+    - Protracker
```
GreyCat: Please list only applications that originated the format, not *all* applications that support it. It's like you list only "Adobe Photoshop" for .psd, not all possible viewers / raster image editors with kind of support .psd.

media/tracker_modules/xm.ksy
```diff
@@ -0,0 +1,280 @@
...
+  encoding: utf-8
+doc-ref: |
+  http://sid.ethz.ch/debian/milkytracker/milkytracker-0.90.85%2Bdfsg/resources/reference/xm-form.txt
+  ftp://ftp.modland.com/pub/documents/format_documentation/FastTracker%202%20v2.04%20(.xm).html
+  
+seq:
+  - id: preheader
+    type: preheader
+  - id: header
+    size: preheader.header_size - 4
```
GreyCat: You don't like prefixes proposed in the style guide that much (i.e. `len_`, `num_`, `ofs_`, etc)?

media/tracker_modules/xm.ksy
```diff
@@ -0,0 +1,280 @@
...
+    - libmodplug
+    - Mikmod
+    - Extended Module Player
+    - VLC
+    - Media Player Classic
+    - MPlayer
+  file-extension: xm
+  license: Unlicense
+  endian: le
+  encoding: utf-8
```
GreyCat: Again, given that xm originated from FastTracker in early 1990s, I highly doubt that any utf-8 support existed back there at all.

KOLANICH: I guess it is quite safe to put utf-8 everywhere where text-ASCII is assumed (without special characters) because it is compatible with ASCII.

media/tracker_modules/xm.ksy
KOLANICH: Yes, I really don't like them:
1 it is a bit weird to put low-level details first and high-level last
2 combining verbose words with non-verbose ones also looks weird

media/tracker_modules/xm.ksy
GreyCat: As far as I see, there are tons of xm modules which make liberal use of non-ASCII names everywhere, usually assuming something like cp437 or some similar IBM codepage.

media/tracker_modules/xm.ksy
GreyCat: Well, could you raise your concerns in https://github.com/kaitai-io/kaitai_struct/issues/140. Current "silent protest" won't get us anywhere — I still think that any standard at all is better than no standard, and until we'll discuss your objections, I would just take time to enforce it manually, i.e. by fixing your contributions. This will, in turn, will probably just annoy you, because:

* It takes longer time to get your ksys into this repository
* After that, "that guy" comes and mangles it in a way you don't like it

I see that you do a lot of interesting stuff here, and it would be bad if all that would go to waste just because we couldn't agree on some basic stuff.

I actually have an idea. Can I contact you using your mail.ru address?

KOLANICH: >I actually have an idea. Can I contact you using your mail.ru address?
Of course you can, but why not to discuss it on GH. Have I made your furious?

>I would just take time to enforce it manually, i.e. by fixing your contributions.

Completely OK.

Also
https://github.com/kaitai-io/kaitai_struct_doc/pull/7

media/tracker_modules/xm.ksy
KOLANICH: I can make it an array then. But it is not very good idea - ksy descriptions are also a documentation, module name is meant to be a string, but we don't have strings of unknown encoding for now.
How about ```encpding: ?``` triggering heuristical encoding detection?

media/tracker_modules/xm.ksy
GreyCat: > If you contact by email, please leave a comment on GH about that.

Confirming, I've sent you the e-mail.

GreyCat:
@KOLANICH Thanks! Finally, I've got to merge this one.

PR #73, opened 2018-01-10
Read GIFs till EOF block

ujjwal96:
This PR makes the gif parser read GIFs only till EOF block. This prevents parsing errors in case a file has bytes after the EOF block.
[example gif](https://user-images.githubusercontent.com/16747982/34779270-b86ca25e-f645-11e7-80e0-1624e90ea0fb.gif)

### DIFF
--- a/image/gif.ksy
+++ b/image/gif.ksy
@@ -34,7 +34,8 @@ seq:
     doc-ref: https://www.w3.org/Graphics/GIF/spec-gif89a.txt - section 18
   - id: blocks
     type: block
-    repeat: eos
+    repeat: until
+    repeat-until: _io.eof or _.block_type == block_type::end_of_file
 types:
   header:
     doc-ref: https://www.w3.org/Graphics/GIF/spec-gif89a.txt - section 17


image/gif.ksy
```diff
@@ -34,7 +34,8 @@ seq:
     doc-ref: https://www.w3.org/Graphics/GIF/spec-gif89a.txt - section 18
   - id: blocks
     type: block
-    repeat: eos
+    repeat: until
+    repeat-until: _.block_type.to_i == 0x3b
```
GreyCat: Um, why don't we just use

    repeat-until: _.block.type == block_types::end_of_file

or, better yet, something like:

    repeat-until: _io.eof or _.block.type == block_types::end_of_file

... to make sure we don't fail on truncated files which lack EOF block?

ujjwal96: Due to my inexperience with ksy :sweat_smile: 
I'll push the new changes :+1:

GreyCat:
Cool, thanks!

PR #76, opened 2018-01-21
Added MagicaVoxel .vox file specification

Arlorean:
Tested with a voxel model with MATT chunks from https://blog.sketchfab.com/complete-guide-importing-magicavoxel-sketchfab/.
Actual test file can be downloaded here: https://www.cyn-prod.com/files/lab_vox.zip (CC-BY license https://sketchfab.com/elbriga).
Other test files can be found in the MagicaVoxel download for Windows or MacOS (https://ephtracy.github.io/).
Generated C# code for testing the runtime and it all seems to work fine. 
I plan to replace my own .vox read (and eventually writing) code with this in https://github.com/Arlorean/Voxels.

### DIFF
--- a/media/magicavoxel_vox.ksy
+++ b/media/magicavoxel_vox.ksy
@@ -1,0 +1,187 @@
+meta:
+  id: vox
+  file-extension: vox
+  application: MagicaVoxel
+  endian: le
+  license: MIT
+  title: MagicaVoxel File
+doc-ref: 'https://ephtracy.github.io/ MagicaVoxel Homepage'
+doc-ref: 'https://github.com/ephtracy/voxel-model/blob/master/MagicaVoxel-file-format-vox.txt Format Description'
+seq:
+  - id: magic
+    contents: 'VOX '
+  - id: version
+    type: u4
+    doc: 150 expected
+  - id: main
+    type: chunk
+types:
+  chunk:
+    seq:
+      - id: chunk_id
+        type: u4be
+        enum: chunk_type
+      - id: num_bytes_of_chunk_content
+        type: u4
+      - id: num_bytes_of_children_chunks
+        type: u4
+      - id: chunk_content
+        if: num_bytes_of_chunk_content != 0
+        size: num_bytes_of_chunk_content
+        type:
+          switch-on: chunk_id
+          cases:
+            'chunk_type::pack': pack
+            'chunk_type::size': size
+            'chunk_type::xyzi': xyzi
+            'chunk_type::rgba': rgba
+            'chunk_type::matt': matt
+      - id: children_chunks
+        if: num_bytes_of_children_chunks != 0
+        type: chunk
+        repeat: eos
+  pack:
+    seq:
+      - id: num_models
+        type: u4
+  size:
+    seq:
+      - id: size_x
+        type: u4
+      - id: size_y
+        type: u4
+      - id: size_z
+        type: u4
+  xyzi:
+    seq:
+      - id: num_voxels
+        type: u4
+      - id: voxels
+        type: voxel
+        repeat: expr
+        repeat-expr: num_voxels
+  rgba:
+    seq:
+      - id: colors
+        type: color
+        repeat: expr
+        repeat-expr: 256
+  voxel:
+    seq: 
+      - id: x
+        type: u1
+      - id: y
+        type: u1
+      - id: z
+        type: u1
+      - id: color_index
+        type: u1
+  color:
+    seq:
+      - id: r
+        type: u1
+      - id: g
+        type: u1
+      - id: b
+        type: u1
+      - id: a
+        type: u1
+  matt:
+    seq:
+    - id: id
+      type: u4
+    - id: material_type
+      type: u4
+      enum: material_type
+    - id: material_weight
+      type: f4
+    - id: property_bits
+      type: u4
+      # .to_i not implemented for C# runtime yet
+      #enum: property_bits_type
+    - id: plastic
+      if: has_plastic
+      type: f4
+    - id: roughness
+      if: has_roughness
+      type: f4
+    - id: specular
+      if: has_specular
+      type: f4
+    - id: ior
+      if: has_ior
+      type: f4
+    - id: attenuation
+      if: has_attenuation
+      type: f4
+    - id: power
+      if: has_power
+      type: f4
+    - id: glow
+      if: has_glow
+      type: f4
+    - id: is_total_power
+      if: has_is_total_power
+      type: f4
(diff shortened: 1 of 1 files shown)


arekbulski:
Is this somehow related to Magica the game? 
http://store.steampowered.com/app/42910/Magicka/

media/magicavoxel_vox.ksy
```diff
@@ -0,0 +1,187 @@
+meta:
+  id: vox
+  file-extension: vox
+  application: MagicaVoxel
+  endian: le
+  license: MIT
+  title: MagicaVoxel File
+doc-ref: 'https://ephtracy.github.io/ MagicaVoxel Homepage'
+doc-ref: 'https://github.com/ephtracy/voxel-model/blob/master/MagicaVoxel-file-format-vox.txt Format Description'
```
GreyCat: Unfortunately, it won't work like that. YAML in general (except when using a very special parser) disallows to have several instances of the same key, thus we'll get only one of these (the latter, if I recall it correctly). There has been a request to support multiple `doc-ref` entires somehow (https://github.com/kaitai-io/kaitai_struct/issues/269), but it's still not implemented.

Not a big deal, of course.

media/magicavoxel_vox.ksy
```diff
@@ -0,0 +1,187 @@
...
+  - id: main
+    type: chunk
+types:
+  chunk:
+    seq:
+      - id: chunk_id
+        type: u4be
+        enum: chunk_type
+      - id: num_bytes_of_chunk_content
+        type: u4
```
GreyCat: Customarily, we advice to use `len_` prefix for attributes that designate number of bytes (i.e. length or size) of some value — see [style guide](http://doc.kaitai.io/ksy_style_guide.html#attr-id).

media/magicavoxel_vox.ksy
```diff
@@ -0,0 +1,187 @@
...
+      type: u4
+    - id: material_type
+      type: u4
+      enum: material_type
+    - id: material_weight
+      type: f4
+    - id: property_bits
+      type: u4
+      # .to_i not implemented for C# runtime yet
+      #enum: property_bits_type
```
GreyCat: It's better to use separate `type` that would look like that for bitmask:

```yaml
types:
  property_bits_type:
    seq:
      - id: reserved1
        type: b1
      - id: has_is_total_power
        type: b1
      - id: has_glow
        type: b1
      - id: has_power
        type: b1
      # ...
```

instead of enums or manual extraction of flags using the value instances.

GreyCat:
Cool, thanks for your contribution! I've commented on a few things, nothing major, yet you might want to consider them. Merging in :)

Arlorean:
It's not related to Magicka the game. It's a file format for storing Voxel
based artwork. MagicaVoxel <https://ephtracy.github.io/> is a free tool
that is used by many game authors to create their Voxel styled games. Crossy
Road <http://www.crossyroad.com/> may be familiar to you for example,
although I think they may have used another tool called Qubicle
<http://minddesk.com/index.php> to do that. I wondered whether to put it in
the CAD folder to be honest. Maybe there should be a new folder for game
asset files or 3D File Formats? STL could be moved into that folder if it
was created.

On Sun, Jan 21, 2018 at 11:23 AM, Arkadiusz Bulski <notifications@github.com

Arlorean:
Thanks. I'll try to make it a single key but multi-line entry.

Arlorean:
I was trying to stay close to the original file format specification with
the naming but I can see the advantage of having a consistent style to the
ksy files. I'll revisit the naming.

https://github.com/ephtracy/voxel-model/blob/master/MagicaVoxel-file-format-vox.txt
2. Chunk Structure
-------------------------------------------------------------------------------
# Bytes | Type | Value
-------------------------------------------------------------------------------
1x4 | char | chunk id
4 | int | num bytes of chunk content (N)
4 | int | num bytes of children chunks (M)
N | | chunk content
M | | children chunks
-------------------------------------------------------------------------------

Arlorean:
I did start that way but since the flag set it 32 bits and the order is
from largest to smallest I thought I'd need to have dummy bits and also
reverse the original sequence from the documentation so I went back to an
enum. Is there a way to start at bit 0 instead of bit 7 (or bit 31)?
I guess I could have an empty placeholder for the first 3 bytes and then
only use the bitset for the 4th byte.
Having an enum with flags would be nice in the C# code in some cases as
well so you can chose to use the bool accessors or access the whole flag
set as a single value.

Arlorean:
Thanks. I really enjoyed using the online web IDE to develop the parser.
I'm going to try some more 3d file formats to see if I can get them
expressed as .ksy files. Thanks for creating and maintaining this project,
it's great!

GreyCat:
@Arlorean 
> Maybe there should be a new folder for game asset files or 3D File Formats? STL could be moved into that folder if it was created.

Actually, there are already quite a few 3D-related formats. AFAIR, @mnakamura1337 was working on some sort of universal viewer that could render various 3D models specified using ksy using the same code (akin to [Kaitai FS](https://github.com/kaitai-io/kaitai_fs), which can allow accessing various filesystems specified using ksy).

> I was trying to stay close to the original file format specification with
the naming but I can see the advantage of having a consistent style to the
ksy files. I'll revisit the naming.

Generally, you can still provide original IDs using `-orig-id` key. There are zero checks on that field — you can keep space, etc — and it is currently just ignored by the compiler, i.e.:

```yaml
  - id: len_chunk
    -orig-id: num bytes of chunk content
    type: u4
```

> and also reverse the original sequence from the documentation

The overall idea is that we are not trying to replicate existing documentation clause-by-clause, but instead we're trying to specify whatever we see in the file format. Specifying bits from highest to lowest or vice versa is generally a matter of choice, so we've opted to stick to "what would be natural to most humans", i.e. big-endian order, AKA most significant digit (bit) first.

> so I went back to an enum.

KS needs to be multi-paradigmal, so we can't really work with flags representation like ones in C/C++ or C# `[Flags]` + `Enum` :(

> Is there a way to start at bit 0 instead of bit 7 (or bit 31)?

Not yet. There is a long-ongoing talk about support little-endian bit reading (see https://github.com/kaitai-io/kaitai_struct/issues/155), and there is proof-of-concept code for what most people would find acceptable, but it's not yet supported in compiler + most runtimes.

> Having an enum with flags would be nice in the C# code in some cases as well so you can chose to use the bool accessors or access the whole flag set as a single value.

We can actually do some sort of special support for C# for these ones (i.e. when we have a type which is clearly a fixed bit mask). Could you file a separate ticket about that?

PR #78, opened 2018-01-27
APM (Apple Partition Map) partition table

SnowFlakey:
Was tested on multi structure CD ( ISO9660 + APM ) and APM created with parted.
I have no access to an old Apple HDD at this point.

### DIFF
--- a/filesystem/apm_partition_table.ksy
+++ b/filesystem/apm_partition_table.ksy
@@ -1,0 +1,69 @@
+meta:
+  id: apm_partition_table
+  title: APM (Apple Partition Map) partition table
+  license: CC0-1.0
+  endian: be
+  encoding: ascii
+doc-ref: Specification taken from https://en.wikipedia.org/wiki/Apple_Partition_Map
+instances:
+  sector_size:
+    value: 0x200
+    # APM only supports 0x200 (512) byte sectors.
+  partition_lookup:
+    io: _root._io
+    pos: _root.sector_size
+    type: partition_entry
+    size: sector_size
+    # Every partition entry contains the number of partition entries.
+    # We parse the first entry, to know how many to parse, including the first one.
+    # No logic is given what to do if other entries have a different number.
+  partition_entries:
+    io: _root._io
+    pos: _root.sector_size
+    type: partition_entry
+    size: sector_size
+    repeat: expr
+    repeat-expr: _root.partition_lookup.number_of_partitions
+types:
+  partition_entry:
+    seq:
+      - id: magic
+        contents: [ 0x50, 0x4d ]
+      - id: reserved_1
+        size: 0x2
+      - id: number_of_partitions
+        type: u4
+      - id: partition_start
+        type: u4
+      - id: partition_size
+        type: u4
+      - id: partition_name
+        type: strz
+        size: 0x20
+      - id: partition_type
+        type: strz
+        size: 0x20  
+      - id: data_start
+        type: u4
+      - id: data_size
+        type: u4
+      - id: partition_status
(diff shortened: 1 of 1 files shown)


filesystem/apm_partition_table.ksy
```diff
@@ -0,0 +1,69 @@
+meta:
+  id: apm_partition_table
+  title: APM (Apple Partition Map) partition table
+  license: CC0-1.0
+  endian: be
+  encoding: ascii
```
GreyCat: I'm not 100% sure that it would be ASCII and not just some raw 8-bit system-dependent encoding. Any hints?

filesystem/apm_partition_table.ksy
```diff
@@ -0,0 +1,69 @@
...
+    value: 0x200
+    # APM only supports 0x200 (512) byte sectors.
+  partition_lookup:
+    io: _root._io
+    pos: _root.sector_size
+    type: partition_entry
+    size: sector_size
+    # Every partition entry contains the number of partition entries.
+    # We parse the first entry, to know how many to parse, including the first one.
+    # No logic is given what to do if other entries have a different number.
```
GreyCat: Technically, you can just do something like to look that up without parsing everything:

```yaml
instances:
  num_partitions:
    pos: sector_size + 4
    type: u4
```

I'm not sure it's any better, though...

filesystem/apm_partition_table.ksy
```diff
@@ -0,0 +1,69 @@
...
+      - id: magic
+        contents: [ 0x50, 0x4d ]
+      - id: reserved_1
+        size: 0x2
+      - id: number_of_partitions
+        type: u4
+      - id: partition_start
+        type: u4
+      - id: partition_size
+        type: u4
```
GreyCat: It would be super cool to add some docs, stating which units are used in these values.

GreyCat:
Thanks for your contribution! I've added some minor comments. Given that there's already 3 partition tables, what would you think about moving them into some separate directory? Like `filesystem/partition_table` or just `partition_table`?

filesystem/apm_partition_table.ksy
SnowFlakey: From Apple's rescue CD:
"Apple", "Apple_partition_map", "Macintosh", "Apple_driver43"

It all looks ASCII to me, I think only IBM is old enough to pre-date ASCII with EBCDIC

filesystem/apm_partition_table.ksy
SnowFlakey: Well, if Apple is using a regular partition entry as the header, then I would say, reuse the correct logic.

filesystem/apm_partition_table.ksy
SnowFlakey: I can do that

filesystem/apm_partition_table.ksy
GreyCat: The problem is that real ASCII is 7-bit. Occasionally, people refer to something as ASCII, but actually it turns out that there's some (usually system config dependent) 8-bit encoding like cp437 or whatever's old Mac equivalent. Trying to parse that as "ASCII" would fail with a runtime error.

SnowFlakey: The document quotes it as "fixed ASCII right-side NULL padded"
I have not found the reference of my reference.

KOLANICH:
`/storage/partition_table`
`/storage/image` for generic images like the ones used in VMs
`/storage/filesystem` for filesystems and images dictating internal filesystem

PR #79, opened 2018-01-27
APM (Apple Partition Map) updated docs

SnowFlakey:


GreyCat:
Thanks!

GreyCat:
I also wondered if that would be a viable idea to add accessors like that:

```yaml
instances:
  partition:
    pos: partition_start * _root.sector_size
    size: partition_size * _root.sector_size
  boot_code:
    pos: boot_code_start * _root.sector_size
    size: boot_code_size
```

etc.

PR #81, opened 2018-02-02
Add partial support for coff symbols

theorbtwo:
This is limited to support for the coff symbols that I happened to come across.
It does support:
 - symbol names (both short and long)
 - looking up section + value to get the actual value
It does not support:
 - types -- all symbols in my file are type 0
 - storage classes -- all symbols in my file are storage class 2 or 3.  (extern or static)
 - aux symbols -- none of mine had any

### DIFF
--- a/executable/microsoft_pe.ksy
+++ b/executable/microsoft_pe.ksy
@@ -53,6 +53,19 @@ types:
         type: u2
       - id: characteristics
         type: u2
+    instances:
+      symbol_table_size:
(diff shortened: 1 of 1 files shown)


GreyCat:
This file lacks the licensing info. I'm going to add CC0-1.0 to it — would it be ok with you to license your contribution with the same license?

theorbtwo:
Good with me.

PR #82, opened 2018-02-03
Iso9660

SnowFlakey:
This is far from done, but it the review should not become too big


filesystem/iso9660.ksy
```diff
@@ -354,25 +354,73 @@ types:
...
+        if: signature == su_signature::susp_indicator
+      - id: susp_st # ST
+        type: susp_unknown
+        if: signature == su_signature::susp_terminator
+      - id: rrip_tf # TF
+        type: susp_unknown
+        if: signature == su_signature::rrip_time_file
+      - id: rrzf_zf # ZF
+        type: susp_unknown
+        if: signature == su_signature::rrzf_zisofs
```
GreyCat: Uhh, why don't you use switch here?

```yaml
- id: body
  type:
    switch-on: signature
    cases:
      'su_signature::rras_amiga_specific': susp_unknown
      'su_signature::susp_continuation_area': susp_unknown
      'su_signature::rrip_child_link': susp_unknown
      # ...
```
?

filesystem/iso9660.ksy
```diff
@@ -1,210 +1,344 @@
...
+#        type: volume_partition
+#        if: descriptor_type == descriptor_type::volume_partition_descriptor
+  boot_record_volume:
+    doc-ref: ecma-119 8.2
+    seq:
+      - id: boot_system_identifier
+        doc-ref: ecma-119 8.2.4
+        type: strz
+        size: 0x20
+        encoding: ASCII
```
KOLANICH: ```encoding: ASCII``` is so widespread here.

As UTF-8 is superset of ASCII, why not to just put ```encoding: UTF-8``` into ```meta```?

filesystem/iso9660.ksy
```diff
@@ -214,60 +348,201 @@ types:
...
+        type: u4bi
+        if: len_dr > 0x0
+      - id: data_len
+        doc-ref: ecma-119 9.1.4
+        type: u4bi
+        if: len_dr > 0x0
+      - id: rec_date_time
+        doc-ref: ecma-119 9.1.5
+        type: recdatetime
+        if: len_dr > 0x0
```
KOLANICH: I gess we need to move the comparison into an instance and use in if that instance.

filesystem/iso9660.ksy
```diff
@@ -1,210 +1,344 @@
...
-      - id: padding
-        type: u1
-        if: len_dir_name % 2 == 1
-  datetime:
+        size: 0x80
+        encoding: ASCII
+      - id: publisher_identifier
+        doc-ref: ecma-119 8.5.14
+        type: str
+        size: 0x80
```
KOLANICH: Consider moving these stings of predefined lengths into own types

filesystem/iso9660.ksy
```diff
@@ -1,210 +1,344 @@
...
-      - id: vol_desc_boot_record
-        type: vol_desc_boot_record
-        if: type == 0
-      - id: vol_desc_primary
-        type: vol_desc_primary
-        if: type == 1
-  vol_desc_boot_record:
+        doc-ref: ecma-119 8.1.3
+        contents: [0x01]
+      - id: boot_record
```
KOLANICH: Switch should be here too.

filesystem/iso9660.ksy
```diff
@@ -1,210 +1,344 @@
 meta:
   id: iso9660
-  title: ISO9660 CD filesystem
+  title: ISO9660
   file-extension: iso
   xref:
     wikidata: Q815645
   license: CC0-1.0
-doc: |
-  ISO9660 is standard filesystem used on read-only optical discs
```
KOLANICH: why have you removed the doc?

filesystem/iso9660.ksy
```diff
@@ -354,6 +354,102 @@ types:
...
+        type: b1
+      - id: creation_short
+        type: datetime_short
+        if: ( creation == true ) and ( long_form == false )
+      - id: creation_long
+        type: datetime_long
+        if: ( creation == true ) and ( long_form == true )
+      - id: modify_short
+        type: datetime_short
+        if: ( modify == true ) and ( long_form == false )
```
KOLANICH: Again this clause should be moved into an instance

filesystem/iso9660.ksy
```diff
@@ -354,6 +354,102 @@ types:
...
+        type: b1
+      - id: access
+        type: b1
+      - id: modify
+        type: b1
+      - id: creation
+        type: b1
+      - id: creation_short
+        type: datetime_short
+        if: ( creation == true ) and ( long_form == false )
```
KOLANICH: I see lots of ```long_form == true``` and ```long_form == false```.
Consider moving long and short forms into a separate type.

filesystem/iso9660.ksy
```diff
@@ -408,48 +456,12 @@ types:
...
-        if: ( effective == true ) and ( long_form == false )
-      - id: effective_long
-        type: datetime_long
-        if: ( effective == true ) and ( long_form == true )
+      - id: datetime_short
+        type: rrip_tf_short
+        if: long_form == false
+      - id: datetime_long
+        type: rrip_tf_long
+        if: long_form == true
```
KOLANICH: I guess ```== true``` is not needed. ```x == false``` is ```not x```.
Use ```switch-on``` and ```cases``` instead of sequential ```if```s.

filesystem/iso9660.ksy
```diff
@@ -5,269 +5,625 @@ meta:
...
-      - id: vol_space_size
+        size: 0x20
+        encoding: ASCII
+      - id: unused02
+        doc-ref: ecma-119 8.4.7
+        contents: [ 0x0, 0x0, 0x0, 0x0,  0x0, 0x0, 0x0, 0x0 ]
+      - id: volume_space_size
+        doc-ref: ecma-119 8.4.8
         type: u4bi
-      - id: unused3
```
KOLANICH: ```reserved``` is IMHO a better name.

KOLANICH:
Please also look at  https://github.com/kaitai-io/kaitai_struct/issues/17

GreyCat:
> Please also look at kaitai-io/kaitai_struct#17

How's that related to this PR?

KOLANICH:
Oh, sory, it's really unrelated. The suggestion is etracted.

filesystem/iso9660.ksy
```diff
@@ -5,269 +5,607 @@ meta:
...
+            'su_signature::rrip_posix_device_number': susp_unknown # PN
+            'su_signature::rrip_posix_file_attributes': rrip_px # PX
+            'su_signature::rrip_relocated_directory': susp_unknown # RE
+            'su_signature::rrip_extensions_in_use_indicator': susp_unknown # RR
+            'su_signature::rrip_sparse_file': susp_unknown # SF
+            'su_signature::rrip_symbolic_link': susp_unknown # SL
+            'su_signature::susp_indicator': susp_unknown # SP
+            'su_signature::susp_terminator': susp_unknown # ST
+            'su_signature::rrip_time_file': rrip_tf # TF
+            'su_signature::rrzf_zisofs': susp_unknown # ZF
```
GreyCat: If `susp_unknown` is supposed to catch all "yet unsupported" types, then why don't we just use `_` (i.e. else/default case)?

```yaml
cases:
  'su_signature::rrip_alternate_name': rrip_nm # NM
  'su_signature::rrip_posix_file_attributes': rrip_px # PX
  'su_signature::rrip_time_file': rrip_tf # TF
  _: susp_unknown
```

KOLANICH: I guess they should be grouped (as a draft for future improvements) and commented out.

filesystem/iso9660.ksy
```diff
@@ -5,269 +5,607 @@ meta:
...
+        type: u4bi
+        if: len_dr > 0x0
+      - id: data_len
+        doc-ref: ecma-119 9.1.4
+        type: u4bi
+        if: len_dr > 0x0
+      - id: rec_date_time
+        doc-ref: ecma-119 9.1.5
+        type: datetime_short
+        if: len_dr > 0x0
```
KOLANICH: Shouldn't it be moved into a subtype?

filesystem/iso9660.ksy
SnowFlakey: That would look nice in code but... it breaks the SU stream on ISO's without SUSP

GreyCat: Understood, thanks. I'll try to think what could be done here...

SnowFlakey: Maybe a defaults for:
in the enum, but not specific overruled
not found in the enum

filesystem/iso9660.ksy
SnowFlakey: I am thinking about this... it was split in the original, but I did like the output from this one more.
This works great, I just don't know the CPU performance of having multiple if statements.

filesystem/iso9660.ksy
KOLANICH: I have read a bit, it really doesn't accept utf-8. But you can move ascii to meta too.

filesystem/iso9660.ksy
```diff
@@ -5,269 +5,607 @@ meta:
...
     seq:
-      - id: type
+      - id: descriptor_type
+        doc-ref: ecma-119 8.1.1
         type: u1
+        enum: descriptor_type
       - id: magic
-        contents: "CD001"
+        doc-ref: ecma-119 8.1.2
+        contents: [0x43, 0x44, 0x30, 0x30, 0x31]
```
KOLANICH: What was the problem with a string signature?

filesystem/iso9660.ksy
SnowFlakey: It is done, we now have directory_record_head and directory_record_body

filesystem/iso9660.ksy
```diff
@@ -495,108 +495,98 @@ types:
...
-  directory_record:
+  directory_record_head:
     doc-ref: ecma-119 9.1
     seq:
       - id: len_dr
         doc-ref: ecma-119 9.1.1
         type: u1
+      - id: body
+        type: directory_record_body
+        if: len_dr > 0x0
```
KOLANICH: ```size: len_dr```?

filesystem/iso9660.ksy
```diff
@@ -495,108 +495,98 @@ types:
...
         size: 0x1
-        if: ( len_dr > 0x0 ) and ( len_fi % 2 == 0 ) # only if even number
+        if: len_fi % 2 == 0 # only if even number
       - id: system_use
         doc-ref: ecma-119 9.1.13
-        size: ( len_dr - 33 ) - len_fi # recheck this logic
+        size: ( _parent.len_dr - 33 ) - len_fi # recheck this logic
         type: su_headers
-        if: ( len_dr > 0x0 ) and ( ( len_dr - 33 ) - len_fi > 1 ) # only if at least 2 bytes are available for magic
+        if: ( _parent.len_dr - 33 ) - len_fi > 1 # only if at least 2 bytes are available for magic
```
KOLANICH: ```size-eos: true``` ?

filesystem/iso9660.ksy
SnowFlakey: Well, it was 3 fields and the first was 0x00, 0x01, 0x02, 0x03 or 0xff
This I found more in line of that
Also... 100% rewrite, I did not look that much at the original in the beginning

filesystem/iso9660.ksy
KOLANICH: use ```-orig-id``` for the original name (and it doesn't need to be a valid KS id) and use the most convenient in ```id```

filesystem/iso9660.ksy
SnowFlakey: size-eos is true with 1 byte, but I want I must be sure there are at least 2 bytes.
I have seen this fail on old CD's

filesystem/iso9660.ksy
SnowFlakey: I can do this, but once I add Juliet then UTF-8 like will be added all over the place.

filesystem/iso9660.ksy
KOLANICH: @GrayCat, can we put restrictions on ```_io``` stream size?

filesystem/iso9660.ksy
KOLANICH: according to wikipedia, joilet uses UCS-2

SnowFlakey: ASCII is now global

filesystem/iso9660.ksy
SnowFlakey: I am sorry, but I am using 4, maybe 6 standards soon, adding my own names just makes a bigger mess. I am willing to change a lot, this one I will not.

filesystem/iso9660.ksy
GreyCat: I'd probably go with `if: _io.size - _io.pos >= 2` here. It's ugly too, but at least it's not manual sizeof counting with lots of magical constants like 33.

filesystem/iso9660.ksy
KOLANICH: I guess the string was better.

filesystem/iso9660.ksy
KOLANICH: Completely OK, it's your work, do as you think is better, just keep in mind that you are allowed and encouraged rename things.

filesystem/iso9660.ksy
```diff
@@ -5,269 +5,589 @@ meta:
...
+      - id: occurrence_of_type_m_path_table
+        doc-ref: ecma-119 8.5.10
+        type: u4le
+      - id: optional_occurrence_of_type_m_path_table
+        doc-ref: ecma-119 8.5.11
+        type: u4le
+      - id: directory_record_for_root_directory
+        doc-ref: ecma-119 8.5.12
+        type: directory_record_head
+        size: 0x22
```
KOLANICH: Yet another magic constant. Can we eliminate it?

filesystem/iso9660.ksy
```diff
@@ -5,269 +5,589 @@ meta:
...
+  rrip_nm:
+    doc-ref: rrip 4.1.4
     seq:
-      - id: len_dir_name
+      - id: length
         type: u1
-      - id: len_ext_attr_rec
+      - id: version
+        contents: [ 0x1 ]
+      - id: flags # ToDo, made it too slow for now
```
KOLANICH: ???

And should flags be an own type of ```b1```s?

filesystem/iso9660.ksy
```diff
@@ -5,269 +5,589 @@ meta:
...
+        doc-ref: ecma-119 9.1.10
+        type: u1
+      - id: file_id_file
+        doc-ref: ecma-119 9.1.11
         type: str
-        size: 2
-        encoding: ASCII
-      - id: sec_hundreds
+        size: len_fi
+        if: not file_flags_directory
```
KOLANICH: ```switch-on```?

filesystem/iso9660.ksy
KOLANICH: Looks like we have a reusable piece which can be moved into a substruct, don't we?

filesystem/iso9660.ksy
SnowFlakey: magic constant 33... I like that, it is actually how many bytes I am into the stream.
The number of bytes that I have to process is a formula, it is not guaranteed till eos.
There is absolutely no logic what will follow.
Example Solaris 2.6 will have "SunOS" as magic with binary items after that.

Most common:
SUSP, SUSP will have 2 byte magic following length + version.

filesystem/iso9660.ksy
SnowFlakey: It already is, a directory_record_head, unless I don't understand you correctly

filesystem/iso9660.ksy
SnowFlakey: I "fixed" this, well I split the logic and added doc to explain it.

filesystem/iso9660.ksy
SnowFlakey: Done

filesystem/iso9660.ksy
SnowFlakey: Done

filesystem/iso9660.ksy
SnowFlakey: Done

filesystem/iso9660.ksy
SnowFlakey: Done

filesystem/iso9660.ksy
SnowFlakey: Maybe I don't understand.
The switch-on would leave me a single variable containing:
- the content of the file
- more nested directory structures

Giving them different variable names actually helps here I think.

filesystem/iso9660.ksy
```diff
@@ -5,269 +5,684 @@ meta:
...
+            'su_signature::rrip_posix_device_number': susp_unknown # PN
+            'su_signature::rrip_posix_file_attributes': rrip_px # PX
+            'su_signature::rrip_relocated_directory': susp_unknown # RE
+            'su_signature::rrip_extensions_in_use_indicator': susp_unknown # RR
+            'su_signature::rrip_sparse_file': susp_unknown # SF
+            'su_signature::rrip_symbolic_link': susp_unknown # SL
+            'su_signature::susp_indicator': susp_sp # SP
+            'su_signature::susp_terminator': susp_st # ST
+            'su_signature::rrip_time_file': rrip_tf # TF
+            'su_signature::rrzf_zisofs': susp_unknown # ZF
```
KOLANICH: Aren't they subtypes? Should we put them into own ```types``` and drop the prefix?

filesystem/iso9660.ksy
KOLANICH: >The switch-on would leave me a single variable containing
    * the content of the file
    * more nested directory structures

Yes, and IMHO it's good.

filesystem/iso9660.ksy
```diff
@@ -5,269 +5,684 @@ meta:
...
+    seq:
+      - id: le
+        type: u4le
+      - id: be
+        type: u4be
+  volume_descriptor:
+    doc-ref: ecma-119 8.1
     seq:
-      - id: type
+      - id: descriptor_type
```
KOLANICH: When you consider the work as finished please do some cleanup involving removing unnecessary prefixes and moving types into local ```types```. I personally prefer to do it earlier as I actively use code folding in editor.

filesystem/iso9660.ksy
SnowFlakey: Done

filesystem/iso9660.ksy
SnowFlakey: Making sure I understand your comment:

I put the prefix in front of it as a reference to the specification.
As I also put that in the doc-ref, I should drop that.

KOLANICH: Not quite.

I mean that it often happens that some types are used only in some type. In this case to improve readability (most of editors allow to fold all the blocks and then only keep the relevant blocks unfolded) we can move it into `types` in the type using these types. Often these types have a prefix in name of the name type they are used in. When inlined their "address" would be ```<parent_type_name>::<parent_type_name>_<subtype_name>```. It's ugly and redundant. In these cases I remove parent type name making the path look like ```<parent_type_name>::<subtype_name>```.

SnowFlakey: I am sorry, but I really do not understand this comment...
Are you talking about: 
- a nested enum
- splitting the enum, and reference multiple?
- moving the enum into su_header
Or just something else?

filesystem/iso9660.ksy
SnowFlakey: In many places I could to a cleanup like this, but changing descriptor_type to type breaks the code, no idea why, i guess it is not a reserved word.

filesystem/iso9660.ksy
SnowFlakey: I made the enums local

KOLANICH: I meant not only enums, but the types too.

SnowFlakey: Okay

filesystem/iso9660.ksy
KOLANICH: It's really strange. In multiple of my definitions I have used `type` as `id` and everything worked OK. What exactly happens?

filesystem/iso9660.ksy
SnowFlakey: Done

filesystem/iso9660.ksy
SnowFlakey: Done! 
When changing this the ide failed in a way it did not even know the id in the global meta block.

filesystem/iso9660.ksy
```diff
@@ -87,6 +87,37 @@ types:
...
+        type: u2le
+      - id: dir_name
+        doc-ref: ecma-119 9.4.5
+        type: str
+        size: len_dir_name
+      - id: padding_field
+        doc-ref: ecma-119 9.4.6
+        doc: |
+          Padding field is added when len_dir_name contains an odd number
+        size: 0x1
```
KOLANICH: ```size: len_dir_name & 1``` ?

filesystem/iso9660.ksy
```diff
@@ -778,6 +809,12 @@ types:
           - id: file_structure_version
             doc-ref: ecma-119 8.4.31
             type: s1
+        instances:
+          path_table:
+            io: _root._io
```
KOLANICH: Are you sure it should positioned manually to that offset?

SnowFlakey: Yes I am sure, I have reached the end of the current stream.
The data offset is documented as a sector on the disk

filesystem/iso9660.ksy
SnowFlakey: "fixed", going with "len_dir_name % 2 == 1"

KOLANICH: ???

I have looked on your "fix" and don't understand why you need a separate field

SnowFlakey:
Testing the code:
Never found... 
- Amiga shareware ISO's that actually use the Amiga extension.
- Any ISO using zisofs
- Any ISO using optional path records, or extented attribute records ( also not clear where they should be stored )

Good sources:
- path records = OS/2
- SUSP + RRIP headers = any Linux ISO
- non SUSP, SU headers = Solaris, can't remember the version, header magic was "SunOS"

Overall logics:
directory records, never follow 0 or 1 ( . and .. ) or last ( end record )

filesystem/iso9660.ksy
SnowFlakey: I just realise that you combined the size and if statement, it was not about the if statement itself.
The spec actually states, that is should be 0x0. ( contents: [ 0x0 ] ) ?
But that would not combine the attributes

KOLANICH: Do you mean that it should look like
```yaml
- id: padding_field
                 doc-ref: ecma-119 9.4.6
                 doc: |
                   Padding field is added when len_dir_name contains an odd number
                content: [0x00]
                if: len_dir_name % 2
```
?

SnowFlakey: That would be my best interpretation of the spec.

9.4.6 Padding Field [BP (9 + LEN_DI)]
This field shall be present in the Path Table Record only if the number in the Length of Directory Identifier field is an odd number.
If present, this field shall be set to (00).

SnowFlakey:
iso9660 with Rock Ridge test

Created tags:
AL+ AS- CE- CL- ER- ES- NM+ PD- PL- PN- PX+ RE- RR- SF- SL- SP- ST- TF+ ZF+

apt-get install xorriso

# Basics

rm -rf iso9660rr
mkdir iso9660rr

# links

mkdir iso9660rr/links
echo "Original" > iso9660rr/links/original.txt
ln iso9660rr/links/original.txt iso9660rr/links/hardlink.txt
ln -s original.txt iso9660rr/links/softlink.txt

  # I see original and softlink in the directoy_records but never in directoy_record

# deep directories

mkdir -p iso9660rr/deepdirs/1/2/3/4/5/6/7/8/9
echo "Found in the deep end" > iso9660rr/deepdirs/1/2/3/4/5/6/7/8/9/deep.txt

# zisofs

mkdir zisofs.org
for i in `seq 0 1023`; do echo "the quick brown fox jumps over the lazy dog" >> zisofs.org/compress.txt; done
mkzftree -v zisofs.org iso9660rr/zisofs

# Identifier files

echo "AbstractIdentifier" > iso9660rr/A.TXT
echo "BibliographicFileIdentifier" > iso9660rr/B.TXT
echo "CopyrightFileIdentifier" > iso9660rr/C.TXT

xorrisofs -abstract A.TXT -biblio B.TXT -copyright C.TXT -volset "Kaitai test ISO" -volid "DISK_01_OF_01" -publisher "Kaitai project" -appid "xorrisofs" -sysid "Kaitai" -preparer "@xorriso@" --hardlinks -R -z -o iso9660rr.iso ./iso9660rr

filesystem/iso9660.ksy
```diff
@@ -5,206 +5,84 @@ meta:
...
-      - id: volume_id
-        type: str
-        size: 32
-        encoding: UTF-8
-      - id: unused2
-        contents: [0, 0, 0, 0, 0, 0, 0, 0]
-      - id: vol_space_size
-        type: u4bi
-      - id: unused3
-        contents: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```
KOLANICH: @GreyCat, it seems we need sequence multiplications and list comprehensions, don't we?

filesystem/iso9660.ksy
```diff
@@ -5,206 +5,84 @@ meta:
...
   Format (HSF), proposed for CD-ROMs in 1985, and, after several
-  revisions, it was accepted as ISO9960:1998.
-
+  revisions, it was accepted as ISO9660:1998.
   The format emphasizes portability (thus having pretty minimal
   features and very conservative file names standards) and sequential
   access (which favors disc devices with relatively slow rotation
   speed).
+doc-ref: |
+  ecma-119 http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-119.pdf
```
KOLANICH: https

filesystem/iso9660.ksy
```diff
@@ -214,60 +92,847 @@ types:
...
+        instances:
+          path_table:
+            io: _root._io
+            pos: _root.sector_size * loc_l_path_table
+            size: path_table_size.le
+            type: path_table_records
+      supplementary:
+        doc-ref: ecma-119 8.5
+        seq:
+          - id: volume_flags_reserved
```
KOLANICH: If I were you I would have moved all the flags into a separate type.

KOLANICH:
The file is large and the commit history is long. So, I guess it may worth to split it into multiple files. And please don't forget to squash all the commits into a single one rewriting the history when the PR will be ready to be merged.

filesystem/iso9660.ksy
```diff
@@ -214,60 +92,847 @@ types:
...
+                type: u1
+              - id: vol_seq_num
+                doc-ref: ecma-119 9.1.9
+                type: u2bi
+              - id: len_fi
+                doc-ref: ecma-119 9.1.10
+                type: u1
+              - id: file_id_file
+                doc-ref: ecma-119 9.1.11
+                type: str
```
SnowFlakey: When called from primary this has ASCII encoding.
When called from supplementary this could have UTF-8 / UFT-16 encoding.
And there are multiple types in between, even in a loop.

How can I do that? ( not really wanting to copy the whole type )

KOLANICH: Using parametrized types, of course.

GreyCat: @SnowFlakey I don't quite understand what do you mean by "UTF-8 / UTF-16" encoding. They are completely incompatible, how would it decide on an encoding?
@KOLANICH No, you can't do that. We don't allow passing arbitrary variables into encoding.

KOLANICH: Unfortunately we cannot pass even expressions.
But we can do some perversions
```yaml
params:
  - id: use_utf8
    type: b1
seq:
  - id: str_a
    type: strz
    encoding: ascii
    if: not use_utf8
  - id: str_u
    type: strz
    encoding: utf-8
    if: use_utf8
instances:
  str:
    value: (use_utf ? str_u : str_a) #doesn't work for now
```

filesystem/iso9660.ksy
```diff
@@ -258,12 +258,15 @@ types:
             type: text32
           - id: unused01
             doc-ref: ecma-119 8.5
-            size: 0x8
+            contents: [ 0x0, 0x0, 0x0, 0x0,  0x0, 0x0, 0x0, 0x0 ]
```
KOLANICH:
```yaml
  reserved:
    seq:
      - id: zero
        contents: [0]
        repeat: eos
```

filesystem/iso9660.ksy
SnowFlakey: Fixed

filesystem/iso9660.ksy
```diff
@@ -7,206 +7,84 @@ meta:
...
   (mostly CD-ROM). The standard was based on earlier High Sierra
   Format (HSF), proposed for CD-ROMs in 1985, and, after several
-  revisions, it was accepted as ISO9960:1998.
-
+  revisions, it was accepted as ISO9660:1998.
   The format emphasizes portability (thus having pretty minimal
   features and very conservative file names standards) and sequential
   access (which favors disc devices with relatively slow rotation
   speed).
+doc-ref: |
```
KOLANICH, line 21: BTW we now have arrays in `doc-ref`

filesystem/iso9660.ksy
```diff
@@ -216,60 +94,846 @@ types:
...
+                          - id: version
+                            contents: [ 0x1 ]
+                          - id: lba_child
+                            type: u4bi
+                      susp_er:
+                        doc-ref: susp 5.5
+                        seq:
+                          - id: length
+                            type: u1
+                          - id: version
```
KOLANICH, line 662:
```yaml
type: u1
valid: 1
```

filesystem/iso9660.ksy
```diff
@@ -216,60 +94,846 @@ types:
...
+                            contents: [ 0x1 ]
+                          - id: virtual_file_size_high
+                            type: u4bi
+                          - id: virtual_file_size_low
+                            type: u4bi
+                          - id: table_debth
+                            type: u1
+                            enum: table_debth
+                        enums:
+                          table_debth:
```
KOLANICH, line 776: de***p***th?

filesystem/iso9660.ksy
```diff
@@ -216,60 +94,846 @@ types:
...
+                            size: length - 5
+                      susp_sp:
+                        doc-ref: susp 5.3
+                        seq:
+                          - id: length
+                            type: u1
+                          - id: version
+                            contents: [ 0x1 ]
+                          - id: check_bytes
+                            contents: [ 0xbe, 0xef ]
```
KOLANICH, line 812:
```yaml
type: u2
valid: 0xbeef
```

SnowFlakey: Just to be sure... would both little and big endian do the same with 0xbeef?
As the iso9660 is a big mix of both

filesystem/iso9660.ksy
SnowFlakey: fixed

filesystem/iso9660.ksy
generalmimon: > Just to be sure... would both little and big endian do the same with 0xbeef?

No, `u2` will be affected by the default endianness. I would leave `contents` here.

filesystem/iso9660.ksy
SnowFlakey: fixed

filesystem/iso9660.ksy
KOLANICH: No, `u2` should use the current endianness (endianness can be switcheable (automatically making all the numbers specified without explicit endianness being of endianness chosen in runtime), though this feature seems to be got broken somewhen and though in this spec it is not used). If one wants it always be `be` for some `u2` field, he can always write `u2be`. Fixed the proposed change to make it correspond to `contents: [0xbe, 0xef]` and make it guaranteedly be `be` even if you would decide to use switcheable endianness.

filesystem/iso9660.ksy
SnowFlakey: The length is not always 1, but I fixed the "version" with the same logic

filesystem/iso9660.ksy
KOLANICH: >Also it's literally called check_bytes, so there's no real reason to convert it into an integer (maybe except saving a few CPU clock ticks).

It is 2 bytes, the `be` endianness is used extensively in this spec, CPUs think in words, and 0x(dead)beef is a well-known hex-speak constant ... I feel like it was meant to be used as `u2` by the standards authors.

SnowFlakey: The kind of strange logic in this format is that many times both little and big endian are stored in different fields.
But I did not read all those specs in 2 years...

SnowFlakey: "(I can see 18 occurences of contents: [ 0x1 ] in the file)"
I replaced all 1 byte contents

SnowFlakey: I think the logic of 0xbe 0xef is, you read it as a string and you also reference it as a u2, then the can see if your machine is le or be

KOLANICH: Such things are usually done in headers when the field is at well-known offset, not in deeply buried types. The only reason I can imagine it be done in a deeply buried type is supporting multiple endiannesses and allowing each piece to have an own endianness, shifting the burden of byte-swapping to the one reading it. What does the standard say?

SnowFlakey: I did document the spec reference so nicely... 
doc-ref: susp 5.3 
susp = http://www.ymi.com/ymi/sites/default/files/pdf/Systems%20Use%20P1281.pdf

And now www.ymi.com is down

SnowFlakey:
Are there any other things I have to look at?
I started to work on this to finally get it merged.

I know this is large, this because so many extensions where written for ISO9660.
Please remember that the current file does not even work.

KOLANICH:
Merging is a big problem here (see the list of PRs). Ones who merge are out of time to review everything. But I guess a non-working spec won't be merged at all. So I propose to do the following.

1. create a separate branch and copy this file there
2. remove everything that prevents it from working
3. Commit, push and create a separate **working** PR.
4. Squash all the commits in this PR into 1. Rebase this branch over that one. So you would get 2 commits, one implementing lot of stuff and working, another one adding the stuff that is broken.

SnowFlakey:
> Merging is a big problem here (see the list of PRs). Ones who merge are out of time to review everything, PRs stay unmerged for years. But I guess a non-working spec won't be merged at all. So I propose to do the following.
> 
> 1. create a separate branch and copy this file there

I did merge kaitai-io/kaitai_struct_formats origin/master back into my branch.

> 2. remove everything that prevents it from working

This ISO9660 spec works, the file in kaitai-io/kaitai_struct_formats does not.

> 3. Commit, push and create a separate **working** PR.
> 4. Squash all the commits in this PR into 1. Rebase this branch over that one. So you would get 2 commits, one implementing lot of stuff and working, another one adding the stuff that is broken.

KOLANICH:
>This ISO9660 spec works, the file in kaitai-io/kaitai_struct_formats does not.

Thank you for the clarification, and especially for the spec.

filesystem/iso9660.ksy
generalmimon: Yes, I can confirm that, I found the same logic in the Android source code (https://android.googlesource.com/platform/external/syslinux/+/refs/heads/oreo-m4-s3-release/core/fs/iso9660/susp_rr.c#299):

```c
    if (sua[0] != 'S' || sua[1] != 'P' || sua[2] != 7 || sua[3] != 1 ||
	sua[4] != 0xbe || sua[5] != 0xef)
	return 0;
```

So no endianness stuff here.

> And now [www.ymi.com](http://www.ymi.com) is down...

Please update the `doc-ref` links that no longer work with the Wayback Machine mirror, so https://web.archive.org/web/20170404132301/http://www.ymi.com/ymi/sites/default/files/pdf/Systems%20Use%20P1281.pdf in this case.

filesystem/iso9660.ksy
```diff
@@ -216,60 +94,867 @@ types:
...
-instances:
-  sector_size:
-    value: 2048
-  primary_vol_desc:
-    pos: 0x010 * sector_size
-    type: vol_desc
+        enum: volume_type
+      - id: magic
+        doc-ref: ecma-119 8.1.2
+        contents: [ 0x43, 0x44, 0x30, 0x30, 0x31 ]
```
KOLANICH, line 113: Just

```yaml
contents: CD001
```

filesystem/iso9660.ksy
```diff
@@ -7,206 +7,84 @@ meta:
...
-      - id: boot_id
-        type: str
-        size: 32
-        encoding: UTF-8
-  vol_desc_primary:
-    doc-ref: 'http://wiki.osdev.org/ISO_9660#The_Primary_Volume_Descriptor'
+      - id: le
+        type: u2le
+      - id: be
+        type: u2be
```
KOLANICH, line 52:
```yaml
valid: le
```

filesystem/iso9660.ksy
```diff
@@ -7,206 +7,84 @@ meta:
...
-      - id: vol_seq_num
-        type: u2bi
-      - id: logical_block_size
-        type: u2bi
-      - id: path_table_size
-        type: u4bi
-      - id: lba_path_table_le
-        type: u4le
-      - id: lba_opt_path_table_le
+      - id: le
```
KOLANICH, line 56: `u4bi`?

filesystem/iso9660.ksy
```diff
@@ -7,206 +7,84 @@ meta:
...
-        encoding: UTF-8
-      - id: abstract_file_id
+        size: 0x2
+      - id: hour
         type: str
-        size: 36
-        encoding: UTF-8
-      - id: bibliographic_file_id
+        size: 0x2
+      - id: minute
```
KOLANICH, line 75: in the previous occurence it was `min`

filesystem/iso9660.ksy
```diff
@@ -216,60 +94,867 @@ types:
...
+      supplementary:
+        doc-ref: ecma-119 8.5
+        seq:
+          - id: volume_flags_reserved
+            doc-ref: ecma-119 8.5.3 b1-b7
+            type: b7
+          - id: volume_flags_not_iso2375
+            doc-ref: ecma-119 8.5.3 b0
+            type: b1
+          - id: system_identifier
```
KOLANICH, line 257: duplication

filesystem/iso9660.ksy
```diff
@@ -216,60 +94,867 @@ types:
...
+                            type: b1
+                          - id: parent
+                            type: b1
+                          - id: current
+                            type: b1
+                          - id: continue
+                            type: b1
+                          - id: content
+                            type: str
+                            size: length - 5
```
KOLANICH, line 820: Move the 5 bytes into a subtype and replace `5` with `sizeof<the_subtype>`

filesystem/iso9660.ksy
```diff
@@ -7,206 +7,84 @@ meta:
     pronom: fmt/468
     wikidata: Q815645
   license: CC0-1.0
+  endian: be
```
generalmimon, line 10:
```suggestion
  ks-version: 0.9
  bit-endian: be
  endian: be
```

The spec uses [big-endian bit-sized integers](https://doc.kaitai.io/user_guide.html#bit-ints-be) and it's better to make this explicit with the [`bit-endian` key](https://doc.kaitai.io/user_guide.html#bit-endian). Also `bit-endian` and `valid` were added in [0.9](https://kaitai.io/news/2020/10/16/kaitai-struct-v0.9-released.html), so let's set that as a minimum required KS version (although not really a big deal, because only a few languages' codegen actually cares about the `ks-version` key and also with every major release we make become all previous versions EOL).

filesystem/iso9660.ksy
```diff
@@ -216,60 +94,867 @@ types:
...
+            type: text32
+          - id: unused02
+            doc-ref: ecma-119 8.4.7
+            contents: [ 0x0, 0x0, 0x0, 0x0,  0x0, 0x0, 0x0, 0x0 ]
+          - id: volume_space_size
+            doc-ref: ecma-119 8.4.8
+            type: u4bi
+          - id: unused03
+            doc-ref: ecma-119 8.4.9
+            contents: [ 0x0, 0x0, 0x0, 0x0,  0x0, 0x0, 0x0, 0x0,  0x0, 0x0, 0x0, 0x0,  0x0, 0x0, 0x0, 0x0,  0x0, 0x0, 0x0, 0x0,  0x0, 0x0, 0x0, 0x0,  0x0, 0x0, 0x0, 0x0,  0x0, 0x0, 0x0, 0x0 ]
```
generalmimon, line 174: Please split it into multiple lines for better readability:

```suggestion
            contents: |
              [
                0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
                0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
                0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
                0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0
              ]
```

filesystem/iso9660.ksy
SnowFlakey: I don't understand your comment.
I expect both fields to have a 16 bit number, one is le and the other is be encoded

filesystem/iso9660.ksy
SnowFlakey: My variables are the same as in the document, but I follow multiple documents.
So the naming style does sometimes change.

If I rename things the validation against the original spec would get much harder.

filesystem/iso9660.ksy
SnowFlakey: Well in the spec they give the total size of the block incl the header, so you have to reduce the size of the header.
And making breaking it into a subtype will just make more of a mess.

filesystem/iso9660.ksy
KOLANICH: `-orig-id: `

filesystem/iso9660.ksy
KOLANICH: Must the 2 numbers be the same?

filesystem/iso9660.ksy
KOLANICH: IMHO `[0x00] * 32` would be much more readable

filesystem/iso9660.ksy
SnowFlakey: Technically yes, but the le = le ordered and the be = be ordered.
If you look at it as a string you will see the 2 bytes are swapped.

filesystem/iso9660.ksy
```diff
@@ -7,206 +7,84 @@ meta:
...
-      - id: data_preparer_id
+  datetime_long:
+    doc-ref: ecma-119 8.4.26.1
+    seq:
+      - id: year
         type: str
-        size: 128
-        encoding: UTF-8
-      - id: application_id
+        size: 0x4
```
generalmimon, line 65: Please don't use hex notation in the `size` key. 

```suggestion
        size: 4
```

Even the `ecma-119` specification uses normally decimal notation, so why do you convert it?

> **8.4.26.1 Date and Time Format**
> 
> The date and time shall be represented by a 17-byte field recorded as follows.
> 
> <table>
> <caption align=center><strong>Table 5 – Date and Time Format</strong></caption>
> <thead>
> <tr>
> <th>RBP</th>
> <th>Interpretation</th>
> <th>Content</th>
> </tr>
> <tbody>
> <tr><td>1 to 4</td><td>Year from I to 9999</td><td>Digits</td></tr>
> <tr><td>5 to 6</td><td>Month of the year from 1 to 12</td><td>Digits</td>
> <tr><td>7 to 8</td><td>Day of the month from 1 to 31</td><td>Digits</td>
> <tr><td>9 to 10</td><td>Hour of the day from 0 to 23</td><td>Digits</td>
> <tr><td>11 to 12</td><td>Minute of the hour from 0 to 59</td><td>Digits</td>
> <tr><td>13 To 14</td><td>Second of the minute from 0 to 59</td><td>Digits</td>
> <tr><td>15 to 16</td><td>Hundredths of a second</td><td>Digits</td>
> <tr><td>17</td><td>Offset from Greenwich Mean Time in number<br>of 15 min intervals from -48 (West) to +52<br> (East) recorded according to 7.1.2</td><td>
> numerical value</td></tr>
> </tbody>
> </table>

I mean, most humans (and even programmers) are still more comfortable with the decimal base than hexadecimal, at least for byte sizes.

filesystem/iso9660.ksy
KOLANICH:
```yaml
valid: le
```

is equivalent to

```yaml
valid:
  eq: le
```

filesystem/iso9660.ksy
SnowFlakey: fixed

filesystem/iso9660.ksy
SnowFlakey: fixed

filesystem/iso9660.ksy
SnowFlakey: fixed

filesystem/iso9660.ksy
SnowFlakey: I don't see it. One is system other is volume, even the doc-ref is different

filesystem/iso9660.ksy
```diff
@@ -7,206 +7,84 @@ meta:
...
   Format (HSF), proposed for CD-ROMs in 1985, and, after several
-  revisions, it was accepted as ISO9960:1998.
-
+  revisions, it was accepted as ISO9660:1998.
   The format emphasizes portability (thus having pretty minimal
   features and very conservative file names standards) and sequential
   access (which favors disc devices with relatively slow rotation
   speed).
+doc-ref:
+  - ecma-119 https://www.ecma-international.org/publications/files/ECMA-ST/Ecma-119.pdf
```
generalmimon, line 22: This URL is now keeps redirecting (with status code [`301 Moved Permanently`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301)), so please link directly to the target.

SnowFlakey: fixed

filesystem/iso9660.ksy
SnowFlakey: Same discussion: https://github.com/kaitai-io/kaitai_struct_formats/pull/82#discussion_r166069793

filesystem/iso9660.ksy
```diff
@@ -7,269 +7,962 @@ meta:
...
+                doc-ref: ecma-119 9.4.5
+                type: str
+                size: len_dir_name
+              - id: padding_field
+                doc-ref: ecma-119 9.4.6
+                doc: |
+                  Padding field is added when len_dir_name contains an odd number
+                type: u1
+                valid: 0
+                if: len_dir_name % 2 == 1
```
generalmimon, line 388: I don't consider this a good idea - if there is a padding field and all you can say about it is that it's filled with 0's, it should use a byte array (and therefore go with `contents`):

```suggestion
              - id: padding_field
                doc-ref: ecma-119 9.4.6
                doc: |
                  Padding field is added when len_dir_name contains an odd number
                contents: [ 0x0 ]
                if: len_dir_name % 2 == 1
```

The difference from the `version` field (https://github.com/kaitai-io/kaitai_struct_formats/pull/82#discussion_r606848014) is in the intent. You want `version` to be an integer, because it really is a meaningful number. On the other hand, magic signatures and paddings should be just byte arrays.

filesystem/iso9660.ksy
KOLANICH: https://github.com/kaitai-io/kaitai_struct_formats/pull/82/files#diff-1e076bd9a91d2da636ba57110890a3857b5ebade3b60b49ed57f9d1d35dc08d2R162

filesystem/iso9660.ksy
SnowFlakey: I had to revert this!
It breaks while testing on https://ide.kaitai.io/

filesystem/iso9660.ksy
SnowFlakey: Removed all size in hex

filesystem/iso9660.ksy
SnowFlakey: fixed

filesystem/iso9660.ksy
generalmimon: > I had to revert this!
> It breaks while testing on https://ide.kaitai.io/

Oh, you're right, thanks for pointing it out! That's actually something predictable, since `contents` only supports YAML arrays, not string expressions as most KS language keys.

In this case, just leave out the `|` character which makes the contents to be interpreted as a string:

```suggestion
            contents: [
              0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
              0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
              0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
              0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0
            ]
```

filesystem/iso9660.ksy
generalmimon: > fixed

This also needs to be done on other padding fields:

https://github.com/kaitai-io/kaitai_struct_formats/blob/3188df4e7a0fe8f834175d7415591f72a455c62c/filesystem/iso9660.ksy#L467-L473

https://github.com/kaitai-io/kaitai_struct_formats/blob/3188df4e7a0fe8f834175d7415591f72a455c62c/filesystem/iso9660.ksy#L158-L161

filesystem/iso9660.ksy
SnowFlakey: The spec was designed to not burden you with le <-> be conversion.
So compairing these values kind of feels like negating the original intent.

On a rational note... why did they ever create extra IO on a slow medium while CPU's can do this so much faster?

generalmimon: > The spec was designed to not burden you with le <-> be conversion.
> So compairing these values kind of feels like negating the original intent.

I don't think so - you already cause the le <-> be conversion to occur when you use `u2le` and `u2be`. It's not that they are not parsed until you access their value: `seq` attributes are parsed immediately (you'd need to use `instances` which are parsed only on demand - then it would make sense). To avoid byte swaps entirely, you need to stop using integers with non-native endianness. However, native endianness can't be determined natively using the KS language, so you have to determine externally and pass it into the KSY via [`params`](https://doc.kaitai.io/user_guide.html#param-types). Some KS runtime libraries even provide this information, so you can use it:

[**kaitai_struct_cpp_stl_runtime** / kaitai/kaitaistream.cpp:9](https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/blob/4da1668426bd67aaf1932628c1c7db76ff77665c/kaitai/kaitaistream.cpp#L9)

```cpp
#define __BYTE_ORDER    BYTE_ORDER
```

[**kaitai_struct_csharp_runtime** / KaitaiStream.cs:39](https://github.com/kaitai-io/kaitai_struct_csharp_runtime/blob/ac9add4de3f5563471981e078f345c174293bf49/KaitaiStream.cs#L39)

```csharp
static readonly bool IsLittleEndian = BitConverter.IsLittleEndian;
```

[**kaitai_struct_javascript_runtime** / KaitaiStream.js:493](https://github.com/kaitai-io/kaitai_struct_javascript_runtime/blob/2de2c4393182b91c344bfadad0583f3694decd48/KaitaiStream.js#L493)

```js
KaitaiStream.endianness = new Int8Array(new Int16Array([1]).buffer)[0] > 0;
```

---

I don't think that detecting native endianness to avoid endian conversions is necessarily worth it, it depends on the use case. However, if you decide to keep the `u2be` + `u2le` pair in `seq`, I guess that checking their values with `valid/eq` (kaitai-io/kaitai_struct#435) makes sense, because you want to ensure that they store the same values.

filesystem/iso9660.ksy
SnowFlakey: Fixed, but... the closing bracket on its own line also broke it.

filesystem/iso9660.ksy
SnowFlakey: fixed

filesystem/iso9660.ksy
SnowFlakey: Added

filesystem/iso9660.ksy
KOLANICH: @SnowFlakey

filesystem/iso9660.ksy
SnowFlakey: Your question is ambiguous.
Do you wonder what u4bi is? If so a combo of both u4le + u4be 
Or do you wonder if I can replace this with u4bi? if so, no as this is the definition of u4bi

filesystem/iso9660.ksy
SnowFlakey: I have implemented these standards exactly!
The fact that ref. 8.4 & 8.5 partially contain the same items and not a reference to example a u4bi is not for me to fix.

```
      primary:
        doc-ref: ecma-119 8.4
```
```
          - id: system_identifier
            doc-ref: ecma-119 8.4.5
            type: text32
          - id: volume_identifier
            doc-ref: ecma-119 8.4.6
            type: text32
```

```
      supplementary:
        doc-ref: ecma-119 8.5
```
```
          - id: system_identifier
            doc-ref: ecma-119 8.5.4
            type: text32
          - id: volume_identifier
            doc-ref: ecma-119 8.5.5
            type: text32
```

KOLANICH: IMHO the fact that it is written so in the original spec shouldn't prevent us from eliminating redundancy. So, it is for you to fix, their spec is their, but the ksy spec is yours!

filesystem/iso9660.ksy
KOLANICH: @SnowFlakey

filesystem/iso9660.ksy
SnowFlakey: Well I guess we conflict on 2 items... this and the naming of some identifiers as that topic came up 2 times.

This PR works and will greatly extent the support of the format.
If you don't want my code, then let me know and I will close the pr.

KOLANICH: Don't close your PR, your PR is welcome here! The notices made are not requirements, but advices.

filesystem/iso9660.ksy
generalmimon: > And making breaking it into a subtype will just make more of a mess.

Yes, I have to agree - it probably will. However, I have a better idea - you use these specific types in `header`:

https://github.com/kaitai-io/kaitai_struct_formats/blob/619e1ba426a52612905049c37adee9462e6f078c/filesystem/iso9660.ksy#L513-L535

and in fact, all of them start with `length` and `version`.

https://github.com/kaitai-io/kaitai_struct_formats/blob/619e1ba426a52612905049c37adee9462e6f078c/filesystem/iso9660.ksy#L958-L964

https://github.com/kaitai-io/kaitai_struct_formats/blob/619e1ba426a52612905049c37adee9462e6f078c/filesystem/iso9660.ksy#L558-L566

https://github.com/kaitai-io/kaitai_struct_formats/blob/619e1ba426a52612905049c37adee9462e6f078c/filesystem/iso9660.ksy#L645-L653

and so on. I checked that this pair appears in the beginning of all fields, with the only exception that `susp_unknown` doesn't have the `valid: 1` constraint on `version` (if this is intentional and not an oversight, we can easily get around it by using `if` or `valid/expr`).

It would be much better if you would extract these common fields right into `header` (because in fact, why not, it makes sense in my opinion - the names `length` and `version` sound like they totally fit into a header structure):

https://github.com/kaitai-io/kaitai_struct_formats/blob/619e1ba426a52612905049c37adee9462e6f078c/filesystem/iso9660.ksy#L508-L513

```diff
 header:
   seq:
     - id: signature
       type: u2be
       enum: signature
+    - id: length
+      type: u1
+    - id: version
+      type: u1
+      valid: 1
     - id: specific
```

The nice thing about it is that now you have the `length` available in the parent struct, so you can simply put the field `specific` in a substream. Then you will be able to throw out all the `size: length - 5` calculations and just use `size-eos: true`. Also, it won't allow the specific types to possibly get out of the designated substream and the parsing will fail on the first field that doesn't fit in there (not by ending up with a negative size from those `size: length - 5` calculations, which will also lead to an error, but fields with these calculations are not even present in all types - not sure if it's explicitly written in the reference spec that the `length` should be disregarded in some specific types, but if it's not, I'd definitely confine every specific type within that substream).

```diff
 header:
   seq:
@@ ... @@
     - id: specific
+      size: length - (signature._sizeof + length._sizeof + version._sizeof) # note: this `_sizeof` sum should probably be in a `value` instance
       type:
         switch-on: signature
         cases:
           'signature::aaip_attribute_list': susp_unknown # AL
@@ ... @@
           'signature::rrzf_zisofs': rrzf_zf # ZF
```

Now, you will just take each specific type, for example:

https://github.com/kaitai-io/kaitai_struct_formats/blob/619e1ba426a52612905049c37adee9462e6f078c/filesystem/iso9660.ksy#L722-L731

..., simply throw out `length` and `version`, and replace `size: length - 4` with `size-eos: true`:

```diff
 susp_pd:
   doc-ref: susp 5.2
   seq:
-    - id: length
-      type: u1
-    - id: version
-      type: u1
-      valid: 1
     - id: padding_area
-      size: length - 4
+      size-eos: true
```

You see how much better it is? You'll replace 18 identical `length`, `version` pairs in every single specific type with 1 common pair in `header`, you'll treat the `length` consistently and correctly, and eliminate the magic numbers in manual `size: length - 5` calculations. What more could we possibly wish for?

SnowFlakey: "What more could we possibly wish for?"
For the feeling to go away that this PR is eternally unmergeable.

You can have it as-is and improve on it in your own time.

filesystem/iso9660.ksy
```diff
@@ -7,269 +7,959 @@ meta:
...
+                            valid: 1
+                          - id: reserved
+                            doc: |
+                              Grouped all 4x reserved into a single reserved
+                            type: b5
+                          - id: parent
+                            type: b1
+                          - id: current
+                            type: b1
+                          - id: continue
```
armijnhemel, line 717: I would recommend changing this to `continued` as `continue` is a reserved keyword in Python and it is causing some issues.

filesystem/iso9660.ksy
```diff
@@ -7,269 +7,959 @@ meta:
...
+                            valid: 1
+                          - id: reserved
+                            type: b4
+                          - id: root
+                            type: b1
+                          - id: parent
+                            type: b1
+                          - id: current
+                            type: b1
+                          - id: continue
```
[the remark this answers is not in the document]
KOLANICH: maybe `continuation`?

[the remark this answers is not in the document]
SnowFlakey: [Please see](https://github.com/kaitai-io/kaitai_struct_formats/pull/82#discussion_r608703391)

filesystem/iso9660.ksy
```diff
@@ -7,269 +7,959 @@ meta:
...
+            types:
+              susp:
+                doc: |
+                  We check if we have at least 2 bytes left.
+                  The SUSP magic is 2 bytes.
+                seq:
+                  - id: header
+                    type: header
+                    if: _io.size - _io.pos >= 2
+                    repeat: eos
```
armijnhemel, line 506: According to IEEE P1281 a valid SUSP entry is at least 4 bytes long:

```
If the remaining allocated space following the last recorded System Use Entry in a System Use field or Continuation Area is less than four bytes long, it cannot contain a System Use Entry and shall be ignored.
```
I have adapted `susp` to the following:

```
              susp:
                doc: |
                  We check if we have at least 4 bytes left.
                  The SUSP magic is 2 bytes. A complete SUSP entry
                  is at least 4 bytes long.
                seq:
                  - id: header
                    type: header
                    if: _io.size - _io.pos >= 4
                    repeat: until
                    repeat-until: _io.size - _io.pos < 4

```

filesystem/iso9660.ksy
```diff
@@ -7,269 +7,959 @@ meta:
...
+        seq:
+          - id: len_dr
+            doc-ref: ecma-119 9.1.1
+            doc: |
+              If len_dr == 0 we do not process the body
+              If len_dr >= 1 we include field len_dr in the size
+            type: u1
+          - id: body
+            type: body
+            if: len_dr > 0x0
```
armijnhemel, line 413:
```
size: len_dr - len_dr._sizeof
```

filesystem/iso9660.ksy
```diff
@@ -7,269 +7,959 @@ meta:
...
+                  Padding field is added when len_fi contains an even number
+                contents: [ 0x0 ]
+                if: len_fi % 2 == 0
+              - id: system_use
+                doc-ref: ecma-119 9.1.13
+                doc: |
+                  The size of the system_use is defined as:
+                  ( len_dr - size of directory_record = 33 bytes ) - len_fi
+                type: susp
+                size: ( _parent.len_dr - 33 ) - len_fi
```
armijnhemel, line 484: This is actually not correct: `susp` is whatever it is from up to that point, to the end of the directory record. If there is a padding byte, then it should be decreased by 1.

By specifying the size it becomes really easy to avoid this problem:

```
                size-eos: true
```

filesystem/iso9660.ksy
```diff
@@ -7,269 +7,959 @@ meta:
...
+                type: directory_record
+      directory_records:
+        doc: |
+          First item "." it points to it self
+          Second item ".." it points to the parent, or also to self if it the root
+        seq:
+          - id: directory_record
+            type: directory_record
+            repeat: until
+            repeat-until: _.len_dr == 0
```
armijnhemel, line 401: This is actually not correct, as there can be additional padding bytes. Section 6.8.1.1 of the ISO9660 spec says:

```
Each Directory  Record  shall  end  in  the  Logical  Sector  in  which  it  begins.  Unused  byte  positions  after  the  last Directory Record in a Logical Sector shall be set to (00).
```

So this means that you also need to look at the remaining bytes in the logical sector as well as the total length of the directory records.

PR #84, opened 2018-03-16
Improved DEX format

silviupopescu:
- replaced partially incorrect implementation of uleb128
- added more human-readable fields for certain sections

### DIFF
--- a/executable/dex.ksy
+++ b/executable/dex.ksy
@@ -6,7 +6,7 @@ meta:
   # Portions of this page are reproduced from work created and shared by Google and used
   # according to terms described in the Creative Commons 3.0 Attribution License.
   imports:
-    - ../common/vlq_base128_le
+    - /common/vlq_base128_le
 doc-ref: https://source.android.com/devices/tech/dalvik/dex-format
 seq:
   - id: header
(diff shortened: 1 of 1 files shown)


executable/dex.ksy
```diff
@@ -5,6 +5,8 @@ meta:
   license: CC-BY-SA-3.0
   # Portions of this page are reproduced from work created and shared by Google and used
   # according to terms described in the Creative Commons 3.0 Attribution License.
+  imports:
+    - ../common/vlq_base128_le
```
GreyCat: I guess it should be `/common/vlq_base128_le`, as this is considered a "standard library" type. It shouldn't depend on where the dex.ksy is located.

executable/dex.ksy
```diff
@@ -352,6 +306,16 @@ types:
...
+    instances:
+      shorty_desc:
+        value: _root.string_ids[shorty_idx].value.data
+      params_types:
+        io: _root._io
+        pos: parameters_off
+        type: type_list
+        if: parameters_off != 0
+      return_type:
+        value: _root.type_ids[return_type_idx].type_name
```
GreyCat: It would be really cool to add some documentation lines on what these are and how they would be useful for end-user.

executable/dex.ksy
silviupopescu: Makes sense. I'll strip the leading periods.

executable/dex.ksy
silviupopescu: I have nothing against it, of course, but I didn't include any further explanations since they are analogous to `type_name` in `type_id_item`. The DEX file format uses indices into the string table and this just exports the string themselves. They are not new fields added to the format, but rather the same information displayed differently. Wouldn't adding docstrings for the instances just duplicate the information in the corresponding seq elements?

GreyCat:
Thanks, @silviupopescu and @koczkatamas!

PR #85, opened 2018-03-22
[network/icmp_packet] add missing data field

mhils:
The ICMP echo messages are missing the final data field. The size information is in the IP header, so we just read until the end of the substream.

Thanks again for the fantastic project! 🍰


GreyCat:
Thanks for both the praise and the contribution!

PR #88, opened 2018-04-13
Added crdb

KOLANICH:


database/crdb.ksy
```diff
@@ -0,0 +1,165 @@
+meta:
+  id: crdb
+  title: GNU/Linux wireless regulation database
+  license: GPL-2.0
```
GreyCat: I believe this is somewhat more complicated. It's amalgam of ISC-licensed data, and CopyLeft-Next-licensed code.

KOLANICH: You are right! I thought that everything on kernel.org is under gpl.

PR #89, opened 2018-04-13
Added MRC

KOLANICH:
The spec is incomplete:
1 the `doc`s are missing because of copyright issues.
2 some parts are missing because untested and not quite finished. Will be added later.

### DIFF
--- a/scientific/TEM/mrc.ksy
+++ b/scientific/TEM/mrc.ksy
@@ -1,0 +1,265 @@
+meta:
+  id: mrc
+  title: MRC/CCP4 2014 file format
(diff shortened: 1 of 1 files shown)


GreyCat:
Could we actually contact these guys and propose some collaboration, and agreeing upon some sort of free license? https://pypi.python.org/pypi/mrcfile is actually BSD-licensed, so I think they shouldn't mind.

KOLANICH:
I have already contacted them but no response for now.

KOLANICH:
Replaced by #340

PR #90, opened 2018-04-16
ELF + Mach-O update

koczkatamas:
- Mach-O: fixes a flag-type field representation (`macho_flags`)
- ELF: parses phdr_type & section_header flags
- ELF: parses dynamic section


GreyCat:
I actually have only one question: why would you do it in a such a weird way? Won't just a plain `seq` with lots of `type: b1` be easier?

PR #92, opened 2018-04-16
elf: skeleton dynsym/dynstr support

rossburton:
This isn't perfect as you need to manually look up the symbol names in dynstr,
but it's a step in the right direction.


GreyCat:
Looks legit, thanks!

PR #95, opened 2018-04-25
microsoft_pe.ksy - Update SubSystem Enum

yuvaljacoby:
https://msdn.microsoft.com/en-us/library/windows/desktop/ms680547(v=vs.85).aspx#Windows_Subsystem


GreyCat:
Thanks!

PR #98, opened 2018-07-29
Mifare support

KOLANICH:


hardware/mifare/mifare_classic.ksy
```diff
@@ -0,0 +1,243 @@
...
+    repeat: eos
+types:
+  key:
+    seq:
+      - id: key
+        size: 6
+  sector:
+    params:
+      - id: idx
+        type: u1
```
GreyCat: The sole purpose of `idx` here is to make sure that before the body of very first sector, a `manufacturer` signature will appear. Why don't just move in outside of the sector, and get rid of this `idx` and whole params thing?

hardware/mifare/mifare_classic.ksy
```diff
@@ -0,0 +1,243 @@
...
+    seq:
+      - id: manufacturer
+        type: manufacturer
+        if: has_manufacturer
+      
+      - id: data_
```
GreyCat: Having `data_` and `data` will break for all languages with non-underscore case, i.e. C#, Java, JavaScript, etc.

hardware/mifare/mifare_classic.ksy
```diff
@@ -0,0 +1,243 @@
...
+        -orig-id: abtData
+        size: _io.size - _io.pos - 16 # sizeof(trailer)
```
GreyCat: You don't specify `size` for a `sector`, so this would essentially mean that there's always one sector, consisting of:

* `manufacturer`
* `data_` (spanning all up to last 16 bytes)
* `trailer` (16 bytes)

Are you sure that this is what you want?

hardware/mifare/mifare_classic.ksy
```diff
@@ -0,0 +1,243 @@
...
+                value: 'addr_[0]'
+                if: valid
+              value:
+                value: value_[0]
+                if: valid
+      filler:
+        doc: "only to create _io"
+        seq:
+          - id: data
+            size: _io.size
```
GreyCat: You can just use `filler: {}` or `filler` with only a `doc` element.

hardware/mifare/mifare_classic.ksy
```diff
@@ -0,0 +1,243 @@
...
+              addr_valid:
+                value: 'addr_[0]==~addr_[1] and addr_[0]==addr_[2] and addr_[1]==addr_[3]'
+              valid:
+                value: 'value_valid and addr_valid'
+              addr:
+                value: 'addr_[0]'
+                if: valid
+              value:
+                value: value_[0]
+                if: valid
```
GreyCat: Again, `valid` vs `valid_` will break for many languages.

hardware/mifare/mifare_classic.ksy
KOLANICH: Because it breaks incapsulation. We wanna access all the sectors data in a uniform way. I mean the 0th sector contains data too, and when we loop over sectors we wanna access the data in it too without handling this situation in our code. I expect the compiler be somewhen smart and featureful enough to optimize this stuff out.

hardware/mifare/mifare_classic.ksy
KOLANICH: I guess we should keep trailing underscores. It's a pretty little addition for name transforming functions.

hardware/mifare/mifare_classic.ksy
GreyCat: The problem right now is that your ksy always describes exactly 1 sector. Are you ok with that? If yes, why would you have an array of sectors?

hardware/mifare/mifare_classic.ksy
GreyCat: No, we should not. This would raise tons of complaints from style checkers in all these languages.

KOLANICH: Gmmm, maybe to replace the final underscore with `Internal<number here for uniquiness>' ?

hardware/mifare/mifare_classic.ksy
KOLANICH: Ouch. This is definitely not OK, I need to think more about it, in fact I only used sectors, because each sector is read separately, maybe we really don't need that array. Thank you.

hardware/mifare/mifare_classic.ksy
GreyCat: I don't think it has much to do with incapsulation. If, by definition, there could be only 1 `manufacturer` per `mifare_classic`, it makes total sense to have it as an attribute in `mifare_classic`.

Dealing with sectors and blocks might make some sense when one's writing a flasher app or something like that, but it does not matter if you're acting on a card dump. If they say that there are "sectors", but first "sector" is totally not like all others, it makes sense either to do a special type for that first "sector", or not define first "sector" as sector, but rather as a distinct list of attributes.

hardware/mifare/mifare_classic.ksy
GreyCat: Sure, that will work.

hardware/mifare/mifare_classic.ksy
KOLANICH: >but first "sector" is totally not like all others

It is like all others, the only difference is manufacturer block (in fact it's just a block of data written on factory and locked, but there exist cards (so called "Mifare Zero") with that block unlocked).

GreyCat: I won't push any strong opinions on this, but, generally, what you're dealing here is a card dump. There is no notion of "sectors" or "blocks" in a stream of bytes — you are free to read arbitrary locatons and define whatever data blocks you want. As far as I understand, this whole "sectors" and "blocks" thing is relevant if you are doing some card reader / writer software — and if you do, you generally don't care what's inside these sectors/blocks, you just use them as ready-made data blocks.

KOLANICH: >As far as I understand, this whole "sectors" and "blocks" thing is relevant if you are doing some card reader / writer software — and if you do, you generally don't care what's inside these sectors/blocks, you just use them as ready-made data blocks.

The logic of apps using nfc cards is closely tied to blocks, values and sectors since the card IC "enforce" (the crypto was broken, see mfcuk and mfoc) access conditions on block granularity. Modern apps just use cards as dumb storage using own digital signatures, but still write specific data on specific offsets.

>you are doing some card reader / writer software

All software working with the cards is sort of reader/writer software. Though most of them are not generic readers/writers but application-specific ones combining in a single app reading a card, decoding its data and showing it to a user. Card dumps are often incomplete because some keys may be unknown (and the tools mentioned are usually incompatible with the nfc modules used in android phones - the most widespread hardware for reading and writing cards).

GreyCat: Ok, may be then we should start with some sort of application that does this "read → decode → show to user" routine and see how can we make KS useful for it? For example, if it only works on sector level, there is no point in having this `sectors` array at all.

KOLANICH: >there is no point in having this sectors array at all.

If I understand right, mfd format is just a sequence of raw sectors. The ones needeng to parse sectors or their parts separately can use the types defined here, as I have done in https://github.com/KOLANICH/MFCTool.py

hardware/mifare/mifare_classic.ksy
KOLANICH: I found that `data` is needed, but was put into a separate type for getting its `_io`

hardware/mifare/mifare_classic.ksy
KOLANICH: fixed

hardware/mifare/mifare_classic.ksy
KOLANICH: fixed

hardware/mifare/mifare_classic.ksy
KOLANICH: fixed. But still, we really should make the ksy names with trailing underscore workable cross-language.

hardware/mifare/mifare_classic.ksy
KOLANICH: Tested on the mfd. Works.

KOLANICH: @GreyCat seems to be ready to be merged.

hardware/mifare/mifare_classic.ksy
```diff
@@ -0,0 +1,244 @@
...
+      
+      - id: data_filler
+        -orig-id: abtData
+        size: _io.size - _io.pos - 16 # sizeof(trailer)
+        type: filler
+      - id: trailer
+        type: trailer
+    instances:
+      block_size:
+        value: 16
```
GreyCat: If this is a global constant, it makes sense to move it to top level type (i.e. `mifare_classic`)?

hardware/mifare/mifare_classic.ksy
GreyCat: Ok, but then the comment in `doc` is not true anymore.

GreyCat:
Ok, I have left a few minor comments, feel free to fix it later or not fix it ;) Merging in, thanks!

hardware/mifare/mifare_classic.ksy
KOLANICH: >If this is a global constant, it makes sense to move it to top level type (i.e. mifare_classic)?

It is completely makes sense to move (in fact initially it was there), but if I do it `sector` will depend on presence of it in `_root` and there will be troubles in using `sector` separately.

It may make sense to a add compiler optimization detecting constants and moving them into global scope and using correzponding dzta types for them.

hardware/mifare/mifare_classic.ksy
KOLANICH: Why not true?

Instead of using filler we could have used its `data` directly with the appropriate size. So it was moved into `filler` only to create `_io`

KOLANICH:
BTW, the WebIDE doesn't compile this format. The source of the problem may be that it has 
"Kaitai compiler version: 0.8-SNAPSHOT (2018-01-27 00:24:06.671)`

@koczkatamas, could you upgrade it?

GreyCat:
Looks like 0.8 release fails here: http://formats.kaitai.io/mifare_classic/index.html
If it depends of ksc 0.9+, we should, at the very least, add `meta/ks-version: 0.9`.

I'll investigate why it fails in more detail later.

hardware/mifare/mifare_classic.ksy
GreyCat: > It may make sense to a add compiler optimization detecting constants and moving them into global scope and using corresponding data types for them.

Actually, this is a valid reason to finally introduce something like `consts` (on the same level with `seq`, `instances`, etc).

KOLANICH:
We have problems: in https://formats.kaitai.io/mifare_classic/index.html there is no chart and compiled sources for the languages except lua

PR #100, opened 2018-08-06
remove redundant fileds - annoyingstring

yuvaljacoby:
The annoyingstring is annoying but we do not need all there members (only one of them which we use it in the "name" field)...

### DIFF
--- a/executable/microsoft_pe.ksy
+++ b/executable/microsoft_pe.ksy
@@ -145,7 +145,6 @@ types:
       name_zeroes:
         pos: 0
         type: u4
-        if: name_zeroes == 0
       name_offset:
         pos: 4
         type: u4
--- a/executable/microsoft_pe.ksy
+++ b/executable/microsoft_pe.ksy
@@ -145,6 +145,7 @@ types:
       name_zeroes:
         pos: 0
         type: u4
+        if: name_zeroes == 0
       name_offset:
         pos: 4
         type: u4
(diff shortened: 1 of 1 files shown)


executable/microsoft_pe.ksy
```diff
@@ -145,6 +145,7 @@ types:
       name_zeroes:
         pos: 0
         type: u4
+        if: name_zeroes == 0
```
GreyCat: This by itself doesn't make much sense. Basically, if compiles into an infinite recursion, i.e. to get value of `name_zeroes`, one must check if expression, and for that we need to get value of `name_zeroes`.

GreyCat:
@yuvaljacoby Thanks for your contribution! However, I don't quite understand what you want to achieve with this patch: (1) adding condition to `name_zeroes` leads to infinite recursion, (2) adding conditions to two other members is possible, but what purpose does it serve? These fields are not supposed to be invoked manually anyway.

yuvaljacoby:
1. sorry, you are correct of course. will fix if we will decide this is relevant... 
2. The way I used the ksy was to dynamically get all the data I can for the pe file. The annoying string is part of the symbol table, which is sometime very large (hundreds - thousands of symbols), which resulted multiple unused fields which I preferred to not extract in the first place...

GreyCat:
Makes sense, thanks! Please fix that `name_zeroes` condition and we'll merge it in :)

GreyCat:
@yuvaljacoby, kindly ping? Are you still interested in this PR?

yuvaljacoby:
Wow I totally forgot!
committed the fix, sorry!!!

GreyCat:
Thanks!

PR #101, opened 2018-09-02
PSP save metadata format

KOLANICH:


KOLANICH:
@GreyCat

game/sony/PS/PSP/psp_save_metadata.ksy
```diff
@@ -0,0 +1,102 @@
+meta:
+  id: psp_save_metadata
+  endian: le
+  encoding: utf-8
+  license: Unlicense
+  file-extension: SFO
```
GreyCat: This file lacks `title` — I'd suggest to move current `doc` contents here and extend them to be more useful for an average user doing a search query, i.e. something like

```
title: Sony PlayStation Portable (PSP) save game metadata file (PARAM.SFO)
```

Ideally, I'd love to have `doc` entry too, outlining what it is all about:

* what "Sony PlayStation Portable" is,
* what is "save metadata",
* when and by whom it is created (is it a particular game? is it OS-wide thing? is it hardware-wide thing?),
* where physically it is created (is it some sort of removable memory card? some sort of flash memory built-in in the console?)
* how does one access this file (just read it from a memory card? dumps RAM of the console? dumps built-in flash? unsolders some chip and dumps it?)

game/sony/PS/PSP/psp_save_metadata.ksy
```diff
@@ -0,0 +1,102 @@
+meta:
+  id: psp_save_metadata
```
GreyCat: File name:

* We don't really have any decent support for nested directories yet
* There is little value in starting system of directories (i.e. `sony`, `PS`, `PSP`) when there's no other files to put there
* I don't see why we should use upper case in dir names here
* Compile results are 100% flat, so you'd probably like to include stuff that removes ambiguolity it into file name anyway, i.e. using full names instead of abbreviations, like `sony_playstation_portable_save_metadata` or at least `sony_psp_save_metadata`. `psp` is [way too ambiguous](https://en.wikipedia.org/wiki/PSP), even just in "Computing".

So, basically, my suggestion: let's use `game/sony_playstation_portable_save_metadat` or `game/sony_psp_save_metadata` for now?

game/sony/PS/PSP/psp_save_metadata.ksy
```diff
@@ -0,0 +1,102 @@
...
+          switch-on: name
+          cases:
+            "'CATEGORY'": category
+            "'PARENTAL_LEVEL'": u4 # I guess it is min allowed age
+            "'SAVEDATA_DETAIL'": strz # the string displayed in save description
+            "'SAVEDATA_DIRECTORY'": strz # the name of the directory where a save is saved
+            "'SAVEDATA_FILE_LIST'": savedata_file_list
+            "'SAVEDATA_PARAMS'": savedata_params # don't know what is it
+            "'SAVEDATA_TITLE'": strz # arbitrary string, usually contains a name of a player or a anme of last played level
+            "'TITLE'": strz # contains a name of a game
```
GreyCat: Mixing strings, integers and user types would actually break on languages like C++, which do not have an equivalent of `Object` or `Any` type. I'm not sure if we care, though.

KOLANICH: We care, but we don't. I mean this is compiler bug and compiler should detect and fix this situation automatically. For example it can for c++ generate a special object keeping the type and a pointer and having the conversoin operators (and conversion move operators!) redefined.

KOLANICH:
@GreyCat

KOLANICH:
@GreyCat

PR #104, opened 2018-11-05
Add TR-DOS filesystem format

kolen:
[TR-DOS](https://en.wikipedia.org/wiki/TR-DOS) is operating system for ZX-Spectrum computer with Beta Disc floppy drive interface. It might be not very popular in the west, but was especially popular in USSR and early post-USSR due to availability of Soviet clones of Beta Disc and WD1793.

`.trd` files are sector-by-sector raw floppy disk dumps (similar to PC floppy `.img` or copies of `/dev/fd0`), so this format file describes actual TR-DOS filesystem rather than `.trd` format itself.

~May not compile for `graphviz` and `go` right now due to https://github.com/kaitai-io/kaitai_struct/issues/491~ (removed `substring` because changed to bytes)

- [x] figure out how to extract first byte from filename
- [x] `List(ArraySeq(IntNum(0))) (of class io.kaitai.struct.exprlang.Ast$expr$List)` for `go` target — This is probably ok, Go support is experimental

### DIFF
--- a/filesystem/trd.ksy
+++ b/filesystem/trd.ksy
@@ -3,9 +3,6 @@ meta:
   file-extension: trd
   title: "TR-DOS flat-file disk image"
   license: CC0-1.0
-  # It's extended ascii (with ZX Spectrum Basic tokens, UDG, etc), but
-  # "unknown-8bit" can't be used for some reason
-  encoding: ascii
   endian: le
 doc: |
   .trd file is a raw dump of TR-DOS (ZX-Spectrum) floppy. .trd files are
@@ -39,18 +36,16 @@ types:
   file:
     seq:
       - id: name
-        type: str
         size: 8
       - id: extension
-        type: str
-        size: 1
+        type: u1
       - id: position_and_length
         type:
           switch-on: extension
           cases:
-            '"B"': position_and_length_basic
-            '"C"': position_and_length_code
-            '"#"': position_and_length_print
+            0x42: position_and_length_basic # 'B'
+            0x43: position_and_length_code  # 'C'
+            0x23: position_and_length_print # '#'
             _: position_and_length_generic
       - id: length_sectors
         type: u1
@@ -59,10 +54,14 @@ types:
       - id: starting_track
         type: u1
     instances:
+      name_first_byte:
+        pos: 0
+        type: u1
+        io: name._io
       is_deleted:
-        value: name.substring(0, 1) == "\1"
+        value: name_first_byte == 0x01
       is_terminator:
-        value: name.substring(0, 1) == "\0"
+        value: name_first_byte == 0x00
       contents:
         pos: starting_track * 256 * 16 + starting_sector * 256
         size: length_sectors * 256
@@ -125,14 +124,12 @@ types:
       - id: unused_2
         size: 2
       - id: password
-        type: str
         size: 9
       - id: unused_3
         size: 1
       - id: num_deleted_files
         type: u1
       - id: label
-        type: str
         size: 8
       - id: unused_4
         size: 3
--- a/filesystem/trd.ksy
+++ b/filesystem/trd.ksy
@@ -36,6 +36,9 @@ types:
   file:
     seq:
       - id: name
+        # It uses custom type due to limitation of streams: there's no way to
+        # extract first byte from byte array directly
+        type: filename
         size: 8
       - id: extension
         type: u1
@@ -54,14 +57,10 @@ types:
       - id: starting_track
         type: u1
     instances:
-      name_first_byte:
-        pos: 0
-        type: u1
-        io: name._io
       is_deleted:
-        value: name_first_byte == 0x01
+        value: name.first_byte == 0x01
       is_terminator:
-        value: name_first_byte == 0x00
+        value: name.first_byte == 0x00
       contents:
         pos: starting_track * 256 * 16 + starting_sector * 256
         size: length_sectors * 256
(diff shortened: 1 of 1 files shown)


filesystem/trd.ksy
```diff
@@ -0,0 +1,138 @@
+meta:
+  id: trd
```
GreyCat: KS file formats collection does not have any namespacing — basically, everything eventually gets dumped into a single bin, and the only way you can reference this format is by using "trd". So, we typically strive very hard to not use abbreviations and other short ambiguous IDs. In this particular case, I can't be sure that in future we won't have a clash with some other "trd" — especially given that [Wikipedia gives us so many alternatives](https://en.wikipedia.org/wiki/TRD).

Can I ask you to use a longer name (like `tr_dos_image` or something like that)?

filesystem/trd.ksy
```diff
@@ -0,0 +1,138 @@
...
+  file-extension: trd
+  title: "TR-DOS flat-file disk image"
+  license: CC0-1.0
+  # It's extended ascii (with ZX Spectrum Basic tokens, UDG, etc), but
+  # "unknown-8bit" can't be used for some reason
+  encoding: ascii
```
GreyCat: This is obviously borderline territory, but have you checked if anyone actually has name for this encoding (i.e. IANA, IBM, Java encodings registry, etc?)

Leaving it as ASCII would obviously break stuff if a non-ASCII char will be encountered. Using no encoding and treating these strings as byte arrays would be ugly in terms of representation.

filesystem/trd.ksy
```diff
@@ -0,0 +1,138 @@
...
+        if: extension == "C"
+      - id: print_extent_num
+        type: u1
+        if: extension == "#"
+      - id: print_unused
+        type: u1
+        if: extension == "#"
+      - id: length1_unused
+        size: 2
+        if: extension != "B" and extension != "C" and extension != "#"
```
GreyCat: A suggestion: why not move this tons of `if` into separate types and just use type switching (i.e. `switch-on: extension`)?

filesystem/trd.ksy
```diff
@@ -0,0 +1,138 @@
...
+          (physical_track << 1) | side, the same way that tracks are stored
+          sequentially in .trd file
+        type: u1
+      - id: disk_type
+        type: u1
+        enum: disk_type
+      - id: number_of_files
+        doc: |
+          Number of non-deleted files. Directory can have more than
+          number_of_files entries due to deleted files
```
GreyCat: Please consider following [style guide](http://doc.kaitai.io/ksy_style_guide.html) and naming these `num_files`, `num_free_sectors`, etc?

filesystem/trd.ksy
kolen: Changed to `tr_dos_image`.

filesystem/trd.ksy
kolen: No, there's no such encoding listed everywhere. Most of [extended characters](https://en.wikipedia.org/wiki/ZX_Spectrum_character_set) are not representable as Unicode code points. AFAIK, in languages with checked unicode strings such as Rust such strings can't be represented as string type. So I think the only way to safely handle this is to make filenames byte arrays, then code that uses decoder generated from this definition can convert (with loss) to string representation if needed. Replacing Basic tokens, etc, is outside of scope of filesystem definition, I think.

However, Kaitai IDE will not show decoded filenames in "object tree" in this case.

IANA character set registry has `UNKNOWN-8BIT` charset, but it causes exception in Kaitai IDE, and AFAIK, it has the same semantics as treating string as byte array (it's not real encoding and not representable in unicode).

filesystem/trd.ksy
kolen: Thanks, it's easier and cleaner with `switch-on`. I didn't know about this feature because it's not listed in KSY reference, only in User guide.
Changed to `switch-on`.

filesystem/trd.ksy
kolen: Changed.

filesystem/trd.ksy
kolen: I changed all strings (file names, disk label, disk password) to byte arrays. However, I had to add indirection to file name (it's now available as `name.name`) because of limitation that byte arrays don't have `_io` ([this was also in recent discussion on gitter](https://gitter.im/kaitai_struct/Lobby?at=5be02324cd2a6b5bdc9b89f9)), and I can't find another way to extract first byte from byte array.

kolen:
It now fails with `go` target, but I heard that Go support is experimental (there's no Go on http://formats.kaitai.io) so that's probably ok

```
% kaitai-struct-compiler -t go trd.ksy
List(ArraySeq(IntNum(0))) (of class io.kaitai.struct.exprlang.Ast$expr$List)
```

filesystem/trd.ksy
GreyCat: Thanks for the explanation. Indeed, this looks pretty much like situation with "ANSI" or "OEM" character set, which essentially means that exact set of characters displayed is unknown and depends on system's settings. Indeed, typically we treat that as encoding-less byte array.

You're totally right, byte arrays nowadays lack `a[x]` operator, and I guess that's something that we should address.

GreyCat:
Thanks for your contribution, just merged in :+1:
Go support is indeed lacking (see [CI report](http://kaitai.io/ci/)), but we can gradually work on it.

PR #105, opened 2018-11-05
TrueType 'post' goes to eos.

bungeman:
The current implementation assumes that there will be a zero length
pascal string at the end of a version 2 'post' table. There is no such
requirement and this will fail to load a font which happens to have such
a table which ends on exactly a four byte boundary.

Note that FreeType determines the number of pascal strings to read by
iterating over the glyph name index values and selecting the highest
index. However, this only shows used strings and not all of the data.

### DIFF
--- a/font/ttf.ksy
+++ b/font/ttf.ksy
@@ -880,8 +880,7 @@ types:
             repeat-expr: number_of_glyphs
           - id: glyph_names
             type: pascal_string
-            repeat: until
-            repeat-until: _.length == 0
+            repeat: eos
     seq:
       - id: format
         type: fixed


bungeman:
I'm not sure what you mean by checking _io.size > _io.pos which does anything different from just going to eos (of the 'post' table). Note that this is only being done when 'if: format.major == 2 and format.minor == 0' (see the base 'post' seq below the proposed change) which is never going to change, so there will never be a place after glyph_names which will have any meaningful data in it (garbage can be placed here, but there currently isn't any way of knowing it's garbage). Note also that with this change things work, without it any OpenType font with a version 2.0 post table simply isn't loaded.

Note that the only way to know the 'right' repeat number for glyph_names is to go through all of the glyph_name_index values, take the maximum of them, and subtract 258. There does not appear to be any way to find a maximum from an array, so this doesn't appear possible.

PR #110, opened 2018-11-07
new format ivf (VP8): still very simple, no frame disassembly, but it…

frankpolte:
… is a start. works with the ivf files from https://github.com/webmproject/vp8-test-vectors

### DIFF
--- a/media/ivf.ksy
+++ b/media/ivf.ksy
@@ -1,0 +1,67 @@
+meta:
+  id: ivf
+  file-extension: ivf
+  endian: le
+  title: VP8 raw file format
+  ks-version: 0.7
+  license: CC0-1.0
+doc-ref: https://wiki.multimedia.cx/index.php/IVF
+seq:
+## header start
+  - id: magic1
+    contents: DKIF
+    doc: Magic Number of IVF Files
+  - id: version
+    type: u2
+    doc: This should be 0
+  - id: headerlen
+    type: u2
+    doc: Normally the header length is 32 byte
+  - id: codec
+    contents: VP80
+    doc: Name of the codec e.g. 'VP80' for VP8 
+  - id: width
+    type: u2
+    doc: The (initial) width of the video, every keyframe may change the resolution 
+  - id: height
+    type: u2
+    doc: The (initial) height of the video, every keyframe may change the resolution
+  - id: framerate
+    type: u4
+    doc: the (framerate * timescale) e.g. for 30 fps -> 30000 
+  - id: timescale
+    type: u4
+    doc: the timescale is a divider of the seconds (VPX is integer math only) mostly 1000
+  - id: framecount
+    type: u4
+    doc: the number of frames (if not a camera stream)
+  - id: unused
+    type: u4
+## header end
+
+## payload start
+  - id: image_data
+    type: blocks
+    repeat: expr
+    repeat-expr: framecount
+## payload end
+
+## type definitions    
+types:
+  # eithder a chunk or list
+  blocks:
+    seq:
+      - id: entries
+        type: block
+        #repeat: eos
+        #size-eos: true
+  block:
+    seq:
+      - id: framesize
+        doc: size of the frame data
+        type: u4
+      - id: timestamp
+        type: u8
+      - id: framedata
+        size: framesize
+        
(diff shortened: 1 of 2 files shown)


media/ivf.ksy
```diff
@@ -0,0 +1,67 @@
...
+## header start
+  - id: magic1
+    contents: DKIF
+    doc: Magic Number of IVF Files
+  - id: version
+    type: u2
+    doc: This should be 0
+  - id: headerlen
+    type: u2
+    doc: Normally the header length is 32 byte
```
GreyCat: Could you make this `len_header` to match [style guide](https://doc.kaitai.io/ksy_style_guide.html#attr-id) recommendation?

media/ivf.ksy
```diff
@@ -0,0 +1,67 @@
+meta:
+  id: ivf
+  file-extension: ivf
+  endian: le
+  title: VP8 raw file format
```
GreyCat: Please remove word "format" from the title — just "VP8 raw file" will work, as in certain contexts this "title" would be used as `$title format`.

media/ivf.ksy
```diff
@@ -0,0 +1,67 @@
...
+    doc: The (initial) height of the video, every keyframe may change the resolution
+  - id: framerate
+    type: u4
+    doc: the (framerate * timescale) e.g. for 30 fps -> 30000 
+  - id: timescale
+    type: u4
+    doc: the timescale is a divider of the seconds (VPX is integer math only) mostly 1000
+  - id: framecount
+    type: u4
+    doc: the number of frames (if not a camera stream)
```
GreyCat: `num_frames`?

media/ivf.ksy
```diff
@@ -0,0 +1,67 @@
+meta:
+  id: ivf
```
GreyCat: Formats repo have global IDs without any namespacing. Given that "ivf" is not a very popular format and there might be other formats identified as "ivf", may I suggest to rename this (and the file) to something more verbose, like `vp8_ivf`, as this is clearly a container for VP8?

media/ivf.ksy
```diff
@@ -0,0 +1,67 @@
...
+  # eithder a chunk or list
+  blocks:
+    seq:
+      - id: entries
+        type: block
+        #repeat: eos
+        #size-eos: true
+  block:
+    seq:
+      - id: framesize
```
GreyCat: `len_frame`?

frankpolte:
@GreyCat : I renamed and cleaned up the stuff. Even the file was renamed, so that we can add other codec ivf files e.g. vp9_ivf and so on. Right now it is for all codecs, but I plan for parsing vp8 specifically. Is there a way to reference the outer stuff from a more global ksy file?
Maybe a `ivf_mediacontainer.ksy` holds the main header stuff and `vp8_ivf.ksy`, `vp9_ivf.ksy` or `xyz_ivf.ksy` only parse the frame content but including  `ivf_mediacontainer.ksy`.

media/vp8_ivf.ksy
```diff
@@ -0,0 +1,64 @@
+meta:
+  id: vp8_ivf
+  file-extension: ivf
+  endian: le
+  title: VP8 raw file
+  ks-version: 0.7
+  license: CC0-1.0
+doc-ref: https://wiki.multimedia.cx/index.php/IVF
+seq:
+## header start
```
KOLANICH: I usually prefer the code be self-documenting. Should you remove these comments and use subtypes instead?

media/vp8_ivf.ksy
```diff
@@ -0,0 +1,64 @@
...
+  - id: magic1
+    contents: DKIF
+    doc: Magic Number of IVF Files
+  - id: version
+    type: u2
+    doc: This should be 0
+  - id: len_header
+    type: u2
+    doc: Normally the header length is 32 byte
+  - id: codec
```
KOLANICH: I don't think it is meant to be hardcoded.

GreyCat:
> Is there a way to reference the outer stuff from a more global ksy file?

Sure, we have  [imports mechanism](http://doc.kaitai.io/user_guide.html#_importing_types_from_other_files) for that.

I guess everything looks good now, so I'm merging it. Thanks again!

PR #111, opened 2018-11-08
Add 'gvar' support to ttf.

bungeman:


font/ttf.ksy
```diff
@@ -242,6 +242,112 @@ types:
               - id: repeat_value
                 type: u1
                 if: repeat
+  gvar:
+    doc: >
+      gvar - Glyph Variations Table
+      This table provides all of the variation data that describe how TrueType glyph outlines in a 'glyf' table change across the font’s variation space.
+    seq:
+      - id: major_version
```
KOLANICH: Use `type: fixed` (I don't know why it is not called `version`) instead of these 2 fields.

font/ttf.ksy
```diff
@@ -242,6 +242,112 @@ types:
...
+        type: u4
+      - id: glyph_count
+        type: u2
+      - id: flags
+        type: u2
+      - id: glyph_variation_data_array_offset
+        type: u4
+      - id: glyph_variation_data_offsets
+        type:
+          switch-on: flags & 0x01
```
KOLANICH: `flags` should be a a type of `b1`

font/ttf.ksy
```diff
@@ -242,6 +242,112 @@ types:
...
+        seq:
+          - id: coordinates
+            type: f2dot14
+            repeat: expr
+            repeat-expr: _parent.axis_count
+      tuple_variation_header:
+        seq:
+          - id: variation_data_size
+            type: u2
+          - id: tuple_index
```
KOLANICH: This should be a bit-type too.

font/ttf.ksy
```diff
@@ -242,6 +242,112 @@ types:
...
+            if: _io.size > 0
+          - id: tuple_variation_headers
+            type: tuple_variation_header
+            parent: _parent
+            repeat: expr
+            repeat-expr: tuple_variation_count & 0x0FFF
+            if: _io.size > 0
+        instances:
+          #The serialized data interpretation depends on each of the tuple_variation_header.tuple_index flags
+          #https://docs.microsoft.com/en-us/typography/opentype/spec/otvarcommonformats#a-nameserializeddata-idserializeddataserialized-dataa
```
KOLANICH: use `doc` and `-doc-ref`

font/ttf.ksy
bungeman: This layout may be bitwise consistent with 'fixed' but it isn't spec'ed that way. The specification is not 'fixed' (as is the case with the version many other tables) but two uint16. I think this may have even been done on purpose since when this table was defined the 'fixed' versions in other tables were often being serialized incorrectly due to using 'fixed'.

font/ttf.ksy
bungeman: flags is defined as a uint16, are you requesting an array of b1? This would actually look rather strange to anyone familiar with the format I think, since this is actually a big-endian value and the specification consistently refers to flag masks by value.

font/ttf.ksy
bungeman: I'm not sure how to do that, there are 4 bits of flags and the 12 other bits are a number (this isn't just a bitfield). Those could be value instances I suppose. I'm not currently using the number here, but it should be used to point to the correct shared tuple when the peak tuple is implied.

font/ttf.ksy
bungeman: This was more an implementation comment as to why this isn't complete, I suppose it could end up in a doc for now.

font/ttf.ksy
bungeman: I changed the flags to a set of reserved flags and named this flag, removing all of the bit masking in the process.

font/ttf.ksy
bungeman: Done

font/ttf.ksy
bungeman: Changed to 4 bit flags and 12 bit index. Removed the bit masking in the process.

font/ttf.ksy
```diff
@@ -242,6 +242,128 @@ types:
...
+        type: u2
+      - id: shared_tuple_count
+        type: u2
+      - id: shared_tuples_offset
+        type: u4
+      - id: glyph_count
+        type: u2
+      - id: reserved_flags
+        type: b15
+      - id: long_offsets
```
KOLANICH: I usually put them into a separaty type of even count of bytes, if it is possible. This makes it easier to calculate offsets in mind and  also makes it easier to edit the type. And don't forget about endianness, it's not handled automatically for now, so if someone wanted `be` it'd be easier to handle that case if you moved the flags into a separate subtype.

PR #114, opened 2018-11-19
Add data descriptor parsing to Zip

koczkatamas:


GreyCat:
Sorry for this merge taking so long. I've somehow missed it and never realized that it was such a small and tidy change. Merged it now.

PR #115, opened 2018-11-19
Constant Databases

Kerrigan29a:
CDB file format

### DIFF
--- a/database/cdb.ksy
+++ b/database/cdb.ksy
@@ -10,14 +10,25 @@ seq:
     type: pointer
     repeat: expr
     repeat-expr: 256
+    doc: |
+      Each of the 256 initial pointers states a position and a length. The
+      position is the starting byte position of the hash table. The length
+      is the number of slots in the hash table.
   - id: records
     type: record
     repeat: until
     repeat-until: _io.pos == pointers[0].pointer_position
+    doc: |
+      Records are stored sequentially, without special alignment. A record
+      states a key length, a data length, the key, and the data.
   - id: hash_tables
     type: hash_table(_index)
     repeat: expr
     repeat-expr: 256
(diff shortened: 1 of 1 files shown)


database/cdb.ksy
```diff
@@ -0,0 +1,54 @@
+meta:
+  id: cdb
+  file-extension: cdb
+  endian: le
+  encoding: ASCII
+  license: CC0-1.0
+doc-ref: https://cr.yp.to/cdb/cdb.txt
```
KOLANICH: it'd be nice if you added `doc`

Kerrigan29a: done

KOLANICH: I'm not sure that djb cannot sue you (or send a DMCA takedown against this repo) for the strange fact that some parts of this spec match literaly the text from his website.

Kerrigan29a: Of course the documentation parts match literally the original specification. I'm just quoting the original words with the mandatory reference to the original source. The problem would be to use the original documentation without citing the source.

Kerrigan29a: I'm ok with that

Kerrigan29a: https://twitter.com/drescalada/status/1065185136327700480

database/cdb.ksy
```diff
@@ -0,0 +1,76 @@
+meta:
+  id: cdb
+  file-extension: cdb
+  endian: le
+  encoding: ASCII
```
GreyCat: There are no strings in this format, so there's nothing to declare encoding for.

Kerrigan29a: In 'record' type. Lines 57 and 60

PR #116, opened 2018-11-20
Add rekordbox file format specs.

brunchboy:
These have been proven to work in the context of my Beat Link Trigger
project, enabling it to retrieve the database over NFS, parse it, and
extract all the track metadata it needs even when it is impossible to
connect to the database server running on the players because there
are four of them in use.


database/pdb.ksy
```diff
@@ -0,0 +1,961 @@
+meta:
+  id: pdb_file
```
KOLANICH: Should you use `rekordbox_pdb`?

database/pdb.ksy
```diff
@@ -0,0 +1,961 @@
...
+
+  The majority of the work in reverse-engineering this format was
+  performed by @henrybetts and @flesniak, for which I am hugely
+  grateful. @GreyCat helped me learn the intricacies (and best
+  practices) of Kaitai far faster than I would have managed on my own.
+
+doc-ref: https://github.com/Deep-Symmetry/crate-digger/blob/master/doc/Analysis.pdf
+
+seq:
+  - contents: [0, 0, 0, 0]
```
KOLANICH: Is it a signature or what is it?

database/pdb.ksy
```diff
@@ -0,0 +1,961 @@
...
+    seq:
+      - id: index
+        type: u4
+        doc: |
+          Identifies the desired page number.
+    instances:
+      body:
+        doc: |
+          When referenced, loads the specified page and parses its
+          contents appropriately for the type of data it contains.
```
KOLANICH: Nice approach in short run (`instance`s are lazy, `seq`s are not), but keep in mind that in long run (when we gonna have everything lazy, https://github.com/kaitai-io/kaitai_struct/issues/65 ) we may want to replace it with indexing of array of pages for better clarity.

database/pdb.ksy
```diff
@@ -0,0 +1,961 @@
...
+        doc: |
+          Identifies the type of information stored in the rows of this page.
+      - id: next_page
+        doc: |
+          Index of the next page containing this type of rows. Points past
+          the end of the file if there are no more.
+        type: page_ref
+      - type: u4
+        doc: |
+          @flesniak said: "sequence number (0->1: 8->13, 1->2: 22, 2->3: 27)"
```
KOLANICH: Could you get this field an `id` and a better `doc`? I don't understand currently (without reading the `doc-ref`) how 6his field is related to the mentioned mapping.

database/pdb.ksy
```diff
@@ -0,0 +1,961 @@
...
+        doc: |
+          Holds the value used for `num_rows` (see below) unless
+          `num_rows_large` is larger (but not equal to `0x1fff`). This
+          seems like some strange mechanism to deal with the fact that
+          lots of tiny entries, such as are found in the
+          `playlist_entries` table, are too big to count with a single
+          byte. But why not just always use `num_rows_large`, then?
+      - type: u1
+        doc: |
+          @flesniak said: "a bitmask (1st track: 32)"
```
KOLANICH: if it is a bitmask, it should he a type of `b1`s.

database/pdb.ksy
brunchboy: No, I can’t. There is no official documentation.

database/pdb.ksy
brunchboy: Thanks, I will look in to that.

database/pdb.ksy
KOLANICH: >There is no official documentation.

It doesn't matter, you can call it like `unkn` and put into the docs your guesses and observations.

database/pdb.ksy
brunchboy: That is an interesting discussion, when might such a change happen? I may make some changes to the mappings as more things are discovered, for example if we can figure out the index tables, which would make laziness even nicer.

database/pdb.ksy
brunchboy: I don’t have guesses or observations about the ones I have unlabeled, and the style guide said to not give IDs to things that are unknown, which is why I took this approach. We might be able to get more ideas from @flesniak whose [Python implementation](https://github.com/flesniak/python-prodj-link) was my main source of information, but these mappings are a small part of a large project, and I have other more pressing user needs to focus on than fleshing out details beyond the ones I need for supporting stage shows, so to be honest I am not likely to have time to pursue such details. I submitted what I know as a courtesy to the KSC project since it might be useful to others with similar needs, but if this is not thorough enough, you can delete the merge request. People looking for such capabilities are frankly more likely to find my projects before they find KSC.

database/pdb.ksy
brunchboy: We are not certain that rekordbox is the  only program that creates files in this format. It is used by a number of embedded databases, evidently.

brunchboy: Although, now that I think about it, only rekordbox would use this precise set of table and column structures, so yes, this would be a good change, I will make it. Thanks!

database/pdb.ksy
brunchboy: This is unknown.

brunchboy: In fact, it may be best to just map it as a 4-byte integer, but this way I will get a parse failure if we see something else there, which is interesting information, but not to my users. So perhaps I will change it.

database/pdb.ksy
brunchboy: I’m sorry, but what? I just scoured the [KSY reference](http://doc.kaitai.io/ksy_reference.html#primitive-data-types) and can find no mention of single bit values or the `b` type indicator.

database/pdb.ksy
brunchboy: All right, this is done; I also renamed `anlz_file` to `rekordbox_anlz` and updated my API documentation and Beat Link’s use of it.

database/pdb.ksy
brunchboy: I have also made this change.

database/pdb.ksy
brunchboy: Ok, thanks for letting me know about the future potential, but I think we should leave this as it is for now.

brunchboy:
This is as far as I am able to take these files at the moment. You are welcome to add them if you like, and it will not hurt my feelings if you decide they are not useful, because they are incredibly useful to me. 😄 Thanks in any case for your excellent tools!

database/pdb.ksy
```diff
@@ -0,0 +1,964 @@
+meta:
+  id: rekordbox_pdb
```
KOLANICH: file name should match its `id`

brunchboy: If this is true, it should be stated in the style guide.

KOLANICH: It is, as far as I remember.

brunchboy: I searched the whole style guide before posting my comment, and could not find anything along those lines. I will rename the files when I have a few minutes though.

brunchboy: All right, I have renamed the files and submitted a PR to the style guide to help future me avoid making the same mistake. 😄 Thanks. I still need to fix something in the `rekordbox_anlz.ksy` file to avoid Java exceptions when the file contains a `fourcc` that we have not previously encountered, as I am currently discussing in the gitter channel.

brunchboy: Ok, I have revised the mapping to stop trying to use enums until the underlying compiler issue is resolved.

GreyCat: Any chance we can do `devicesql_pdb` here, as we know that it is not exactly specific to rekordbox?

brunchboy: There was another thread on this topic, @GreyCat. The filename started out not being specific to rekordbox, band @KOLANICH suggested the current rename. I was hesitant at first but ended up agreeing because, while there are certainly aspects of the format which are generally applicable to other DeviceSQL implementations, the specific column types which are used by this mapping (e.g. tracks, artists, playlists, etc.) are clearly used only by rekordbox.

GreyCat: Ok, I agree with `rekordbox_pdb`, but please rename the file as `database/rekordbox_pdb.ksy` then?

database/pdb.ksy
GreyCat: I'll skip this so far. There are *way* too many obstacles on that path — i.e. alignment issues, eventual performance issues (allocating an ArrayList of booleans vs just a single 16-bit integer), etc.

database/pdb.ksy
```diff
@@ -0,0 +1,964 @@
...
+      - type: u2
+        doc: |
+          @flesniak said: "1004 for strange blocks, 0 otherwise"
+      - type: u2
+        doc: |
+          @flesniak said: "always 0 except 1 for history pages, num
+          entries for strange pages?"
+      - id: heap
+        size-eos: true
+        if: heap_pos < 0  # never true, but stores pos
```
GreyCat: `if: false` would be just as good

database/pdb.ksy
KOLANICH: >while there are certainly aspects of the format which are generally applicable to other DeviceSQL implementations

Can they be moved into a separate ksy and used via import mechanisms?

database/pdb.ksy
brunchboy: I made that change.

KOLANICH:
BTW, you can rewrite history and ```--amend``` changes to a single commit.

brunchboy:
Only if the commit hasn’t been pushed anywhere, of course. I do that sometimes. `magit` makes that sort of thing very easy in Emacs. 😄

database/rekordbox_pdb.ksy
```diff
@@ -375,8 +376,10 @@ types:
     doc: |
       A row that holds an artist name and ID.
     seq:
-      - id: magic
-        contents: [0x60, 0x00]
+      - type: u2
+        doc: |
+          Some kind of magic word? Usually 0x60, 0x00 but have seen
+          0x64, 0x00.
```
KOLANICH: Can it be format version?

database/rekordbox_pdb.ksy
```diff
@@ -390,19 +390,34 @@ types:
...
         type: device_sql_string
+        if: subtype == 0x60
         pos: _parent.row_base + ofs_name
         doc: |
           The name of this artist.
         -webide-parse-mode: eager
+      ofs_long_name:
+        type: u2
+        if: subtype == 0x64
+        pos: _parent.row_base + 0x0a
```
KOLANICH:
```yaml
    ofs_real_name:
        value: _parent.row_base + (subtype == 0x64 ? ofs_long_name : ofs_name)
    name:
        type: device_sql_string
        if: subtype == 0x60
        pos: ofs_real_name
```

KOLANICH: Also pleas follow the conv3ntion `pos [io] type desc`

brunchboy: Around the time you were writing this, I was at my lunchtime swim, and in the pool is often when I review recent code and make improvements and optimizations, and I came up with the same idea, it will make dealing with the value much easier. Thanks! I will fix the order of the fields as well.

brunchboy: Rats, when I try that I am getting a KSC error in the generate phase: `mapping values are not allowed here  in 'reader', line 410, column 66:  ... (subtype == 0x64)? ofs_name_far : ofs_name_near)` (column 66 is right after the `:`. Any ideas?

brunchboy: Ah sorry, that’s a FAQ, need to escape it as a string for the silly YAML format.

brunchboy: I ended up doing it a slightly different way, but it has the same net effect, the struct is easier to understand and _much_ easier to use; the name is always accessed using the same instance regardless of how it is actually stored.
https://github.com/brunchboy/kaitai_struct_formats/commit/a2898a03a65db4f1f7e19f7e3bf84d16ba338f65

database/rekordbox_pdb.ksy
```diff
@@ -391,33 +392,26 @@ types:
...
+      ofs_name_far:
+        pos: _parent.row_base + 0x0a
+        type: u2
+        doc: |
+          For names that might be further than 0xff bytes from the
+          start of this row, this holds a two-byte offset, and is
+          signalled by the subtype value.
+        if: subtype == 0x64
       name:
+        pos: '_parent.row_base + (subtype == 0x64? ofs_name_far : ofs_name_near)'
```
KOLANICH:
```python
ofs_name_far:
        pos: _parent.row_base + 0x0a
```

```
pos: '_parent.row_base + (subtype == 0x64? (_parent.row_base + 0x0a) : ofs_name_near)'

was it intended?

brunchboy: Yes, that is really the weird layout that they have set up.

brunchboy: Note that your substitution is not quite correct. In the case where it is a “far” string the offset from `_parent.row_base` is the two-byte value found at `_parent.row_base + 0x0a`. So row-base gets added to the value found there, it is not added twice to the resulting offset.

brunchboy:
I just made some small documentation changes based on things I noticed while finishing the [detailed explanation article](
https://github.com/Deep-Symmetry/crate-digger/blob/master/doc/Analysis.pdf). These mappings have been used successfully to run shows by a number of people over the past few months, so if you would like to include them in your collection, they are probably ready. If you are not interested in them I can close the pull request as well; just let me know.

KOLANICH:
>This value seems to be redundant, because it can be calculated by dividing the offset
of the start of the page by $len\_page$, but perhaps it serves as a
sanity check.

Have you tried to replace these values and check the consequences?

>_root.len_page * index

I think that there is no reason to link pages into a list if they are organized into an array. So it may be that array is just an impl detail of memory allocation. I mean that you should check if moving the pages and their `page_ref`s don't spoil the format. If it is the case, the list may be just a strange key-value storage with `index` as a key and looking up a page by its index is crawling the list and checking the index of every page.

brunchboy:
I’m sorry, I am again not sure what you mean. I can’t move pages around, I can only examine them as they are given to me. The code that creates them is in third-party proprietary software designed to run in low-power processors. The pages of a given table are linked together because there might be pages of other tables interspersed between them.

In all files that any of my users have seen the index is equal to the array index of the page. There are many examples of far worse redundancy and inconsistency in the format.

KOLANICH:
Thank you for the clarification, now I understand your spec better.

PR #119, opened 2018-12-16
asn1: handle also 1-byte lengths

BenGardiner:
There was already handling of two-byte lengths (`0x82`); extend the handling to include 1 byte lengths also (`0x81`)


GreyCat:
Thanks!

PR #121, opened 2019-01-07
Added rtp protocol support

pedoc:


network/rtp_packet.ksy
```diff
@@ -0,0 +1,85 @@
...
+    26: jpeg
+    27: unassigned6
+    28: nv
+    29: unassigned7
+    30: unassigned8
+    31: h261
+    32: mpv
+    33: mp2t
+    34: h263
+    96: mpeg_ps
```
GreyCat: End of file newline missing here.

network/rtp_packet.ksy
```diff
@@ -0,0 +1,85 @@
...
+    rfc: 3550
+    wikidata: https://en.wikipedia.org/wiki/Real-time_Transport_Protocol
+  license: Unlicense
+  endian: be
+
+doc: |
+  The Real-time Transport Protocol (RTP) is a network protocol for delivering audio and video over IP networks. RTP is used in communication and entertainment systems that involve streaming media, such as telephony, video teleconference applications including WebRTC, television services and web-based push-to-talk features.
+  RTP typically runs over User Datagram Protocol (UDP). RTP is used in conjunction with the RTP Control Protocol (RTCP). While RTP carries the media streams (e.g., audio and video), RTCP is used to monitor transmission statistics and quality of service (QoS) and aids synchronization of multiple streams. RTP is one of the technical foundations of Voice over IP and in this context is often used in conjunction with a signaling protocol such as the Session Initiation Protocol (SIP) which establishes connections across the network.
+  RTP was developed by the Audio-Video Transport Working Group of the Internet Engineering Task Force (IETF) and first published in 1996 as RFC 1889, superseded by RFC 3550 in 2003.
+
```
GreyCat: 1. This description seems to be copy-pasted from Wikipedia, it's licensed as [https://en.wikipedia.org/wiki/Wikipedia:Text_of_Creative_Commons_Attribution-ShareAlike_3.0_Unported_License](CC BY-SA 3.0 Unported), you either need to mention that in `meta/license`, or rewrite this text to be able to use `Unlicense`.

2. Please don't use such long lines, as they will be literally pasted into docstring comments this way. It's better to wrap it around ~80th column.

network/rtp_packet.ksy
```diff
@@ -0,0 +1,85 @@
...
+    type: u2
+  - id: timestamp
+    type: u4
+  - id: ssrc
+    type: u4
+  - id: header_extension
+    type: header_extention
+    if: extension==true
+  - id: data
+    size: _io.size-12
```
GreyCat: 1. Please use spaces in such expressions, i.e. `_io.size - 12`, not `_io.size-12`
2. This seems to just consume everything up to end of stream anyway, why don't just use `size-eos: true`?

network/rtp_packet.ksy
```diff
@@ -0,0 +1,85 @@
...
+    enum: payload_type_enum
+  - id: sequence_number
+    type: u2
+  - id: timestamp
+    type: u4
+  - id: ssrc
+    type: u4
+  - id: header_extension
+    type: header_extention
+    if: extension==true
```
GreyCat: `if: extension` will be sufficient. Or, better yet, rename that attribute to `has_extension`, and use `if: has_extension`?

network/rtp_packet.ksy
```diff
@@ -0,0 +1,85 @@
...
+  - id: data
+    size: _io.size-12
+types:
+  header_extention:
+    seq:
+      - id: id
+        type: u2
+      - id: length
+        type: u2
```
GreyCat: I actually wonder if these should be followed up by data of this length?

network/rtp_packet.ksy
pedoc: @GreyCat 
thanks for reply,i will be fixed.
sorry my bad :)

network/rtp_packet.ksy
GreyCat: My point is that right now `size-eos: true` and `size: _io.size - 12` is exactly the same. Even if there is some padding appended after `data`, you're still capturing it in `data` right now.

GreyCat:
I guess most of the things were addressed, I'm merging this in. Thanks, @pedoc!

network/rtp_packet.ksy
pedoc: Yes, but I originally wanted to remove the padded data.

GreyCat: Take a look, I think I've got it nailed in this commit: d8aaa74a8e66bfab937fc1666ba3d7889b4885a2

PR #123, opened 2019-02-13
Added support for the DIGITAL TERRAIN ELEVATION DATA (DTED) file format

djkoberstein:


geospatial/dted.ksy
```diff
@@ -0,0 +1,399 @@
+meta:
+  id: dted
+  title: DIGITAL TERRAIN ELEVATION DATA (DTED)
+  file-extension:
+    - dt0
+    - dt1
+    - dt2
+  endian: be
```
KOLANICH: license is missing

geospatial/dted.ksy
```diff
@@ -0,0 +1,399 @@
...
+    type: acc_header
+  - id: records
+    type: data_record
+    repeat: expr
+    repeat-expr: uhl.number_of_longitude_lines.value
+types:
+  uhl_header:
+    seq:
+      - id: recognition_sentinel
+        contents: UHL
```
KOLANICH: Could you add `doc`s to everything?

geospatial/dted.ksy
```diff
@@ -0,0 +1,399 @@
...
+      - id: seconds
+        type: str
+        size: 4
+        encoding: ASCII
+      - id: hemisphere
+        type: str
+        size: 1
+        encoding: ASCII
+    instances:
+      degree_decimal:
```
KOLANICH: just `value`?

geospatial/dted.ksy
```diff
@@ -0,0 +1,399 @@
...
+        value: (degrees.to_i + minutes.to_i/60.0 + seconds.to_i/3600.0) * (hemisphere == 'N'? 1 : -1)
+  angle_dddmmss_sh:
```
KOLANICH: What does `_sh` mean?

geospatial/dted.ksy
```diff
@@ -0,0 +1,399 @@
...
+      - id: absolute_vertical_accuracy
+        type: ascii_int_ssss
+      - id: relative_horizontal_accuracy_of_product_in_meters
+        type: ascii_int_ssss
+      - id: relative_vertical_accuracy_or_product_in_meters
+        type: ascii_int_ssss
+      - id: reserved_for_future_use1
+        type: str
+        size: 4
+        encoding: ASCII
```
KOLANICH: Encoding can be moved into `meta`

geospatial/dted.ksy
```diff
@@ -0,0 +1,399 @@
...
+  data_record_elevation:
+    seq:
+      - id: is_negative
+        type: b1
+      - id: value
+        type: b15
+    instances:
+      meters:
+        value: value * (is_negative? -1 : 1)
+  ascii_int_ssss:
```
KOLANICH: either just `ascii_int` or `ascii_int(4)`. And maybe not `ascii` but `text`?

geospatial/dted.ksy
djkoberstein: The added docs explain what it is, but it represents the hemisphere of the degrees minutes seconds representation.

geospatial/dted.ksy
djkoberstein: I'll add a license to this.

geospatial/dted.ksy
djkoberstein: Docs will be added

geospatial/dted.ksy
djkoberstein: The purpose of this type is to do the conversion between the ASCII representation and Degree Decimal. It's a convenience thing.  This is because any elevation attribute is meaningless unless it is programmatically calculated what the latitude and longitude of that elevation point is.

geospatial/dted.ksy
djkoberstein: Cool! I didn't realize you could do that.

geospatial/dted.ksy
djkoberstein: the spec explicitly mentions that the text is ascii so I'm inclined more towards ascii_int. I'll update accordingly.

geospatial/dted.ksy
djkoberstein: I don't know if this is right, but even single line comments I placed on a newline because it makes the web ids syntax highlighting easier to read. For example
```
      - id: longitude_interval_seconds
        type: ascii_int
        doc: |
          Longitude data interval in tenths of seconds.
```
instead of
```
      - id: longitude_interval_seconds
        type: ascii_int
        doc: Longitude data interval in tenths of seconds.
```

KOLANICH: IMHO: completely OK.

geospatial/dted.ksy
KOLANICH: I mean if that `instance` should be named just `value`

geospatial/dted.ksy
```diff
@@ -0,0 +1,596 @@
...
+          a full one-degree cell. ( e.g. 1201 for DTED1, 3601.for DTED2).
+      - id: multiple_accuracy
+        type: str
+        size: 1
+        encoding: ASCII
+        doc: |
+          0 - Single |
+          1 - Multiple
+      - id: reserved
+        type: str
```
KOLANICH: If it is reserved, why str?

geospatial/dted.ksy
djkoberstein: The actual format of the field this type represents is DDDMMSS.SH where:
* DDD is 000 - 180 degrees.
* MM is 00 - 59 minute.
* SS is 00 - 59 seconds.
* H is the hemisphere, N,S,E or W.

In that case is it correct of me to include this formatting into the type name?

geospatial/dted.ksy
```diff
@@ -0,0 +1,596 @@
...
+        type: coordinate_pair_ddmmss_sh
+        doc: |
+          Latitude of origin of data— leading zero for values less than 10; H |
+          is the hemisphere of the data (DDMMSS.SH). |
+          Longitude of origin of data— leading zeroes for values less than |
+          100; H is the hemisphere of the data (DDDMMSS.SH).
+      - id:  sw_corner_of_data
+        type: coordinate_pair_ddmmssh
+        doc: |
+          Latitude of SW corner of data, bounding rectangle— leading zero for |
```
KOLANICH: I wonder if part of this docs should be moved into type doc.

geospatial/dted.ksy
```diff
@@ -0,0 +1,596 @@
...
+          for values less than 100; H is the hemisphere of the data.
+      - id:  clockwise_orientation_angle
+        type: angle_dddmmss_sh
+        doc: |
+          Clockwise orientation angle of data with respect to true North. |
+          (Will usually be all zeros.)
+      - id:  latitude_interval
+        type: ascii_int
+        doc: |
+          Latitude interval in tenths of seconds between rows of elevation |
```
KOLANICH: The same, looks like it also can be moved into type doc.

geospatial/dted.ksy
```diff
@@ -0,0 +1,596 @@
...
+          Number of Longitude lines. For magnetic tape, this is the actual |
+          count of the number of longitude points (columns that contain data). |
+          For CD-ROM, this is the count of the number of longitude points in a |
+          full one-degree cell. The count is based on the level of DTED and |
+          the latitude zone of the cell. (See Table II and III) (0000-9999). 
+      - id:  partial_cell_indicator
+        type: str
+        size: 2
+        doc: |
+          Partial Cell Indicator
```
KOLANICH: I propose to rename it and create an instance getting actual coverage.

geospatial/dted.ksy
djkoberstein: Done.  I'll add a doc tag which specifies that it is in the degree decimal format.

geospatial/dted.ksy
```diff
@@ -0,0 +1,596 @@
...
+          Reserved for free text comments. (Free text or Blank filled.)
+  acc_header:
+    doc: |
+      Accuracy Description (ACC) Record. The ACC is used for magnetic tape and |
+      CD-ROM. Fixed Length = 2700 ASCII Characters |
+      Note: *If Product has subregional accuracies, the overall accuracy of |
+      the product will be the worst accuracy.
+    seq:
+      - id: recognition_sentinel
+        contents: ACC
```
KOLANICH: I wonder if it is a marker of block type, so if a generic block type should be created and specific block type is selected with
```yaml
type:
    switch-on: recognition_sentinel
    cases:
      '"ACC"': acc_header
```

geospatial/dted.ksy
KOLANICH: >In that case is it correct of me to include this formatting into the type name?

It is, but better to avoid it, if there is no ambiguity without it.

geospatial/dted.ksy
```diff
@@ -0,0 +1,603 @@
...
+        value: value * (is_negative? -1 : 1)
+  ascii_int:
+    seq:
+      - id: text
+        type: str
+        size: 4
+    instances:
+      value:
+        value: text.to_i
+  coordinate_pair_ddmmss_sh:
```
KOLANICH: Is `Not Available (NA)` properly handled?

geospatial/dted.ksy
djkoberstein: It is implied by the spec as both the DSI and ACC records specify that the entire section is ASCII.  The UHL record as the spec refers to them is a similar in nature and all fields that have been defined are in ASCII.  The only record type which is different is  the Data Record which follow the metadata inside the UHL, DSI, and ACC records.  However if you recommend that I change it I can change it.

geospatial/dted.ksy
djkoberstein: That is a good idea. The documentation will be different than what is in the spec but since I chose to  add this type instead of repeating my self updating the documentation to reflect that I think would be a good idea.

geospatial/dted.ksy
djkoberstein: An interesting suggestion.  It certainly could be implemented that way.  The spec doesn't seem to allow for that as the spec reads as follow (page 13)
```
						Record starts at byte
a. User Header Label (UHL: 80 bytes)			1
b. Data Set Identification Record (DSI: 648 bytes)	81
c. Accuracy Record (ACC: 2700 bytes)			729
d. Data Records (Each data record is 2414 bytes)*	3429
							5843
							8257
							etc
```

geospatial/dted.ksy
djkoberstein: Ok I'll rename the original field to partial_cell_indicator_text and add a calculated instance field which looks as follows
```
    instances:
      partial_cell_indicator:
        value: (partial_cell_indicator_text.to_i == 0) ? | 100.0 : partial_cell_indicator_text.to_i
        doc: |
          Partial Cell Indicator (% of data coverage)
```

geospatial/dted.ksy
```diff
@@ -0,0 +1,596 @@
...
+        size: 2
+      - id: seconds
+        type: str
+        size: 4
+      - id: hemisphere
+        type: str
+        size: 1
+    instances:
+      degree_decimal:
+        value: (degrees.to_i + minutes.to_i/60.0 + seconds.to_i/3600.0) * (hemisphere == 'N'? 1 : -1)
```
djkoberstein: On this field the `seconds.to_i` is technically incorrect.  The underlying text is actually a floating -point number.   However I didn't see any methods in the docs to be able to parse the text as such.  I was expecting something like `seconds.to_f`.  Unless there is a better way I'll change it as follows:
```
  angle_dddmmss_sh:
    doc: |
      Typed used created to convert to degree decimal representation
    seq:
      - id: degrees
        type: str
        size: 3
      - id: minutes
        type: str
        size: 2
      - id: seconds
        type: str
        size: 2
      - id: decimal_point
        type: str
        size: 1
      - id: tenths_of_seconds
        type: str
        size: 1
      - id: hemisphere
        type: str
        size: 1
    instances:
      value:
        value: (degrees.to_i + minutes.to_i/60.0 + seconds.to_i/3600.0 + tenths_of_seconds.to_i/36000.0) * (hemisphere == 'N'? 1 : -1)
        doc: |
          Degree decimal floating point representation     
```

geospatial/dted.ksy
djkoberstein: Personally I think it is fairly clear with the docs what the type actually represents. An alternative would be something like angle_tenth_of_seconds, and angle_even_seconds.

geospatial/dted.ksy
djkoberstein: No it is not.  I'm not sure how to handle it.  How about something like this:
```
  ascii_int_maybe:
    seq:
      - id: text
        type: str
        size: 4
    instances:
      has_value:
        value: text != "NA"
      value:
        value: text.to_i
```

KOLANICH:
```yaml
value:
  value: text.to_i
  if: has_value
```

geospatial/dted.ksy
KOLANICH: `partial_cell_indicator_text.to_i` is calculated twice. It may make sense to create an instance for that.

geospatial/dted.ksy
KOLANICH: https://github.com/kaitai-io/kaitai_struct/issues/524

geospatial/dted.ksy
```diff
@@ -0,0 +1,626 @@
...
+        size: 2
+      - id: hemisphere
+        type: str
+        size: 1
+    instances:
+      value:
+        value: (degrees.to_i + minutes.to_i/60.0 + seconds.to_i/3600.0) * (hemisphere == 'N'? 1 : -1)
+        doc: |
+          Degree decimal floating point representation
+  date_yymm:
```
KOLANICH: just `date`?

geospatial/dted.ksy
```diff
@@ -0,0 +1,626 @@
...
+    seq:
+      - id: year
+        type: str
```
KOLANICH: Is it an int?

geospatial/dted.ksy
djkoberstein: Cool until then I'll leave it as shown above.

geospatial/dted.ksy
djkoberstein: I'll update it to be named just date as I think the format is always YYMM.

geospatial/dted.ksy
djkoberstein: It is an int encoded as ASCII.  As such I'll change it as follows:
```
  date:
    seq:
      - id: year_text
        type: str
        size: 2
      - id: month_text
        type: str
        size: 2
    instances:
      year:
        value: year_text.to_i
      month:
        value: month_text.to_i
```

geospatial/dted.ksy
djkoberstein: Done. I'll update to add the `if: has_value`

geospatial/dted.ksy
djkoberstein: What do you think?

geospatial/dted.ksy
KOLANICH: Why not to use the existing type, just parameterizing it with string length?

geospatial/dted.ksy
KOLANICH: IMHO: for `angle` fields it may make sense to leave as it is. The remark wqs about other fields having the same pattern, like `date_yymm`.

djkoberstein: The odd part is that each of the following types
```
angle_ddmmss_sh
angle_dddmmss_sh
angle_ddmmssh
angle_dddmmssh
```
is that they really represent the same thing, an angle.  The only difference is how the data is parsed. If I had exactly 1 way to parse that field then I would shorten it down to just "angle".  Or if I could have the same type parsed in 4 different ways then I would do that.  Actually looking at the docs, maybe this would be a good usage of a Parameteric type.  Perhaps something like this:
```
  angle:
    params:
      - id: len_degrees
        type: u1
      - id: has_decimal_point
        type: u1
    seq:
      - id: degrees
        type: str
        size: len_degrees
      - id: minutes
        type: str
        size: 2
      - id: seconds
        type: str
        size: 2
      - id: decimal_point
        type: str
        size: 1
        if: has_decimal_point == 1
      - id: tenths_of_seconds
        type: str
        size: 1
        if: has_decimal_point == 1
      - id: hemisphere
        type: str
        size: 1
    instances:
      value:
        value: (degrees.to_i + minutes.to_i/60.0 + seconds.to_i/3600.0 + tenths_of_seconds.to_i/36000.0) * (hemisphere == 'N'? 1 : -1)
        doc: |
          Degree decimal floating point representation
```

djkoberstein: Unfortunately this means that I need to upgrade to 0.8 and the online repl won't work this this type anymore.

geospatial/dted.ksy
```diff
@@ -34,19 +34,19 @@ types:
       - id: recognition_sentinel
         contents: UHL
         doc: |
-          Recognition sentinel
+          Recognition sentinel ("UHL")
```
KOLANICH: It is unneeded to put the strings into docs. The `ksy` file is docs itself.

geospatial/dted.ksy
```diff
@@ -486,68 +485,47 @@ types:
...
+        if: has_value
+  coordinate_pair_precise:
     doc: |
-      Typed used to make types consistent once parsed
+      Typed used group 2 angles to form a coordinate.  It is accurate to the |
+      tenth of a second.
     seq:
       - id: latitude
-        type: angle_ddmmss_sh
+        type: angle(2,1)
```
KOLANICH: 👍

geospatial/dted.ksy
djkoberstein: Ok

geospatial/dted.ksy
djkoberstein: Glad you like the change

djkoberstein:
Is there something else i am supposed to do to get this included?

geospatial/dted.ksy
```diff
@@ -0,0 +1,587 @@
...
+          latitude zone of the cell. (See Table I, II and III).
+      - id: number_of_latitude_points
+        type: ascii_int
+        doc: |
+          Count of the number of latitude points per longitude line for |
+          a full one-degree cell. ( e.g. 1201 for DTED1, 3601.for DTED2).
+      - id: multiple_accuracy
+        type: str
+        size: 1
+        encoding: ASCII
```
KOLANICH: seems to be unneeded.

geospatial/dted.ksy
```diff
@@ -0,0 +1,587 @@
...
+        doc: |
+          0 - Single |
```
KOLANICH: does it make sense to make it a enum?

geospatial/dted.ksy
```diff
@@ -0,0 +1,587 @@
...
+        doc: |
+          Reserved for future use. (Blank filled.)
+      - id: multiple_accuracy_outline_flag
+        type: str
+        size: 2
+        doc: |
+          Multiple Accuracy Outline Flag. |
+          00 = No accuracy subregions provided. |
+          02-09 = Number of accuracy subregions per 1° cell (maximum 9).
+      - id: subregion
```
KOLANICH: isn't count of this meant to depend on `multiple_accuracy_outline_flag`?

geospatial/dted.ksy
```diff
@@ -0,0 +1,587 @@
...
+        size: 1
+        doc: |
+          Reserved for NIMA use only.
+      - id: reserved_for_future_use2
+        type: str
+        size: 31
+        doc: |
+          Reserved for future use. (Blank filled.)
+      - id: multiple_accuracy_outline_flag
+        type: str
```
KOLANICH: shouldn't it be a number?

geospatial/dted.ksy
```diff
@@ -0,0 +1,587 @@
...
+        doc: |
+          Relative (Point-to-Point) Horizontal Accuracy of Sub region in |
+          meters. (0000-9999 or Not Available (NA))
+      - id: relative_vertical_accuracy_of_subregion_in_meters
+        type: ascii_int_maybe
+        doc: |
+          Relative (Point-to-Point) Vertical Accuracy of Subregion in meters. |
+          (0000-9999 or Not Available (NA))
+      - id: number_of_coordinates_in_accuracy_subregion_outline
+        type: str
```
KOLANICH: The same issuues.

geospatial/dted.ksy
```diff
@@ -0,0 +1,587 @@
...
+        size: 2
+    instances:
+      year:
+        value: year_text.to_i
+      month:
+        value: month_text.to_i
+enums:
+  security_classification_enum:
+    83: secrete
```
KOLANICH: `secret`?

PR #125, opened 2019-03-06
Add hashcat restore format

GH0st3rs:


GreyCat:
Thanks! Some top-level documentation would be nice, but still a solid contribution!

PR #126, opened 2019-03-08
Add the different variants of the Unix ar format

dgelessus:
The ar format is not properly standardized, so there are a few different variants. Each one has its own spec - all variants use the same basic structure, but the details are different enough that they can't be merged into a single spec very well.

Some parts of the ar format(s) don't translate very well to Kaitai (for example I had to use some substream and instance trickery to parse the name fields properly). If there's a cleaner or more efficient way to implement something, please let me know. (These are also the first larger Kaitai specs I've written, so there might be useful features that I don't know about 😃)

### DIFF
--- a/archive/ar_bsd.ksy
+++ b/archive/ar_bsd.ksy
@@ -154,5 +154,3 @@ types:
       An archive member's header and data.
       
       By default, modern ar implementations set the modification timestamp, user ID and group ID to 0 and the mode to 644 (octal), regardless of the file's original metadata, to make archive creation reproducible.
-      
-      Rarely, the modification timestamp, user ID, group ID and mode fields may be blank (only spaces). This is the case in particular for the '//' member (the long name list) of SysV archives.
--- a/archive/ar_bsd.ksy
+++ b/archive/ar_bsd.ksy
@@ -33,6 +33,31 @@ seq:
     repeat: eos
     doc: List of archive members. May be empty.
 types:
+  regular_member_name:
+    seq:
+      - id: name
+        terminator: 0x20
+        pad-right: 0x20
+        doc: The member name, right-padded with spaces.
+    doc: |
+      A regular (or "short") member name, stored directly in the name field.
+      
+      Note: Since regular names in BSD archives are terminated using spaces, file names that contain spaces cannot be stored as regular names. Such names must be stored as long names, even if they are not longer than 16 bytes.
+  long_member_name:
+    seq:
+      - id: magic
+        contents: '#1/'
+        doc: Indicates a long member name.
+      - id: name_size_dec
+        type: str
+        terminator: 0x20
+        pad-right: 0x20
+        doc: The size of the long member name in bytes, in ASCII decimal, right-padded with spaces.
+    instances:
+      name_size:
+        value: name_size_dec.to_i
+        doc: The size of the long member name in bytes, parsed as an integer.
+    doc: A long member name, stored at the start of the member's data.
   member_name:
     seq:
       - id: first_three_bytes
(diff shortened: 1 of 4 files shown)


KOLANICH:
How about `deb`?

KOLANICH:
1. I guess that different variants should go into `ar` dir.
2. should numbers stored as strings be parsed into numbers
3. types different only by strings lengths can be parametrized using `params`.
4. chunks the same for several files can be moved into separate files and imported.

dgelessus:
Thank you for the feedback 😃

> I guess that different variants should go into `ar` subdir.

Good idea, will do.

> types different only by strings lengths can be parametrized using `params`.

I assume you mean the space-padded number fields (`modified_timestamp_dec`, `user_id_dec`, etc.)? I could make a separate type for that, but that would create an extra nesting level, right? (`members[0].user_id_dec` would become `members[0].user_id_dec.text` for example.) Or does Kaitai have something like typedefs?

Actually on second thought a separate type for "space-padded number literal" might not be a bad idea, because I could move the string-to-integer conversion into the type as well.

> chunks the same for several files can be moved into separate files and imported.
> 
> If only a part of type can be parameterized and reused, it can ge moved into an own type, cannot it?

Again I'm guessing you mean the metadata part (modification time, user ID, etc.)? I don't know what else could be moved into a shared type/file, because of the subtle differences between the variants.

> Please learn to use `git commit --amend` and `git push --force` and squash all the commits into a single one.

Yeah, the commit history here is not very nice, I can clean it up a little once I've fixed the other issues. (I don't like squashing away history when I'm still working on the PR - it makes it hard for others to see what the previous comments/reviews were talking about.)

If this repo has it enabled, GitHub also lets you [squash when merging the PR](https://help.github.com/en/articles/configuring-commit-squashing-for-pull-requests). That way a squashed commit is merged into the repo, but the real history is kept in the PR.

>> The Debian package format (.deb) is also based on the ar format.
> 
> Which one? Could you add deb to extension lists of the relevant formats and remove this note from irrelevant ones?

Both variants are supported actually 😃 `deb`s don't require any "long file name" extensions (all file names are 15 bytes or shorter and don't contain any spaces), and trailing slashes are optional in `deb`s. (See the [`deb(5)`](https://manpages.debian.org/stretch/dpkg-dev/deb.5.en.html) man page.)

But you're right, perhaps the text shouldn't be copy-pasted into each variant. Maybe I should change the `ar_sysv` and `ar_bsd` descriptions to say "see the `ar_common` description for general info" and only explain what the variants do differently.

I wanted to try writing a `deb` spec as well - based on the man page, it looks doable. But if that doesn't work, I'll list the `deb` extension in the `ar` specs.

KOLANICH:
>I could make a separate type for that, but that would create an extra nesting level, right? 

Right. There is a proposal to fix it: https://github.com/kaitai-io/kaitai_struct/issues/88

>If this repo has it enabled, GitHub also lets you squash when merging the PR.

Thanks for the info.

>Again I'm guessing you mean the metadata part (modification time, user ID, etc.)?

Yes, I mean mostly it for now.

>I wanted to try writing a deb spec as well - based on the man page, it looks doable. But if that doesn't work, I'll list the deb extension in the ar specs.

I don't understand the purpose of writing a separate spec for deb, if its binary format is idenntical to `ar`s.

dgelessus:
> I don't understand the purpose of writing a separate spec for deb, if its binary format is idenntical to `ar`s.

The `deb` format is basically a subset of `ar` format with very specific requirements on what archive members can/must exist, how they are named, and in which order they are stored. I want to see if those requirements can be specified in a useful way using Kaitai. If the `deb` spec ends up being too complicated or not useful, I'll leave it be, but I want to give it a try at least.

KOLANICH:
https://github.com/kaitai-io/kaitai_struct/issues/81 may be useful for it when it is implemented.

archive/ar/member_metadata.ksy
```diff
@@ -0,0 +1,22 @@
+meta:
+  id: member_metadata
+  title: Unix ar archive member metadata
+  license: CC0-1.0
```
KOLANICH: Shouldn't it import `space_padded_number` too?

dgelessus: Yup, it should. Why does this even compile? This seems like a bug...

dgelessus:
[Here](https://github.com/dgelessus/ksf_stuff/blob/4f260081795dd2abd1ceb7c082b546ff3a54a8c0/archive/ar.py) is a very basic CLI tool to read ar archives using these specs. I mostly wrote it as a proof-of-concept to see if it's possible to write code that works generically with multiple specs. In this case it worked out very well - the differences between the format variants are almost completely abstracted away by the specs, and the application code only needs very few variant-specific branches.

This sort of generic code is currently only possible with dynamically-typed languages like Python. Similar code in statically-typed languages wouldn't compile, because the classes generated by ksc for each spec don't implement a common base class/interface - this would require support from ksc, see kaitai-io/kaitai_struct#314.

PR #129, opened 2019-03-21
little fixes

lwerdna:
the gif.ksy one looks like nothing is changed thru github, but those dashes were made of some strange 0xE2 byte which, after compilation, would end up in compiled files and the python interpreter would complain

the wmf.ksy one would compile and have an enum named "not" which is a reserved python keyword and thereful wouldn't run - changing the .ksy works but doesnt feel right, feel free to reject this one

### DIFF
--- a/image/wmf.ksy
+++ b/image/wmf.ksy
@@ -252,7 +252,7 @@ enums:
     0x0003: masknotpen
     0x0004: notcopypen
     0x0005: maskpennot
-    0x0006: not
+    0x0006: not_
     0x0007: xorpen
     0x0008: notmaskpen
     0x0009: maskpen
--- a/image/gif.ksy
+++ b/image/gif.ksy
@@ -18,7 +18,7 @@ doc: |
 
   GIF format allows encoding of palette-based images up to 256 colors
   (each of the colors can be chosen from a 24-bit RGB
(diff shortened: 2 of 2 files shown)


image/wmf.ksy
```diff
@@ -252,7 +252,7 @@ enums:
     0x0003: masknotpen
     0x0004: notcopypen
     0x0005: maskpennot
-    0x0006: not
+    0x0006: not_
```
GreyCat: This one won't really work well. This is currently a problem with compiler with allows that, but in reality, for languages that use lowerCamelCase, this `not_` will still be rendered as `not`, so, ideally, we should prohibit trailing / extra underscores in identifiers :(

Feel free to rename with some sort of prefix?

lwerdna: hehe, _not?

But I think a better solution is that the backend for each language has an identifier filter that transforms any reserved words into legal words.

GreyCat: `_not` would result in the same, and, moreover, we've reserved all `_` prefixed identifiers for internal KS stuff (like `_root`, `_parent`, `_index`, etc).

Yeah, https://github.com/kaitai-io/kaitai_struct/issues/90 would have helped...

PR #131, opened 2019-03-28
macOS '.DS_Store' files

SkypLabs:
Hello,

This PR aims to add a description for the macOS `.DS_Store` files generated by the Finder. You can test the description on these files: [ds_store_samples.zip](https://github.com/kaitai-io/kaitai_struct_formats/files/3015783/ds_store_samples.zip)

There are two things I would like to improve though:

* The position of certain instances need to be calculated according to `alignment_header` but I didn't succeed to make it work. I had to manually add 4 to the result as in this example: `buddy_allocator_header.ofs_bookkeeping_info_block + 4`
* When I try to use `_index` lines 65 and 66, I get the following error: `undefined TypeError: this.directoryEntries[i] is undefined`. I think it is the same issue as described in https://github.com/kaitai-io/kaitai_struct/issues/436.

Could you please give me a hand on these two issues?

Thank you in advance.

### DIFF
--- a/macos/ds_store.ksy
+++ b/macos/ds_store.ksy
@@ -17,9 +17,9 @@ seq:
     type: buddy_allocator_header
 instances:
   buddy_allocator_body:
+    pos: buddy_allocator_header.ofs_bookkeeping_info_block + 4
     type: buddy_allocator_body
     size: buddy_allocator_header.size_bookkeeping_info_block
-    pos: buddy_allocator_header.ofs_bookkeeping_info_block + 4
 types:
   buddy_allocator_header:
     seq:
@@ -62,8 +62,8 @@ types:
     instances:
       directories:
         io: _root._io
-        size: 1 << block_addresses[directory_entries[0].block_id] & 0x1f
         pos: (block_addresses[directory_entries[0].block_id] >> 0x05 << 0x05) + 4
+        size: 1 << block_addresses[directory_entries[0].block_id] & 0x1f
         type: master_block
         repeat: expr
         repeat-expr: directory_count
--- a/macos/ds_store.ksy
+++ b/macos/ds_store.ksy
@@ -61,6 +61,26 @@ types:
         repeat: expr
         repeat-expr: num_free_lists
     types:
+      block_descriptor:
+        seq:
+          - id: address_raw
+            type: u4
+        instances:
+          mask:
+            value: 0x1f
+          offset:
+            value: (address_raw & ~mask) + 4
+          size:
+            value: 1 << address_raw & mask
+      directory_entry:
+        seq:
+          - id: len_name
+            type: u1
+          - id: name
+            size: len_name
+            type: str
+          - id: block_id
+            type: u4
       free_list:
         seq:
           - id: counter
@@ -80,26 +100,6 @@ types:
         repeat: expr
         repeat-expr: num_directories
         doc: Master blocks of the different B-trees
-  block_descriptor:
-    seq:
-      - id: address_raw
-        type: u4
-    instances:
-      mask:
-        value: 0x1f
-      offset:
-        value: (address_raw & ~mask) + 4
-      size:
-        value: 1 << address_raw & mask
-  directory_entry:
-    seq:
-      - id: len_name
-        type: u1
-      - id: name
-        size: len_name
-        type: str
-      - id: block_id
-        type: u4
   master_block_ref:
     params:
       - id: idx
--- a/macos/ds_store.ksy
+++ b/macos/ds_store.ksy
@@ -19,18 +19,18 @@ instances:
   buddy_allocator_body:
     type: buddy_allocator_body
     size: buddy_allocator_header.size_bookkeeping_info_block
-    pos: buddy_allocator_header.offset_bookkeeping_info_block + 4
+    pos: buddy_allocator_header.ofs_bookkeeping_info_block + 4
 types:
   buddy_allocator_header:
     seq:
       - id: magic
         contents: [0x42, 0x75, 0x64, 0x31]
         doc: Magic number 'Bud1'
-      - id: offset_bookkeeping_info_block
+      - id: ofs_bookkeeping_info_block
         type: u4
       - id: size_bookkeeping_info_block
         type: u4
-      - id: copy_offset_bookkeeping_info_block
+      - id: copy_ofs_bookkeeping_info_block
         type: u4
         doc: Needs to match 'offset_bookkeeping_info_block'
       - size: 16
--- a/macos/ds_store.ksy
+++ b/macos/ds_store.ksy
@@ -18,8 +18,8 @@ seq:
 instances:
   buddy_allocator_body:
     pos: buddy_allocator_header.ofs_bookkeeping_info_block + 4
-    type: buddy_allocator_body
     size: buddy_allocator_header.len_bookkeeping_info_block
+    type: buddy_allocator_body
 types:
   buddy_allocator_header:
     seq:
--- a/macos/ds_store.ksy
+++ b/macos/ds_store.ksy
@@ -62,9 +62,7 @@ types:
     instances:
       directories:
         io: _root._io
-        pos: (block_addresses[directory_entries[0].block_id] >> 0x05 << 0x05) + 4
-        size: 1 << block_addresses[directory_entries[0].block_id] & 0x1f
-        type: master_block
+        type: master_block_ref(_index)
         repeat: expr
         repeat-expr: num_directories
         doc: Master blocks of the different B-trees
@@ -86,6 +84,15 @@ types:
         type: u4
         repeat: expr
         repeat-expr: counter
+  master_block_ref:
+    params:
+      - id: idx
+        type: u8
+    instances:
+      master_block:
+        pos: (_parent.block_addresses[_parent.directory_entries[idx].block_id] >> 0x05 << 0x05) + 4
+        size: 1 << _parent.block_addresses[_parent.directory_entries[idx].block_id] & 0x1f
+        type: master_block
   master_block:
     seq:
       - id: block_id
--- a/macos/ds_store.ksy
+++ b/macos/ds_store.ksy
@@ -1,9 +1,13 @@
 meta:
   id: macos_ds_store
-  endian: be
-  ks-version: 0.8
+  title: macOS '.DS_Store' format
   license: CC-BY-SA-4.0
+  ks-version: 0.8
+  endian: be
+doc: |
+  Apple macOS '.DS_Store' file format.
 doc-ref: |
+  https://en.wikipedia.org/wiki/.DS_Store
   https://metacpan.org/pod/distribution/Mac-Finder-DSStore/DSStoreFormat.pod
   https://0day.work/parsing-the-ds_store-file-format
 seq:
--- a/macos/ds_store.ksy
+++ b/macos/ds_store.ksy
@@ -1,5 +1,5 @@
 meta:
-  id: macos_ds_store
+  id: ds_store
   title: macOS '.DS_Store' format
   license: CC-BY-SA-4.0
   ks-version: 0.8
--- a/macos/ds_store.ksy
+++ b/macos/ds_store.ksy
@@ -72,10 +72,12 @@ types:
       - id: address_raw
         type: u4
     instances:
+      mask:
+        value: 0x1f
       offset:
-        value: (address_raw & ~0x1f) + 4
+        value: (address_raw & ~mask) + 4
       size:
-        value: 1 << address_raw & 0x1f
+        value: 1 << address_raw & mask
   directory_entry:
     seq:
       - id: len_name
--- a/macos/ds_store.ksy
+++ b/macos/ds_store.ksy
@@ -44,7 +44,7 @@ types:
       - size: 4
         doc: Unknown field
       - id: block_addresses
-        type: u4
+        type: block_descriptor
         repeat: expr
         repeat-expr: 256
         doc: Addresses of the different blocks
@@ -67,6 +67,15 @@ types:
         repeat: expr
         repeat-expr: num_directories
         doc: Master blocks of the different B-trees
+  block_descriptor:
+    seq:
+      - id: address_raw
+        type: u4
+    instances:
+      offset:
+        value: (address_raw & ~0x1f) + 4
+      size:
+        value: 1 << address_raw & 0x1f
   directory_entry:
     seq:
       - id: len_name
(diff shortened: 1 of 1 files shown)


GreyCat:
Thanks for this PR! I haven't yet done a through review, but I've noticed that you use `ofs_xxx`, which follows [style guide](http://doc.kaitai.io/ksy_style_guide.html), but other IDs are not very consistent — i.e. you use `xxx_size`, `xxx_len`, `length`, etc, instead of `len_xxx`, `xxx_count`, `counter`, `counter`, etc, instead of `num_xxxs`. Could you consider using consistent names, as per style guide?

> The position of certain instances need to be calculated according to `alignment_header` but I didn't succeed to make it work. I had to manually add 4 to the result as in this example: `buddy_allocator_header.ofs_bookkeeping_info_block + 4`

Um, sorry, I don't quite follow what you mean here. You want some offsets to be automatically adjusted by size of `alignment_header`, which is a fixed 4 bytes? If yes, probably you can benefit from https://github.com/kaitai-io/kaitai_struct/issues/84, when it will be available.

> When I try to use `_index` lines 65 and 66...

Try splitting it into separate type, i.e.:

```yaml
    instances:
      directories:
        io: _root._io
        type: master_block_ref(_index)
        repeat: expr
        repeat-expr: directory_count
        doc: Master blocks of the different B-trees
types:
  master_block_ref:
    params:
      - id: idx
        type: u8
    instances:
      master_block:
        pos: (block_addresses[directory_entries[idx].block_id] >> 0x05 << 0x05) + 4
        size: 1 << block_addresses[directory_entries[idx].block_id] & 0x1f
        type: master_block
```

SkypLabs:
> I haven't yet done a through review, but I've noticed that you use `ofs_xxx`, which follows style guide, but other IDs are not very consistent — i.e. you use `xxx_size`, `xxx_len`, `length`, etc, instead of `len_xxx`, `xxx_count`, `counter`, `counter`, etc, instead of `num_xxxs`. Could you consider using consistent names, as per style guide?

Of course, thanks for catching that. Actually, I read the style guide and renamed some of the fields accordingly. For example, I searched for occurrences of `length` and renamed `xxx_length` to `len_xxx`. Same for `offset`. I missed the `xxx_size` ones and similar fields. It should be fixed now.

> Um, sorry, I don't quite follow what you mean here. You want some offsets to be automatically adjusted by size of `alignment_header`, which is a fixed 4 bytes? If yes, probably you can benefit from kaitai-io/kaitai_struct#84, when it will be available.

Actually, `alignment_header` has a static size. This is why it is fine to use `+ 4` in my calculus. However, I was wondering if there is a more elegant way to explicitly refer to this field in `pos`.

> Try splitting it into separate type

I will try and let you know. Thanks.

GreyCat:
> if there is a more elegant way

There will be, once we'll get sizeof-style things done.

SkypLabs:
> Try splitting it into separate type

It works well now :+1:

GreyCat:
Ready for to be merged?

SkypLabs:
If you're satisfied of my work, yes :relaxed:

macos/ds_store.ksy
```diff
@@ -0,0 +1,195 @@
...
+instances:
+  buddy_allocator_body:
+    pos: buddy_allocator_header.ofs_bookkeeping_info_block + 4
+    size: buddy_allocator_header.len_bookkeeping_info_block
+    type: buddy_allocator_body
+types:
+  buddy_allocator_header:
+    seq:
+      - id: magic
+        contents: [0x42, 0x75, 0x64, 0x31]
```
KOLANICH:
```yaml
contents: ["Bud1"]
```

macos/ds_store.ksy
```diff
@@ -0,0 +1,195 @@
...
+        repeat-expr: num_directories
+        doc: Master blocks of the different B-trees
+  directory_entry:
+    seq:
+      - id: len_name
+        type: u1
+      - id: name
+        size: len_name
+        type: str
+        encoding: UTF-8
```
KOLANICH: Should be moved into `meta`

macos/ds_store.ksy
```diff
@@ -0,0 +1,195 @@
...
+        type: u4
+        repeat: expr
+        repeat-expr: counter
+  master_block_ref:
+    params:
+      - id: idx
+        type: u8
+    instances:
+      master_block:
+        pos: (_parent.block_addresses[_parent.directory_entries[idx].block_id] >> 0x05 << 0x05) + 4
```
KOLANICH: 1. `& ~0x1f` ?
2. snould bit-sized types be used?

macos/ds_store.ksy
SkypLabs: Done.

macos/ds_store.ksy
SkypLabs: Done.

macos/ds_store.ksy
SkypLabs: 1. Yes you're right. It's cleaner and shorter :+1:
2. I'm not sure to see how you would use bit-sized types within this case. I use bitwise operations to calculate the position and the size of some blocks which have their own type.

macos/ds_store.ksy
```diff
@@ -0,0 +1,193 @@
...
+        doc: Unused block
+  buddy_allocator_body:
+    seq:
+      - id: num_blocks
+        type: u4
+        doc: Number of blocks in the allocated-blocks list
+      - size: 4
+        doc: Unknown field
+      - id: block_addresses
+        type: u4
```
KOLANICH:
```yaml
types:
  block_descriptor:
    seq:
       offset:
          type: u27
       size:
          type: u5
```

?

SkypLabs: Right. I didn't get that you meant adding a new type for the block addresses. I like this idea. However, it is a bit more complex than in your example. `offset_raw` needs to be a `u4` with the least-significant 5 bits set to 0 and the size needs to be equal to `2 ^ size`.

Consequently, it can be written like below:

```
  block_descriptor:
    seq:
      - id: address_raw
        type: u4
    instances:
      offset:
        value: (address_raw & ~0x1f) + 4
      size:
        value: 1 << address_raw & 0x1f
```

And then:

```
instances:
  master_block:
    pos: _parent.block_addresses[_parent.directory_entries[idx].block_id].offset
    size: _parent.block_addresses[_parent.directory_entries[idx].block_id].size
    type: master_block
```

Nevertheless, and despite that it works perfectly well, the offset and size values displayed in the Web IDE are incorrect:

Screenshot from 2019-04-09 01-02-31

Any idea why? I pushed these changes to https://github.com/SkypLabs/kaitai_struct_formats/tree/macos_ds_store_block_descriptor for your information.

macos/ds_store.ksy
```diff
@@ -0,0 +1,193 @@
+meta:
+  id: ds_store
+  title: macOS '.DS_Store' format
+  license: CC-BY-SA-4.0
```
KOLANICH: CC-BY-* are not very suitable licenses for source code. They are more for art.

macos/ds_store.ksy
KOLANICH: >bit more complex than in your example. offset_raw needs to be a u4 with the least-significant 5 bits set to 0 and the size needs to be equal to 2 ^ size.

You are right.

>despite that it works perfectly well, the offset and size values displayed in the Web IDE are incorrect

I have stopped using WebIDE because it has some bugs fixed by noone. In my case (another file format) it also visualizes incorrectly, so I had to write a convertor of parsed objects into tag files for wxHexEditor (IMHO wXHexEditor needs support for nested tags desperately).

SkypLabs: Ok I see. I will create an issue in the Web IDE's Git repository to report the bug later on.

macos/ds_store.ksy
SkypLabs: Right. I've found a lot of CC-licensed KSY files in the file format gallery. I thought it was a kind of convention to use such licenses. I changed to MIT.

macos/ds_store.ksy
```diff
@@ -0,0 +1,202 @@
...
+    seq:
+      - id: block_id
+        type: u4
+        if: mode > 0
+      - id: record
+        type: record
+    instances:
+      block:
+        io: _root._io
+        pos: _root.buddy_allocator_body.block_addresses[block_id].offset
```
KOLANICH: 1. Why is `size` unused? Should it be?
2. I guess it may make sense to make `block` a child of `block_descriptor` and make it available via `block` instance of the descriptor.

macos/ds_store.ksy
KOLANICH: CC-0 is different from other CC licences, it is just a public domain disclaimer with some legal words making it suitable for the jurisdictions like Germany.

SkypLabs: Interesting, I didn't know. Thanks for the explanation.

macos/ds_store.ksy
SkypLabs: 1. `size` is not used because otherwise I get the following error message (or a similar one):

```
Parse error: undefined
Call stack: undefined KaitaiEOFError: requested 34 bytes, but only -12 bytes available
```

This is because the whole structure is a B-tree. A block can contain other blocks which in their turn can contain other blocks, etc.

2. It's possible but if you do so, you won't see the hierarchy of the B-tree, only a flat list of blocks.

macos/ds_store.ksy
```diff
@@ -0,0 +1,202 @@
...
+        repeat: expr
+        repeat-expr: num_directories
+        doc: Master blocks of the different B-trees
+  block_descriptor:
+    seq:
+      - id: address_raw
+        type: u4
+    instances:
+      offset:
+        value: (address_raw & ~0x1f) + 4
```
KOLANICH: the mask can be moved into an own value  instance

macos/ds_store.ksy
```diff
@@ -0,0 +1,202 @@
...
+  master_block_ref:
+    params:
+      - id: idx
+        type: u8
+    instances:
+      master_block:
+        pos: _parent.block_addresses[_parent.directory_entries[idx].block_id].offset
+        size: _parent.block_addresses[_parent.directory_entries[idx].block_id].size
+        type: master_block
+  master_block:
```
KOLANICH: should be nested into `master_block_ref`.

macos/ds_store.ksy
```diff
@@ -0,0 +1,202 @@
...
+  directory_entry:
+    seq:
+      - id: len_name
+        type: u1
+      - id: name
+        size: len_name
+        type: str
+      - id: block_id
+        type: u4
+  free_list:
```
KOLANICH: Should be nested into `buddy_allocator_body`

macos/ds_store.ksy
```diff
@@ -0,0 +1,202 @@
...
+        doc: Indicates the number of directory entries
+      - id: directory_entries
+        type: directory_entry
+        repeat: expr
+        repeat-expr: num_directories
+        doc: Each directory is an independent B-tree
+      - id: free_lists
+        type: free_list
+        repeat: expr
+        repeat-expr: 32
```
KOLANICH: magic constant can be moved into an own instance

macos/ds_store.ksy
```diff
@@ -0,0 +1,202 @@
...
+    seq:
+      - id: num_blocks
+        type: u4
+        doc: Number of blocks in the allocated-blocks list
+      - size: 4
+        doc: Unknown field
+      - id: block_addresses
+        type: block_descriptor
+        repeat: expr
+        repeat-expr: 256
```
KOLANICH: yet another magic constant

macos/ds_store.ksy
```diff
@@ -0,0 +1,202 @@
...
+        type: u4
+        doc: Needs to match 'offset_bookkeeping_info_block'
+      - size: 16
+        doc: Unused block
+  buddy_allocator_body:
+    seq:
+      - id: num_blocks
+        type: u4
+        doc: Number of blocks in the allocated-blocks list
+      - size: 4
```
KOLANICH: Feel free to add any guesses and observations you have into the docs on unknown fields.

macos/ds_store.ksy
KOLANICH: >It's possible but if you do so, you won't see the hierarchy of the B-tree, only a flat list of blocks.

Not sure in this statement. IMHO nestedness will be preserved: nearly everywhere you use blocks, you use their descriptors. So It is descriptors who define the blocks. BTW: `value` instances can be references to structs.

macos/ds_store.ksy
SkypLabs: Good idea. Done.

macos/ds_store.ksy
KOLANICH: >size is not used because otherwise I get the following error message (or a similar one):

>This is because the whole structure is a B-tree. A block can contain other blocks which in their turn can contain other blocks, etc.

Largely depends on the structure of the stuff. If these sizes are meaningful, they probably can be enforced by using the right `io`

macos/ds_store.ksy
SkypLabs: Done.

macos/ds_store.ksy
SkypLabs: Done.

macos/ds_store.ksy
SkypLabs: Interesting. I didn't know it was possible to nest types into others. Done.

macos/ds_store.ksy
SkypLabs: Done.

macos/ds_store.ksy
SkypLabs: Does it actually make sense to expose it to the user in the parsing result?

macos/ds_store.ksy
SkypLabs: Done.

macos/ds_store.ksy
KOLANICH: IMHO - should be moved to root level

SkypLabs: Yes I think it's much better this way. Done.

macos/ds_store.ksy
SkypLabs: > Not sure in this statement. `value` instances can be references to structs.

Yes but you won't be able to know which upper-level block the one you are currently looking at is linked to, only the sublevel ones. However, a `block` child value can be added to `block_descriptor` in addition to the B-tree structure starting with `master_block`.

SkypLabs: > Largely depends on the structure of the stuff. If these sizes are meaningful, they probably can be enforced by using the right `io`.

Certainly, but I haven't found a way to make it work so far.

SkypLabs:
Thanks @KOLANICH and @GreyCat for your awesome review! It was also really instructive. I've learnt a lot about your project.

KOLANICH:
Thank you for your contribution ;)

GreyCat:
Just merged in, thanks for this awesome work!

PR #132, opened 2019-03-29
Add WebSocket dataframe format

bspeice:
Not sure if there's a better way to handle payload length and opcode overlaps, but this should get things started.


network/websocket.ksy
```diff
@@ -0,0 +1,77 @@
+meta:
```
GreyCat: You haven't specified an open source license under which you license your work. Please add `meta/license` tag.

network/websocket.ksy
```diff
@@ -0,0 +1,77 @@
...
+  id: websocket_dataframe
```
GreyCat: This should match file name. Could you please rename the file to `websocket_dataframe.ksy`?

network/websocket.ksy
```diff
@@ -0,0 +1,77 @@
...
+  The WebSocket Protocol enables two-way communication between a client
+  running untrusted code in a controlled environment to a remote host
+  that has opted-in to communications from that code.  The security
+  model used for this is the origin-based security model commonly used
+  by web browsers.  The protocol consists of an opening handshake
+  followed by basic message framing, layered over TCP.  The goal of
+  this technology is to provide a mechanism for browser-based
+  applications that need two-way communication with servers that does
+  not rely on opening multiple HTTP connections (e.g., using
+  XMLHttpRequest or <iframe>s and long polling).
```
GreyCat: This is a copy-paste of copyrighted text from IETF standard: https://tools.ietf.org/id/draft-ietf-hybi-thewebsocketprotocol-09.html

Unfortunately, it's "All rights reserved", so you can't just copy-paste it here.

network/websocket.ksy
```diff
@@ -0,0 +1,77 @@
...
+    
+instances:
+  fin:
+    value: (b1 & 0b10000000) >> 7
+  rsv1:
+    value: (b1 & 0b01000000) >> 6
+  rsv2:
+    value: (b1 & 0b00100000) >> 5
+  rsv3:
+    value: (b1 & 0b00010000) >> 4
```
GreyCat: You can use bit-level types (i.e. `type: b1`) instead of doing all these arithmetics.

network/websocket.ksy
```diff
@@ -0,0 +1,77 @@
...
+  opcode:
+    enum: opcode
+    value: (b1 & 0b00001111)
```
GreyCat: Again,

```
seq:
  # ...
  - id: opcode
    type: b4
    enum: opcode
```

will be much shorter & easier to read.

GreyCat:
Thanks for your contribution! I've added some questions / proposals, please take a look?

bspeice:
In progress, the review is greatly appreciated!

network/websocket.ksy
bspeice: My default is Apache 2.0, but I'm happy to license under different terms if it's easier for the maintainers.

network/websocket.ksy
bspeice: Done! I was unable to apply the bit types to the mask/payload values given the need for comparison in `mask_key`.

bspeice:
First pass should be done, let me know if there are any further issues!

network/websocket_dataframe.ksy
```diff
@@ -0,0 +1,68 @@
...
+  xref:
+    rfc: 6455
+  endian: be
+  license: APACHE-2.0
+doc: |
+  The WebSocket protocol establishes a two-way communication channel via TCP.
+  Messages are made up of one or more dataframes, and are delineated by
+  frames with the `fin` bit set.
+seq:
+  - id: fin
```
KOLANICH: could you use clearer `id`s, using the names specified in the rfc in `-orig-id`?

network/websocket_dataframe.ksy
```diff
@@ -0,0 +1,68 @@
...
+  - id: extended_len_1
+    type: u2
+    if: payload_len == 126
+  - id: extended_len_2
+    type: u4
+    if: payload_len == 127
+  - id: mask_key
+    type: u4
+    if: is_masked == 1
+  - id: payload
```
KOLANICH: Should be parsed depending on opcode

network/websocket_dataframe.ksy
```diff
@@ -0,0 +1,68 @@
...
+  endian: be
+  license: APACHE-2.0
+doc: |
+  The WebSocket protocol establishes a two-way communication channel via TCP.
+  Messages are made up of one or more dataframes, and are delineated by
+  frames with the `fin` bit set.
+seq:
+  - id: fin
+    type: b1
+  - id: rsv1
```
KOLANICH: Can be merged. And probably should.

KOLANICH:
Also, could you use `git commit --amend` further and merge all the existing commits in this PR into a single one?

network/websocket.ksy
GreyCat: We are in somewhat gray legal situation here: it's not super clear if assigning a license to a .ksy file somehow transfers to results of compilation of that file. For that reason, many formats are listed as CC0, to promote maximum usage possibility in both open and closed source software: http://formats.kaitai.io/xref.html

If Apache 2.0 is your choice, then it should be probably spelled as `license: Apache-2.0`, as per https://spdx.org/licenses/Apache-2.0.html

GreyCat:
> Could you further use `git commit --amend` and merge all the existing commits in this PR into a single one?

Don't do that and don't advise people to do that. Squashing can be easily done during merge, and using amends + force pushes hurts reviews a lot, as we have to start from scratch every time.

network/websocket_dataframe.ksy
```diff
@@ -0,0 +1,68 @@
...
+    type: u4
+    if: is_masked == 1
+  - id: payload
+    size: len
+    
+instances:
+  is_masked:
+    value: (b1 & 0b10000000) >> 7
+  payload_len:
+    value: (b1 & 0b01111111)
```
GreyCat: The only way `b1` is used is to split up into these two, so it would make sense to implement is using `bX` as well:

```yaml
seq:
  # ...
  - id: is_masked
    type: b1
  - id: len_payload
    type: b7
```

Also please note the [style guide](http://doc.kaitai.io/ksy_style_guide.html) — the general guidance is to use `len_XXX` identifiers for attributes that carry size information for `XXX` attribute.

network/websocket_dataframe.ksy
GreyCat: Also, as per [style guide](http://doc.kaitai.io/ksy_style_guide.html), it's better to stick to `reserved` name:

> For reserved fields which are known to be unused, use `reserved` name (or `reserved1`, `reserved2`, etc, if there are many of them)

KOLANICH:
>using amends + force pushes hurts reviews a lot, as we have to start from scratch every time.

AFAIK, GH computes differences between force-pushes, allowing to review only changes.

GreyCat:
The very idea of force push is that the previous state of files does not exist anymore, so I doubt that.

KOLANICH:
https://git-scm.com/docs/git-reflog

network/websocket_dataframe.ksy
bspeice: Not sure what you mean here; the masking key is resolved based on an earlier flag, and is unrelated to the opcode.

network/websocket_dataframe.ksy
bspeice: What do you mean by merging these? Moving them to a separate types block?

bspeice:
Git `reflog` and `gc` only make sense from a local context. If you want me to squash before merge, happy to do that, but I'll wait until such time as everything is approved.

network/websocket_dataframe.ksy
bspeice: Happy to change the names, but there were issues with `if: is_masked == 1` being used to either include the masking bits or not. Given that, I opted to leave it as a byte that gets parsed later. If there's a better solution, let me know.

network/websocket.ksy
bspeice: Changed to CC0!

network/websocket_dataframe.ksy
KOLANICH:
```yaml
- id: reserved
  type: b3
```

network/websocket_dataframe.ksy
KOLANICH:
```yaml
- id: reserved
  type: b3
```

KOLANICH:
>Git `reflog` and `gc` only make sense from a local context.

The data structures of remote repos are the same. So GH has access to previous versions of histories, and this can be checked. The commands are usually explicitly called in a local context by most of users. On servers `gc` is usually called automatically, [some software has features to call it explicitly](https://docs.gitlab.com/ee/administration/housekeeping.html).

network/websocket_dataframe.ksy
bspeice: Ah. Still issues with that; `continuation` opcodes are ambiguous, as the first frame will have determined the message format. Is it worth changing this from a dataframe parser to a message parser?

network/websocket_dataframe.ksy
GreyCat: They're all still independent attributes? I mean, why don't we just do:

```yaml
  - id: is_masked
    type: b1
  - id: len_payload_primary
    type: b7
  - id: len_payload_extended_1
    type: u2
    if: len_payload_primary == 126
  - id: len_payload_extended_2
    type: u4
    if: len_payload_primary == 127
  - id: mask_key
    type: u4
    if: is_masked
```

bspeice: That works super well! I hadn't thought of it at the time, but it's now done.

bspeice:
OK, so far the final issue is figuring out if there's a way to handle the payload type based on the opcode. Websocket transfers `TEXT` opcodes as UTF-8, but we're unable to figure that out for dataframes that are continuation opcodes. If we have a way of accessing the opcode of the first instance in a repeat we could decode entire messages and change the payload type based on the first being TEXT or not, but I haven't found anything to suggest making use of repeats like that is possible.

websocket_dataframe.py
```diff
@@ -0,0 +1,66 @@
+# This is a generated file! Please edit source .ksy file and use kaitai-struct-compiler to rebuild
```
KOLANICH: @GreyCat, should we add all the extensions of target language files into .gitignore?

network/websocket_dataframe.ksy
KOLANICH: Continuation will be really pain. [`Chains` proposal](https://github.com/kaitai-io/kaitai_struct/issues/196) can solve it, but it is not yet implemented.

websocket_dataframe.py
bspeice: Certainly wouldn't be opposed to it, but this was removed in the meantime.

GreyCat:
> If we have a way of accessing the opcode of the first instance in a repeat we could decode entire messages and change the payload type based on the first being TEXT or not, but I haven't found anything to suggest making use of repeats like that is possible.

I see several options here:

1. Go from the parsing of a dataframe only to the parsing of a whole message. Probably in that case you'll have that top-level `repeat`, in there you can reach to the very first frame of a message via something akin to `_root.frames[0]`.
2. Make dataframe parameterized: propagate that flag picked from the first frame through a `params` section of top-level type.

KOLANICH:
I feel like it may won't help much. I mean that without chains we won't be able to parse contents of multiframe payloads.

bspeice:
Had no idea that the `[0]` subscript would work, but the compiler doesn't appear to complain. I can switch to parsing based on propagating the param, but my guess is that anyone who cares about that distinction is also fine taking just the dataframe definition from this project and tweaking what they need.

Anything else?

websocket_dataframe.py
GreyCat: @KOLANICH Might make sense, albeit we currently have some UNIX shell/Ruby-based automation machinery in this repo (in `/_build`).

Want to start a PR?

GreyCat:
Cool, thanks. Looks like it's all good now, merging in.

bspeice:
Took another look at this; we're currently not un-masking the payload if the masking key is set. Is there a good way to conditionally apply a `process` step?

GreyCat:
Not really. A common pattern that converges them looks like that:

```yaml
seq:
  - id: body_unprocessed
    size: len_body
    type: body
    if: not must_process
  - id: body_processed
    size: len_body
    type: body
    process: some.processing(arg1, arg2, ...)
    if: must_process
instances:
  body:
    value: 'must_process ? body_processed : body_unprocessed'
```

PR #133, opened 2019-04-14
Add Sudoers Time Stamp file format

fishilico:
Hello,

After I read https://github.com/nongiach/sudo_inject, I was wondering whether the format of time stamp files used by `sudo` was implemented in Kaitai Struct. As it is not, here is a PR to add it.

This format is quite simple, it consists of a sequence of entries defined by `struct timestamp_entry` in `sudo`'s source code. This is similar to `utmp` files, which is why I created a format in `log/`.

As of today, there are two versions of this format:
* version 1 has only been "documented" in sudo's source code, https://www.sudo.ws/repos/sudo/file/SUDO_1_8_10/plugins/sudoers/check.h
* version 2 is officially documented on https://www.sudo.ws/man/1.8.27/sudoers_timestamp.man.html. Its main difference is the field `struct timespec start_time`, which have been introduced.

Could you please review this PR?

### DIFF
--- a/log/sudoers_ts.ksy
+++ b/log/sudoers_ts.ksy
@@ -17,11 +17,12 @@ types:
       - id: version
         doc: version number of the timestamp_entry struct
         type: u2
-      - id: size
+      - id: len_record
         doc: size of the record in bytes
         type: u2
+        -orig-id: size
       - id: payload
-        size: size - 4
+        size: len_record - 4
         type:
           switch-on: version
           cases:
--- a/log/sudoers_ts.ksy
+++ b/log/sudoers_ts.ksy
@@ -1,0 +1,114 @@
+meta:
+  id: sudoers_ts
+  title: Sudoers Time Stamp file
+  license: CC0-1.0
+  endian: le
+doc: |
+  This spec can be used to parse sudo time stamp files located in directories
+  such as /run/sudo/ts/$USER or /var/lib/sudo/ts/$USER.
+doc-ref: https://www.sudo.ws/man/1.8.27/sudoers_timestamp.man.html
+seq:
+  - id: records
+    type: record
+    repeat: eos
+types:
+  record:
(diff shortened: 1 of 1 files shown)


log/sudoers_ts.ksy
```diff
@@ -0,0 +1,114 @@
+meta:
+  id: sudoers_ts
+  title: Sudoers Time Stamp file
+  license: CC0-1.0
+  endian: le
```
GreyCat: As per original sudo implementation, it seems to be machine-dependent, not truly little endian?

log/sudoers_ts.ksy
```diff
@@ -0,0 +1,114 @@
...
+  - id: records
+    type: record
+    repeat: eos
+types:
+  record:
+    seq:
+      - id: version
+        doc: version number of the timestamp_entry struct
+        type: u2
+      - id: size
```
GreyCat: Minor nitpick: [style guide](http://doc.kaitai.io/ksy_style_guide.html) recommends using `len_payload` for size of certain attribute (`payload`) in bytes.

log/sudoers_ts.ksy
fishilico: Indeed. What is the recommended way to define a machine-dependent endianness? Removing this line, or something else?

log/sudoers_ts.ksy
fishilico: It is the length of the record, so `len_record` would be better. I will change it. Thanks for your review!

log/sudoers_ts.ksy
KOLANICH: You can use `version` of the first record to determine endianness. Then there is a way to set endiannes in `meta` using  `switch-on`-`cases`.

GreyCat:
Merging this in, thanks!

PR #102, opened 2018-10-22
Add bitcoin transaction

SkypLabs:
Hi,

This PR contains a KSY file for decoding raw Bitcoin transactions (see an [example](https://github.com/kaitai-io/kaitai_struct_formats/files/2503578/raw_bitcoin_transaction.zip)).

Cheers

PR #135, opened 2019-04-20
Improve bitcoin transaction

SkypLabs:
Hello,

I would like to apply what I've learnt from #131 to #102.

This PR brings the following improvements:

* Change the license (see https://github.com/kaitai-io/kaitai_struct_formats/pull/131#discussion_r273331680)
* Nest types and enumerations into others when it's relevant
* Use `len` as a prefix and not as a suffix as defined in the [style guide](https://doc.kaitai.io/ksy_style_guide.html#attr-id).

Thanks

### DIFF
--- a/network/bitcoin_transaction.ksy
+++ b/network/bitcoin_transaction.ksy
@@ -101,6 +101,12 @@ types:
         type: public_key
         doc: |
           Public key (bitcoin address of the recipient).
+    enums:
+      sighash_type:
+        1: sighash_all
+        2: sighash_none
+        3: sighash_single
+        80: sighash_anyonecanpay
   der_signature:
     seq:
       - id: sequence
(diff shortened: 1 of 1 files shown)


network/bitcoin_transaction.ksy
```diff
@@ -62,88 +62,90 @@ types:
...
+            doc: |
+              Type of signature.
+          - id: len_pubkey_stack
+            type: u1
+          - id: pubkey
+            type: public_key
+            doc: |
+              Public key (bitcoin address of the recipient).
+        types:
+          der_signature:
```
KOLANICH: there is already a spec for ASN.1

GreyCat:
Relatively simple change, looks legit, thanks!

network/bitcoin_transaction.ksy
SkypLabs: Right. I will have a look when I get a chance. Thanks!

PR #136, opened 2019-04-21
Add UEFI db and dbx Signature List format

fishilico:
This new format parses UEFI variables `db` and `dbx` that contain signatures, certificates and hashes. On a Linux system using UEFI, these variables are readable from `/sys/firmware/efi/efivars/db-d719b2cb-3d3a-4596-a3bc-dad00e67656f`, `/sys/firmware/efi/efivars/dbDefault-8be4df61-93ca-11d2-aa0d-00e098032b8c`, `/sys/firmware/efi/efivars/dbx-d719b2cb-3d3a-4596-a3bc-dad00e67656f` and `/sys/firmware/efi/efivars/dbxDefault-8be4df61-93ca-11d2-aa0d-00e098032b8c`.

I added this new format in `security/`, next to OpenPGP message and SSH pubkey formats. If another directory is more suitable, this format can be moved.

### DIFF
--- a/security/efivar_signature_list.ksy
+++ b/security/efivar_signature_list.ksy
@@ -1,0 +1,151 @@
+meta:
+  id: efivar_signature_list
+  title: UEFI Variable with Signature List
+  license: CC0-1.0
+  endian: le
+doc: |
+  Parse UEFI variables db and dbx that contain signatures, certificates and
+  hashes. On a Linux system using UEFI, these variables are readable from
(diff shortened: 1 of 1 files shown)


GreyCat:
Thanks for this contribution, looks very good!

I'm kind of indecisive between whether it should go into `security` or into `firmware`, so, given that you've already put it into `security`, let's stick with that. Actually, it looks like our categorization system needs some kind of revamp to allow multiple tags or categories or something like that.

PR #142, opened 2019-04-29
Add ZX Spectrum tape file format

ciarant:
80s home computer emulator file format.


game/tap.ksy
```diff
@@ -0,0 +1,49 @@
...
+    seq:
+      - id: length
+        contents: [0x13, 0x00]
+      - id: flag
+        type: u1
+        enum: flag_enum
+      - id: header
+        type: header_block
+      - id: data
+        size: header.data_length + 4
```
dgelessus: If I understand this correctly, the `block` type currently parses a header block together with the following data block. This works fine if the data is complete, consistent and uses the standard structure, but would break in a few unusual cases:

* If the data length stored in the header block doesn't match the actual length of the following block, the `data` field will parse an incorrect amount of data, and any following blocks will not be parsed correctly.
* If the header block is for some reason followed by another header block, that header will be parsed as data instead of a header, which will also likely break parsing of the following blocks.
* If a data block is not preceded by a header, the data will be incorrectly parsed as a header instead (or more likely parsing will fail because the data length is not exactly 0x13).

The third case is probably the most important - I've seen some custom loaders use data blocks in the standard format, but without a preceding header block. These are valid tapes and should be accepted by the spec. The other two error cases are less important, because they only appear when the tape data is incomplete or out of order.

I think the best way to handle this would be to parse each TAP block on its own, without trying to link the header blocks with their corresponding data blocks. This should be fairly easy to do - each TAP block is preceded by its length (so there's no need to read the data block length from its header block) and includes a header/data type flag (so you can tell which blocks are headers and should be parsed).

dgelessus:
Also, a more general idea: the format of the TAP block contents (that is, everything in a block after the length bytes) is also used for [standard-speed blocks](http://www.worldofspectrum.org/TZXformat.html#STDSPEED) in the more advanced [TZX format](http://www.worldofspectrum.org/TZXformat.html). It could be useful to extract the inner block format into its own spec (like `zx_spectrum_tape_block`) and have this spec include it. This could be reused if someone writes a TZX spec in the future. (This isn't *necessary* of course, this refactoring could also be done later when someone writes a TZX spec.)

dgelessus:
Also, a more general idea: the format of the TAP block contents (that is, everything in a block after the length bytes) is also used for [standard-speed blocks](http://www.worldofspectrum.org/TZXformat.html#STDSPEED) in the more advanced [TZX format](http://www.worldofspectrum.org/TZXformat.html). It could be useful to extract the inner block format into its own spec (like `zx_spectrum_tape_block`) and have this spec include it. This could be reused if someone writes a TZX spec in the future. (This isn't *necessary* of course, this refactoring could also be done later when someone writes a TZX spec.)

game/tap.ksy
```diff
@@ -0,0 +1,49 @@
+meta:
+  id: tap
```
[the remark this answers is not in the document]
ciarant: Sounds good. Will do.

game/tap.ksy
```diff
@@ -0,0 +1,49 @@
...
+  file-extension: tap
+  endian: le
+  license: CC0-1.0
+  title: ZX Spectrum tape file format
+doc-ref: https://faqwiki.zxnet.co.uk/wiki/TAP_format
```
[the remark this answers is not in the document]
ciarant: Just to be sure, is this the correct syntax?
```
doc-ref:
  - https://faqwiki.zxnet.co.uk/wiki/TAP_format
  - http://www.zx-modules.de/fileformats/tapformat.html
```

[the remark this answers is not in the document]
dgelessus: You only need the `-`s, not the `|`. The `|` syntax is used for multiline strings, in this case you just need a list of strings.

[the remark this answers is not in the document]
ciarant: Sounds good. Just fyi, I'm getting an error on the Web IDE when I omit the '|'.
```
/doc-ref: expected string, got List
```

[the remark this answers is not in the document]
dgelessus: Ah, the WebIDE uses the stable 0.8 version of Kaitai Struct. Lists in `doc-ref` are only supported in the unstable 0.9 version. I don't know if there's a version of the WebIDE that uses the unstable version of Kaitai Struct (there is https://ide.kaitai.io/devel/, but that's the development version of the WebIDE itself, I don't know if that uses a newer Kaitai Struct version...). I'm also not sure if we're supposed to use unstable features in kaitai_struct_formats, or if the specs are supposed to be compatible with the current stable release...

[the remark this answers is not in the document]
ciarant: Hmm, just to be safe it's probably best to stick to using stable features. How about I do something like this:
```
doc-ref: https://faqwiki.zxnet.co.uk/wiki/TAP_format
doc-ref: http://www.zx-modules.de/fileformats/tapformat.html
```

game/tap.ksy
```diff
@@ -0,0 +1,49 @@
+meta:
+  id: tap
+  file-extension: tap
+  endian: le
+  license: CC0-1.0
+  title: ZX Spectrum tape file format
```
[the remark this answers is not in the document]
dgelessus: You can also add `pronom: fmt/801` to the `xref` block (for the [PRONOM format database entry](http://www.nationalarchives.gov.uk/PRONOM/Format/proFormatSearch.aspx?status=detailReport&id=1601)).

game/tap.ksy
ciarant: Great info. Would you have a sample TAP file demonstrating the third case?

game/tap.ksy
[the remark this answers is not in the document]
dgelessus: That doesn't work, or at least not like you might think. `doc-ref` is a key in a YAML dictionary, and specifying the same key twice is not allowed in YAML, though some parsers silently allow duplicates and only keep one of the values.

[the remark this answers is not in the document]
ciarant: Ahh. OK, I'll just keep a single `doc-ref` for now.

game/tap.ksy
dgelessus: One example is the game [Full Throttle](http://www.worldofspectrum.org/infoseekid.cgi?id=0001908) - World of Spectrum has a [TAP download](http://www.worldofspectrum.org/pub/sinclair/games/f/FullThrottle.tap.zip) available. It has a custom loader (standard format, header and data) followed by a data block with no header.

game/tap.ksy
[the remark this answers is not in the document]
GreyCat: I'm actually wondering is `game` is a good primary category as well. ZX Spectrum is not just a gaming appliance — and this looks like it's a general purpose tape emulator image format. May be `filesystem` will be a better candidate?

game/tap.ksy
```diff
@@ -0,0 +1,49 @@
...
+        size: header.data_length + 4
+  header_block:
+    seq:
+      - id: header_type
+        type: u1
+        enum: header_type_enum
+      - id: filename
+        size: 10
+        type: str
+        encoding: ASCII
```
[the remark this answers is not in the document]
GreyCat: Probably this warrants certain extension to character sets that will automatically decode that into Unicode where possible, but this is definitely a separate big feat. If wonder if any big organization have already assigned ZX Spectrum encoding some kind of codepage/official name, and if that has support in something like iconv?

Please add this as a reminder to https://github.com/kaitai-io/kaitai_struct/issues/393 ?

game/tap.ksy
```diff
@@ -0,0 +1,49 @@
...
+      - id: data_length
+        type: u2
```
GreyCat: You might want to consider sticking to [style guide](http://doc.kaitai.io/ksy_style_guide.html) and naming it `len_data`.

game/tap.ksy
```diff
@@ -0,0 +1,49 @@
...
+      - id: param1
+        type: u2
+      - id: param2
+        type: u2
+      - id: checksum
```
GreyCat: It would be great to add some documentation on how this checksum is being formed / checked.

game/tap.ksy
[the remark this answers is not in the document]
dgelessus: Right, `game` may not be the best category. `filesystem` isn't entirely accurate either since the TAP format is a wrapper around the actual data found on the tape, but on the other hand it's a *very* thin wrapper (it just adds length bytes) and I can't think of any better category right now.

game/tap.ksy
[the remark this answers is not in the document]
dgelessus: This may be of interest: [L2/19-025: Proposal to add characters from legacy computers and teletext to the UCS](https://www.unicode.org/L2/L2019/19025-terminals-prop.pdf).

One thing to keep in mind is that the ZX Spectrum character set includes user-defined graphics characters (basically characters whose glyphs you can change as you like) and BASIC tokens (`PRINT`, `LOAD`, `IF`, etc. - these are actually single characters, but are normally displayed as multiple characters on screen). There's no obvious answer to transcoding these - especially with the BASIC keywords it depends on what you want to do with the text. For displaying you want to "expand" them to their text equivalents, but when comparing or storing them you want to keep them in their original byte form.

game/tap.ksy
dgelessus: (For reference: the checksum byte is simply all data bytes XORed together, see https://faqwiki.zxnet.co.uk/wiki/TAP_format#Format_Description.)

game/tap.ksy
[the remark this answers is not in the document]
ciarant: Everyone happy to move to `filesystem` then?

game/zx_spectrum_tap.ksy
```diff
@@ -0,0 +1,52 @@
...
+        type: u1
+        enum: header_type_enum
+      - id: filename
+        size: 10
+        pad-right: 0x20
+      - id: len_data
+        type: u2
+      - id: param1
+        type: u2
+      - id: param2
```
KOLANICH: How about `doc`? And why not
```yaml
- id: params
  type: u2
  repeat: expr
  repeat-expr: 2
```
?

game/zx_spectrum_tap.ksy
```diff
@@ -0,0 +1,52 @@
...
+  header_type_enum:
+    0: program
+    1: num_array
+    2: char_array
+    3: bytes
+types:
+  block:
+    seq:
+      - id: len_header
+        contents: [0x13, 0x00]
```
KOLANICH: Is it really meant to be hardcoded? This spec doesn't look futureproof.

game/tap.ksy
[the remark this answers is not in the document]
KOLANICH: https://github.com/kaitai-io/kaitai_struct/issues/269

game/zx_spectrum_tap.ksy
ciarant: That's certainly an option. I'll review.

game/zx_spectrum_tap.ksy
ciarant: All ROM headers are 0x13 bytes long (see http://www.zx-modules.de/fileformats/tapformat.html). Does need some rework to allow for non-header data blocks.

KOLANICH: If they always were, there would be no need to put that number into files. This stuff is usually done if the number can change in future. I know that spectrum is an obs9pete platfom

game/zx_spectrum_tap.ksy
dgelessus: Not sure if it's a good idea to make this an array. Semantically, the two "param" values are not an array, they are simply two `u2` values that have nothing to do with each other.

By the way, the meaning of the "param" values is known, but it depends on the `header_type`. It would be possible to create individual structures for each interpretation of the "param" values and select the correct one using a `switch-on` based on `header_type`.

ciarant: Agreed. Working on having separate param structures - just having trouble trying to avoid the `body_*` scenario described in the documentation.

game/zx_spectrum_tap.ksy
ciarant: Resolved

game/zx_spectrum_tap.ksy
ciarant: Done. Would appreciate a review.

game/zx_spectrum_tap.ksy
```diff
@@ -0,0 +1,82 @@
...
+  program_params:
+    seq:
+      - id: autostart_line
+        type: u2
+      - id: len_program
+        type: u2
+  array_params:
+    seq:
+      - type: u1
+        doc: Unused
```
dgelessus: According to [the style guide](http://doc.kaitai.io/ksy_style_guide.html#attr-id), "unused" fields should have an `id`.

game/zx_spectrum_tap.ksy
```diff
@@ -0,0 +1,82 @@
...
+      - id: var_name
+        type: u1
+        doc: Variable name (1..26 meaning A$..Z$ +192)
+      - type: u2
+        doc: Unused (0x8000)
```
dgelessus: Should we specify this value as `contents`, since it's always supposed to be that value? On the other hand it's not an important part of the format, so maybe we shouldn't fail parsing if the value doesn't match.

game/zx_spectrum_tap.ksy
ciarant: Yeah. The issue I have though is that there are multiple unused fields - is there a naming convention I should use other than `unused1`, `unused2`...

game/zx_spectrum_tap.ksy
ciarant: I guess it might make more sense to specify the expected value as it will help flag an issue with a corrupt file (or the parser grammar!). I think I'll add it. Sound OK?

game/zx_spectrum_tap.ksy
dgelessus: Yep, that's the convention:

> For reserved fields which are known to be unused, use `reserved` name (or `reserved1`, `reserved2`, etc, if there are many of them)

game/zx_spectrum_tap.ksy
dgelessus: I really don't know what would be the best option here... You're right, it would help with detecting cases where something went wrong with the parsing. But I don't know how the Spectrum handles these unused fields when loading, or if they are even checked at all. If they aren't, there are *probably* tapes somewhere that have different values in those fields, but are otherwise valid. If that is the case, we should allow any value there.

It's always fun working with old formats where nobody really knows what's valid and how edge cases should be handled 😄

ciarant: I've just added the contents check for now.

game/zx_spectrum_tap.ksy
ciarant: I've added ids.

game/zx_spectrum_tap.ksy
KOLANICH: Much better, thanks.

ciarant:
> Also, a more general idea: the format of the TAP block contents (that is, everything in a block after the length bytes) is also used for [standard-speed blocks](http://www.worldofspectrum.org/TZXformat.html#STDSPEED) in the more advanced [TZX format](http://www.worldofspectrum.org/TZXformat.html). It could be useful to extract the inner block format into its own spec (like `zx_spectrum_tape_block`) and have this spec include it. This could be reused if someone writes a TZX spec in the future. (This isn't _necessary_ of course, this refactoring could also be done later when someone writes a TZX spec.)

Great idea. I think I'll hold off for now and attack it later as a TZX support and refactoring exercise.

ciarant:
Anything more I need to do with this?

GreyCat:
I think everyone's consensus that we're ready with this one, so merging in. Thanks for this contribution, @ciarant!

ciarant:
> I think everyone's consensus that we're ready with this one, so merging in. Thanks for this contribution, @ciarant!

Thank you for Kaitai Struct!

PR #145, opened 2019-05-03
network/websocket: Fix invalid `frame_header` ref

bspeice:
Should we add a CI check that just runs the compiler on every format when opening a PR? Should hopefully catch issues like this in the future.


GreyCat:
@bspeice That would be most awesome, but I never got to get something like that done.

KOLANICH:
@bspeice, you may want read the discussion in https://github.com/kaitai-io/kaitai_struct/issues/165

PR #146, opened 2019-05-05
Support TAP files that don't have 0x8000 in bytes block reserved field

ciarant:
Contrary to what the file format spec describes the reserved field in a bytes block doesn't always contain 0x8000.


GreyCat:
Thanks!

PR #150, opened 2019-05-19
Add Python Pickle format

moreati:
Remaining tasks

- [x] Add protocol 4 types, opcodes etc

### DIFF
--- a/serialization/pickle.ksy
+++ b/serialization/pickle.ksy
@@ -143,7 +143,9 @@ types:
         type: str
         encoding: ASCII
         terminator: 0x0a # "\n"
-    doc: Double float, encoded with the ASCII characters [0-9.eE-], or 'nan'.
+    doc: |
+      Double float, encoded with the ASCII characters [0-9.e+-], '-inf', 'inf',
+      or 'nan'.
   long1:
     seq:
       - id: len
--- a/serialization/pickle.ksy
+++ b/serialization/pickle.ksy
@@ -5,9 +5,6 @@ meta:
   file-extension:
     - pickle
     - pkl
-    #- pkl0
-    #- pkl1
-    #- pkl2
   license: CC0-1.0
   endian: le
 doc: |
--- a/serialization/pickle.ksy
+++ b/serialization/pickle.ksy
@@ -116,11 +116,15 @@ types:
   stringnl:
     seq:
       - id: val
+        type: str
+        encoding: ASCII
         terminator: 0x0a # "\n"
     doc: Quoted string, possibly containing Python string escapes.
   stringnl_noescape:
     seq:
       - id: val
+        type: str
+        encoding: ASCII
         terminator: 0x0a # "\n"
     doc: Unquoted string, does not contain string escapes.
   stringnl_noescape_pair:
--- a/serialization/pickle.ksy
+++ b/serialization/pickle.ksy
@@ -51,30 +51,30 @@ types:
             'opcode::short_binstring': string1
             'opcode::binbytes': bytes4
             'opcode::short_binbytes': bytes1
-            #'opcode::none': no_arg
-            #'opcode::newtrue': no_arg
-            #'opcode::newfalse': no_arg
+            'opcode::none': no_arg
+            'opcode::newtrue': no_arg
+            'opcode::newfalse': no_arg
             'opcode::unicode': unicodestringnl
             'opcode::binunicode': unicodestring4
             'opcode::float': floatnl
             'opcode::binfloat': f8
-            #'opcode::empty_list': no_arg
-            #'opcode::append': no_arg
-            #'opcode::appends': no_arg
-            #'opcode::list': no_arg
-            #'opcode::empty_tuple': no_arg
-            #'opcode::tuple': no_arg
-            #'opcode::tuple1': no_arg
-            #'opcode::tuple2': no_arg
-            #'opcode::tuple3': no_arg
-            #'opcode::empty_dict': no_arg
-            #'opcode::dict': no_arg
-            #'opcode::setitem': no_arg
-            #'opcode::setitem2': no_arg
-            #'opcode::pop': no_arg
-            #'opcode::dup': no_arg
-            #'opcode::mark': no_arg
-            #'opcode::pop_mark': no_arg
+            'opcode::empty_list': no_arg
+            'opcode::append': no_arg
+            'opcode::appends': no_arg
+            'opcode::list': no_arg
+            'opcode::empty_tuple': no_arg
+            'opcode::tuple': no_arg
+            'opcode::tuple1': no_arg
+            'opcode::tuple2': no_arg
+            'opcode::tuple3': no_arg
+            'opcode::empty_dict': no_arg
+            'opcode::dict': no_arg
+            'opcode::setitem': no_arg
+            'opcode::setitems': no_arg
+            'opcode::pop': no_arg
+            'opcode::dup': no_arg
+            'opcode::mark': no_arg
+            'opcode::pop_mark': no_arg
             'opcode::get': decimalnl_short
             'opcode::binget': u1
             'opcode::long_binget': u4
@@ -85,15 +85,15 @@ types:
             'opcode::ext2': u2
             'opcode::ext4': u4
             'opcode::global':  stringnl_noescape_pair
-            #'opcode::reduce': no_arg
-            #'opcode::build': no_arg
+            'opcode::reduce': no_arg
+            'opcode::build': no_arg
             'opcode::inst': stringnl_noescape_pair
-            #'opcode::obj': no_arg
-            #'opcode::newobj': no_arg
+            'opcode::obj': no_arg
+            'opcode::newobj': no_arg
             'opcode::proto': u1
-            #'opcode::stop': no_arg
+            'opcode::stop': no_arg
             'opcode::persid': stringnl_noescape
-            #'opcode::binpersid': no_arg
+            'opcode::binpersid': no_arg
         doc: |
           Optional argument for the operation. Data type and length
           are determined by the value of the opcode.
@@ -204,8 +204,8 @@ types:
         encoding: utf8
         size: len
     doc: Length prefixed string, between 0 and 2**32-1 bytes long
-  # Some opcodes have no associated argument
-  # no_arg: {}
+  no_arg:
+    doc: Some opcodes take no argument, this empty type is used for them.
 
 enums:
   opcode:
--- a/serialization/pickle.ksy
+++ b/serialization/pickle.ksy
@@ -1,0 +1,443 @@
+meta:
+  id: pickle
+  title: Python pickle serialiazation format (protocol 3)
+  application: Python
+  file-extension:
+    - pickle
+    - pkl
+    #- pkl0
+    #- pkl1
+    #- pkl2
+  license: CC0-1.0
+  endian: le
+doc: |
+  Python Pickle format serializes Python objects to a byte stream, as a sequence
+  of operations to run on the Pickle Virtual Machine.
+
+  The format is mostly implementation defined, there is no formal specification.
+  Pickle data types are closely coupled to the Python object model.
+  Python singletons, and most builtin types (e.g. `None`, `int`,`dict`, `list`)
+  are serialised using dedicated Pickle opcodes.
+  Other builtin types, and all classes  (e.g. `set`, `datetime.datetime`) are
+  serialised by encoding the name of a constructor callable.
+  They are deserialised by importing that constructor, and calling it.
+doc-ref:  https://github.com/python/cpython/blob/3.3/Lib/pickletools.py
+seq:
+  # TODO is there a way to declare PROTO is optional, but only valid at position 0?
+  - id: ops
+    type: op
+    repeat: eos
+  # TODO is there a way to declare a trailing STOP is required?
+types:
+  op:
+    seq:
+      - id: code
+        type: u1
+        enum: opcode
+        doc: |
+          Operation code that determines which action should be
+          performed next by the Pickel Virtual Machine. Some opcodes
+          are only available in later versions of the Pickle protocol.
+      - id: arg
+        type:
+          switch-on: code
+          cases:
+            'opcode::int': decimalnl_short
+            'opcode::binint': s4
+            'opcode::binint1': u1
+            'opcode::binint2': u2
+            'opcode::long': decimalnl_long
+            'opcode::long1': long1
+            'opcode::string': stringnl
+            'opcode::binstring': string4
+            'opcode::short_binstring': string1
+            'opcode::binbytes': bytes4
+            'opcode::short_binbytes': bytes1
+            #'opcode::none': no_arg
+            #'opcode::newtrue': no_arg
+            #'opcode::newfalse': no_arg
+            'opcode::unicode': unicodestringnl
+            'opcode::binunicode': unicodestring4
+            'opcode::binfloat': f8
+            #'opcode::empty_list': no_arg
+            #'opcode::append': no_arg
+            #'opcode::appends': no_arg
+            #'opcode::list': no_arg
+            #'opcode::empty_tuple': no_arg
+            #'opcode::tuple': no_arg
+            #'opcode::tuple1': no_arg
+            #'opcode::tuple2': no_arg
+            #'opcode::tuple3': no_arg
+            #'opcode::empty_dict': no_arg
+            #'opcode::dict': no_arg
+            #'opcode::setitem': no_arg
+            #'opcode::setitem2': no_arg
+            #'opcode::pop': no_arg
+            #'opcode::dup': no_arg
+            #'opcode::mark': no_arg
+            #'opcode::pop_mark': no_arg
+            'opcode::get': decimalnl_short
+            'opcode::binget': u1
+            'opcode::long_binget': u4
+            'opcode::put': decimalnl_short
+            'opcode::binput': u1
+            'opcode::long_binput': u4
+            'opcode::ext1': u1
+            'opcode::ext2': u2
+            'opcode::ext4': u4
+            'opcode::global':  stringnl_noescape_pair
+            #'opcode::reduce': no_arg
+            #'opcode::build': no_arg
+            'opcode::inst': stringnl_noescape_pair
+            #'opcode::obj': no_arg
+            #'opcode::newobj': no_arg
+            'opcode::proto': u1
+            #'opcode::stop': no_arg
+            'opcode::persid': stringnl_noescape
+            #'opcode::binpersid': no_arg
+        doc: |
+          Optional argument for the operation. Data type and length
+          are determined by the value of the opcode.
+
+  decimalnl_short:
+    seq:
+      - id: val
+        type: str
+        encoding: ASCII
+        terminator: 0x0a # "\n"
+    doc: Integer, encoded with the ASCII characters [0-9-].
+  decimalnl_long:
+    seq:
+      - id: val
+        type: str
+        encoding: ASCII
+        terminator: 0x0a # "\n"
+    doc: Integer, encoded with the ASCII chracters [0-9-], followed by 'L'.
+  # TODO Can kaitai express constraint that these are quoted?
+  stringnl:
+    seq:
+      - id: val
+        terminator: 0x0a # "\n"
+    doc: Quoted string, possibly containing Python string escapes.
+  stringnl_noescape:
+    seq:
+      - id: val
+        terminator: 0x0a # "\n"
+    doc: Unquoted string, does not contain string escapes.
+  stringnl_noescape_pair:
+    seq:
+      - id: val1
+        type: stringnl_noescape
+      - id: val2
+        type: stringnl_noescape
+    doc: Pair of unquoted, unescaped strings.
+  unicodestringnl:
+    seq:
+      - id: val
+        type: str
+        encoding: ASCII
+        terminator: 0x0a # "\n"
+    doc: Unquoted string, containing Python Unicode escapes.
+  floatnl:
+    seq:
+      - id: val
+        type: str
+        encoding: ASCII
+        terminator: 0x0a # "\n"
+    doc: Double float, encoded with the ASCII characters [0-9.eE-], or 'nan'.
+  long1:
+    seq:
+      - id: len
+        type: u1
+      - id: val
+        size: len
+    doc: |
+      Large signed integer, in the range -2**(8*255-1) to 2**(8*255-1)-1,
+      encoded as two's complement.
+  long4:
+    seq:
+      - id: len
+        type: u4
+      - id: val
+        size: len
+    doc: |
+      Large signed integer, in the range -2**(8*2**32-1) to 2**(8*2**32-1)-1,
+      encoded as two's complement.
+  string1:
+    seq:
+      - id: len
+        type: u1
+      - id: val
+        type: str
+        encoding: latin1
+        size: len
+    doc: Length prefixed string, between 0 and 255 bytes long.
+  string4:
+    seq:
+      - id: len
+        # Not a typo, the length really is a signed integer
+        type: s4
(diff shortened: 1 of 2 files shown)


serialization/pickle.ksy
```diff
@@ -0,0 +1,443 @@
+meta:
+  id: pickle
+  title: Python pickle serialiazation format (protocol 3)
+  application: Python
+  file-extension:
+    - pickle
+    - pkl
+    #- pkl0
```
KOLANICH: Why not?

serialization/pickle.ksy
```diff
@@ -0,0 +1,443 @@
...
+  Other builtin types, and all classes  (e.g. `set`, `datetime.datetime`) are
+  serialised by encoding the name of a constructor callable.
+  They are deserialised by importing that constructor, and calling it.
+doc-ref:  https://github.com/python/cpython/blob/3.3/Lib/pickletools.py
+seq:
+  # TODO is there a way to declare PROTO is optional, but only valid at position 0?
+  - id: ops
+    type: op
+    repeat: eos
+  # TODO is there a way to declare a trailing STOP is required?
```
KOLANICH: [`contents`](https://doc.kaitai.io/user_guide.html#_checking_for_magic_signatures)

serialization/pickle.ksy
moreati: Sorry, I meant to take those out. `.pkl[0-4]` are some personal conventions I use. They aren't adopted anywhere else that I know of.

serialization/pickle.ksy
```diff
@@ -0,0 +1,443 @@
...
+  The format is mostly implementation defined, there is no formal specification.
+  Pickle data types are closely coupled to the Python object model.
+  Python singletons, and most builtin types (e.g. `None`, `int`,`dict`, `list`)
+  are serialised using dedicated Pickle opcodes.
+  Other builtin types, and all classes  (e.g. `set`, `datetime.datetime`) are
+  serialised by encoding the name of a constructor callable.
+  They are deserialised by importing that constructor, and calling it.
+doc-ref:  https://github.com/python/cpython/blob/3.3/Lib/pickletools.py
+seq:
+  # TODO is there a way to declare PROTO is optional, but only valid at position 0?
```
KOLANICH: There is, but it would require some `_io` hacks.

serialization/pickle.ksy
dgelessus: This doesn't work in combination with `repeat: eos` though, right? (The array will always consume all of the data, so the `contents` field will never have any data left and will always fail.)

If `STOP` can only appear at the end of the data, you could use `repeat: until` to terminate the array after the first `STOP`.

serialization/pickle.ksy
```diff
@@ -0,0 +1,443 @@
...
+        encoding: ASCII
+        terminator: 0x0a # "\n"
+    doc: Unquoted string, containing Python Unicode escapes.
+  floatnl:
+    seq:
+      - id: val
+        type: str
+        encoding: ASCII
+        terminator: 0x0a # "\n"
+    doc: Double float, encoded with the ASCII characters [0-9.eE-], or 'nan'.
```
dgelessus: How are infinities handled?

serialization/pickle.ksy
```diff
@@ -0,0 +1,443 @@
...
+    doc: |
+      Large signed integer, in the range -2**(8*2**32-1) to 2**(8*2**32-1)-1,
+      encoded as two's complement.
+  string1:
+    seq:
+      - id: len
+        type: u1
+      - id: val
+        type: str
+        encoding: latin1
```
dgelessus: Where does this encoding come from? Is it known to always be Latin-1? I'm asking because some formats, especially older ones, appear to use Latin-1 as their string encoding, but they actually use a system-/locale-dependent default encoding (which might often be Latin-1, but could also be something else), or treat all strings as raw bytes without assuming any encoding.

serialization/pickle.ksy
```diff
@@ -0,0 +1,443 @@
...
+            'opcode::global':  stringnl_noescape_pair
+            #'opcode::reduce': no_arg
+            #'opcode::build': no_arg
+            'opcode::inst': stringnl_noescape_pair
+            #'opcode::obj': no_arg
+            #'opcode::newobj': no_arg
+            'opcode::proto': u1
+            #'opcode::stop': no_arg
+            'opcode::persid': stringnl_noescape
+            #'opcode::binpersid': no_arg
```
dgelessus: All of the `no_arg` cases are commented out here - how are no-arg operations handled then? As far as I can tell, they would currently cause errors, because none of the cases would match.

serialization/pickle.ksy
```diff
@@ -0,0 +1,443 @@
+meta:
+  id: pickle
+  title: Python pickle serialiazation format (protocol 3)
```
dgelessus: There's a small typo in "seriali**a**zation".

serialization/pickle.ksy
```diff
@@ -0,0 +1,443 @@
+meta:
+  id: pickle
```
dgelessus: I think it would be a good idea to rename this spec to `python_pickle` - the name `pickle` on its own is not that clear if you don't already know how Python's serialization format is called.

Also I just noticed that when you compile this spec to Python with the default ksc options, it produces a file called `pickle.py`, which conflicts with Python's `pickle` module 😃

serialization/pickle.ksy
KOLANICH: `_io` magic is to the rescue. One can determine the stream size and create a substream containing all but the last bytes. So the last bytes are to be used for `contents` and the rest is occupied by a substream containing `repeat: eos`

serialization/pickle.ksy
KOLANICH: this type seems to be never referenced

serialization/pickle.ksy
```diff
@@ -0,0 +1,443 @@
...
+        type: str
+        encoding: ASCII
+        terminator: 0x0a # "\n"
+    doc: Integer, encoded with the ASCII chracters [0-9-], followed by 'L'.
+  # TODO Can kaitai express constraint that these are quoted?
+  stringnl:
+    seq:
+      - id: val
+        terminator: 0x0a # "\n"
+    doc: Quoted string, possibly containing Python string escapes.
```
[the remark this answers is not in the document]
moreati: > How exactly is this string quoted?

The opening and closing quote characters will either be a pair of `'`' or a pair of `"`. No triple quoted (`'''abc'''`), raw (`r'abc'`), unicode (u'abc'), or byte (`b'abc'`) strings.

> Is this (and stringnl_noescape below) a byte string or a Unicode string?

The field was designed to store a Python 2.x `str` object. Python 2.x is a bit fuzzy on the byte string/unicode string distinction. The contents are the result of `repr(some_str) + '\n'`. So when produced by Python 2.x the encoding will be ASCII.

I need to confirm what contortions Python 3.x performs.

serialization/pickle.ksy
moreati: > How are infinities handled?

Good question, almost certainly as `inf\n`. I'll confirm, consider other corner cases, and update the doc.

moreati: > this type seems to be never referenced. pickling a float results in an IEEE float being stored.

`cases:` above is missing a case. Nice spot.

serialization/pickle.ksy
moreati: I was unsure of the best way to handle these. Having an empty arg field seemed a bit redundant, on the other hand having a uniform shape for the op type has some appeal. I'm open to either approach.


> As far as I can tell, they would currently cause errors

Not that I've seen so far. I left it this way after getting no errors from the web IDE, when parsing a few samples.

serialization/pickle.ksy
moreati: Will do.

serialization/pickle.ksy
KOLANICH: >As far as I can tell, they would currently cause errors, because none of the cases would match.

No, it is a completely valid situation. If none of `cases` matches, the field is assigned to `None` and its analogues and consumes `0` space

dgelessus: Actually it seems that if no case matches, the field is not parsed at all, which (in Python) results in a missing attribute:

```python
In [1]: !ksc --target=python python_pickle.ksy 

In [2]: import python_pickle                                                                        

In [3]: import pickle                                                                               

In [4]: thing = python_pickle.PythonPickle.from_bytes(pickle.dumps(123))                            

In [5]: [op.code for op in thing.ops]                                                               
Out[5]: [<Opcode.proto: 128>, <Opcode.binint1: 75>, <Opcode.stop: 46>]

In [7]: thing.ops[0].arg                                                                            
Out[7]: 3

In [8]: thing.ops[1].arg                                                                            
Out[8]: 123

In [9]: thing.ops[2].arg                                                                            
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-9-edc885b8c4bf> in <module>
----> 1 thing.ops[2].arg

AttributeError: 'Op' object has no attribute 'arg'
```

I think having explicit `no_arg` cases would be better, so that the code using the spec can always safely access `arg`. You could do this with a default case (`_: no_arg`), but I think listing each no-arg opcode separately would be better, so that you can tell apart opcodes that really have no arguments and ones that just haven't been implemented.

serialization/python_pickle.ksy
```diff
@@ -0,0 +1,448 @@
+meta:
+  id: python_pickle
+  title: Python pickle serialization format (protocol 3)
```
dgelessus: Since the Pickle data format is backwards-compatible, maybe the "protocol 3" part should be changed to "protocol 1 through 3".

Or perhaps this information should go into the `doc` block rather than the title. If the id or title of a spec mentions a specific format version, that usually indicates that there are other versions of the format that are incompatible enough to need their own spec. For the Pickle format this is not the case - because it is backwards-compatible, a single spec should be enough for all versions of the format, so there's no need to differentiate between versions in the title.

serialization/python_pickle.ksy
```diff
@@ -0,0 +1,448 @@
...
+  application: Python
+  file-extension:
+    - pickle
+    - pkl
```
dgelessus: The Pickle format has a [Wikidata entry](https://www.wikidata.org/wiki/Q7190889) and a [JSTFFP page](http://fileformats.archiveteam.org/wiki/Pickle), these could be added in an [`xref` block](http://doc.kaitai.io/user_guide.html#meta-xref).

serialization/python_pickle.ksy
```diff
@@ -0,0 +1,448 @@
...
+  # TODO is there a way to declare a trailing STOP is required?
+types:
+  op:
+    seq:
+      - id: code
+        type: u1
+        enum: opcode
+        doc: |
+          Operation code that determines which action should be
+          performed next by the Pickel Virtual Machine. Some opcodes
```
dgelessus: Small typo: Pickel should be Pickle.

serialization/pickle.ksy
moreati: I'm investigating this, but it may be a few days the answer deserves citations

serialization/python_pickle.ksy
moreati: > perhaps this information should go into the doc block

Agreed, modulo protocol 0 is the earliest version.

serialization/pickle.ksy
moreati: Boy were you right to question this

moreati: @dgelessus I hope https://github.com/kaitai-io/kaitai_struct_formats/pull/150/commits/31ea16e013c2627d3c9c80e86e428d63b34441b9#diff-b8c96cd5bf4bd59885b667523ba9466bR184 answers this to your satisfaction

dgelessus: Ah, typical Python 2 encoding handling I see 😃 I agree with changing the type here from `str` to a plain byte array - when nothing is known for certain about the encoding, it makes sense to keep it as bytes and let the user of the spec deal with it.

Thank you for doing the research on this! It's good to have edge cases like this documented properly, especially when other sources are unclear on the topic.

serialization/python_pickle.ksy
```diff
@@ -0,0 +1,588 @@
...
+
+  no_arg:
+    doc: Some opcodes take no argument, this empty type is used for them.
+
+enums:
+  opcode:
+    0x28: # "("
+        id: "mark"
+        -orig-id: MARK
+        doc: push special markobject on stack
```
GreyCat: Looks like an extra indentation in all these enums?

dgelessus:
By the way, [PEP 574](https://www.python.org/dev/peps/pep-0574/), which proposes version 5 of the Pickle protocol, has been accepted for Python 3.8. This may be interesting in the future, but right now it's way too early to support version 5 - it hasn't even been implemented in CPython yet, and the PEP is missing some important information, like concrete byte values for the new opcodes.

moreati:
> PEP 574, which proposes version 5 of the Pickle protocol, has been accepted for Python 3.8

Huh, thank you. Thought I'd seen that a a while back, and it had been rejected. Must have confused it with another proposal.

serialization/python_pickle.ksy
```diff
@@ -0,0 +1,588 @@
...
+  * Protocol 0: ASCII clean, no explicit version, fields are '\n' terminated.
+  * Protocol 1: Binary, no explicit version, first length prefixed types.
+  * Protocol 2: Python 2.3+. Explicit versioning, more length prefixed types.
+    https://www.python.org/dev/peps/pep-0307/
+  * Protocol 3: Python 3.0+. Dedicated opcodes for `bytes` objects.
+  * Protocol 4: Python 3.4+. Opcodes for 64 bit strings, framing, `set`.
+    https://www.python.org/dev/peps/pep-3154/
+doc-ref: https://github.com/python/cpython/blob/v3.7.3/Lib/pickletools.py
+seq:
+  # TODO is there a way to declare PROTO is optional, but only valid at position 0?
```
KOLANICH: @moreati,  are you going to fix this and that? I have given you info you have requested.

More precisely, I meant something like
```yaml
seq:
  pos:
    - id: first_op
      type: u1
instances:
  after_or_just:
    value: first_op == 0x80
  main_after:
    pos: 1
    type: pickle_main
    if: after_or_just
  main_just:
    pos: 0
    if: not after_or_just
  main:
    value: "(after_or_just?main_after:main_just)"
types:
   pickle_main:
      ....
```

serialization/python_pickle.ksy
```diff
@@ -0,0 +1,588 @@
...
+  - id: ops
+    type: op
+    repeat: eos
+  # TODO is there a way to declare a trailing STOP is required?
```
KOLANICH: As I have said, there is. Somethkng like these, though there may be better ways to express this, I just have no time to play with it now:

```yaml
seq:
  - id: placeholder # to create a substream
    size-eos: true # if I remember right
instances:
  main:
    io: placeholder._io
    pos: 0
    size: placeholder._size -1
  assert_stop:
    pos: placeholder._size -1
    type: assert_stop
types:
  placeholder: {}
  assert_stop:
    seq:
      - contents: ["."]
```

serialization/python_pickle.ksy
```diff
@@ -0,0 +1,588 @@
...
+            'opcode::binpersid': no_arg
+        doc: |
+          Optional argument for the operation. Data type and length
+          are determined by the value of the opcode.
+
+  decimalnl_short:
+    seq:
+      - id: val
+        type: str
+        encoding: ASCII
```
KOLANICH: 6 `ASCII` vs 3 `utf8`. May make sense to move `ascii` into `meta`.

serialization/python_pickle.ksy
```diff
@@ -0,0 +1,588 @@
...
+    0x75: # "u"
+      id: "setitems"
+      -orig-id: SETITEMS
+      doc: modify dict by adding topmost key+value pairs
+    0x47: # "G"
+      id: "binfloat"
+      -orig-id: BINFLOAT
+      doc: push float; arg is 8-byte float encoding
+
+    #'I01\n':
```
KOLANICH: ?

serialization/python_pickle.ksy
dgelessus: > 6 `ASCII` vs 3 `utf8`. May make sense to move `ascii` into `meta`.

Hm, I think in this case it's better to put the encodings on the individual fields. I would use a global `encoding` only if there is one obvious "main" encoding with very few or no exceptions. A ratio of 2:1 is not clear enough IMHO, and since there aren't *that* many string fields in the spec, it wouldn't make a big difference in terms of compactness.

> BTW, the convention is to use lowercase in encodings.

Is it? I searched through the ksf repo and found both uppercase and lowercase used often, with uppercase being a bit more common than lowercase. All the examples in the [user guide](http://doc.kaitai.io/user_guide.html) use uppercase encoding names too. The [style guide](http://doc.kaitai.io/ksy_style_guide.html) doesn't say anything about how encoding names should be written, maybe we should change that?

KOLANICH: >The style guide doesn't say anything about how encoding names should be written, maybe we should change that?

Yes, we should. The discussion about encoding casing is https://github.com/kaitai-io/kaitai_struct/issues/391

serialization/python_pickle.ksy
moreati: This came from an implementation detail in pickle.py. Before `True` and `False` got their own opcodes (`NEWTRUE` and `NEWFALSE`), booleans were encoded by a special value in an `INT` op. I'll remove these, and a note to this effect in the `doc:`.

serialization/python_pickle.ksy
moreati: I'll use lowercase throughout.

serialization/python_pickle.ksy
moreati: I'm leaning toward not doing this here. I'm starting to think of validating inclusion/placement of `PROTO` and `STOP` as semantic concerns - to be checked/enforced at a layer above the parser, along with other op ordering requirements. Encoding _just_ these two checks here makes the format deceleration harder to understand, without sufficient benefit.

Does that sound okay?

serialization/python_pickle.ksy
moreati: Similarly to `PROTO` I now think this is better handled outside this format declaration (and parsers generated from it). Sorry for leaving it hanging so long that you had to write the above.

If you would prefer to enforce this here, then I will try to make it work.

serialization/python_pickle.ksy
KOLANICH: It is definitely sounds OK,  but because of another reasons: the opcodes validated via `contents` are not true objects. IMHO the KS spec should have this encoded because it is a spec, it should be self-sufficient in the sense it should be possible to throw away the original spec and rely only on Kaitai one. Also we want assertions functionality (run-time toggleable) and a serializer (this implies a symbolic engine and an SMT-solver in order to automatically resolve the assertions and expressions). And IMHO the constraints on the first and last opcode should be encoded as assertions against objects, not raw bytes, but they are not yet implemented.

serialization/python_pickle.ksy
```diff
@@ -0,0 +1,585 @@
...
+      -orig-id: SETITEMS
+      doc: modify dict by adding topmost key+value pairs
+    0x47: # "G"
+      id: "binfloat"
+      -orig-id: BINFLOAT
+      doc: push float; arg is 8-byte float encoding
+
+    # Protocol 2
+    0x80:
+      id: "proto"
```
KOLANICH: quoting is unneeded

serialization/python_pickle.ksy
```diff
@@ -0,0 +1,585 @@
...
+      -orig-id: PROTO
+      doc: identify pickle protocol
+    0x81:
+      id: "newobj"
```
KOLANICH: Since the original names are preserved in `-orig-id` how about changing the ids? For example adding words delimiters?

serialization/python_pickle.ksy
```diff
@@ -0,0 +1,585 @@
...
+    doc: |
+      Integer or boolean, encoded with the ASCII characters [0-9-].
+
+      The values '00' and '01' encode the Python values `False` and `True`.
+      Normally a value would not contain leading '0' characters.
+
+  decimalnl_long:
+    seq:
+      - id: val
+        type: str
```
KOLANICH: Shouldn't they be `strz`? Does the compiler accept them? @GreyCat, isn't it a bug?

serialization/python_pickle.ksy
moreati: My mistake. I thought `id: none` needed protection from the YAML parser, and I quoted the others for consistency. However the special case is `null`.

serialization/python_pickle.ksy
moreati: Not sure I understand, do you mean something like `id: new_object`?

KOLANICH: for this var, exactly this.

serialization/python_pickle.ksy
moreati: The values are definitely not terminated by `\0`. In general Python doesn't use NUL terminated strings, and `\0` is allowed anywhere in a `bytes`, or `str`/`unicode` object.

serialization/python_pickle.ksy
moreati: Should I treat the type identifiers similarly? e.g. `unicodestring8` -> `unicode_string8`

serialization/python_pickle.ksy
KOLANICH: `strz` is for any terminated strings, `\0` is just the default terminator. `str` is for strings with explicitly specified size.

moreati: Okay, but please be aware that's not the impression I got from http://doc.kaitai.io/ksy_reference.html#_strings

> To designate attribute as string type, use `type: str` and provide encoding info [...] `type: strz` can be also used as a shortcut to define a null-terminated string (C-style). I.e. these foo and bar attributes are equivalent:
> 
> ```
> - id: foo
>   type: str
>   terminator: 0
> - id: bar
>   type: strz
> ```

KOLANICH:
you are completely right and I was wrong: `strz` is just an alias to `str` with zeto terminator:

https://github.com/kaitai-io/kaitai_struct_compiler/blob/21ec2e7b91b5e2f68465e74813f20b1da4fd5d8c/shared/src/main/scala/io/kaitai/struct/datatype/DataType.scala#L347L351

So don't change the stuff, it is OK in the current state.

serialization/python_pickle.ksy
KOLANICH: you are completely right and I was wrong: `strz` is just an alias to `str` with zeto terminator:

https://github.com/kaitai-io/kaitai_struct_compiler/blob/21ec2e7b91b5e2f68465e74813f20b1da4fd5d8c/shared/src/main/scala/io/kaitai/struct/datatype/DataType.scala#L347L351

So don't change the stuff, it is OK in the current state.

serialization/python_pickle.ksy
KOLANICH: Of course you should!

>For the current type identifiers I used the `ArgumentDescriptor.name` attributes in `pickletools.py`

We have `-orig-id` for that, right?

moreati: > We have `-orig-id` for that, right?

That covers the opcode names in the enum. I was asking about the names of the argument _types_, further up. Those are of the form

```
unicodestring8:
  seq:
    id: len
    ...
```

from my reading of the KSY spec I could add an `-orig-id` to `len`, but not to `unicodestring8` as a whole. Did I miss something?

KOLANICH: Ahy attr beginning from a dash can be added anywhere - they are ignored by KSC.

GreyCat:
Looks like vast majority of pressing issues was resolved, so I'll go ahead and merge this in. Thanks for this tremendous effort, everyone, especially @moreati!

PR #159, opened 2019-06-11
Added spider.sav

KOLANICH:


KOLANICH:
@GreyCat

common/random/lcg.ksy
```diff
@@ -0,0 +1,42 @@
+meta:
+  id: lcg
+  title: Linear Congruential Generator
+  xref:
+    iso: "28640:2010"
+    wikidata: Q1190228
+  license: Unlicense
+  endian: le
+doc: |
+  A simple non-secure pseudorandom generator.
```
GreyCat: A very interesting idea (and certainly very clever hack with this recursive approach to keep state). What we'll definitely need is:

* A better name. "LCG" is "London Capital Group", "Living Card Game", "Large Hadron Collider Computing Grid", "London Computer Group", certain airports, train stations, venues, and probably tons of other things. Even `random_lcg` will be a better name that won't clash with all that.
* More documentation — at least a very basic doc that explains what's going on and how to use it both from KSY author perspective and from end user's perspective.

common/random/lcg.ksy
```diff
@@ -0,0 +1,42 @@
...
+params:
+  - id: a
+    type: u4
+  - id: b
+    type: u4
+  - id: value_bits
+    type: u1
+  - id: initial_seed
+    type: u4
```
GreyCat: Technically, this implementation in 32 bits + 8-bit value is specific to Spider Solitaire. I'm not 100% sure, but may be it warrants rebadging it as `random_lcg_32_8` or something like that?

game/spider_sav.ksy
```diff
@@ -0,0 +1,148 @@
...
+  title: Windows XP Spider Solitaire save file
+  application:
+    - Windows Spider Solitaire
+  file-extension: sav
+  xref:
+    wikidata: Q2512977
+  license: Unlicense
+  endian: le
+  imports:
+    - ../common/random/lcg
```
GreyCat: Please use `/common/...` to avoid dependency on position with common components.

KOLANICH:
@generalmimon

PR #164, opened 2019-06-12
Added smbios with some tables.

tinrodriguez8:
Added [smbios format](https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.1.1.pdf).

The current .ksy contains few tables like:

- Bios information
- System information
- Processor information
- Portable Battery information

To add more tables, create a new file with the table specification.
Also, there is a special name '**skipper**', that is used to parse any table who is not specified.

### DIFF
--- a/firmware/smbios/bios_information.ksy
+++ b/firmware/smbios/bios_information.ksy
@@ -1,0 +1,42 @@
+meta:
+  id: bios_information
+  endian: le
+  imports:
+    - strings
+seq:
+  - id: length
+    type: u1
+  - id: handle_index
+    type: u2
+  - id: vendor_index
+    type: u1
+  - id: bios_version
+    type: u1
+  - id: bios_starting_address_segment
+    type: u2
+  - id: bios_release_date_index
+    type: u1
+  - id: bios_rom_size
+    type: u1
+  - id: bios_characteristics
+    type: u8
+  - id: bios_characteristics_ext_bytes
+    size: length - 0x12
+  - id: strings_array
+    type: strings
+    # Repeat until [00, 00] is matched, this mean, double terminator byte (0)
+    # the trick here is _.string.length is the last string we have parsed and
+    # strings_array has to contain at least 2 elements.
+    repeat: until
+    repeat-until: _.string.length == 1 and strings_array.size > 1
+instances:
+  handle:
+    value: strings_array[handle_index - 1]
+    if: handle_index > 0
+  vendor:
+    value: strings_array[vendor_index - 1]
+    if: vendor_index > 0
+  bios_release_date:
+    value: strings_array[bios_release_date_index - 1]
+    if: bios_release_date_index > 0
+
--- a/firmware/smbios/smbios.ksy
+++ b/firmware/smbios/smbios.ksy
@@ -1,0 +1,37 @@
+meta:
+  id: smbios
+  endian: le
+  imports:
+    - bios_information
+    - system_information
+    - processor_information
+    - portable_battery
+    - skipper
+    - strings
+seq:
+  - id: tables
+    type: table
+    repeat: eos
+types:
+  table:
+    seq:
+      - id: type
+        type: u1
+        enum: type_enum
+      - id: table
+        type:
+          switch-on: type
+          cases:
+            # TODO: Add rest of the tables
+            'type_enum::table0': bios_information
+            'type_enum::table1': system_information
+            'type_enum::table4': processor_information
+            'type_enum::table22': portable_battery
+            _: skipper
+    enums:
+      type_enum:
+        0x00: table0
+        0x01: table1
+        0x04: table4
+        0x16: table22
+
--- a/firmware/smbios/portable_battery.ksy
+++ b/firmware/smbios/portable_battery.ksy
@@ -1,0 +1,71 @@
+meta:
+  id: portable_battery
+  endian: le
+  imports:
+    - strings
+seq:
+  - id: length
+    type: u1
+  - id: handle
+    type: u2
+  - id: location_index
+    type: u1
+  - id: manufacturer_index
+    type: u1
+  - id: manufacture_date_index
+    type: u1
+  - id: serial_number_index
+    type: u1
+  - id: device_name_index
+    type: u1
+  - id: device_chemistry
+    type: u1
+  - id: design_capacity
+    type: u2
+  - id: design_voltaje
+    type: u2
+  - id: sbds_version_number_index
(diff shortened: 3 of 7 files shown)


firmware/smbios/bios_information.ksy
```diff
@@ -0,0 +1,42 @@
+meta:
+  id: bios_information
```
GreyCat: Please note that we effectively have a flat ID name space, so this `bios_information` will likely clash with other formats. Please consider being more specific and using some kind of a prefix, such as `smbios_bios_info`.

tinrodriguez8: Changed! but still having tests failed on CircleCI.

firmware/smbios/smbios.ksy
```diff
@@ -0,0 +1,37 @@
+meta:
+  id: smbios
+  endian: le
+  imports:
+    - smbios_bios_info
+    - smbios_system_info
+    - smbios_processor_info
+    - smbios_portable_battery
+    - smbios_skipper
+    - smbios_strings
```
GreyCat: This is obviously the main entry point to this project. It really warrants at least some documentation, explaining:

* What SMBIOS is
* How it's used
* How to get SMBIOS file dumps / access it in memory, etc

firmware/smbios/smbios.ksy
```diff
@@ -0,0 +1,37 @@
...
+seq:
+  - id: tables
+    type: table
+    repeat: eos
+types:
+  table:
+    seq:
+      - id: type
+        type: u1
+        enum: type_enum
```
GreyCat: From what I've observed so far, it looks like all SMBIOS tables have common header parts, like length + handle. Why not parse them here, along with `type`?

firmware/smbios/smbios.ksy
```diff
@@ -0,0 +1,37 @@
...
+            'type_enum::table1': smbios_system_info
+            'type_enum::table4': smbios_processor_info
+            'type_enum::table22': smbios_portable_battery
+            _: smbios_skipper
+    enums:
+      type_enum:
+        0x00: table0
+        0x01: table1
+        0x04: table4
+        0x16: table22
```
GreyCat: Enums are supposed to be human-readable, i.e. `type_enum::portable_battery`, not `type_enum::table22`. There's  no value in having enum like this — if you want to skip it, it's better to just use raw integers.

firmware/smbios/smbios.ksy
```diff
@@ -0,0 +1,37 @@
...
+      - id: table
+        type:
+          switch-on: type
+          cases:
+            # TODO: Add rest of the tables
+            'type_enum::table0': smbios_bios_info
+            'type_enum::table1': smbios_system_info
+            'type_enum::table4': smbios_processor_info
+            'type_enum::table22': smbios_portable_battery
+            _: smbios_skipper
```
GreyCat: Ideally, "skipper" should go away, being just the byte array in this case, if all the other common parts of all tables would be extracted to this level.

firmware/smbios/smbios_bios_info.ksy
```diff
@@ -0,0 +1,42 @@
...
+  - id: bios_starting_address_segment
+    type: u2
+  - id: bios_release_date_index
+    type: u1
+  - id: bios_rom_size
+    type: u1
+  - id: bios_characteristics
+    type: u8
+  - id: bios_characteristics_ext_bytes
+    size: length - 0x12
```
GreyCat: Again, if this will be size-bound substream, you could have just used `size-eos: true` on this attribute instead of introducing manual calculations.

firmware/smbios/smbios_portable_battery.ksy
```diff
@@ -0,0 +1,71 @@
...
+  - id: manufacturer_index
+    type: u1
+  - id: manufacture_date_index
+    type: u1
+  - id: serial_number_index
+    type: u1
+  - id: device_name_index
+    type: u1
+  - id: device_chemistry
+    type: u1
```
GreyCat: Probably this is some kind of enum?

firmware/smbios/smbios_portable_battery.ksy
```diff
@@ -0,0 +1,71 @@
...
+  - id: design_capacity
+    type: u2
+  - id: design_voltaje
+    type: u2
```
GreyCat: voltage?

firmware/smbios/smbios_portable_battery.ksy
```diff
@@ -0,0 +1,71 @@
...
+  - id: manufacture_date_index
+    type: u1
+  - id: serial_number_index
+    type: u1
+  - id: device_name_index
+    type: u1
+  - id: device_chemistry
+    type: u1
+  - id: design_capacity
+    type: u2
```
GreyCat: Some documentation on what units are used (i.e. kWh, something else?) would be nice.

firmware/smbios/smbios_portable_battery.ksy
GreyCat: Also, documentation mentioning units would be great too ;)

firmware/smbios/smbios_processor_info.ksy
```diff
@@ -0,0 +1,88 @@
...
+  - id: part_number_index
+    type: u1
+  - id: core_count
+    type: u1
+  - id: core_enabled
+    type: u1
+  - id: thread_count
+    type: u1
+  - id: processor_characteristics
+    type: u2
```
GreyCat: These are probably a bitmask?

firmware/smbios/smbios_strings.ksy
```diff
@@ -0,0 +1,9 @@
+meta:
+  id: smbios_strings
+  endian: le
+seq:
+  - id: string
+    type: strz
+    encoding: UTF-8
+    include: true
+
```
GreyCat: If these will be part of `smbios::table` type, this file won't be needed too.

firmware/smbios/bios_information.ksy
GreyCat: Thanks! This is actually a very good question why these CircleCI jobs get to run and what it is attempting to do. There should be no PR-related jobs running at the CI in this repo, to my knowledge.

KOLANICH:
You may also want to take a look at https://github.com/KOLANICH/kaitai_struct_formats/tree/pnp_optionrom_incomplete , it is incomplete because it intentionally misses `doc`

PR #173, opened 2019-06-16
Add PHP serialized value and phar archive format

dgelessus:
This adds specs for the following two formats, both from the PHP ecosystem:

* `serialization/php_serialized_value`: PHP's serialized data format, as used by the standard `serialize` and `unserialize` functions.
* `archive/phar_without_stub`: The phar (PHP Archive) format, used to package a PHP application or library into a single, self-contained, and optionally executable archive.
  * This spec does not handle the executable PHP stub that precedes every phar archive, because it is currently not possible to express multi-byte string terminators in Kaitai Struct (see kaitai-io/kaitai_struct#158 and kaitai-io/kaitai_struct#538). All other parts of the format are fully specced, and the format of the executable PHP stub is described in text form in the `doc` block.

I'm submitting these as one PR, because the phar format loosely depends on the PHP serialization format (some optional fields in the phar format contain serialized PHP data).

### DIFF
--- a/archive/phar_without_stub.ksy
+++ b/archive/phar_without_stub.ksy
@@ -66,7 +66,7 @@ enums:
       id: openssl
       -orig-id: PHAR_SIG_OPENSSL
       doc: |
-        Indicates an OpenSSL signature. Available since API version 1.1.1. This type is not documented in the phar extension's documentation of the phar format.
+        Indicates an OpenSSL signature. Available since API version 1.1.1, PHP_Archive 0.12.0 (even though it claims to only support API version 1.1.0) and phar extension 1.3.0. This type is not documented in the phar extension's documentation of the phar format.
         
         Note: In older versions of the phar extension, this value was used for an undocumented and unimplemented "PGP" signature type (`PHAR_SIG_PGP`).
 types:
--- a/archive/phar_without_stub.ksy
+++ b/archive/phar_without_stub.ksy
@@ -1,0 +1,217 @@
+meta:
+  id: phar_without_stub
+  title: PHP phar archive (without stub)
+  application: PHP
+  file-extension: phar
+  xref:
+    wikidata: Q1269709
+  license: CC0-1.0
+  ks-version: 0.9
+  imports:
+    - /serialization/php_serialized_value
+  endian: le
+doc: |
+  A phar (PHP archive) file. The phar format is a custom archive format from the PHP ecosystem that is used to package a complete PHP library or application into a single self-contained archive. All phar archives start with an executable PHP stub, which can be used to allow executing or including phar files as if they were regular PHP scripts. PHP 5.3 and later include the phar extension, which adds native support for reading and manipulating phar files.
+  
+  The phar format was originally developed as part of the PEAR library PHP_Archive, first released in 2005. Later, a native PHP extension named "phar" was developed, which was first released on PECL in 2007, and is included with PHP 5.3 and later. The phar extension has effectively superseded the PHP_Archive library, which has not been updated since 2010. The phar extension is also no longer released independently on PECL; it is now developed and released as part of PHP itself.
+  
+  Because of current limitations in Kaitai Struct (see kaitai-io/kaitai_struct#158 and kaitai-io/kaitai_struct#538), the executable PHP stub that precedes the rest of the archive is not handled by this spec. Before parsing a phar using this spec, the stub must be removed manually.
+  
+  A phar's stub is terminated by the special token `__HALT_COMPILER();` (which may be followed by at most one space, the PHP tag end `?>`, and an optional line terminator). The stub termination sequence is immediately followed by the remaining parts of the phar format, as described in this spec.
+  
+  The phar stub usually contains code that loads the phar and runs a contained PHP file, but this is not required. A minimal valid phar stub is `<?php __HALT_COMPILER();` - such a stub makes it impossible to execute the phar directly, but still allows loading or manipulating it using the phar extension.
+doc-ref:
+  - 'https://www.php.net/manual/en/phar.fileformat.php'
+  - 'https://github.com/php/php-src/tree/master/ext/phar'
+  - 'https://svn.php.net/viewvc/pecl/phar/'
+  - 'https://svn.php.net/viewvc/pear/packages/PHP_Archive/'
+seq:
+  - id: manifest
+    type: manifest
+    doc: The archive's manifest, containing general metadata about the archive and its files.
+  - id: files
+    size: manifest.file_entries[_index].compressed_size
+    repeat: expr
+    repeat-expr: manifest.file_count
+    doc: The contents of each file in the archive (possibly compressed, as indicated by the file's flags in the manifest). The files are stored in the same order as they appear in the manifest.
+  - id: signature
+    type: signature
+    size-eos: true
+    if: manifest.flags.has_signature
+    doc: |
+      The archive's signature - a digest of all archive data before the signature itself.
+      
+      Note: Almost all of the available "signature" types are actually hashes, not signatures, and cannot be used to verify that the archive has not been tampered with. Only the OpenSSL signature type is a true cryptographic signature.
+enums:
+  signature_type:
+    0x1:
+      id: md5
+      -orig-id: PHAR_SIG_MD5
+      doc: Indicates an MD5 hash.
+    0x2:
+      id: sha1
+      -orig-id: PHAR_SIG_SHA1
+      doc: Indicates a SHA-1 hash.
+    0x4:
+      id: sha256
+      -orig-id: PHAR_SIG_SHA256
+      doc: Indicates a SHA-256 hash. Available since API version 1.1.0, PHP_Archive 0.12.0 and phar extension 1.1.0.
+    0x8:
+      id: sha512
+      -orig-id: PHAR_SIG_SHA512
+      doc: Indicates a SHA-512 hash. Available since API version 1.1.0, PHP_Archive 0.12.0 and phar extension 1.1.0.
+    0x10:
+      id: openssl
+      -orig-id: PHAR_SIG_OPENSSL
+      doc: |
+        Indicates an OpenSSL signature. Available since API version 1.1.1. This type is not documented in the phar extension's documentation of the phar format.
+        
+        Note: In older versions of the phar extension, this value was used for an undocumented and unimplemented "PGP" signature type (`PHAR_SIG_PGP`).
+types:
+  serialized_value:
+    seq:
+      - id: raw
+        size-eos: true
+        doc: The serialized value, as a raw byte array.
+    instances:
+      parsed:
+        pos: 0
+        type: php_serialized_value
+        doc: The serialized value, parsed as a structure.
+  file_flags:
+    seq:
+      - id: value
+        type: u4
+        doc: The unparsed flag bits.
+    instances:
+      permissions:
+        value: value & 0x1ff
+        -orig-id: PHAR_ENT_PERM_MASK
+        doc: The file's permission bits.
+      zlib_compressed:
+        value: (value & 0x1000) != 0
+        -orig-id: PHAR_ENT_COMPRESSED_GZ
+        doc: Whether this file's data is stored using zlib compression.
+      bzip2_compressed:
+        value: (value & 0x2000) != 0
+        -orig-id: PHAR_ENT_COMPRESSED_BZ2
+        doc: Whether this file's data is stored using bzip2 compression.
+  file_entry:
+    seq:
+      - id: filename_size
+        type: u4
+        doc: The size of the file name, in bytes.
+      - id: filename
+        size: filename_size
+        doc: The name of this file. If the name ends with a slash, this entry represents a directory, otherwise a regular file. Directory entries are supported since phar API version 1.1.1. (Explicit directory entries are only needed for empty directories. Non-empty directories are implied by the files located inside them.)
+      - id: uncompressed_size
+        type: u4
+        doc: The size of the file's data when uncompressed, in bytes.
+      - id: timestamp
+        type: u4
+        doc: The time at which the file was added or last updated, as a Unix timestamp.
+      - id: compressed_size
+        type: u4
+        doc: The size of the file's data when compressed, in bytes.
+      - id: crc32
+        type: u4
+        doc: The CRC32 checksum of the file's uncompressed data.
+      - id: flags
+        type: file_flags
+        doc: Flags for this file.
+      - id: metadata_size
+        type: u4
+        doc: The size of the metadata, in bytes, or 0 if there is none.
+      - id: metadata
+        size: metadata_size
+        type: serialized_value
+        if: metadata_size != 0
+        doc: Metadata for this file, in the format used by PHP's `serialize` function. The meaning of the serialized data is not specified further, it may be used to store arbitrary custom data about the file.
+  api_version:
+    meta:
+      endian: be
+    seq:
+      - id: release
+        type: b4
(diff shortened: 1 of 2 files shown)


archive/phar_without_stub.ksy
```diff
@@ -0,0 +1,217 @@
...
+        pos: 0
+        type: php_serialized_value
+        doc: The serialized value, parsed as a structure.
+  file_flags:
+    seq:
+      - id: value
+        type: u4
+        doc: The unparsed flag bits.
+    instances:
+      permissions:
```
KOLANICH: Why not a bit field type?

archive/phar_without_stub.ksy
```diff
@@ -0,0 +1,217 @@
...
+        -orig-id: PHAR_ENT_COMPRESSED_BZ2
+        doc: Whether this file's data is stored using bzip2 compression.
+  file_entry:
+    seq:
+      - id: filename_size
+        type: u4
+        doc: The size of the file name, in bytes.
+      - id: filename
+        size: filename_size
+        doc: The name of this file. If the name ends with a slash, this entry represents a directory, otherwise a regular file. Directory entries are supported since phar API version 1.1.1. (Explicit directory entries are only needed for empty directories. Non-empty directories are implied by the files located inside them.)
```
KOLANICH: Is it a `str`?

archive/phar_without_stub.ksy
dgelessus: The official documentation specifies them as integer bit masks (see [here](https://www.php.net/manual/en/phar.fileformat.flags.php) and [here](https://www.php.net/manual/en/phar.fileformat.manifestfile.php)). Translating hex masks to bit fields is already not obvious because of bit order issues. In addition, here the flags are stored as a *little-endian* integer, so you also have to deal with the bytes being swapped (compared to the hex masks, which are naturally written in big-endian order).

The `permissions` field is a particular problem, because it spans two bytes. In combination with the little-endian byte swapping, this means that the `permissions` field would need to be split into two bit fields (the 8 least significant bits, then 7 unused bits, then the 1 most significant bit), so you would need to reconstruct it using a value instance anyway. kaitai-io/kaitai_struct#155 is probably relevant here.

This explanation might not be completely accurate, I might be misremembering some things. A while ago I tried implementing this using bit fields and couldn't get it to work after many tries, so I changed it to the integer and bit mask combination and it worked on the first try. Since the official format documentation and the two major implementations treat these fields as integers with bit masks, I think it makes sense for our spec to do the same, especially since the alternative with bit fields would be a pain to write and hard to understand.

archive/phar_without_stub.ksy
dgelessus: Not really. The phar format doesn't specify an encoding for file names, and the PHP phar extension's API lets you use arbitrary byte sequences as file names (because PHP's strings are plain bytes with no specific encoding). In practice, almost all phar filenames are ASCII, but the spec shouldn't assume that.

KOLANICH: Thanks for the info. Could you add it into `doc`?

archive/phar_without_stub.ksy
dgelessus: I don't think using bit fields here makes sense until Kaitai Struct allows switching to LSB-first bit order. Processing little-endian data in MSB-first bit order is too confusing IMHO, and as far as I can tell doesn't offer any big advantage over the current integer mask approach.

archive/phar_without_stub.ksy
```diff
@@ -0,0 +1,219 @@
...
+  
+  A phar's stub is terminated by the special token `__HALT_COMPILER();` (which may be followed by at most one space, the PHP tag end `?>`, and an optional line terminator). The stub termination sequence is immediately followed by the remaining parts of the phar format, as described in this spec.
+  
+  The phar stub usually contains code that loads the phar and runs a contained PHP file, but this is not required. A minimal valid phar stub is `<?php __HALT_COMPILER();` - such a stub makes it impossible to execute the phar directly, but still allows loading or manipulating it using the phar extension.
+  
+  Note: The phar format does not specify any encoding for text fields (stub, alias name, and all file names), so these fields may contain arbitrary binary data. The actual text encoding used in a specific phar file usually depends on the application that created the phar, and on the standard encoding of the system on which the phar was created.
```
GreyCat: Please [wrap these lines to have 80 columns limit](http://doc.kaitai.io/ksy_style_guide.html#doc). Using long lines causes problems with many pieces of our tooling (i.e. GitHub, some text editors, and, ultimately, these result in the source code generated with very long comments, which typically breaks target languages style guides).

archive/phar_without_stub.ksy
```diff
@@ -0,0 +1,219 @@
...
+        value: (value & 0x1000) != 0
+        -orig-id: PHAR_ENT_COMPRESSED_GZ
+        doc: Whether this file's data is stored using zlib compression.
+      bzip2_compressed:
+        value: (value & 0x2000) != 0
+        -orig-id: PHAR_ENT_COMPRESSED_BZ2
+        doc: Whether this file's data is stored using bzip2 compression.
+  file_entry:
+    seq:
+      - id: filename_size
```
GreyCat: Please consider using `len_filename`, as per style guide.

archive/phar_without_stub.ksy
```diff
@@ -0,0 +1,219 @@
...
+        doc: The name of this file. If the name ends with a slash, this entry represents a directory, otherwise a regular file. Directory entries are supported since phar API version 1.1.1. (Explicit directory entries are only needed for empty directories. Non-empty directories are implied by the files located inside them.)
+      - id: uncompressed_size
+        type: u4
+        doc: The size of the file's data when uncompressed, in bytes.
+      - id: timestamp
+        type: u4
+        doc: The time at which the file was added or last updated, as a Unix timestamp.
+      - id: compressed_size
+        type: u4
+        doc: The size of the file's data when compressed, in bytes.
```
GreyCat: Please consider using `len_compressed`, as per style guide.

archive/phar_without_stub.ksy
```diff
@@ -0,0 +1,219 @@
...
+        doc: Whether this file's data is stored using bzip2 compression.
+  file_entry:
+    seq:
+      - id: filename_size
+        type: u4
+        doc: The size of the file name, in bytes.
+      - id: filename
+        size: filename_size
+        doc: The name of this file. If the name ends with a slash, this entry represents a directory, otherwise a regular file. Directory entries are supported since phar API version 1.1.1. (Explicit directory entries are only needed for empty directories. Non-empty directories are implied by the files located inside them.)
+      - id: uncompressed_size
```
GreyCat: Please consider using `len_uncompressed`, as per style guide.

archive/phar_without_stub.ksy
```diff
@@ -0,0 +1,219 @@
...
+        doc: The size of the file's data when compressed, in bytes.
+      - id: crc32
+        type: u4
+        doc: The CRC32 checksum of the file's uncompressed data.
+      - id: flags
+        type: file_flags
+        doc: Flags for this file.
+      - id: metadata_size
+        type: u4
+        doc: The size of the metadata, in bytes, or 0 if there is none.
```
GreyCat: Please consider using `len_metadata`, as per style guide.

archive/phar_without_stub.ksy
```diff
@@ -0,0 +1,219 @@
...
+        value: (value & 0x2000) != 0
+        -orig-id: PHAR_HDR_COMPRESSED_BZ2
+        doc: Whether any of the files in this phar are stored using bzip2 compression.
+      has_signature:
+        value: (value & 0x10000) != 0
+        -orig-id: PHAR_HDR_SIGNATURE
+        doc: Whether this phar contains a signature.
+  manifest:
+    seq:
+      - id: manifest_size
```
GreyCat: `len_manifest`?

archive/phar_without_stub.ksy
```diff
@@ -0,0 +1,219 @@
...
+        type: u4
+        doc: |
+          The size of the manifest, in bytes.
+          
+          Note: The phar extension does not allow reading manifests larger than 100 MiB.
+      - id: file_count
```
GreyCat: `num_files`?

archive/phar_without_stub.ksy
```diff
@@ -0,0 +1,219 @@
...
+        type: u4
+        doc: The number of files in this phar.
+      - id: api_version
+        type: api_version
+        doc: The API version used by this phar manifest.
+      - id: flags
+        type: global_flags
+        doc: Global flags for this phar.
+      - id: alias_size
```
GreyCat: `len_alias`?

archive/phar_without_stub.ksy
```diff
@@ -0,0 +1,219 @@
...
+        type: u4
+        doc: The size of the alias, in bytes.
+      - id: alias
+        size: alias_size
+        doc: The phar's alias, i. e. the name under which it is loaded into PHP.
+      - id: metadata_size
```
GreyCat: `len_metadata`?

serialization/php_serialized_value.ksy
```diff
@@ -0,0 +1,260 @@
+meta:
+  id: php_serialized_value
+  title: Serialized PHP value
+  application: PHP
+  file-extension: metadata.bin # Metadata stored as files in tar-based phar archives
```
GreyCat: Um, do they really routinely have `metadata.bin` extension? If not, it's better not to use this field, as this will create more confusion, rather just mention that in the docs.

serialization/php_serialized_value.ksy
```diff
@@ -0,0 +1,260 @@
...
+  # This encoding is only used to parse numbers. All strings, class names, etc. are treated as raw byte arrays, because PHP strings are byte strings with no particular encoding.
+  encoding: ASCII
+doc: |
+  A serialized PHP value, in the format used by PHP's built-in `serialize` and `unserialize` functions. This format closely mirrors PHP's data model: it supports all of PHP's scalar types (`NULL`, booleans, numbers, strings), associative arrays, objects, and recursive data structures using references. The only PHP values not supported by this format are *resources*, which usually correspond to native file or connection handles and cannot be meaningfully serialized.
+  
+  There is no official documentation for this data format; this spec was created based on the PHP source code and the behavior of `serialize`/`unserialize`. PHP makes no guarantees about compatibility of serialized data between PHP versions, but in practice, the format has remained fully backwards-compatible - values serialized by an older PHP version can be unserialized on any newer PHP version. This spec supports serialized values from PHP 7.3 or any earlier version.
```
GreyCat: Again, all `doc` in these file should be ideally wrapper to 80 columns.

serialization/php_serialized_value.ksy
```diff
@@ -0,0 +1,260 @@
...
+      - id: value
+        type: php_serialized_value
+        doc: The value of the entry.
+    doc: A mapping entry consisting of a key and a value.
+  size_prefixed_mapping:
+    seq:
+      - id: size_dec
+        type: str
+        terminator: 0x3a # ':'
+        doc: The number of key-value pairs in the mapping, in ASCII decimal.
```
GreyCat: It's not just some "size", it's "number of entries", thus please consider using `num_entries_dec` and `num_entries` for clarity.

serialization/php_serialized_value.ksy
KOLANICH: Hmm. How about deprecating `file-extension` and instead introducing `filename-regexp`?

archive/phar_without_stub.ksy
dgelessus: Right, I need to fix the names, sorry. (As someone who doesn't usually develop for Windows, I'm not used to Hungarian notation 😃)

archive/phar_without_stub.ksy
dgelessus: re. @KOLANICH: If you disagree with the style guide's length limit, can we discuss this in a separate issue? FWIW, I personally prefer soft wraps and no hard line limit, because it's compatible with all display sizes/layouts (unless you have to use an editor/tool that doesn't support soft wraps...), but if the style guide says 80 columns I'm going to follow that.

archive/phar_without_stub.ksy
GreyCat: `foo_size` is Hungarian notation too ;)

serialization/php_serialized_value.ksy
dgelessus: Good point, "size" isn't the best term here. (I'm guessing this was a copy-paste artifact.)

archive/phar_without_stub.ksy
GreyCat: @KOLANICH has once started a PR with changed to style guide: https://github.com/kaitai-io/kaitai_struct_doc/pull/7 — which I've mostly disagreed with.

We can have a discussion, but, unfortunately, I don't think there's much room to argue here. We're more or less tied to how the infrastructure around us works: i.e. how target languages work, how GitHub / text editors / similar tools work.

Just about *every* programming language I know insists on having a certain max line length. Even if programming language does not enforce it (like with C/C++), typically all the style guides that I know do. I haven't yet seen a single programming language so far that says "format all your code / documentation in a long single line".

And this is exactly what we're doing here: we're creating these docstrings to be inserted into target language source code. It's much easier to stick with the common practices of all other programming languages that to invent our own rules and then advocate for "proper" tools which support these.

serialization/php_serialized_value.ksy
GreyCat: "File extension" is a well-known thing, and many systems depend on that. Regexp might be a good idea, but I won't go as far as deprecating file extension, at least until majority of software out there will not treat it specially.

dgelessus: To clarify, the files in question are called ".metadata.bin" - the name starts with a dot, as per the Unix "hidden" file convention. Depending on how you look at it, this can be considered a file with the extension "metadata.bin" (though you could argue that a leading dot doesn't count and the extension is just "bin").

It's not a general-purpose file name/extension for serialized PHP data, i. e. if you would serialize some data yourself and write it to a file, you wouldn't call it "mydata.metadata.bin".

You're probably right that having this in the `file-extension` field is more confusing than useful, I think I'll remove it.

GreyCat: `file-extension` is supposed to be used by various tools like ones that analyze system file extension to action bindings. I agree that it's kind of vague, so let's leave it out for now.

I generally agree with @KOLANICH's idea on having a regexp — can I ask you to create an issue to track that?

archive/phar_without_stub.ksy
dgelessus: Yeah, I'm not suggesting that the official style should be changed. I have my reasons why I prefer no line breaks, but the project's code style takes precedence - it's more important to have a consistent code style at all (otherwise you have a mix of styles, and no matter which one you choose it will sometimes be wrong).

The Kaitai Struct code style is quite conservative with the 80 character limit, but at least this way it causes the least possible issues with other tools and code styles. Though to be exact, having an 80 character line limit in KSY files doesn't guarantee that the generated code will also fit into 80 columns - if the text is indented deeper in the generated code than in the KSY file, it will slightly exceed 80 characters. But I don't think we should try to fix that, because the only thing worse  
than a
small line length limit is text that was badly rewrapped into 80  
columns.

serialization/php_serialized_value.ksy
dgelessus: Done: kaitai-io/kaitai_struct#591

GreyCat:
Looks good to me, merging in!

GreyCat:
Thanks, @dgelessus!

PR #176, opened 2019-06-18
python pickle: Fix endianess of binfloat argument

moreati:
Contrary to every other data type in a Pickle, doubles are stored big endian [1]

[1]: https://github.com/python/cpython/blob/v3.7.3/Lib/pickletools.py#L820-L831


KOLANICH:
IEEE 754 floats and doubles have no endianness, do they?

GreyCat:
@KOLANICH They do, why not. That's exactly the reason why we have `f8be` and `f8le`.

@moreati Thanks, merging in!

PR #177, opened 2019-06-23
Added csv

KOLANICH:
I'd like to have it possible to set `terminator` to variables, allowing to repurpose this format for `tsv`.

### DIFF
--- a/database/csv.ksy
+++ b/database/csv.ksy
@@ -1,0 +1,33 @@
+meta:
+  id: csv
+  title: Comma-separated values
+  application: 
+  file-extension: csv
+  xref:
+    fileformat: csv
+    justsolve: CSV
+    loc: fdd000323
+    mime: text/csv
+    pronom: x-fmt/18
+    rfc: 4180
+    wikidata: Q935809
+  encoding: UTF-8
+  license: Unlicense
+
+doc: |
+  a widespread format where values and lines are separated by something. For now it is only CSV because of Kaitai limitations.
+
+seq:
+  - id: rows
+    type: row
+    terminator: 10
+    eos-error: false
+    repeat: eos
+types:
+  row:
+    seq:
+      - id: cells
+        type: str
+        terminator: 44
+        eos-error: false
+        repeat: eos


LikeLakers2:
Erm... apologies for butting in, but isn't Kaitai for binary formats? I don't see how csv is a format we'd need.

Besides, if you need a visual representation of a CSV, you can use a spreadsheet app, like Google Sheets or Excel. And it's pretty easy already to implement a CSV parser.

KOLANICH:
>but isn't Kaitai for binary formats? I don't see how CSV/TSV is a format we'd need.

It is. I have implemented it because csv spec impl is present in https://github.com/synalysis/Grammars ,  and I am developing a converter from that grammars to kaitai specs. One part of its test suite are `real-life tests` working the following way: I compile a spec from this repo and a spec transpiled from a synalysis grammar for the same format. Then I parse the same file with both specs. I also have a yaml file specifying which fields of one spec must be equal to which fields of another spec, my code iterates it and verifies that everything specified matches.

GreyCat:
From what I see so far, the proposed implementation has big deviations from what is usually considered csv, in particular, [as described in RFC](https://tools.ietf.org/html/rfc4180#section-2). The most challenging part of parsing csv is quoting, and this is clearly a thing that requires ambig handling, which is not available right now in KS, and will likely be never implemented.

So, in a way, I agree with @LikeLakers2 that it's not a binary format, and there's little value in having this not-quite-matching-RFC implementation in our repo.

KOLANICH:
>The most challenging part of parsing csv is quoting, and this is clearly a thing that requires ambig handling, which is not available right now in KS, and will likely be never implemented.

Certainly. I have just implemented the subset it was easy to implement. The goal was not to male the most correct impl, but the one working fine  on some files only.

LikeLakers2:
>The goal was not to male the most correct impl, but the one working fine on some files only.

That... makes no sense. Why would we *want* to make Kaitai specs that don't actually work 100% of the time?

KOLANICH:
>Why would we want to make Kaitai specs that don't actually work 100% of the time?

Because Kaitai csv spec is in fact not needed - csv is not a binary grammar, but a text one and there are better CSV parsers than the one generated from this spec. In fact there exists a BNF for csv, which probably can be translated into CoCo/R, which can target multiple languages.

For this spec I know only 2 use cases:
1. be used as a test for Synalysis2Kaitai
2. be just a tiny very basic example (everyone knows what csv is, everyone feels what is the simplest possible variant of csv)

Feel free to close this PR if you don't want this spec in this repo. My job is to propose, your job is to decline ;)

KOLANICH:
@GreyCat

GreyCat:
Let's not endorse the confusion. The proposed format is essentially *not* csv and does not address one of the main csv challenges. What we can do is that we can say that it is rather [tsv](https://en.wikipedia.org/wiki/Tab-separated_values) format, replace commas with tabs and then it will be good to go.

KOLANICH:
>The proposed format is essentially not csv and does not address one of the main csv challenges. 

Do you mean commas in the text?

>What we can do is that we can say that it is rather tsv format, replace commas with tabs and then it will be good to go.

In fact I targeted both dialects, but KS is quite limited that we cannot put a param into `contents`. I guess it is possible to use black magick again, but I want to
* keep this spec as small and as simple as possible
* keep it csv because in synalisis repo there is csv, not tsv. I am OK with tsv, in fact I usually use TSV if i need a text format for table data, but ... there is no tsv in synalisis repo. I can try to push tsv there, or just copy this csv in mybown repo and transform this file into tsv. Or just discard the first requirement and use some magic to handle quotes, though magic comes with overhead.

jhgorse:
I don't understand what happened here. I am interested in Kaitai Struct and am relatively new. I came searching for .ksy file describing csv to learn, which is almost this:
https://github.com/kaitai-io/kaitai_struct_formats/pull/177/commits/f3c1c5ce7106bff4b387bec13236f9a08624ce0d#diff-ee998fee64b04107f36d7b39f314028cd8a660922f930f80b6fb91679c87797b

What is the issue? Quoting commas in csv column values as strings?

KOLANICH:
1. here is the file in my repo https://github.com/KOLANICH-specs/kaitai_struct_formats/blob/csv/database/csv.ksy
2. this spec doesn't support quotes at all. It is mainly an example to be an analogue of https://github.com/synalysis/Grammars/blob/master/csv.grammar (in order to test my converter script).

PR #181, opened 2019-07-20
Add spec for the Unix pack compression format

dgelessus:
`pack` is a very early (and very basic) Unix file compression utility.

[Here](https://github.com/dgelessus/ksf_stuff/blob/master/archive/unix_pack.py) is a simple pack decompressor based on this spec.

### DIFF
--- a/archive/unix_pack.ksy
+++ b/archive/unix_pack.ksy
@@ -1,0 +1,107 @@
+meta:
+  id: unix_pack
+  title: Unix pack(1)-compressed data
+  application:
+    - pack(1)
+    - unpack(1)
+    - pcat(1)
+  # The extension is specifically lowercase.
+  # Uppercase .Z is used for compress(1)-compressed data.
+  file-extension: z
+  xref:
+    justsolve: Pack
+    wikidata: Q11239298
+  license: CC0-1.0
+  endian: be
+doc: |
+  The Unix pack compression format, a very basic compression format based on
+  Huffman coding. The Unix `pack(1)` utility could compress files into
+  this format, and could be decompressed using the utilities `unpack(1)`
+  and `pcat(1)`.
+  
+  The pack format is long obsolete - it was superseded in the 1980s
+  by the LZW-based compression format of `compress(1)` (which is itself
+  almost never used anymore on modern Unix derivatives). Despite this,
+  the `gzip(1)` utility still supports decompressing (but not compressing)
+  data in the pack compression format.
+doc-ref:
+  - 'https://www.vidarholen.net/contents/blog/?p=691'
+  - 'https://github.com/koalaman/pack/blob/master/pack.hs'
+  - 'https://git.savannah.gnu.org/cgit/gzip.git/tree/unpack.c'
+  # The source code for the original Unix pack and unpack utilities
+  # (as of Research Unix, Edition 8) can be found in the archive v8.tar.bz2
+  # in the directory usr/src/cmd/pack.
+  - 'https://www.tuhs.org/Archive/Distributions/Research/Dan_Cross_v8/'
+seq:
+  - id: magic
+    contents: [0x1f, 0x1e]
+  - id: len_uncompressed
+    type: u4
+    doc: The length of the data when uncompressed, in bytes.
+  - id: tree
+    type: tree
+    doc: The Huffman tree used to compress the data.
+  - id: data
+    size-eos: true
+    doc: |
+      The compressed data. This is a sequence of Huffman codes (variable-length
+      bit sequences) stored in most-significant bit first order without any
+      padding/alignment between individual codes. If the coded data does not
+      end on a byte boundary, it is padded with zeroes at the end (i. e. in the
+      least significant bits).
+types:
+  tree_level:
+    params:
+      - id: level_number
+        type: u1
+        doc: This level's number.
+    seq:
+      - id: leaves
+        type: u1
+        repeat: expr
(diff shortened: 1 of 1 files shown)


archive/unix_pack.ksy
```diff
@@ -0,0 +1,107 @@
+meta:
+  id: unix_pack
+  title: Unix pack(1)-compressed data
```
KOLANICH: is `1` a man section? If it is, I don't think it is a good idea to put it into name of a program because it is not a part of it.

archive/unix_pack.ksy
```diff
@@ -0,0 +1,107 @@
...
+seq:
+  - id: magic
+    contents: [0x1f, 0x1e]
+  - id: len_uncompressed
+    type: u4
+    doc: The length of the data when uncompressed, in bytes.
+  - id: tree
+    type: tree
+    doc: The Huffman tree used to compress the data.
+  - id: data
```
KOLANICH: I wonder if we can implement decoding right in KS.

archive/unix_pack.ksy
```diff
@@ -0,0 +1,107 @@
...
+          Note: the last leaf count is stored *minus two*. However, the last
+          leaf always stands for EOF and does not have an associated
+          byte value, so for the purposes of parsing this format, the last
+          level's leaf count is considered *minus one*.
+      - id: levels
+        type: tree_level(_index)
+        repeat: expr
+        repeat-expr: num_levels
+        doc: The individual levels of the tree.
+    doc: |
```
KOLANICH: shouldn;t it be before `seq`?

archive/unix_pack.ksy
```diff
@@ -0,0 +1,107 @@
...
+  tree_level:
+    params:
+      - id: level_number
+        type: u1
+        doc: This level's number.
+    seq:
+      - id: leaves
+        type: u1
+        repeat: expr
+        # See doc of tree.level_leaf_counts for an explanation of
```
KOLANICH: Why not to just move the explaination into the `doc` of this field?
Also it may be beneficial to move the calculation into an instance.

archive/unix_pack.ksy
```diff
@@ -0,0 +1,107 @@
...
+  - id: data
+    size-eos: true
+    doc: |
+      The compressed data. This is a sequence of Huffman codes (variable-length
+      bit sequences) stored in most-significant bit first order without any
+      padding/alignment between individual codes. If the coded data does not
+      end on a byte boundary, it is padded with zeroes at the end (i. e. in the
+      least significant bits).
+types:
+  tree_level:
```
KOLANICH: shouldn't it be a subtype of `tree`?

archive/unix_pack.ksy
```diff
@@ -0,0 +1,107 @@
...
+      follows these rules:
+      
+      * The tree is always left-aligned, i. e. on each level of the tree,
+        all inner (non-leaf) nodes are on the left side of the tree,
+        and all leaves are on the right side.
+      * The total number of nodes on each tree level is two times the number
+        of inner nodes on the previous level (since this is a binary tree,
+        each inner node has exactly two children).
+      * There is a single implicit root node.
+      * The number of inner nodes on each tree level can be calculated from
```
KOLANICH: it would be nice to remove this sentence and just encode it as an `instance`

archive/unix_pack.ksy
dgelessus: Correct, this is how man pages usually refer to command-line tools. It seems that this is done less often outside man pages (and even in man pages the section number is often only included the first time a tool is mentioned, not every time), so I'll probably remove the section numbers in most places.

archive/unix_pack.ksy
dgelessus: I don't think this is possible. To read the data, you need to iterate through the bits one by one and walk through the tree as indicated by each bit. This would require keeping state through multiple iterations of a `repeat`ed field, which isn't possible with Kaitai Struct. Even decoding the tree into a usable form would be difficult (or even impossible) because of how compactly it is stored in byte form.

For comparison, [here](https://github.com/dgelessus/ksf_stuff/blob/master/archive/unix_pack.py#L9-L66) is my Python code for decoding the tree and decompressing the data. If you have any idea how that could be translated to Kaitai Struct, feel free to try.

archive/unix_pack.ksy
dgelessus: You're right. (For some reason I thought putting `doc` first was only done for the top-level type, and that nested types would be handled like fields, with `doc` at the bottom.)

archive/unix_pack.ksy
dgelessus: I'm putting this in a YAML comment so that it only appears in the spec itself, and not in the code generated by ksc. I think the explanation is only relevant to people reading the source code of the spec, and not to people using the generated code (they will only access the `leaves` list itself and won't care about the internal calculations).

Good idea to move this into an instance though, will do that.

I don't think it makes sense to put this into the `doc`, because it's not relevant to users of the spec - they will only use the `leaves` list and won't have to deal with internal calculations like this. By putting the explanation in a YAML comment, it will be visible to people reading the spec, but it will be discarded when the spec is compiled.

archive/unix_pack.ksy
dgelessus: I don't think this is possible. The calculation for each level's inner node count requires the previous level's inner node count, and ksc throws an exception if I try to recursively access a value instance from inside itself (which makes sense, because it's hard/impossible to infer the type in this case).

PR #182, opened 2019-07-21
Drafted CBOR format

KOLANICH:
Undefined-length sequences are not yet implemented.

### DIFF
--- a/serialization/cbor.ksy
+++ b/serialization/cbor.ksy
@@ -1,0 +1,328 @@
+meta:
+  id: cbor
+  title: Concise Binary Object Representation
+  endian: be
+  license: Unlicense
+  xref:
+    mime: application/cbor
+    rfc:
+      - 7049
+      - 8152
+      - 8610
+      - 8392
+    wikidata: Q28455556
+doc: |
+  A binary format capable of serialization of superset of JSON.
+seq:
+  - id: ib
+    type: ib
+  - id: payload
+    type:
+      switch-on: ib.major_type
+      cases:
+        'type::uint': uint
+        'type::neg_uint': uint
+        'type::byte_str': byte_str
+        'type::text_str': text_str
+        'type::array': array
+        'type::map': map
+        'type::tag': tag
+        'type::simple_or_float': simple_or_float
+types:
+  ib:
+    seq:
+      - id: major_type
+        type: b3
+        enum: type
+      - id: additional_info
+        type: b5
+    instances:
+      is_raw_value:
+        value: additional_info < 24
+      value:
+        value: additional_info
+        if: is_raw_value
+      is_extension:
+        value: additional_info == 24
+      is_integer_length:
+         value: "additional_info >= 24 and additional_info <= 27"
+      integer_length:
+        value: additional_info - 24
+        if: is_integer_length
+      is_reserved:
+        value: "additional_info >= 28 and additional_info <= 30"
+      is_indefinite_length:
+        value: additional_info == 31
+  uint:
+    seq:
+      - id: value_with_length
+        type:
+          switch-on: _parent.as<cbor>.ib.integer_length
+          cases:
+            0: u1
+            1: u2
+            2: u4
+            3: u8
+        if: _parent.as<cbor>.ib.is_integer_length
+    instances:
+      is_negative:
+        value: "_parent.as<cbor>.ib.major_type == type::neg_uint"
+      raw_value:
+        value: "(_parent.as<cbor>.ib.is_raw_value?_parent.as<cbor>.ib.value:value_with_length)"
+      modulus:
+        value: "(is_negative?1+raw_value:raw_value)"
+      value:
+        value: "(is_negative?-modulus:modulus)"
+  byte_str:
+    seq:
+      - id: length
+        type: uint
+      - id: value
+        size: length.value
+    instances:
+      ib:
+        value: _parent.ib
+  text_str:
+    seq:
+      - id: length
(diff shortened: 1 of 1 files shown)


KOLANICH:
@GreyCat

KOLANICH:
@GreyCat

serialization/cbor.ksy
```diff
@@ -0,0 +1,328 @@
...
+            value: "bfc.mantissa*scale.value"
+            if: bfc.valid
+        types:
+          pow2:
+            params:
+              - id: log
+                type: u8
+            instances:
+              is_positive:
+                value: log>=0
```
generalmimon: `is_positive` will always be `true`, because `log` is of type `u8`, which means unsigned integer.

serialization/cbor.ksy
```diff
@@ -0,0 +1,328 @@
...
+            params:
+              - id: log
+                type: u1
+            instances:
+              is_positive:
+                value: log>=0
+              mod:
+                value: (is_positive?log:-log)
+              positive_pow:
+                value: "(((mod & 1).as<b1> ? 5 : 1) * ((mod & 2).as<b1> ? 25 : 1) * ((mod & 4).as<b1> ? 625 : 1) * ((mod & 8).as<b1> ? 390625 : 1) * ((mod & 16).as<b1> ? 0x2386f26fc1 : 1) * ((mod & 32).as<b1> ? 0x4ee2d6d415b85acef81 : 1) * ((mod & 64).as<b1> ? 0x184f03e93ff9f4daa797ed6e38ed64bf6a1f01 : 1) * ((mod & 128).as<b1> ? 0x24ee91f2603a6337f19bccdb0dac404dc08d3cff5ec2374e42f0f1538fd03df99092e953e01 : 1)) << mod"
```
generalmimon: Simply put, casting `CalcIntType` to `b1` is evil. The type casting in KS is currently quite dumb: in dynamically typed languages it's effectively noop (just fools the type checking engine in the compiler), in static typed languages it blindly does the explicit casting without checking whether the types are compatible or not. We really need to fix that (https://github.com/kaitai-io/kaitai_struct/issues/696) 

And I just checked that `(boolean) 4` fails at least in Java, because `incompatible types: int cannot be converted to boolean`. So please use `(mod & 1) != 0`, as it is the only reliable way.

Another thing is that we don't have integers with infinite precision (`BigInteger`) in KS (that's also the reason why the power operation is not yet implemented: https://github.com/kaitai-io/kaitai_struct/issues/216), so the number can easily overflow in most languages (except Python, of course). Even if we had power operation that takes and returns `double`s, it apparently isn't enough in this case. _(Hmm, I can't find an dedicated issue for bignums, it's shortly mentioned in https://github.com/kaitai-io/kaitai_struct/issues/104#issuecomment-358261907 and support for 64-bit integers in JS is related: https://github.com/kaitai-io/kaitai_struct/issues/183 but not quite. No one seems to consider it so important.)_

Actually, I'm surprised that the compiler written in Scala is capable of dealing with such big integers like `0x4ee2d6d415b85acef81`, that's nice.

KOLANICH: Thanks.

serialization/cbor.ksy
KOLANICH: Nice catch, thanks.

serialization/cbor.ksy
```diff
@@ -0,0 +1,328 @@
...
+        type:
+          switch-on: type
+          cases:
+            "type::simple": simple_or_float
+            #"type::f2": f2
+            "type::f4": f4
+            "type::f8": f8
+    instances:
+      type:
+        value: _parent.as<cbor>.ib.additional_info
```
generalmimon: I keep saying the same: **don't use casting unless you're sure what you're doing**. In ideal case, you shouldn't need to use casting at all. There are only few exotic cases for valid use of typecasting, in which the compiler has no way of knowing that the attribute will have some specific type. _(One of such cases is for example https://github.com/kaitai-io/kaitai_struct/issues/200.)_

Otherwise the casting will just suppress the errors when you're trying to do something clearly unreachable, and it will give both the compiler and the developer a false sense of security that the expression will be always of some type. Therefore you should always think twice if you need to use casting and **why** is it necessary.

This is also the case. Apparently you added the cast `.as<cbor>`, because simple `_parent.ib.additional_info` doesn't work. That makes sense, because the `_parent` in type `simple_or_float` can be of both types `cbor` and `simple_or_float` (you are using the type recursively in the `value` field a few lines above). Next time, please think about why is it necessary to use casting and if it works in all cases.

If you really need use `simple_or_float` recursively, I suggest to make it parametric and pass the `type` attribute to it from the parent struct. You'll get rid of `_parent` usage and if the `type` is `"type::simple"`, you can tell the descendant which type it should use.

serialization/cbor.ksy
```diff
@@ -0,0 +1,328 @@
...
+      switch-on: ib.major_type
+      cases:
+        'type::uint': uint
+        'type::neg_uint': uint
+        'type::byte_str': byte_str
+        'type::text_str': text_str
+        'type::array': array
+        'type::map': map
+        'type::tag': tag
+        'type::simple_or_float': simple_or_float
```
GreyCat: `uint`, `array`, `map` would currently collide with similar-named constructs in many languages (see https://github.com/kaitai-io/kaitai_struct/issues/90). Can you consider adding certain prefix to your type names to make sure we escape the collision?

KOLANICH:
#265

PR #183, opened 2019-07-21
Added numpy npy format

KOLANICH:


KOLANICH:
@GreyCat

serialization/numpy_npy.ksy
```diff
@@ -0,0 +1,157 @@
...
+  ks-opaque-types: true
+  xref:
+    wikidata: Q197520
+doc: |
+  Serialization format used by numpy. An antipattern, because:
+    * uses dicts in python syntax:
+      * in order to parse the array one has to have a parser for python language syntax. This cripples interoperability: noone wants to create a parser for python syntax to just parse a numpy array.
+      * it is tempting to `exec` these dicts, which could have been a security issue.
+      * they had to use `ast.parse` and manual walking of the tree in order to mitigate that. Fortunately now it is built into python as `ast.literal_eval`.
+    The schema of the dicts is fixed, the values in the dicts are integers, the better solution would be to put these numbers as standardized variable-length integers.
```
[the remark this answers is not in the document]
KOLANICH: >The reason for this is explained in the numpy docs - it makes it easier to reverse-engineer the format from scratch in the future, because a structured text representation is easier for humans to understand than packed integer fields. It also makes it easier to add more fields to the header in the future.

Yes, it is. But they can have used json for that.

serialization/numpy_npy.ksy
```diff
@@ -0,0 +1,157 @@
...
+    * it uses pickle, which is a remote code execution. The better solution is to say that serializing anything except arrays of literal types is out of scope of the format. If one needs pickle, he is already fucked up, so he should use pickle instead.
```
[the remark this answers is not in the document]
KOLANICH: >It's worth noting that pickle is only used when serializing arrays of Python objects. Arrays of primitive types are safe, as they are always stored in a flat binary representation. Recent versions of numpy also disable pickle support by default when reading (see CVE-2019-6446).

IMHO it was a mistake to rely on pickle in the format. It can be disabled in an impl, but it doesn't mean that the format itself was designed right.

KOLANICH:
@dgelessus

serialization/numpy_npy.ksy
```diff
@@ -0,0 +1,232 @@
...
+            '">i8"': s8be
+            '">f2"': f2be
+            '">f4"': f4be
+            '">f8"': f8be
+            '">f10"': f10be
+            '">f12"': f12be
+            '">f16"': f16be
+            '">f32"': f32be
+
+            # machine endianness, assumming le
```
dgelessus: I don't think numpy ever writes "machine-endian" dtypes into the file. When writing the dtype string, numpy looks up `.str` on the dtype object, and `numpy.dtype("|u2").str` is either `"<u2"` or `">u2"` depending on your machine's endianness (but never `"|u2"`).

Since these dtypes are ambiguous and cannot appear in npy files under normal circumstances, I think we should remove them completely instead of assuming a (possibly incorrect) byte order. That way if they do appear in a npy file for some reason, the user will get an error message instead of possibly invalid data.

serialization/numpy_npy.ksy
```diff
@@ -0,0 +1,232 @@
...
+        type: u1
+  array:
+    params:
+      - id: idx
+        type: u8
+    seq:
+      - id: data
+        doc: "S<n> for n > 1 is currently out of scope"
+        type:
+          switch-on: _root.parsed_header.item_type.as<str>
```
dgelessus: On the topic of dtypes, while reading the numpy docs, I noticed that numpy supports [many more complex dtypes](https://docs.scipy.org/doc/numpy/reference/arrays.dtypes.html#specifying-and-constructing-data-types) than just single scalars. For these dtypes, `"descr"` in the header dictionary is a list or dictionary rather than a string.

For example, calling `numpy.save` with this array:

```python
numpy.array([(12, 3.4), (56, 7.8)], dtype=[("", "u2"), ("", "f8")])
```

produces a file with this header (wrapped for readability):

```python
{
    'descr': [('foo', '<u2'), ('bar', '<f8')],
    'fortran_order': False,
    'shape': (2,),
}
```

I don't think there's any way to support arbitrary complex dtypes in this spec, so we should probably stick to simple dtypes only and document this limitation.

dgelessus:
Strange, GitHub didn't notify me about the force-pushes even though I'm subscribed to this PR - I only got a notification for the comment where you pinged me.

With complex dtypes in mind, it makes more sense why the header is stored as a Python literal instead of a binary format. Complex dtypes can be nested arbitrarily deeply, and such a structure is less straightforward to serialize in a packed binary form. (It's possible of course, but it would make reverse-engineering more complex again, because it's no longer just length-prefixed strings.)

And while today it would make sense to use JSON for the text representation of the data, the npy format was [designed in 2007](https://numpy.org/neps/nep-0001-npy-format.html). The current Python version at that point was Python 2.5, which had no built-in support for JSON or similar formats (the standard `json` module was added in Python 2.6, released in 2008). The only real alternative supported by Python at that point would have been XML, and you can be glad that they didn't choose that :P

serialization/numpy_npy.ksy
KOLANICH: >I don't think numpy ever writes "machine-endian" dtypes into the file.

I also don't think that they appear in real files serialized by numpy. Types coded with letters  also don't appeat. But in npy spec nothing is said about that it is not allowed to use machine endianness and such npy files are really parsed correctly (on machines with the same endianness). Since the most widespread (and natural in the sense that i-th chunk uses i-th power of the base) endianness (on both desktops, mobile devices and embedded, because x86 dominates desktop, arm dominates everything else and RISCV is LE ) is currently `le` (more precisely `el8` in my notation) IMHO we should support them too.

serialization/numpy_npy.ksy
KOLANICH: Yes, they are currently out of scope.

KOLANICH:
Good point, though that can be solved by a dependency on a third-party JSON library, probably even a native one (for example I usually use `ujson` when the data is needed to occasionally be shown to humans (it has no pretty-printing)).

KOLANICH:
Added the docs

KOLANICH:
@GreyCat

PR #210, opened 2019-08-04
Added ieee754 floats

KOLANICH:
It seems that merging numpy_npy takes a bit long, so I have splited the floats into a separate PR


common/ieee754_float/f1.ksy
```diff
@@ -0,0 +1,22 @@
+meta:
+  id: f1
+  title: IEEE 754 Min-float
```
GreyCat: * This particular entry is *not* IEEE754-compliant in any way, and illustrates *one* of possible one-byte long floating point numbers representation, [as per entry in Wikipedia](https://en.wikipedia.org/wiki/Minifloat).
* The same entry is named "minifloat", not "min-float"

GreyCat:
Thanks for this effort, it is really massive!

I've got several high-level comments:

* These "non-native" implementations should be likely named in a certain way (i.e. with some kind of prefix) that clearly distinguishes them from built-in types. There are several reasons for that:
  * Clearly separating "native" vs these "academical"-purpose types:
    * they have different API,
    * they have different performance expectations,
    * academical ones (as you've very much correctly pointed out) offer really mostly approximate representations
  * In some point in future, we'd likely introduce some machinery that will not allow creation of user types with names which match built-in type names (i.e. `f4`, `f4le`, `u4`, `str`, `bytes`, etc). I heavily suspect that even now `f4.ksy` is unusable, as `f4` would be always matched as built-in type name first.
* "float.ksy" would create `float` / `Float` data types, which would clash with "float" type in many languages. May be we'll solve that automatically in some point in future, but for now — a prefix / suffix would avoid all these problems at all.

common/ieee754_float/f1.ksy
```diff
@@ -0,0 +1,22 @@
...
+  license: Unlicense
+  endian: be
```
GreyCat: I don't think this `endian: be` is used in this type.

common/ieee754_float/f1.ksy
KOLANICH: * Yes, it is not defined in IEEE 754. But it uses the same principles. How about `IEEE 754-like`?
* I don't remember why I have called it this way.

common/ieee754_float/f1.ksy
KOLANICH: I expect https://github.com/kaitai-io/kaitai_struct/issues/155 be fixed someone, so bit-sized types are expected to be laid out correctly. That's why I have marked all the types as `be` because currently they are `be` in that sense and `le` is not supported at all. Anyway, for 1-byte byte-aligned field it shouldn't make any difference.

KOLANICH:
>These "non-native" implementations should be likely named in a certain way (i.e. with some kind of prefix) that clearly distinguishes them from built-in types. There are several reasons for that:

>    Clearly separating "native" vs these "academical"-purpose types:
        * they have different API,

The clash was intentional. I expect [these numbers be supported natively](https://github.com/kaitai-io/kaitai_struct/issues/101) somewhen. I also expect [inlining](https://github.com/kaitai-io/kaitai_struct/issues/88) be implemented somewhen, and when it is implemented, it will be possible to transform the types into drop-in replacements. I mean that in order to use them a user would just need to add an import of them.

>
        * they have different performance expectations,
        * academical ones (as you've very much correctly pointed out) offer really mostly approximate representations
 
These are really an issue, but there is a big IMHO: working somehow is better than not-working at all for most of cases, and for the cases where `somehow` is worse than `not working at all` the persons responsible for creating such a responsible system must already know what they are doing.

>In some point in future, we'd likely introduce some machinery that will not allow creation of user types with names which match built-in type names (i.e. f4, f4le, u4, str, bytes, etc). 

I have another idea. We can create a special dir in KSF and make the compiler to use the specs from it when native impls are not available. Maybe with issuing a warning. Or a critical error that can be disabled by a special flag.

>I heavily suspect that even now f4.ksy is unusable, as f4 would be always matched as built-in type name first.

Yes, it is probably unusable. It is mainly here for educational purposes: I have started with `f4`s, debugged them, then used them as boilerplates to create all the other types. As I have said, users should use the natively available impls, if it is possible.

>"float.ksy" would create float / Float data types, which would clash with "float" type in many languages. May be we'll solve that automatically in some point in future, but for now — a prefix / suffix would avoid all these problems at all.

Would `ieee754_float` be OK?

PR #213, opened 2019-08-04
Added Lytro Light Field Picture

KOLANICH:
Fixes: #175


image/lytro_lfp.ksy
```diff
@@ -0,0 +1,158 @@
...
+  Sample files can be downloaded from:
+    https://github.com/behnam/python-lfp-reader/tree/master/samples
+    https://www.irisa.fr/temics/demos/IllumDatasetLF/index_files/LytroIllum_Dataset_INRIA_SIROCCO.zip
+
+doc-ref:
+  - https://github.com/behnam/python-lfp-reader
+  - https://github.com/hahnec/plenopticam
+  - https://github.com/nrpatel/lfptools
+  - https://eclecti.cc/computervision/reverse-engineering-the-lytro-lfp-file-format
+  - http://optics.miloush.net/lytro/TheFileFormat.aspx
```
GreyCat: This requires `ks-version: 0.9` in `meta`.

image/lytro_lfp.ksy
```diff
@@ -0,0 +1,158 @@
...
+                      pic = self.json['picture']
+                      self.frames = []
+                      self.metadata_index = {}
+                      
+                      for i, f in enumerate(pic['frameArray']):
+                        fDescr = Frame(f['frame'], index)
+                        self.frames.append(fDescr)
+                        self.metadata_index[fDescr.metadata] = SectionLink(i, ChunkType.metadata)
+                        self.metadata_index[fDescr.private_metadata] = SectionLink(i, ChunkType.private_metadata)
+                        self.metadata_index[fDescr.image] = SectionLink(i, ChunkType.image)
```
GreyCat: It's kind of pointless to merge this into main KS repo, as this clearly requires these external modules, and (1) all the current infrastructure does not really support that, (2) I doubt that it's a good way to go for future development — the whole point of having KS specs was having cross-language support, and this particular implementation will be constrained to a single language.

KOLANICH: >and this particular implementation will be constrained to a single language.

It is an example of missing pieces in python. It is pointless to insert there impls in all other langs. A user should look at python code and implement something similar in the lang he uses. The rest of the spec is suitable for other langs (not necessarily all of them).

hakanai: I don't know if Python is a good language for that. There's a library I found where only a Python implementation existed, and it is becoming the _hardest thing ever_ to port to Java.

KOLANICH: As we already know, the proper solution requres changes in KSC. But for now a temporary solution working at least somewhere is better than nothing.

PR #216, opened 2019-08-09
Add enum for section flags in PE file.

joelkevinjones:
Based on
https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#section-flags

### DIFF
--- a/executable/microsoft_pe.ksy
+++ b/executable/microsoft_pe.ksy
@@ -351,10 +351,224 @@ types:
         type: u2
       - id: characteristics
         type: u4
+        enum: section_flags
     instances:
       body:
         pos: pointer_to_raw_data
         size: size_of_raw_data
+    enums:
+      section_flags:
+        0x00000000:
+          id: reserved1
+          -orig-id: Reserved1
+          doc: |
+            Reserved for future use.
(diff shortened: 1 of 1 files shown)


GreyCat:
These are flags, not enum values. Enum values in KS do not carry flags semantics and are not supposed to be ORed together, for example. Please implement them as a user type with separate bits in these 4 bytes assigned to different attributes of this type.

joelkevinjones:
Is this what you had in mind? 
```
    types:
      section_flags:
        seq:
          - id: reserved1
            type: b2
          - id: image_scn_cnt_code
            type: b1
            -orig-id: IMAGE_SCN_CNT_CODE
            doc: |
              The section contains executable code.  0x00000020
          - id: reserved2
            type: b22
          - id: image_scn_mem_write
            type: b1
            -orig-id: IMAGE_SCN_MEM_WRITE
            doc: The section can be written to
          - id: image_scn_mem_read
            type: b1
            -orig-id: IMAGE_SCN_MEM_READ
            doc: The section can be read
          - id: reserved3
            type: b5

```

KOLANICH:
yes, something like that.

PR #217, opened 2019-08-10
Potential fix for issue #612 - Buffer Underflow Exception

nsoft:
(https://github.com/kaitai-io/kaitai_struct/issues/612)

I haven't figured out how to test this change yet, so please take this PR with a grain of salt. Some instructions (perhaps in a BUILDING.txt) for how to do a full build of your project prior to PR's would be helpful. If I've missed it in the docs sorry... all I see is lots about how to run from binaries, and not much about how to build the project and run whatever test suites you have.

### DIFF
--- a/media/stl.ksy
+++ b/media/stl.ksy
@@ -55,7 +55,8 @@ types:
           In practice, software dealing with STL either expected to
           see 0 here, or uses this 16-bit field per se to store
           additional attributes (such as RGB color of a vertex or
-          color index).  vec3d:
+          color index).
+  vec3d:
     seq:
       - id: x
         type: f4
--- a/media/stl.ksy
+++ b/media/stl.ksy
@@ -45,14 +45,6 @@ types:
         type: vec3d
         repeat: expr
         repeat-expr: 3
-  vec3d:
-    seq:
-      - id: x
-        type: f4
-      - id: y
-        type: f4
-      - id: z
-        type: f4
       - id: abr
         type: u2
         doc: |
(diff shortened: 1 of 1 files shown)


nsoft:
Finally figured out how to use your web ide to generate the parser class, and now things seem to work with this adjustment.

media/stl.ksy
```diff
@@ -45,14 +45,6 @@ types:
...
         repeat-expr: 3
-  vec3d:
-    seq:
-      - id: x
-        type: f4
-      - id: y
-        type: f4
-      - id: z
-        type: f4
       - id: abr
```
KOLANICH: This belongs to every vertex: https://github.com/FreeCAD/FreeCAD/blob/6cc1524a0ddbfbcec76b3ef4407aeb79b7fbdc57/src/Mod/Mesh/App/Core/MeshIO.cpp#L1315

GreyCat:
Hi Patrick, and sorry for the delayed response.

This particular repository has nothing to do with building "whole project", and, in fact, there's no designated test suite for this repo. Larger formats typically has one-repo-per-format model and store their test suites along with the .ksy file developed. This particular one is very short and simple, so it doesn't have a test suite at all, unfortunately :(

The basic way to test it is compiling it manually with ksc — i.e. `ksc -t your_language stl.ksy` — and then using library it generates in some simple test program. Alternatively, one can use ksv / ksdump / WebIDE with it.

Thank you for getting down to this fix, merging it in 👍

PR #229, opened 2019-09-22
Fix WAV and add RIFF

generalmimon:
Wave sample files: http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/Samples.html
WAV with bext chunk: http://www.noiseofnorway.net/wp-content/uploads/2011/11/170139-Printer-Scanner-close-scanner-lid-Schoeps.wav

RIFF description can be tested with any of its subtypes: WAV, AVI, RMID, DLS, SF2, ...


common/riff.ksy
```diff
@@ -0,0 +1,90 @@
...
+  The Microsoft implementation is mostly known through container formats
+  like AVI, ANI and WAV, which use RIFF as their basis.
+seq:
+  - id: riff_chunk
+    type: chunk
+types:
+  parent_chunk_data:
+    seq:
+      - id: form_type
+        type: u4be
```
KOLANICH: how about `le` or even a fixed-length `str` here? IMHO there is no need in a enum. For individual formats we need a way to require a chunk to be of a certain type, but it is currently impossible, needs a fix in KSC, probably even more than 1.

So I guess the following architecture can be beneficial:
1. separate files for chunkspecs in a `riff` subdir of the dir where the target format spec lies.
2. a `riff` spec importing them all and having an optional param of allowed chunk type. Ideally it should assert than chunk type matches, but for now I suggest to just check that and if it doesn't, then don't parse using `if`.
2. a spec file for formats importing `riff` spec and accepting the arg which

for example the following tree
```yaml
repo_root:
  - media:
    - riff: 
      - wave_riff_chunk
      - umid_riff_chunk
    - riff.ksy
    - wav.ksy
    - midi.ksy
  - image
    - riff:
      - cdr_riff_chunk 
    - corel_draw_cdr.ksy
```

KOLANICH:
Thank you for the contribution. It is a long-lasting tech debt to unify all riff-based formats under a riff.ksy spec, but it requires multiple features lacking from KSC currently: https://github.com/kaitai-io/kaitai_struct/issues/81, https://github.com/kaitai-io/kaitai_struct/issues/458, and maybe https://github.com/kaitai-io/kaitai_struct/issues/135 and https://github.com/kaitai-io/kaitai_struct/issues/196.

You ma6nalso want to add http://www.johnloomis.org/cpe102/asgn/asgn1/riff.html into `doc-ref`.

common/riff.ksy
generalmimon: > how about `le` or even a fixed-length `str` here? IMHO there is no need in a enum.

Yes, fixed-length `str` will be probably better. I thought that it's the recommended approach for four-character codes to be represented as integers because of [the user guide](https://doc.kaitai.io/user_guide.html#_fourcc), but here it doesn't make much sense. On top of that, strings are much more readable.


> So I guess the following architecture can be beneficial: (...)

If I understand what you're suggesting:
```YAML
# /media/riff.ksy
meta:
  id: riff
  endian: le
  imports:
    - media/riff/wav_riff_chunk
    - media/riff/rmid_riff_chunk
params:
  - id: allowed_chunk_type
    type: str
seq:
  - id: chunk_id
    contents: RIFF
  - id: chunk_size
    type: u4
  - id: chunk_data
    type:
      switch-on: allowed_chunk_type
      cases:
        '"WAVE"': wav_riff_chunk
        '"RMID"': rmid_riff_chunk
        _: unknown_chunk
    size: chunk_size
  - id: pad_byte
    size: chunk_size % 2
types:
  unknown_chunk: {}
```
```YAML
# /media/riff/wav_riff_chunk.ksy
meta:
  id: wav_riff_chunk
  endian: le
seq:
  - id: form_type
    contents: WAVE
  - id: subchunks
    type: chunk_type
    repeat: eos
types:
  # copy & paste `chunk_type` definition from `media/wav.ksy` 
  # and its dependencies here (those >500 lines wouldn't fit here 😃)
```
```YAML
# /media/riff/rmid_riff_chunk.ksy
meta:
  id: rmid_riff_chunk
  encoding: ASCII
  endian: le
  imports:
    - media/standard_midi_file
  doc: |
    A RIFF wrapper around SMF.
seq:
  - id: form_type
    contents: RMID
  - id: subchunks
    type: subchunk
    repeat: eos
types:
  subchunk:
    seq:
      - id: chunk_id
        type: str
        size: 4
        pad-right: 0x20
      - id: chunk_size
        type: u4
      - id: chunk_data
        type:
          switch-on: chunk_id
          cases:
            '"data"': standard_midi_file
        size: chunk_size
      - id: pad_byte
        size: chunk_size % 2
```
```YAML
# /media/wav.ksy
meta:
  id: wav
  imports:
    - media/riff
seq:
  - id: riff_chunk
    type: riff('WAVE')
```
```YAML
# /media/rmid.ksy
meta:
  id: rmid
  imports:
    - media/riff
seq:
  - id: riff_chunk
    type: riff('RMID')
```

I don't know. It's definitely some improvement, but I'm not quite satisfied with it.

1. We're basically moving all chunkspecs to `riff/*_riff_chunk.ksy` files, so the particular format specs become just empty wrappers around `/media/riff.ksy`.
2. Imagine that there are dozens of chunkspecs in `riff/` folder. RIFF spec `/media/riff.ksy` imports all of them, but only one at a time is used. It adds many unnecessary dependencies.
3. There is still quite a lot of repetition. It is necessary to define basic RIFF `chunk` and `parent_chunk_data` over and over again.

Yet it is IMHO the best thing what we can do in KS right now. If we had generics (kaitai-io/kaitai_struct#135) and validation (kaitai-io/kaitai_struct#435 - we already have, but not in all target languages), we would have been able to do this (syntax from https://github.com/kaitai-io/kaitai_struct/issues/273#issue-260926722):
```YAML
# /common/riff/chunk.ksy
meta:
  id: chunk
  encoding: ASCII
  endian: le
templates:
  - expected_chunk_id
  - data_generic
seq:
  - id: chunk_id
    type: str
    valid: expected_chunk_id
    size: 4
    pad-right: 0x20
  - id: chunk_size
    type: u4
  - id: chunk_data
    type: data_generic
    size: chunk_size
  - id: pad_byte
    size: chunk_size % 2
```
```YAML
# /common/riff/parent_chunk_data.ksy
meta:
  id: parent_chunk_data
  encoding: ASCII
  endian: le
templates:
  - expected_form_type
  - subchunk_generic
seq:
  - id: form_type
    type: str
    valid: expected_form_type
    size: 4
    pad-right: 0x20
  - id: subchunks
    type: subchunk_generic
    repeat: eos
```
```YAML
# /common/wav.ksy
meta:
  id: wav
  imports:
    - common/riff/chunk
    - common/riff/parent_chunk_data
seq:
  - id: riff_chunk
    type: >-
      chunk<
        'RIFF',
        parent_chunk_data<
          'WAVE',
          chunk<
            _, # somehow omit the `expected_chunk_id` field
            chunk_data
          >
        >
      >
types:
  chunk_data:
    seq:
      - id: data
        type:
          switch-on: _parent.as<chunk>.chunk_id
          cases:
            '"fmt"': format_chunk_data
            '"bext"': bext_chunk_data
            '"cue"': cue_chunk_data
            '"data"': data_chunk_data
            '"LIST"': >-
              parent_chunk_data<
                _, # omit the `expected_form_type` field
                list_subchunk
              >
  list_subchunk:
    seq:
      id: data
      type:
          switch-on: _parent.as<parent_chunk_data>.form_type
          cases:
            '"INFO"': strz
# ...
```
I like the brevity, simplicity and clarity so much! I can help with the implementation in KSC if you want.

common/riff.ksy
```diff
@@ -0,0 +1,93 @@
...
+seq:
+  - id: riff_chunk
+    type: chunk
+types:
+  parent_chunk_data:
+    seq:
+      - id: form_type
+        type: str
+        size: 4
+        pad-right: 0x20
```
KOLANICH: Not sure that we need this `pad-right`. `pad-right` is the contract that all the strings have padding stripped here. But the strings are 4 bytes, 4 bytes is 1 dword, one dword is a unit for reading memory on 32-bit archs and is faster than reading 3 bytes. The strings are fixed length. There is no chance that there will be arbitrary long padding. So IMHO `pad-right` here doesn't make any sense and would have prevened KSC from optimizing the code if such optimizations ever were introduced.

generalmimon:
@KOLANICH
Could you please take a look at the new approach I take? I've already explained it on https://github.com/kaitai-io/kaitai_struct_formats/issues/230#issuecomment-540217029, just split generic RIFF types to those with assertions and without assertions (sometimes you want to force chunk to be of specific chunk ID, sometimes not).

I think it's now quite readable and less error-prone. Silly mistakes like [this one](https://github.com/kaitai-io/kaitai_struct_formats/blob/master/media/wav.ksy#L29) (the marked line also causes https://github.com/kaitai-io/kaitai_struct/issues/604) should no longer happen if all RIFF-based formats use the common classes as containers for specific data. I would continue with the other RIFF formats if nobody objects.

The riff.ksy format can be used for exploring chunk structure of any RIFF-compliant binary file. It is generic, it only implements INFO list chunk for which RIFF specification says the following: _"The ‘INFO’ list is a registered global chunk that can be used within any RIFF file."_ I don't expect that someone would use it in production, but it is quite handy for development.

KOLANICH:
I wonder if we really need to assert chunk types instead of just switching on them everywhere? I mean just read a chunk id and interpret the blob based on read chunk id only, without any checks wheter this chunk type is allowed to be here.

generalmimon:
You are right, thank you the suggestion! I haven't realized that the assertions are not vital. Is it now OK?

generalmimon:
@KOLANICH, can this PR be already merged?

KOLANICH:
I'm sorry, but I am not the one merging PRs in this repo. @GreyCat is the one who merges.

generalmimon:
Actually, I can merge it myself, because @GreyCat trusts me that I won't do silly things :smirk:, but I hesitate to do so when I'm not 100% sure that it is OK. I wonder why you don't want to enter the Kaitai team, you belong there more than everyone else. I understand that you might not want to feel the pressure to somehow deserve the membership (e.g. at least one contribution per week :joy:), but I think everyone understands that you're busy person and you can participate only once in a while.

KOLANICH:
>I wonder why you don't want to enter the Kaitai team, you belong there more than everyone else.

* I cannot be sure in cybersecurity of my devices. In fact I am sure in their insecurity (Android KitKat from 2016 year cannot be even considered secure). I don't want other's repos being compromised because of this.
* GreyCat doesn't trust me because I am a nonconformist and don't follow the style guide I personally consider defective :) It is probably the reason why my PRs are ignored for such a long time. It is completely OK ;)

>at least one contribution per week

AFAIK there are no such requirements.

generalmimon:
> I cannot be sure in cybersecurity of my devices. (...) I don't want other's repos being compromised because of this.

Nothing is ever secure, but I can't imagine the degree of boredom of a hacker that would bother with screwing up an open-source repo like this, but thanks for the info 🙂

KOLANICH:
>open-source repo like this

access to an org gives access to all the repos. this repo is probably safe, ksy has no io, but if an adversary have managed to backdoor the KSC repo or builds, it would be devastating.

generalmimon:
@GreyCat, can it be finally merged? It's blocking my `master` branch (which is of course my fault, but it quite sucks), every time I want to contribute something I must be careful because the branch `master` isn't `master`. And I actually don't see a place for improvement here.

common/riff/chunk.ksy
```diff
@@ -0,0 +1,17 @@
+meta:
+  id: chunk
```
GreyCat, line 2: It's definitely not OK to name very format-specific things "chunk", as we're ending up getting `Chunk` class at top level, which will be shared across whole gallery. Please consider renaming this to `riff_chunk`?

common/riff/chunk.ksy
```diff
@@ -0,0 +1,17 @@
...
+  encoding: ASCII
+  endian: le
+seq:
+  - id: id
+    type: str
+    size: 4
```
GreyCat, line 8: In general, it's a bad idea to have chunk 4-byte type identifiers to be handles as strings:

* String conversion is comparatively slow and inefficient. All production parsers will use an 4-byte integer enum here, while KS-generated parser will be slower due to string usage / allocation / deallocation / string comparisons.
* Strings have encoding, and in general case, there is no guarantee that it will be anything actually readable by human. Quite a few RIFF-based formats use arbitary byte sequences (like `0xdeadc0de`), which is neither valid UTF-8, nor valid ASCII.

common/riff/parent_chunk_data.ksy
```diff
@@ -0,0 +1,13 @@
+meta:
+  id: parent_chunk_data
```
GreyCat, line 2: Same thing: this RIFF-specific piece will become repo-wide `ParentChunkData` after compilation.

GreyCat:
Apologies for not reviewing this earlier. I totally agree on the motivation and approach, but there are two things that I've flagged:

* Naming/file placement - that's minor and likely easy to fix. Actually, nowadays we might even just keep these pieces as subtypes inside `riff.ksy` and address them using something like `type: 'riff::chunk'`
* Using strings to work with RIFF chunk identifiers. That's probably not too major too - we can have it declared as an u4, as then in every specific format we can convert it to .wav-specific, .avi-specific, etc, map of enum values.

What do you think?

generalmimon:
@GreyCat 

> Actually, nowadays we might even just keep these pieces as subtypes inside `riff.ksy` and address them using something like `type: 'riff::chunk'`

At first, I meant the `riff` spec to be primarily for debugging and reverse-engineering (and it also can be used for extracting metadata from the `INFO` chunk), but yeah, I agree with this approach. It makes perfect sense to include the `chunk` and `parent_chunk_data` as subtypes, we get rid of two import-only files which are not meaningful on their own. And it's better to write `riff::parent_chunk_data` than `riff_parent_chunk_data`.

> * Using strings to work with RIFF chunk identifiers. That's probably not too major too - we can have it declared as an u4, as then in every specific format we can convert it to .wav-specific, .avi-specific, etc, map of enum values.

At time of writing the specs, I was experimenting with the enum implementation as well. The problem was it didn't come to my mind that I can use `enum` key in value instances (the docs don't mention it at all), so I thought that the only use enums is to introduce an enum on the `riff` KSY level with all possible RIFF chunk IDs that can ever occur, which just isn't possible. So I chose the only viable approach besides enums with respect to people (people typically don't know that number `0x45564157` means "WAVE" in little-endian), and that are plain text strings.

But yeah, you're right, string conversion and comparison is inefficent compared to a primitive integer read, and the inconvenience of working with plain numbers can be gracefully solved with converting the integer to format-specific (wav, avi, ...) chunk ID dictionary enum just using a value instance with the `enum` key.

I'll go ahead and try to incorporate the suggestions.

GreyCat:
There are currently 2 conflicting changes. Can I ask you to take a look and resolve these conflicts, so we can merge this one?

generalmimon:
@GreyCat 
> However, there are currently 2 conflicting fragment. Can I ask you to take a look and resolve these conflicts, so we can merge this one?

Yeah, I no longer hoped that this PR can be merged in a reasonable amount of time, so I've recently fixed the WAV format manually so that it at least works (it was broken due to https://github.com/kaitai-io/kaitai_struct/issues/604 and merging this PR would also fix it).

Thanks for the review, please merge it finally if all looks good to you 🙂

GreyCat:
Thanks for this great work!

PR #233, opened 2019-10-12
Extend BMP format support

generalmimon:
Current BMP spec in formats repository is somewhat rudimentary, it is not very much usable. I tried to improve that.

Every BMP file contains a DIB header (bitmap information header) part. There are several different variants. Every of them has a `header_size` field in the beginning, which is used for determining which header is present in the image. 

The existing implementation advised to switch over this field and select appropriate type representing the particular header. Although it seems to make sense, it complicates the access to common properties (e.g. width and height). To access width in the header, you would have to do this:
```yaml
instances:
  image_width:
    value: >-
      len_dib_header == header_type::bitmap_core_header.to_i ?
        header.dib_header.as<bitmap_core_header>.width :
        len_dib_header == header_type::bitmap_info_header.to_i ?
          header.dib_header.as<bitmap_info_header>.width :
          len_dib_header == header_type::bitmap_v4_header.to_i ?
            header.dib_header.as<bitmap_v4_header>.width :
            ... ?
              ... :
              0
```
The headers don't contain completely distinct fields, they are extending each other. The new implementation takes advantage of this: instead of rebuilding each header from nothing, it consists of several extensions that form together the required header. For example, if the present header is of type `BITMAPV4HEADER`, it will contain the fields from `BITMAPCOREHEADER` (these are inlined in `bitmap_header` type), `bitmap_info_extension` and `bitmap_v4_extension`.

Here is a diagram which shows supported DIB headers and their KSY type counterpart. There is a dashed line from `BITMAPCOREHEADER` to `BITMAPINFOHEADER`, because `BITMAPCOREHEADER` uses 16-bit integers for image width and height, while `BITMAPINFOHEADER` and all headers extending it use 32-bit integers (that's where the "+ 4" came from).
The DIB headers diagram

### DIFF
--- a/image/bmp.ksy
+++ b/image/bmp.ksy
@@ -91,10 +91,10 @@ doc: |
       * BITMAPV4HEADER (WIN4XBITMAPHEADER)
 
 seq:
-  - id: file_header
+  - id: file_hdr
     type: file_header
   - id: dib_info
-    size: file_header.ofs_bitmap - 14 # 14 = file_header._sizeof (TODO: replace when KSC 0.9 is released)
+    size: file_hdr.ofs_bitmap - 14 # 14 = file_hdr._sizeof (TODO: replace when KSC 0.9 is released)
     type: bitmap_info
   - id: bitmap
     type: bitmap
@@ -406,7 +406,7 @@ types:
           or _parent.bitmap_v4_ext.color_space_type == color_space::profile_embedded
       profile_data:
         io: _root._io
-        pos: 14 + ofs_profile # 14 = _root.file_header._sizeof
+        pos: 14 + ofs_profile # 14 = _root.file_hdr._sizeof
         size: len_profile
         type:
           switch-on: _parent.bitmap_v4_ext.color_space_type == color_space::profile_linked
--- a/image/bmp.ksy
+++ b/image/bmp.ksy
@@ -104,6 +104,8 @@ types:
     doc: |
       Replace with an opaque type if you care about the pixels.
       You can look at an example of a JavaScript implementation: https://github.com/generalmimon/bmptool/blob/master/src/Bitmap.js
+
+      There is a proposal for adding bitmap data type to Kaitai Struct: https://github.com/kaitai-io/kaitai_struct/issues/188
   file_header:
     -orig-id: BITMAPFILEHEADER
     doc-ref: https://docs.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapfileheader
--- a/image/bmp.ksy
+++ b/image/bmp.ksy
@@ -17,26 +17,95 @@ meta:
   endian: le
   license: CC0-1.0
   ks-version: 0.8
+doc: |
+  The **BMP file format**, also known as **bitmap image file** or **device independent
+  bitmap (DIB) file format** or simply a **bitmap**, is a raster graphics image file
+  format used to store bitmap digital images, independently of the display
+  device (such as a graphics adapter), especially on Microsoft Windows
+  and OS/2 operating systems.
+
+  ## Samples
+
+  Great collection of various BMP sample files:
+  [**BMP Suite Image List**](
+    http://entropymine.com/jason/bmpsuite/bmpsuite/html/bmpsuite.html
+  ) (by Jason Summers)
+
+  If only there was such a comprehensive sample suite for every file format! It's like
+  a dream for every developer of any binary file format parser. It contains a lot of
+  different types and variations of BMP files, even the tricky ones, where it's not clear
+  from the specification how to deal with them (marked there as "**q**uestionable").
+
+  If you make a program which will be able to read all the "**g**ood" and "**q**uestionable"
+  BMP files and won't crash on the "**b**ad" ones, it will definitely have one of the most
+  extensive support of BMP files in the universe!
+
+  ## BITMAPV2INFOHEADER and BITMAPV3INFOHEADER
+
+  A beneficial discussion on Adobe forum (archived):
+  [**Invalid BMP Format with Alpha channel**](
+    https://web.archive.org/web/20150127132443/https://forums.adobe.com/message/3272950
+  )
+
+  In 2010, someone noticed that Photoshop generated BMP with an odd type of header. There wasn't
+  any documentation available for this header at the time (and still isn't).
+  However, Chris Cox (former Adobe employee) claimed that they hadn't invented any type
+  of proprietary header and everything they were writing was taken directly
+  from the Microsoft documentation.
+
+  It showed up that the unknown header was called BITMAPV3INFOHEADER.
+  Although Microsoft has apparently requested and verified the use of the header,
+  the documentation on MSDN has probably got lost and they have probably
+  forgotten about this type of header.
+
+  This is the only source I could find about these structures, so we could't rely
+  on it so much, but I think supporting them as a read-only format won't harm anything.
+  Due to the fact that it isn't documented anywhere else, most applications don't support it.
+
+  All Windows headers at once (including mentioned BITMAPV2INFOHEADER and BITMAPV3INFOHEADER):
+  ![Bitmap headers overview](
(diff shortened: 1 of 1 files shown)


generalmimon:
I wanted to try the updated BMP spec in some real application, so I created [BMP Tool](https://github.com/generalmimon/bmptool), which is a debugging tool for BMP images. Currently it can only convert BMP to PNG (with some additional dumps to the screen), but I intend it to also throw some warnings and repair ill-formed files.

I made a rigorous testing against all images from [BMP Suite](https://entropymine.com/jason/bmpsuite/bmpsuite/html/bmpsuite.html), which is an excellent collection of samples which test reading support of BMP files quite in depth. Happy to announce that there are only 4 files whose parsing failed (all from the folder marked as "clearly invalid"), all good and questionable samples are passing. You can actually run the tests yourself, I made it quite straightforward, just install [BMP Tool](https://github.com/generalmimon/bmptool) and follow the steps in Testing section of README.

I did my best to produce a spec to the format gallery that will be easy to use as much as possible, ensuring high format compliance with minimal effort. I think it's hard to find a BMP parser implementation that will excel more.

What do you think?

image/bmp.ksy
```diff
@@ -17,31 +17,104 @@ meta:
...
+     )
+      * BITMAPFILEHEADER (WINBMPFILEHEADER)
+      * BITMAPCOREHEADER (WIN2XBITMAPHEADER)
+      * BITMAPINFOHEADER (WINNTBITMAPHEADER)
+      * BITMAPV4HEADER (WIN4XBITMAPHEADER)
+
 seq:
-  - id: file_hdr
+  - id: file_header
     type: file_header
```
GreyCat: Technically, this [breaks C# code generation](https://github.com/kaitai-io/kaitai_struct/issues/86), but that's fine — it's still a code generation problem, not a format problem.

image/bmp.ksy
```diff
@@ -17,31 +17,104 @@ meta:
...
+    size: file_header.ofs_bitmap - 14 # 14 = file_header._sizeof (TODO: replace when KSC 0.9 is released)
+    type: bitmap_info
+  - id: bitmap
+    type: bitmap
+    size-eos: true
 types:
+  bitmap:
+    doc: |
+      Replace with an opaque type if you care about the pixels.
+      You can look at an example of a JavaScript implementation: https://github.com/generalmimon/bmptool/blob/master/src/Bitmap.js
```
GreyCat: Might want to add a link to https://github.com/kaitai-io/kaitai_struct/issues/188 with a lengthy discussion about exactly what you're talking about here.

image/bmp.ksy
```diff
@@ -52,53 +125,175 @@ types:
         -orig-id: bfOffBits
         type: s4
         doc: Offset to actual raw pixel data of the image
-  bitmap_core_header:
-    -orig-id: BITMAPCOREHEADER
-    doc-ref: https://msdn.microsoft.com/en-us/library/dd183372.aspx
+  bitmap_info:
+    -orig-id: BITMAPINFO
+    doc-ref: https://docs.microsoft.com/cs-cz/windows/win32/api/wingdi/ns-wingdi-bitmapinfo
```
GreyCat: Please reference en-us language version to keep all documentation in English (as it is in English anyway).

image/bmp.ksy
```diff
@@ -52,53 +125,175 @@ types:
...
-        type: u2
-        doc: Number of planes for target device, must be 1
-      - id: bits_per_pixel
-        -orig-id: bcBitCount
-        type: u2
-        doc: Number of bits per pixel that image buffer uses (1, 4, 8, or 24)
-  bitmap_info_header:
-    -orig-id: BITMAPINFOHEADER
-    doc-ref: https://msdn.microsoft.com/en-us/library/dd183376.aspx
+      - id: header_size
```
GreyCat: `len_header`?

image/bmp.ksy
```diff
@@ -52,53 +125,175 @@ types:
...
+            ? color_mask_given.alpha_mask
+            : 0
+
+  bitmap_header:
+    -orig-id:
+      - BITMAPCOREHEADER
+      - OS21XBITMAPHEADER
+    doc-ref:
+      - https://docs.microsoft.com/cs-cz/windows/win32/api/wingdi/ns-wingdi-bitmapcoreheader
+      - https://www.fileformat.info/format/os2bmp/egff.htm#OS2BMP-DMYID.3.1
```
GreyCat: Array in `doc-ref` is KS 0.9+ feature.

image/bmp.ksy
```diff
@@ -52,53 +125,175 @@ types:
...
+          is_color_mask_given and color_mask_given.has_alpha_mask
+            ? color_mask_given.alpha_mask
+            : 0
+
+  bitmap_header:
+    -orig-id:
+      - BITMAPCOREHEADER
+      - OS21XBITMAPHEADER
+    doc-ref:
+      - https://docs.microsoft.com/cs-cz/windows/win32/api/wingdi/ns-wingdi-bitmapcoreheader
```
GreyCat: Again, please replace `cs-sz` -> `en-us`

image/bmp.ksy
```diff
@@ -107,13 +302,194 @@ types:
...
+      profile_data:
+        io: _root._io
+        pos: 14 + ofs_profile # 14 = _root.file_header._sizeof
+        size: len_profile
+        type:
+          switch-on: _parent.bitmap_v4_ext.color_space_type == color_space::profile_linked
+          cases:
+            true: strz
+        if: has_profile
+        doc-ref: https://docs.microsoft.com/cs-cz/previous-versions/windows/desktop/wcs/using-structures-in-wcs-1-0 "If the profile is embedded,
```
GreyCat: Again, please replace `cs-cz` -> `en-us`.

GreyCat:
Thanks for this tremendous effort and, again, sorry for taking so long to review it. I've added some minor comments — please take a look? Otherwise, I think it's ready to be merged in.

image/bmp.ksy
generalmimon: OK, I haven't realized it. I thought about throwing the lists in `doc-ref` out to make it compatible with 0.8, but I anyway have to use `.as<f4>` (which is a 0.9 feature I suppose, doesn't work even in devel Web IDE) in `fixed_point_*` decimal types and I unfortunately  haven't figured out how to do it in 0.8. So I updated the `ks-version` to 0.9.

image/bmp.ksy
generalmimon: I renamed it back to `file_hdr`, it might not be a format problem, but it's anyway probably a good thing to have distinct names for type and field because one can instantly see what belongs to what.

image/bmp.ksy
```diff
@@ -16,7 +16,7 @@ meta:
     wikidata: Q192869
   endian: le
   license: CC0-1.0
-  ks-version: 0.8
+  ks-version: 0.9
```
GreyCat: Ok, this change makes it slightly more complicated: if we'll merge it as is, it will fail compilation on https://formats.kaitai.io/ at all, as it currently only fetches "latest stable" (v0.8) to use.

Ideal solution would be (finally) releasing 0.9 to make it "officially stable", and continue using that — otherwise, we'll need to have something clever like bringing multiple ksc versions into build...

image/bmp.ksy
GreyCat: Actually, this is another great point: we currently don't have a solid way to check these PRs before merging them. Implementing CI-style checks that will clearly show that this is not ok due to (following error messages) would be great...

GreyCat: ... and I completely agree, there's little point in clinging to 0.8. 0.9 is so much more advanced...

GreyCat:
Ok, so it turned out that we have this problem with 0.9 on major scale already, so one more format with version set to 0.9 doesn't make a difference — we'll need to solve it on a higher level.

Merging in, thanks for this massive effort again!

PR #244, opened 2019-11-06
Adding Terse Executable file format

MauriSanfi:
From UEFI Platform Initialization spec document:
"The Terse Executable (TE) image format was created as a mechanism to reduce the overhead of the PE/COFF headers in PE32/PE32+ images, resulting in a corresponding reduction of image sizes for executables running in the PI Architecture environment. Reducing image size provides an opportunity for use of a smaller system flash part."

So the TE format is basically a stripped version of PE.

This type of executables could be found inside the UEFI firmware. The UEFI firmware is stored in SPI flash memory, which is a chip soldered on a system’s motherboard. UEFI firmware is very modular: it usually contains dozens, if not hundreds, of executables. To store all these separates files, the firmware is laid out in volumes using the Firmware File System (FFS), a file system specifically designed to store firmware images. The volumes contain files that are identified by GUIDs and each of these files contain one or more sections holding the data. One of these sections contains the actual executable image. Most of the executable images follow the PE format. However, some of them follow the TE format.

If you can get a dump from a SPI, then you can use a tool like chipsec or uefi-firmware-parser to decode it and extract the executable images from the BIOS section inside the SPI dump. After the extraction you should check which of those files are TE binaries.
If you can not do all this to get the TE binaries, I can provide samples of them.

### DIFF
--- a/executable/te.ksy
+++ b/executable/te.ksy
@@ -1,6 +1,6 @@
 meta:
-  id: te
-  title: TE (Terse Executable) file format. A stripped version of PE (Portable Executable) format.
+  id: uefi_te
+  title: TE (Terse Executable) file.
   application: UEFI
   file-extension:
     - efi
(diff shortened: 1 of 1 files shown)


executable/te.ksy
```diff
@@ -0,0 +1,125 @@
...
+  id: te
+  title: TE (Terse Executable) file format. A stripped version of PE (Portable Executable) format.
+  application: UEFI
+  file-extension:
+    - efi
+    - te
+  ks-version: 0.7
+  license: CC0-1.0
+  endian: le
+doc-ref: https://uefi.org/sites/default/files/resources/PI_Spec_1_6.pdf
```
GreyCat: Can I ask you to add the description you already have in this PR as top-level `doc` element?

executable/te.ksy
```diff
@@ -0,0 +1,125 @@
+meta:
+  id: te
```
GreyCat: We have no namespaces or anything for these, so it's preferable not use very short acronyms, especially if they can be ambiguous for top-level IDs. Please consider something like `uefi_te`?

executable/te.ksy
```diff
@@ -0,0 +1,125 @@
...
+  title: TE (Terse Executable) file format. A stripped version of PE (Portable Executable) format.
```
GreyCat: "... format" and "Stripped version of ..." should not part of the title. The latter should be part of top-level `doc`.

executable/te.ksy
```diff
@@ -0,0 +1,125 @@
...
+    repeat-expr: te.number_of_sections
+types:
+  te_header:
+    seq:
+      - id: signature
+        contents: "VZ"
+      - id: machine
+        type: u2
+        enum: machine_type
+      - id: number_of_sections
```
GreyCat: Please consider following [style guide](https://doc.kaitai.io/ksy_style_guide.html) here and using something like `num_sections`?

GreyCat:
File needs to be renamed too, I'll do it myself. Sorry for long resolution time, and thanks for your contribution!

PR #247, opened 2019-11-20
add padding in udp_datagram.ksy

H4rryK4ne:
I propose to add padding to the `udp_datagram.ksy` since the `body` of `udp_datagram` is only of a specific length. Everything after the udp_datagram.body with size `length - 8` is actually no data.

### DIFF
--- a/network/udp_datagram.ksy
+++ b/network/udp_datagram.ksy
@@ -21,4 +21,7 @@ seq:
   - id: checksum
     type: u2
   - id: body
+    size: length - 8
+  - id: padding
     size-eos: true
+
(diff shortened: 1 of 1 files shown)


GreyCat:
Makes sense, through as per [style guide](https://doc.kaitai.io/ksy_style_guide.html#_trailing_padding) we normally specifically avoid to add padding just for sake of consumption.

Would if be ok if I'll propose to add `size: length - 8`, but skip padding?

H4rryK4ne:
Thanks for this hint --> Done! We will keep this in mind.

GreyCat:
Thanks for this contribution!

PR #248, opened 2019-12-02
Added some structures of Visual Basic

KOLANICH:
@williballenthin, @dzzie, @VBGAMER45, @bontchev, @skyfish4tb, @trietptm, @xRayDev, @3H54N, @Hostn4me, @qwas0514, @decalage2, @Big5-sec

### DIFF
--- a/executable/vba.ksy
+++ b/executable/vba.ksy
@@ -1,0 +1,1209 @@
+meta:
+  id: vba
+  title: Visual Basic exe structures
+  application: Visual Basic
+  file-extension: exe
+  xref:
+    wikidata: Q2378
+  endian: le
+  license: Apache-2.0
+  encoding: ascii
+  ks-opaque-types: true
+doc: |
+  Converted from https://github.com/williballenthin/python-vb/blob/master/vb/__init__.py with addition of some structures described in various reverse-engineered documentation
+  
+  # an external script converting raw virtual addresses to offsets in a file is needed, save it as resolve_offset.py
+  from kaitaistruct import KaitaiStruct, KaitaiStream
+  
+  class ResolveOffset(KaitaiStruct):
+    __slots__ = ("ofs",)
+    def __init__(self, rawAddress: int, root, _io: KaitaiStream, _parent=None, _root=None):
+      _root = root
+      self._parent = _parent
+      self.ofs = _root.pe.get_offset_from_rva(rawAddress - _root.pe.OPTIONAL_HEADER.ImageBase)
+
+doc-ref:
+  - https://github.com/williballenthin/python-vb/blob/master/vb/__init__.py
+  - https://web.archive.org/web/20071020232030if_/http://www.alex-ionescu.com/vb.pdf
+  - https://www.vb-decompiler.org/pcode_decompiling.htm
+  - https://sysenter-eip.github.io/VBParser#visual-basic-5-6-file-format-information
+params:
+  - id: pe
+  - id: offset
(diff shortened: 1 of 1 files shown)


executable/vba.ksy
```diff
@@ -0,0 +1,1209 @@
...
+      - id: ptr
+        type: u4
+    instances:
+      valid:
+        value: ptr != 0
+      resolved:
+        pos: 0
+        type: "resolve_offset(ptr, _root)" # otherwise it won't be passed
+      ofs:
+        value: resolved.ofs.as<u4>
```
generalmimon: That unfortunately won't work. Opaque types (such as `resolve_offset`) are called _opaque_, because you can't see what's inside, it's just a black box (from the perspective of KSY). Therefore you can't reference its properties from the Kaitai Struct language, because the compiler is unable to derive the type of the property or check if the property even exists. It makes sense, the opaque type is implemented only in the target language, the compiler hasn't access to the it while compiling KSY and even if it had, it couldn't understand it. _(I see it's unfortunately quite bad for your spec because you're counting on it in many places.)_

You may suggest _"so just turn off the property checking for the opaque types and trust the KSY dev"_, but then the type of the property would be just `AnyType`, and though you could in theory cast it to the arbitrary type you want, I'm still not sure it's wise. It's a quick solution though.

But I think the right thing to do is not to be lazy and think about the [interfaces](https://github.com/kaitai-io/kaitai_struct/issues/314), which would solve it quite nicely. You could define an interface in the KSY, where you specify an `ofs` property of type `u4` and then tell the KSC that the opaque type `resolve_offset` implements this interface.

And it's much more general than just turning off the checks on opaque types, it would also make the KSY development much more convenient. For example, you could have several versions of the same file format in different types and declare that all implement one interface. Then you'd switch them in some huge type `switch`, which makes accessing the common properties a nightmare, but not in that case, because the combined type would be our common interface.

executable/vba.ksy
```diff
@@ -0,0 +1,1209 @@
...
+      self._parent = _parent
+      self.ofs = _root.pe.get_offset_from_rva(rawAddress - _root.pe.OPTIONAL_HEADER.ImageBase)
+
+doc-ref:
+  - https://github.com/williballenthin/python-vb/blob/master/vb/__init__.py
+  - https://web.archive.org/web/20071020232030if_/http://www.alex-ionescu.com/vb.pdf
+  - https://www.vb-decompiler.org/pcode_decompiling.htm
+  - https://sysenter-eip.github.io/VBParser#visual-basic-5-6-file-format-information
+params:
+  - id: pe
```
generalmimon: Omitting `type` means byte array in Kaitai Struct (https://github.com/kaitai-io/kaitai_struct/issues/686), but I see you're doing things like `_root.pe.get_offset_from_rva()`. Even if you specified some opaque type, this again won't work from the same reason.

executable/vba.ksy
KOLANICH: >That unfortunately won't work. Opaque types (such as resolve_offset) are called opaque, because you can't see what's inside, it's just a black box (from the perspective of KSY).

For python this already works. For statically typed languages we strictly need interfaces proposal. Unfortunately noone has implemented it and I don't want to touch to Scala code.

executable/vba.ksy
KOLANICH: Thanks.

KOLANICH:
#264

PR #257, opened 2020-01-08
Added check-errors to GH Actions workflow

GreyCat:
This adds rudimentary error checking (true or false => breaking or not breaking the build) into our CI PR process.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,6 +49,9 @@ jobs:
         with:
           name: log.json
           path: _build/target/log.json
+      - name: Check build
+        working-directory: _build
+        run: ./check-errors target/log.json
       - name: Build diagrams
         working-directory: _build
         run: ./build-diagrams target html


_build/install-kscs
```diff
@@ -0,0 +1,32 @@
...
+for VER in 0.7 0.8; do
+	echo "Installing $VER"
+	rm -rf "$VER" "kaitai-struct-compiler-$VER"
+	wget "https://dl.bintray.com/kaitai-io/universal/$VER/kaitai-struct-compiler-$VER.zip"
+	unzip "kaitai-struct-compiler-$VER.zip"
+	mv "kaitai-struct-compiler-$VER" "$VER"
+done
+
+# Download latest unstable
+wget https://bintray.com/kaitai-io/universal_unstable/kaitai-struct-compiler/_latestVersion
```
KOLANICH: shouldn't `apt` be used for that?

GreyCat: No, the whole reason is that we download and use multiple versions simultaneously. There's little point in using apt repo for that.

PR #258, opened 2020-01-21
Eurocontrol Asterix format definitions

LmpOSX:
This is the first version of Asterix format KSY definition files.
This infrastructure gives the tools to define every category of Asterix with just two files, a Catalog of items, called cat_xxx_catalog.ksy, with just the Item Reference Number and the format of the field, and an UAP, called cat_xxx.ksy. These were all tested in python using real life capture file os the categories included.


asterix/ast_file.ksy
```diff
@@ -0,0 +1,43 @@
+meta:
+  id: ast_file
```
KOLANICH: the name is ambigous. AST can stand for "abstract syntax tree". Note that KS has no namespaces currently. Creating a new dir in tue root for such a niche application is strange. Probably `surveillance` dir should be created in the root and `asterix` should be moved there.

asterix/ast_data_block.ksy
```diff
@@ -0,0 +1,32 @@
+meta:
+  id: ast_data_block
+  file-extension: ast
+  endian: be
+  license: GPL-3.0-only
```
KOLANICH: Surveillance is evil, copyright is also evil. But since it is already under copyright, **how about AGPL here**? GPL doesn't prevent parties involved into surveillance from privately using this spec.

asterix/ast_file.ksy
```diff
@@ -0,0 +1,43 @@
...
+  license: GPL-3.0-only
+  imports:
+    - ast_message
+  ks-version: 0.9
+
+doc: |
+  Implementantion of ASTERIX format.
+
+  From Eurocontrol Asterix web page at
+  https://www.eurocontrol.int/asterix
```
KOLANICH: doc-ref

asterix/cat_000.ksy
```diff
@@ -0,0 +1,9 @@
+meta:
+  id: cat_000
```
KOLANICH: I guess more human-readable names are needed

asterix/ast_file.ksy
LmpOSX: I agree asterix probably is not to bee in root, but surveillance brings me CCTV and other things to my mind. Probably aeronautical or aero is a good name because there are other protocols in that field, even for radar data, you have also flight plan, etc.
About AST, in the class names it is a convenience name so you not write asterix, so it could be changed, no problem with that.

asterix/COPYING
```diff
@@ -0,0 +1,674 @@
+GNU GENERAL PUBLIC LICENSE
```
KOLANICH: I guess we shouldn't put a licence into this dir, instead we should to submodule https://github.com/IQAndreas/markdown-licenses to the root.

asterix/cat_001.ksy
```diff
@@ -0,0 +1,93 @@
...
+  - id: fspec
+    type: field_spec
+
+  - id: data
+    type:
+      switch-on: is_plot
+      cases:
+        true:  uap_type(fspec, [10,20,40,70,90,130,141,50,120,131,80,100,60,30,150].as<s2[]>.as<str>)
+        false: uap_type(fspec, [10,20,161,40,42,200,70,90,141,130,131,120,170,210,50,80,100,60,30].as<s2[]>.as<str>)
+
```
KOLANICH: should the abgreviation be expanded?

asterix/cat_002.ksy
```diff
@@ -0,0 +1,166 @@
...
+
+instances:
+
+  category:
+    value: '"002"'
+
+  version:
+    value: '"1.00"'
+
+  i010:
```
KOLANICH: I guess the names should be more meaningful

asterix/compound.ksy
```diff
@@ -0,0 +1,86 @@
...
+
+  The compound field format is a variable lenght and content field.
+  This uses four parameters:
+
+  item, which is just the item name, as I001/010, or I062/390. It is just
+  informative, and is not used during parsing, but permits consult the item name
+  in the target language.
+
+  f_fmt, is a string with a list of the formats of the subfields,
+  The available formats are:
```
KOLANICH: Each field can have a `doc`. Probably this info should be moved there

asterix/ast_file.ksy
LmpOSX: OK, I'll move there.

asterix/cat_000.ksy
LmpOSX: All names reply how they are referenced in the standards and in jargon, as much as Kaitai permits. Other names would be difficult to be correctly referenced by an hipothetical user. It is supposed if you use this files you already read the standards and know what the meaning of block, fspec CAT030 or I020/290 is.

asterix/cat_001.ksy
LmpOSX: What do you mean with abbreviations?

asterix/compound.ksy
```diff
@@ -0,0 +1,86 @@
...
+        '"F".as<u1>': fixed("",flags.octects[_index/7].bits[_index % 7],f_pri_sizes[_index]) #F
+        '"R".as<u1>': repetitive("",flags.octects[_index/7].bits[_index % 7],f_pri_sizes[_index]) #R
+        '"X".as<u1>': extended("",flags.octects[_index/7].bits[_index % 7],f_pri_sizes[_index],f_sec_sizes[_index]) #X
+        _: empty
+    repeat: expr
+    repeat-expr: total_flags
+
+instances:
+  fmts:
+    value: f_fmt.as<u1[]>
```
KOLANICH: https://github.com/kaitai-io/kaitai_struct/issues/686

asterix/cat_002.ksy
LmpOSX: i010, is ITEM 010, from Catalog, as named by the Corresponding Category definition. i010 is the only name, there is no other. Then once you access the item, you could decode various possible values encoded in it. for example Item 10, is a 2 octect field, that has encoded two values SIC and SAC, this has to be decoded by calling program or in instance value. But the important thing here is that if you need to decode a category that is not in the repository you could define new one in just 10 minutes, and possible contribute it.

asterix/compound.ksy
```diff
@@ -0,0 +1,86 @@
...
+  - id: f_sec_sizes
+
+seq:
+  - id: flags
+    type: field_spec
+  - id: data
+    type:
+      switch-on: fmts[_index]
+      cases:
+        '"E".as<u1>': explicit("",flags.octects[_index/7].bits[_index % 7]) #E
```
KOLANICH: shouldn't an instances within a custom type be used to encapsulate 7-related logic?

asterix/explicit.ksy
```diff
@@ -0,0 +1,46 @@
...
+  Implementantion of ASTERIX format.
+  
+  From Eurocontrol Asterix web page at
+  https://www.eurocontrol.int/asterix
+
+  Explicit Standard Field format.
+
+  The explicit field format is a variable lenght format.
+
+  It uses two parameters:
```
KOLANICH: `doc` can be used for params too.

asterix/compound.ksy
LmpOSX: I preferred to explain a basic use form doc, imagined the field docs to have more specific information.

asterix/compound.ksy
KOLANICH: Fixed thanks to @generalmimon ! @LmpOSX, feel free to pass bytes as a param.

asterix/cat_001.ksy
KOLANICH: uap, isp, irfs,... Though feel free not to change if you consider that the abbreviations are better here.

asterix/compound.ksy
LmpOSX: I tried, it is not possible, because the instance is calculated only the first time you call it, and the index is not declared which throws an error in python, see https://stackoverflow.com/questions/59653098/using-repeat-expr-index-in-kaitai-struct-instance-calculated-value-use-not-def

asterix/cat_001.ksy
LmpOSX: I try to keep it consistent with vocabulary in the standard. Although it could be have changes in the future. Take into account that the names are restricted to Kaitai impositions, for example I would use UAP, instead of uap, but it is not possible.

asterix/compound.ksy
LmpOSX: Future versions would have additional explanations, for example an README with code examples is one of my objectives.

asterix/compound.ksy
LmpOSX: It worked, but with negative and bigger than 255 it doesn't work

LmpOSX:
Is there anything else left to have this merged to master?

aeronautical/asterix/extended.ksy
```diff
@@ -0,0 +1,60 @@
...
+  Implementantion of ASTERIX format.
+
+  From Eurocontrol Asterix web page at
+  https://www.eurocontrol.int/asterix
+
+  Extended Standard Field format.
+
+  The extended field format is a variable lenght format.
+
+  It uses four parameters:
```
KOLANICH: still not fixed

aeronautical/asterix/asterix_file.ksy
```diff
@@ -0,0 +1,46 @@
...
+  https://www.eurocontrol.int/asterix
+
+  "The all-purpose structured EUROCONTROL surveillance information exchange (ASTERIX)
+  is a set of documents defining the low level ('down to the bit') implementation
+  of a data format used for exchanging surveillance-related information and other ATM applications.
+  ASTERIX is designed for communication media with limited bandwidth. This is why it follows rules
+  that enable it to transmit all the information needed, with the smallest data load possible.
+
+  The ASTERIX library consists of several parts, each of them describing the encoding of information
+  related to a specific application.
```
GreyCat: This message is copied verbatim from https://www.eurocontrol.int/asterix — are you sure that we're allowed to copy it?

aeronautical/asterix/cat_000.ksy
```diff
@@ -0,0 +1,9 @@
+meta:
+  id: cat_000
+  license: GPL-3.0-only
+  endian: be
+
+seq:
+  - id: bytes
+    type: u1
+    repeat: eos
```
GreyCat: Why not `size-eos: true`? What's the value in having a true array of bytes here?

aeronautical/asterix/asterix_data_block.ksy
```diff
@@ -0,0 +1,35 @@
+meta:
+  id: asterix_data_block
+  file-extension: ast
+  endian: be
+  license: GPL-3.0-only
```
GreyCat: Can you explain (1) what's the expected consequence of applying GPL3 to these files? (2) why did you make that decision — i.e. are there any reasons which force GPL3, or is it your personal choice?

aeronautical/asterix/repetitive.ksy
```diff
@@ -0,0 +1,55 @@
...
+
+doc: |
+  Implementantion of ASTERIX format.
+
+  From Eurocontrol Asterix web page at
+  https://www.eurocontrol.int/asterix
+
+  Repetitive Standard Field format.
+
+  The repetitive field format is a variable lenght format.
```
GreyCat: length?

aeronautical/asterix/fixed.ksy
```diff
@@ -0,0 +1,47 @@
+meta:
+  id: fixed
```
GreyCat: We can't really have these generic names, as they will be repo-global. Please consider something like `asterix_fixed`?

aeronautical/asterix/compound.ksy
```diff
@@ -0,0 +1,86 @@
...
+  - id: flags
+    type: field_spec
+  - id: data
+    type:
+      switch-on: fmts[_index]
+      cases:
+        '"E".as<u1>': explicit("",flags.octects[_index/7].bits[_index % 7]) #E
+        '"F".as<u1>': fixed("",flags.octects[_index/7].bits[_index % 7],f_pri_sizes[_index]) #F
+        '"R".as<u1>': repetitive("",flags.octects[_index/7].bits[_index % 7],f_pri_sizes[_index]) #R
+        '"X".as<u1>': extended("",flags.octects[_index/7].bits[_index % 7],f_pri_sizes[_index],f_sec_sizes[_index]) #X
```
GreyCat: If these are essentially fixed set of *identifiers*, why not create a enum with these 4 values and use meaningful names here instead of `"E".as<u1>` (which would actually fail to compile in many languages).

aeronautical/asterix/cat_034_catalog.ksy
```diff
@@ -0,0 +1,52 @@
...
+  license: GPL-3.0-only
+  endian: be
+  imports:
+#    - bit_map
+    - fixed
+    - explicit
+    - extended
+    - repetitive
+    - compound
+  ks-version: 0.9
```
GreyCat: Normally, all top-level types would at bare minimum require the `title` to be set.

aeronautical/asterix/cat_034_catalog.ksy
```diff
@@ -0,0 +1,52 @@
...
+
+  From Eurocontrol Asterix web page at
+  https://www.eurocontrol.int/asterix
+
+  Category 034 Catalog definition file
+
+  ASTERIX Part 2b Category 34
+  Transmission of Monoradar Service Messages
+
+  SUR.ET1.ST05.2000-STD-02b-01
```
GreyCat: This actually belongs to `doc-ref` — it's possible to put non-URL references there, and nowadays it's possible to put multiple, i.e.

```
doc-ref:
  - https://www.eurocontrol.int/publication/cat034-eurocontrol-specification-surveillance-data-exchange-part-2b
  - SUR.ET1.ST05.2000-STD-02b-01, Edition Number: 1.27
```

Another thing you might want to consider is `meta/xref/...` and reserving special tag for these SUR references there.

aeronautical/asterix/cat_001.ksy
```diff
@@ -0,0 +1,99 @@
...
+  Category 001 definition file
+  RADAR DATA EXCHANGE Part 2a
+  Transmission of Monoradar Data Target Reports
+
+  SUR.ET1.ST05.2000-STD-02a-01
+
+  Edition : 1.2
+
+doc-ref: |
+  https://www.eurocontrol.int/publication/cat001-eurocontrol-standard-document-radar-data-exchange-part-2a
```
KOLANICH: What is the sense to make a multiline string here?

generalmimon:
Binary file [aeronautical/.DS_Store](https://github.com/kaitai-io/kaitai_struct_formats/pull/258/files/8cb8ac7a47d3a8fe42630941a51dfc64b460f99d#diff-1f70807e53d9d1b2f4cdb903c396783a) should never be tracked by Git. It's best to add `.DS_Store` to your global .gitignore.

aeronautical/asterix/repetitive.ksy
```diff
@@ -0,0 +1,55 @@
...
+  flag: this is a bool parameter, which determines if there have to be parsed or
+  just skipped, mainly used by compound. In Catalog files, the flag is always true
+
+  f_size, which is the size of the repetitive parts.
+
+params:
+  - id: name
+    type: str
+  - id: flag
+    type: b1
```
generalmimon: `b1` as a param type is weird, I think `bool` type is more recognizable.

KOLANICH: Do we really have `bool` type? If we do, why not to remove it because `b1` seems to be enough?

aeronautical/asterix/asterix_file.ksy
LmpOSX: As I don't know, I've just sent an email to Eurocontrol on this, waiting for response.

aeronautical/asterix/cat_000.ksy
LmpOSX: Is there any drawbacks with this? Or a performance issue?. I opted to leave this way, because during testings I could change u1 with other types or integers changing just type.

aeronautical/asterix/asterix_data_block.ksy
LmpOSX: I want it remain open source. Kolanich suggested AGPL, I think I will switch, but have no opportunity to make it yet.

aeronautical/asterix/repetitive.ksy
LmpOSX: yes

aeronautical/asterix/fixed.ksy
LmpOSX: This will require changing explicit, repetitive, extended and compound types also. And will give a bunch of boiler plate asterix_ all over the category catalogs and definitions. But If this is an impediment the I'll change it.

aeronautical/asterix/asterix_data_block.ksy
GreyCat: The usual disclaimer that I'm not a lawyer applies.

The problem with GPL or AGPL is that it's very unclear what that means in relation to KSY. KSY specifications are not exactly "code" — they're more like documentation. There is no single straightforward way to translate "source code" into "object code", it's unclear what constitutes a "modification" or "modified version", what consitutes "propagation", etc.

For example, if your original intent was to make sure it's "viral", i.e. everyone who uses your specs will need to apply GPLv3 to their whole codebase, then how would you deal with:

* Somebody using reference ksc compiler to generate code based on your .ksy files?
* Somebody reading your specs as documentation and implementing their own parser code based on that?
* Somebody rendering your specs as a chart using reference ksc compiler + then implementing their own parser code looking at the resulting charts?
* Somebody using alternative compiler / totally alternative approach to generate some code based on your .ksy files?

This is pretty gray legal area.

aeronautical/asterix/compound.ksy
LmpOSX: I need to pass a list of formats in a simple and effective way. So I pass a string, of F,R,E,X which indicate the formats of the subfields, for example "FFREFFX".

Using an enum will impose the name of enum, asterix_formats, for example, then a name for every value. Those names have to be in lower case, so there is no asterix_formats::R for "R", and then have to be asterix_format::r, or perhaps, asterix_format::repetitive, but then, the meaning and eloquence of F, R, E, F is lost.

Also, I could not find the way yo indicate 'E', or 'R', instead of "R" or "E", because ksc think it is a variable name.

On the other hand, I recognize that having .as<u1> is cumbersome I tried to avoid it in every form, but it is the price for using [_index], instead of .substring(_index,_index+1) which I think is worst, at least as far as I had tested. Also, it compiles in almost every language.

aeronautical/asterix/repetitive.ksy
LmpOSX: User guide at 6.1. Basic data types, say type spec for booleans is b1.

asterix/COPYING
LmpOSX: Will try, don't know how exactly to do that. If you could give me a help it is Wellcome.

aeronautical/asterix/extended.ksy
LmpOSX: Will change it and upload new version.

aeronautical/asterix/repetitive.ksy
generalmimon: @KOLANICH, @LmpOSX:

Seems that the user guide is too long, we should apparently strive to make it less boring. But please read at least the following excerpt from https://doc.kaitai.io/user_guide.html#param-types:

> `type` specification in `params` definition is slightly different from normal `type` specification in `seq` or `instances`. It specifies pure concept of "type", as would be used to represent data in a programming language, without any serialization details (like endianness, sizes, conversions, encodings, etc). Thus, one can't use any extra keys to specify that, but instead one can use:
> 
> * no type to specify byte arrays
> * `str` to specify strings
> * `bool` to specify booleans
> * `struct` to allow arbitrary KaitaiStruct-compatible user types
> * `io` to specify KaitaiStream-compatible IO streams
> * `any` to allow any type (if target language supports that)


I completely agree with this: `b1` type means "one bit", which is more like a serialization detail than specifying a data type. It might not be clear to someone who has no experience with Kaitai Struct, that the possible values of `b1` are booleans `true` or `false`, not integers `0` or `1`. There's nothing we can do with it in `seq` fields and instances, but I think for params `bool` type should be used instead. It makes more sense: you're passing a boolean, not a "one bit of data". And it's legible for all programmers, not just for the handful of people that work with Kaitai Struct every day.

Additionaly, I'd prefer to ban `b[0-64]` in params `type` completely, because it's IMHO exclusively a serialization detail . It doesn't really say which type will be used in the target language (FYI the current KSC uses `uint64` type for all bit types - except for `b1` of course). On top of that, it suggests that for example one byte (`s1`) won't fit in `b6` param, which apparently does not apply.

Feel free to leave `b1` there if you insist, but I personally think it harms specification readability and intelligibility, which I perceive as one of the great things on Kaitai Struct.

aeronautical/asterix/compound.ksy
generalmimon: > but it is the price for using [_index], instead of .substring(_index,_index+1) which I think is worst, at least as far as I had tested.

Relying on such tricky dirty hacks doesn't worth it. Both typecastings `"E".as<u1>` and `f_fmt.as<u1[]>` are just irresponsible toys for fooling the whole type system in the compiler and making mostly wrong presumptions about how the data types in the target languages work.

We're talking about luck here that it works _in JavaScript_, that's only because syntax for indexing string is the same as indexing an array and it doesn't have type checks. But `f_fmt.as<u1[]>` won't compile in Java, for example, because it justly doesn't know how to execute `((ArrayList<Integer>) ("some string"))` and says `error: incompatible types: String cannot be converted to ArrayList<Integer>` .

I don't understand why you have such a problem with `.substring(_index,_index+1)`, but these dirty hacks are not the way to go.

aeronautical/asterix/repetitive.ksy
KOLANICH: >Additionaly, I'd prefer to ban b[0-64] in params type completely,

I don't think so. In future we may want to support languages like verilog, for them `b1` or `b3` as a parameter perfectly make sense.

aeronautical/asterix/asterix_data_block.ksy
LmpOSX: Is ksc a compiler?

KOLANICH: We usually abbreviate `Kaitai Struct Compiler` (in fact it is a transpiler) as `ksc`.

aeronautical/asterix/fixed.ksy
GreyCat: Hang on, we'll need to (1) resolve the license question first, and once/if we'll do that, then (2) my next question will be "what is the value in keeping this as many .ksy files?"

I mean, there *is* a namespacing mechanism. For example, you do a single top-level type called `asterix` and then all types inside it will be namespaced:

```yaml
meta:
  id: asterix
types:
  fixed: {}
  repetitive: {}
  extended: {}
  compound: {}
```

From the outside world, these will be seen as `Asterix.Fixed`, `Asterix.Repetitive`, etc. From the inside of the file, you can still refer to them as `fixed`, `repetitive`, etc.

aeronautical/asterix/compound.ksy
GreyCat: > I need to pass a list of formats in a simple and effective way.

Enums *are* the simple, effective and performant way to do so in binary formats world. This is exactly the place where enums should be used — when we encode some *meaning* with constants, either it's "R" for "repetitive" or 0x52 for "repetitive", it is exactly what enums are made for.

The whole point of writing a format specification is exactly to hide particular encoding scheme from an end user. There should be no "R" or 0x52 or any other magical constants that one needs to understand the format.

For the typecasting of strings to `u1`, that's actually a good point to make sure that ksc itself checks that it should not be allowed.

LmpOSX: Typecasting is in User guide, I would prefer not using this probably.
On the other hand, Substring method documentation says:  Extracts a portion of a string between character at offset from and character at offset to (inclusive of characters at from and to offsets).
So, since it is inclusive of characters at from and to offsets, to get a one char substring, to and from are the same, which is weird but ok. But then you end up with an empty substring in python in the form a[3:3], so I realized that this was inconsistent and then, I wondered what the upper limit for a counter would be. Then when you reach the upper limit, _index+1 is the limit, or it is _index, and _index+1 is what substring needs on second parameter to return only one char. That is why I don't like  substring(i,i+1), any one reading this code won't know exactly what the counter is. That is why I prefer using this cast. But I agree it could be better to have a cleaner method for indexing substrings.

aeronautical/asterix/asterix_data_block.ksy
LmpOSX: Ksc is a compiler, it translates some source file (ksy) into some compiled code for target you want.
The Ksy files in this folder could be enhanced, I still expect this, and I want those enhanced and new ksy remain open source.
Answering your questions:
1) They have to attach those ksy files, as a source code.
2) If they implement their own I could not stop them.
3) same as above.
4) same as 1.

KOLANICH: 2. and 3. contradicts 4

aeronautical/asterix/fixed.ksy
LmpOSX: Well, it is a file for a type, don't understand why it is bad since in general this is preferable. In any case. I could have an asterix file, and define message, block, record, and also fixed, repetitive extended and compound. But then Asterix will import cat_002 (similar to asterix_record.ksy) , which in turn imports cat_002_catalog, which in turn need to import asterix, to have access to fixed, repetitive etc. ¿Is this supported? because I get an error when doing this circular reference imports.

aeronautical/asterix/compound.ksy
generalmimon: > Substring method documentation says: Extracts a portion of a string between character at offset `from `and character at offset `to `(**inclusive** of characters at from and to offsets).

This is apparently a bug in the documentation. I created a test [ExprStrOps](https://github.com/kaitai-io/kaitai_struct_tests/blob/cb4edb86971f714bf56eef66925827015016a77b/formats/expr_str_ops.ksy) (the results are in the [CI dashboard](https://ci.kaitai.io/)) which tests the `substring` method. All languages use inclusive `from` and **exclusive `to`** _(except Ruby, where the current codegen for `substring` is incorrect, but indeed as the only language uses inclusive `to`, so `"0123456789"[2..7]` yields `"234567"`)_ and this is the **actual behavior** of the Kaitai Struct language. Sorry for the confusion, I'll fix the docs ASAP.

> That is why I don't like substring(i,i+1), any one reading this code won't know exactly what the counter is.

I don't think so. It's quite readable for everyone who knows what `substring` does. And iterating individual characters of a string is a common thing, so I think that most people will know _what the counter is_.

> That is why I prefer using this cast.

You are apparently still not convinced. So again, doing typecasting like `"E".as<u1>` or `"str".as<u1[]>` is not an option. String and number (or array of numbers) are incompatible types and thus casting is undefined behavior. It's unpredictable how this behaves in the target languages until you think out each individually, and it's just a coincidence if it works (I already described why it works in JavaScript). However, it doesn't make any sense to allow doing this, because it's impossible to convert between number and string without knowing [the encoding](https://en.wikipedia.org/wiki/Character_encoding). You seem to silently propose to assume that the string contains only ASCII characters, but what if not?

Simply put, it's a bug in the compiler that it's not yet forbidden (@GreyCat already opened an issue kaitai-io/kaitai_struct#696 tracking this) and will be fixed in the future. Right now, relying on this is strongly discouraged. Feel free to exploit this bug in your own in-house repository at your own risk, but it just can't be in the official format gallery.

jblestang:
Really sad to see this branch disappear. Any reason why?

LmpOSX:
I preferred to retire it until I could comply with al request from the revisions made, but I had not enough time to finish it. Actually I knew from the beginning I could not give much attention if some time it got merged, so my original intention was to have a good working structure that other people could use, enhance, and define new categories. Perhaps it is better to open an independent project so all formal requisites are not mandatories and anyone could modify it to include in Kaitai format gallery.

PR #259, opened 2020-02-01
mp4 (bit dirty, incomplete)

petr-briza:
Parser for MP4 container based on ISO/IEC 14496-14:2003 specs. There are some comments at the start of .ksy. Not all of "box" (format atomic blocks) types are implemented but enough to get some media metadata (like lenght, track types)

### DIFF
--- a/media/mp4.ksy
+++ b/media/mp4.ksy
@@ -1,0 +1,408 @@
+meta:
+  id: mp4
+  file-extension: mp4
+  endian: be
+  license: CC0-1.0
+  encoding: ascii
+  
+#
+# The container is a tree-like structure of boxes.
+# There are various types of boxes. Boxes of some types can contain nested boxes.
+# You can find some interesting information in those boxes:
+#
+# -> moov -> mvhd -> duration / timescale    :: this is how you calculate media length in seconds
+# -> moov -> trak                            :: those are tracks in the container (video/audio/etc.)
+# trak -> tkhd -> @width x @height           :: resolution of the video track
+# trak -> mdia -> minf -> stbl -> stsd.entries[0].type :: fourcc of codec used for the track
+# trak -> mdia -> mdhd -> langCode           :: language code of the track
+#
+# trak -> mdia -> stb -> stts                :: table to get sample index for given time
+# trak -> mdia -> stb -> stsc                :: table to get chunk index for given time
+# trak -> mdia -> stb -> stco                :: table to get offset in file for given chunk index
+#
+# See also:
+# * http://mp4parser.com/
+# * https://archive.codeplex.com/?p=mp4explorer
+# * https://github.com/axiomatic-systems/Bento4
+
+seq:
+  - id: boxes
+    type: box
+    repeat: eos
+
+types:
+
+  boxes:
+    seq:
+      - id: boxes
+        type: box
+        repeat: eos
+
+  box:
+    seq:
+      - id: size
+        type: u4
+      - id: type
+        type: str
+        size: 4
+      - id: largesize
+        type: u8
+        if: size == 1
+      - id: body
+        type:
+          switch-on: type
+          cases:
+            '"moov"': boxes
+            '"trak"': boxes
+            '"edts"': boxes
+            '"mdia"': boxes
+            '"stbl"': boxes
+            '"minf"': boxes
+            '"dinf"': boxes
+            '"ftyp"': box_body_ftyp
+            '"hdlr"': box_body_hdlr
+            '"mvhd"': box_body_mvhd
+            '"vmhd"': box_body_vmhd
+            '"mdhd"': box_body_mdhd
+            '"tkhd"': box_body_tkhd
+            '"elst"': box_body_elst
+            '"stsc"': box_body_stsc
+            '"stsd"': box_body_stsd
+            '"stco"': box_body_stco
+            '"stts"': box_body_stts
+            '"co64"': box_body_co64
+            _: box_body_other
+        size: '(size == 1) ? size - 12 : size - 8'
+
+  # -----------
+  # -- FTYP  --
+  # -----------
+  box_body_ftyp:
+    seq:
+      - id: major_brand
+        type: str
+        size: 4
+      - id: minor_version
+        type: str
+        size: 4
+      - id: compatible_brands
+        type: str
+        size: 4
+        repeat: eos
+#        size-eos: true
+
+  # -----------
+  # -- MVHD  --
+  # -----------
+  box_body_mvhd:
+    seq:
+      - id: version
+        type: u4
+
+      - id: creation_time
+        type:
+          switch-on: version
+          cases:
+            1: u8
+            _: u4
+      - id: modification_time
+        type:
+          switch-on: version
+          cases:
+            1: u8
+            _: u4
+      - id: timescale
+        doc: |
+          Timescale is an integer that specifies the time-scale for the entire presentation; this is the number of time units that pass in one second.
+          For example, a time coordinate system that measures timein sixtieths of a second has a time scale of 60.
+        type: u4
+      - id: duration
+        doc: |
+          duration is an integer that declares length of the presentation (in the indicated timescale).
+          This property is derived from the presentation's tracks: the value of this field corresponds to the duration of the 
+          longest track in the presentation. If the duration cannot be determined then duration is set to all 1s.
+        type:
+          switch-on: version
+          cases:
+            1: u8
+            _: u4
+      - id: rate
+        doc: Rate is a fixed point 16.16 number that indicates the preferred rate to play the presentation; 1.0 (0x00010000) is normal forward playback
+        type: u4
+      - id: volume
+        doc: Volume is a fixed point 8.8 number that indicates the preferred playback volume; 1.0 (0x0100) is full volume
+        type: u2
+      - id: reserved
+        type: u2
+        repeat: expr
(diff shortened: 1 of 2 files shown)


GreyCat:
Thanks for this submission! Can I ask you to review the [style guide](http://doc.kaitai.io/ksy_style_guide.html) and apply some changes to this to match? I'll add more specific comments to the file itself.

media/mp4.ksy
```diff
@@ -0,0 +1,408 @@
...
+#
+# -> moov -> mvhd -> duration / timescale    :: this is how you calculate media length in seconds
+# -> moov -> trak                            :: those are tracks in the container (video/audio/etc.)
+# trak -> tkhd -> @width x @height           :: resolution of the video track
+# trak -> mdia -> minf -> stbl -> stsd.entries[0].type :: fourcc of codec used for the track
+# trak -> mdia -> mdhd -> langCode           :: language code of the track
+#
+# trak -> mdia -> stb -> stts                :: table to get sample index for given time
+# trak -> mdia -> stb -> stsc                :: table to get chunk index for given time
+# trak -> mdia -> stb -> stco                :: table to get offset in file for given chunk index
```
GreyCat: This block of text looks like a good canididate for top-level `doc` element?

media/mp4.ksy
```diff
@@ -0,0 +1,408 @@
...
+#
+# See also:
+# * http://mp4parser.com/
+# * https://archive.codeplex.com/?p=mp4explorer
+# * https://github.com/axiomatic-systems/Bento4
```
GreyCat: And these look like candidates for `doc-ref`?

media/mp4.ksy
```diff
@@ -0,0 +1,408 @@
+meta:
+  id: mp4
+  file-extension: mp4
+  endian: be
+  license: CC0-1.0
```
GreyCat: At the very least, we'll need `title` here so it would be possible to find this file on the website.

media/mp4.ksy
```diff
@@ -0,0 +1,408 @@
...
+    type: box
+    repeat: eos
+
+types:
+
+  boxes:
+    seq:
+      - id: boxes
+        type: box
+        repeat: eos
```
GreyCat: This looks like the very same as root type, i.e. `mp4`?

media/mp4.ksy
```diff
@@ -0,0 +1,408 @@
...
+        type: str
+        size: 4
+      - id: largesize
+        type: u8
+        if: size == 1
+      - id: body
+        type:
+          switch-on: type
+          cases:
+            '"moov"': boxes
```
GreyCat: You can use root type, i.e. `mp4` instead of `boxes` everywhere.

media/mp4.ksy
```diff
@@ -0,0 +1,408 @@
...
+        type: u4
+      - id: entry_count
+        type: u4
+      - id: entries
+        type: box_body_stsd_entry
+        repeat: expr
+        repeat-expr: entry_count
+        
+  box_body_stsd_entry:
+    seq:
```
GreyCat: For subtypes like these (i.e. which are clearly contained to one "box"), please consider using nested types, i.e.

```yaml
seq:
  # ...
  - id: entries
    type: entry
    repeat: expr
    repeat-expr: entry_count
types:
  entry:
# ...
```

media/mp4.ksy
```diff
@@ -0,0 +1,408 @@
...
+            '"vmhd"': box_body_vmhd
+            '"mdhd"': box_body_mdhd
+            '"tkhd"': box_body_tkhd
+            '"elst"': box_body_elst
+            '"stsc"': box_body_stsc
+            '"stsd"': box_body_stsd
+            '"stco"': box_body_stco
+            '"stts"': box_body_stts
+            '"co64"': box_body_co64
+            _: box_body_other
```
GreyCat: You can just skip this as get it automatically captured as byte array if a match fails.

media/mp4.ksy
```diff
@@ -0,0 +1,408 @@
...
+  boxes:
+    seq:
+      - id: boxes
+        type: box
+        repeat: eos
+
+  box:
+    seq:
+      - id: size
+        type: u4
```
GreyCat: `len_body` to match style guide?

petr-briza:
Hi @GreyCat actually I noticed now that QuickTime .mov is almost the same format. Is there any possibility of "including" between several .ksy files? Cannot find it in doc. Maybe good idea would be to merge the same structures between those formats into one "included" file.

media/mp4.ksy
```diff
@@ -0,0 +1,408 @@
...
+  # -----------
+  box_body_mvhd:
+    seq:
+      - id: version
+        type: u4
+
+      - id: creation_time
+        type:
+          switch-on: version
+          cases:
```
KOLANICH: I see pretty many records like tyis one, shoult tyis selector be encapsulated into a type?

GreyCat:
@petr-briza 
> actually I noticed now that QuickTime .mov is almost the same format.

Yeah, I've just realized that too. Actually, I have a very stupid question: are they actually different file formats, and is there any value in keeping them separate?

> Is there any possibility of "including" between several .ksy files? Cannot find it in doc. Maybe good idea would be to merge the same structures between those formats into one "included" file.

Sure, we have [imports](http://doc.kaitai.io/user_guide.html#_importing_types_from_other_files), and, in 0.9 you can even reuse internal structures of the other file. But before jumping to that, let's think if we *really* want to have them separate?

petr-briza:
@GreyCat

> But before jumping to that, let's think if we _really_ want to have them separate?

Actuallly I think: yes. For the very same reason I didn't noticed it's the same format as QuickTime .mov – people might now know it's the same format. If you want to keep only single .ksy file, what options do you have for the file name?

- Something like _ISO-14496-12-2015.ksy_ – most precise but most confusing. Who is no expert on the format (yet) might just not know that this ISO means .mov or .mp4. Such developer will just look to the list of formats, he won's see the desired format and move on (maybe will start writing his own parser, just like me).
- You can have both file formats in the name, like _mp4_mov_iso.ksy_ but this is messy, cannot be sorted in alphabetical order
- My suggestion is just to have two separate .ksy files, one for .mov and one for .mp4, and inside you would have "imports" for the shared part – which could be entire definition, int his case. MP4 is extension of the ISO, I just didn't reach level of implementation yet where I actually used any of the extended boxes.

You might also noticed there is the same problem for example with Microsoft PE format – which is used for .exe, .dll, .sys but also .ax, .drv, .src etc. If somebody wants to parse .ax to find a company name in the file etc, he might not realize there is already a parser called microsoft_pe.ksy.

Another approach would be to create "index" of parsers where you would have a table of "file extensions" → parser, another could be "mime type" → parser. This would help to find a parser easily and in such cases I would recommend to go for the "precise" file names like "ISO-14496-12-2015.ksy"

In all cases I would also recommend separating "file parsers" from "network protocol parsers", "memory block" whatever, etc. For other categories you could have different "index tables" based on different keys.

What do you think?

GreyCat:
Ok, so there are two separate topics:

* Is there any real difference between these two formats that warrants separation? For example, if one clearly includes things which should be treated differently in the other?
* How to make these formats easier to find if people are searching for "mov", "mp4" and other popular terms?

I believe the 2nd question is more universal and is not specific to this particular case (as you've already demonstrated with .exe, .dll, .ax etc examples). Let's focus on the first issue, get our fundamentals perfect and then we'll improve navigation/search experience separately.

petr-briza:
Answer to the second question will have influence on decisions for the first question, because:

- If you want to keep the names according to firmat standard itself, you should rename the "QuickTime MOV" to more generic "ISO" name. And the 1st answer would be "no, this should be just one format". At least at the current implementation stage when we don't have any .MOV-specific or any .MP4-specific boxes implemented yet.
- If you want naming system rarther according to "file extensions" or "mime types", in such case I would answer to the 1st question "yes, let's separate the file formats, call them MP4 and MOV but le's put the shared structures to the shared place"

In both cases you would end up with zero(or almost zero) code duplicity. But you should answer second question first.

petr-briza:
Btw this is what we are talking about: https://en.wikipedia.org/wiki/ISO/IEC_base_media_file_format
I based my parser on the last edition of the spec and as you can see, it's actually based on original .mov and a bunch of format are extensions of this format: MP4, 3GP, 3G2, .mj2, .dvb, .dcf, .m21, .f4v. Of course  mp4 is most common.

GreyCat:
Considering everything above, my proposal is then to:

- [ ] Rename current `quicktime_mov` to `iso_iec_base_media`
- [ ] Add documentation wording that will make it super clear that "mp4", "quicktime", "3gp", etc, are all the same format.
- [ ] Add relevant file extensions to `meta/file-extensions`
- [ ] Merge whatever is implemented in this PR and is missing in `iso_iec_base_media` there.
- [ ] Start a [big format-style](https://github.com/kaitai-io/kaitai_struct/issues/165) repo for this project, move the development there
- [ ] Enhance formats.kaitai.io to make sure it actually has way to find a file format by extension and catches all the popular asks like "mp4 format", "3gp format", etc.

petr-briza:
OK I just found a time to move this forward. I don't think there is a reason to create separate repositories for "big formats" and also I don't think this is such a big format. So I just merged my improvements to the original "mov" format, renamed it as you recommended and created a new pull request: https://github.com/kaitai-io/kaitai_struct_formats/pull/272

PR #262, opened 2020-02-09
pickle: Terminate ops sequence with stop opcode

moreati:
This allows multiple pickles to be read from a single stream, or for a
stream to contain data following a pickle. It also flags invalid pickles
that lack a stop op.


GreyCat:
Makes sense, thanks!

PR #265, opened 2020-02-11
Drafted CBOR format

KOLANICH:


serialization/cbor.ksy
```diff
@@ -0,0 +1,408 @@
...
+        types:
+          iteration:
+            params:
+              - id: idx
+                type: u1
+            instances:
+              prev:
+                value: "idx == 0 ? 0. : (_parent.reduction[idx - 1].as<iteration>.res).as<f8>"
+              res:
+                value: "prev * 256. + _parent.tagged_arr_obj.value[idx]"
```
generalmimon: I'm absolutely shocked that this doesn't end up with an compile error. This is closely related to kaitai-io/kaitai_struct#291 and kaitai-io/kaitai_struct#512. It usually leads to deriving the type of the recursive value instance (it is referencing itself, which is not currently implemented in Kaitai Struct). _But I hope it could be solved with introducing an algorithm for type inference, here is a small taste of it: [**Function type inference with recursive calls**](https://stackoverflow.com/q/15428037). I want to eventually try implementing it, when I get to it._

You're lucky that the compiler's type deriving engine is apparently satisfied with the type casting (and directly returns the type) and doesn't try to examine the casted expression (which would certainly fail). I must admit that you're very inventive :smiley:

I hate to say it, but I'm unfortunately quite sure that this will break when we add the precompile checks for typecasting (https://github.com/kaitai-io/kaitai_struct/issues/696), because we will need to know the type that one tries to cast from. So we'll still need to solve kaitai-io/kaitai_struct#512.

Do you really need to do this? I think it's probably better to solve this even with an opaque type because this is very unstable soil.

KOLANICH: >I'm absolutely shocked that this doesn't end up with an compile error.

Without the cast it has ended up with a very strange compile error. But the cast is IMHO valid here:

* `_parent.reduction[0].prev` is `f8` by definition (`0.` is CalcFloatType, should be cast to `f8`)
* if `_parent.reduction[i].prev`, then `_parent.reduction[i].res` is also `f8`
* if `_parent.reduction[i].res` is `f8`, then  `_parent.reduction[i+1].{res,prev}` is also `f8`
* so all further  `_parent.reduction[i+1].{res,prev}` by induction
□

>Do you really need to do this? I think it's probably better to solve this even with an opaque type because this is very unstable soil.

It's not as enjoyable as solving it natively in Kaitai. Though it is more usable.

generalmimon: > But the cast is IMHO valid here:

Now I see that you're right. I thought that when we add the checks if the casting is valid, the compiler will need to derive the type of the expression the casting applies to, so it won't really help with deriving the instance type and we'll get the well-known problem with deriving the type of a instance that uses itself (i.e. the type of the instance depends on the resulting type of itself, which isn't known at the moment). That would mean we had still needed the algorithm for type inference, with or without the help of the casting.

But I've just realized that I'm wrong. Checking if the casting is valid won't affect the way how the type of the value instance is derived, and thanks to the help of casting it will still work fine.

Thanks for bringing this up! It's an elegant solution of the problem, and I wouldn't have figured it out. Like I said, you're very inventive :smiley:

Actually, the compiler could suggest this solution in the exception when it detects the case of deriving the type of a recursive value instance, which is it currently unable to solve without the help of casting. Like you said, currently the cryptic error `java.util.NoSuchElementException: None.get` is thrown, and the developer has often no clue what is wrong. We should strive to make the compiler smarter and more helpful.

generalmimon (changes requested):

PR #267, opened 2020-02-19
Fix DNS record parsing and add SRV record support.

NN---:
https://en.wikipedia.org/wiki/SRV_record

### DIFF
--- a/network/mdns_packet.ksy
+++ b/network/mdns_packet.ksy
@@ -2,9 +2,10 @@ meta:
   id: dns_packet
   title: DNS (Domain Name Service) packet
   xref:
-    rfc: 1035
+    rfc: 6762
   license: CC0-1.0
   endian: be
+  encoding: utf-8
 doc: |
   (No support for Auth-Name + Add-Name for simplicity)
 seq:
@@ -90,7 +91,6 @@ types:
         if: "not is_pointer"
         doc: "Otherwise its a string the length of the length value"
         type: str
-        encoding: "ASCII"
         size: length
     instances:
       is_pointer:
@@ -155,7 +155,6 @@ types:
       - id: text
         type: str
         size: length
-        encoding: "ASCII"
   txt_body:
     seq:
       - id: data


network/dns_packet.ksy
```diff
@@ -60,19 +60,24 @@ types:
...
+            'type_type::txt': txt_body
+
   domain_name:
     seq:
       - id: name
         type: label
         repeat: until
         doc: "Repeat until the length is 0 or it is a pointer (bit-hack to get around lack of OR operator)"
-        repeat-until: "_.length == 0 or _.length == 0b1100_0000"
+        repeat-until: "_.length == 0 or _.length >= 192"
```
GreyCat: Shouldn't it be `_.length == 0 or _.is_pointer` then, as we already have this `is_pointer` set up?

network/dns_packet.ksy
```diff
@@ -89,7 +94,7 @@ types:
         size: length
     instances:
       is_pointer:
-        value: length == 0b1100_0000
+        value: length >= 192
```
GreyCat: Can you explain why this change was necessary?

network/dns_packet.ksy
GreyCat: Shouldn't it be `_.length == 0 or _.is_pointer` then, as we already have this `is_pointer` set up?

network/dns_packet.ksy
NN---: https://tools.ietf.org/html/rfc1035
Offset can be non zero:

```
    +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
    | 1  1|                OFFSET                   |
    +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
```

network/dns_packet.ksy
NN---: Fixed.

NN---:
Ping. Can this be merged ?

network/mdns_packet.ksy
generalmimon, line 1: Why did you add this duplicate of `dns_packet.ksy`?

generalmimon:
> Ping. Can this be merged ?

Not with merge conflicts.

network/mdns_packet.ksy
generalmimon: I see that you only changed the `encoding: ASCII` to `encoding: UTF-8`, but that is certainly not a sufficient reason to create a duplicate file. Given that UTF-8 is a superset of ASCII, why not just switch to `encoding: UTF-8` in the main `dns_packet.ksy`?

NN---: If this acceptable then it can be one file.

generalmimon:
@NN--- Are you no longer interested in getting this merged?

NN---:
> @NN--- Are you no longer interested in getting this merged?

I see there a re already several changes in the file.
Not sure whether my change is needed anymore.

PR #272, opened 2020-03-07
MOV renamed to ISO/IEC and more boxes/atoms are parsed

petr-briza:
Improvements as discussed in https://github.com/kaitai-io/kaitai_struct_formats/pull/259

### DIFF
--- a/media/iso_iec_base_media.ksy
+++ b/media/iso_iec_base_media.ksy
@@ -1,8 +1,19 @@
 meta:
-  id: quicktime_mov
+  id: iso_iec_base_media
+  file-extension:
+    # the list of extensions is according to https://en.wikipedia.org/wiki/ISO/IEC_base_media_file_format
+    - mp4
+    - 3gp
+    - 3g2
+    - mj2
+    - dvb
+    - dcf
+    - m21
+    - f4v
   application: QuickTime, MP4 ISO 14496-14 media
   license: CC0-1.0
   endian: be
+  encoding: ascii
 doc-ref: 'https://developer.apple.com/library/content/documentation/QuickTime/QTFF/QTFFChap1/qtff1.html#//apple_ref/doc/uid/TP40000939-CH203-BBCGDDDF'
 seq:
   - id: atoms
@@ -42,6 +53,20 @@ types:
             'atom_type::ftyp': ftyp_body
             'atom_type::tkhd': tkhd_body
             'atom_type::mvhd': mvhd_body
+            'atom_type::co64': co64_body
+            'atom_type::dref': dref_body   # Data Reference Box
+            'atom_type::elst': elst_body
+            'atom_type::hdlr': hdlr_body
+            'atom_type::mdhd': mdhd_body
+            'atom_type::smhd': smhd_body   # Sound Media Header
+            'atom_type::stco': stco_body
+            'atom_type::stsc': stsc_body
+            'atom_type::stsd': stsd_body
+            'atom_type::stts': stts_body
+            'atom_type::stss': stss_body   # Sync Sample Box
+            'atom_type::stsz': stsz_body   # Sample Size Box
+            'atom_type::stz2': stz2_body   # Compact Sample Size Box
+            'atom_type::vmhd': vmhd_body
     instances:
       len:
         value: 'len32 == 0 ? (_io.size - 8) : (len32 == 1 ? len64 - 16 : len32 - 8)'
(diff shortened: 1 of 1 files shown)


media/iso_iec_base_media.ksy
```diff
@@ -1,8 +1,19 @@
...
+    # the list of extensions is according to https://en.wikipedia.org/wiki/ISO/IEC_base_media_file_format
+    - mp4
+    - 3gp
+    - 3g2
+    - mj2
+    - dvb
+    - dcf
+    - m21
+    - f4v
   application: QuickTime, MP4 ISO 14496-14 media
```
KOLANICH: I guess it should be either an array, or removed at all.

media/iso_iec_base_media.ksy
```diff
@@ -153,6 +178,261 @@ types:
...
+        type: u4
+      - id: entries
+        type: entry
+        repeat: expr
+        repeat-expr: entry_count
+    types:
+      entry:
+        seq:
+          - id: segment_duration
+            type:
```
KOLANICH: can be encapsulated.

media/iso_iec_base_media.ksy
```diff
@@ -153,6 +178,261 @@ types:
...
+  stsz_body:
+    seq:
+      - id: version
+        type: u4
+      - id: sample_size
+        type: u4
+      - id: sample_count
+        type: u4
+      - id: entry_sizes
+        if: (sample_size==0)
```
KOLANICH: The braces are unneeded

media/iso_iec_base_media.ksy
```diff
@@ -153,6 +178,261 @@ types:
...
+      entry:
+        seq:
+          - id: size
+            type: u4
+          - id: type
+            type: str
+            size: 4
+          - id: body
+            type: entry_body
+            size: 'size - 8'
```
KOLANICH: `sizeof<u4> + 4` (when https://github.com/kaitai-io/kaitai_struct/issues/84 is implemented: `sizeof(size, type)`)

media/iso_iec_base_media.ksy
```diff
@@ -153,6 +178,261 @@ types:
...
+        type: u4
+      - id: chunk_offsets
+        type: u8
+        repeat: expr
+        repeat-expr: entry_count
+  dref_body: 
+    seq:
+      - id: version
+        type: u1
+      - id: flags
```
KOLANICH: If they are flags, should they be a type with bit-sized fields?

media/iso_iec_base_media.ksy
generalmimon: What? The `file-extension` list? I think it's definitely good to keep it here, otherwise the people which extensions this format is associated with. And in the future, we may add some search functionality to the Formats Gallery. Most people won't search for `iso_iec_base_media`, they'll search `mp4` or `3gp`.

And `it is` an array, just written in the multiline format. But I think it's not wrong at all, it's clearly organized and it plays well with the VCS too. And in KSY we generally prefer _Block (multiline) YAML style_ to single line syntax.

media/iso_iec_base_media.ksy
```diff
@@ -153,6 +178,261 @@ types:
...
+        size: 3
+      - id: entry_count
+        type: u4
+      - id: chunk_offsets
+        type: u8
+        repeat: expr
+        repeat-expr: entry_count
+  dref_body: 
+    seq:
+      - id: version
```
KOLANICH: Feels like a common header. Should it be put into a separate type?

media/iso_iec_base_media.ksy
KOLANICH: `application`

media/iso_iec_base_media.ksy
```diff
@@ -153,6 +178,261 @@ types:
...
+        type: u1
+      - id: sample_count
+        type: u4
+      - id: entry_sizes
+        repeat: expr
+        repeat-expr: '(field_size == 4) ? (sample_count/2) : sample_count'
+        type:
+          switch-on: field_size
+          cases:
+            4:  u1   # 4 bits
```
KOLANICH: Should it be put into `doc`?

petr-briza:
Guys sorry, my contribution to the community is as it is, I believe I improved the previous version, take it or leave it. I would recommend to focus rather on improtant stuff like having more boxes/atoms implemented (this is still just a fraction of the whole format) and to have some doc for the fields. Your comments seems like really minor issues and I don't have time to deal with it. Sorry.

PR #273, opened 2020-03-30
Add android boot image format

evdenis:
docs: https://source.android.com/devices/bootloader/boot-image-header
google implementation: https://android.googlesource.com/platform/system/core/+/refs/heads/android10-c2f2-release/mkbootimg/unpack_bootimg.py#45

Signed-off-by: Denis Efremov <efremov@linux.com>

### DIFF
--- a/archive/android_img.ksy
+++ b/archive/android_img.ksy
@@ -1,0 +1,120 @@
+meta:
+  id: android_img
+  title: Android Boot Image
+  license: CC0-1.0
+  file-extension: img
+  endian: le
+doc-ref: https://source.android.com/devices/bootloader/boot-image-header
+seq:
+  - id: magic
+    contents: ANDROID!
+  - id: kernel
+    type: load
+  - id: ramdisk
+    type: load
+  - id: second
+    type: load
+  - id: tags_load
+    type: u4
+  - id: page_size
+    type: u4
+  - id: header_version
+    type: u4
+  - id: os_version
+    type: os_version
+  - id: name
+    type: strz
+    size: 16
+    encoding: ASCII
+  - id: cmdline
+    type: strz
+    size: 512
+    encoding: ASCII
+  - id: sha
+    size: 32
+  - id: extra_cmdline
+    type: strz
+    size: 1024
+    encoding: ASCII
+  - id: recovery_dtbo
+    type: size_offset
+    if: header_version > 0
+  - id: boot_header_size
+    type: u4
+    if: header_version > 0
+  - id: dtb
+    type: load_long
+    if: header_version > 1
+instances:
+  base:
+    value: kernel.addr - 0x00008000
+    doc: base loading address
+  kernel_offset:
+    value: kernel.addr - base
+    doc: kernel offset from base
+  ramdisk_offset:
+    value: 'ramdisk.addr > 0 ? ramdisk.addr - base : 0'
+    doc: ramdisk offset from base
+  second_offset:
+    value: 'second.addr > 0 ? second.addr - base : 0'
+    doc: 2nd bootloader offset from base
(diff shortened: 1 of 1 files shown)


archive/android_img.ksy
```diff
@@ -0,0 +1,116 @@
...
+      major:
+        value: (version >> 25) & 0x7f
+      minor:
+        value: (version >> 18) & 0x7f
+      patch:
+        value: (version >> 11) & 0x7f
+      year:
+        value: (version >> 4) & 0x7f
+      month:
+        value: version & 0xf
```
generalmimon: This can be written much more conveniently using [bit-sized integers](https://doc.kaitai.io/user_guide.html#_bit_sized_integers). It'll be more readable and less error-prone:
```suggestion
    seq:
      - id: major
        type: b7
      - id: minor
        type: b7
      - id: patch
        type: b7
      - id: year
        type: b7
      - id: month
        type: b4
```

archive/android_img.ksy
```diff
@@ -0,0 +1,116 @@
...
+  kernel_offset:
+    value: kernel.addr - base
+  ramdisk_offset:
+    value: ramdisk.addr - base
+  second_offset:
+    value: second.addr - base
+  tags_offset:
+    value: tags_load - base
+  recovery_dtbo_offset:
+    value: recovery_dtbo.offset
```
generalmimon: I tried your spec on a public sample file [`boot.img`](https://github.com/cfig/Android_boot_image_editor/blob/master/src/test/resources/boot.img) and the parsing failed due to nullability of `recovery_dtbo` field (you are missing the version check ensuring that the `recovery_dtbo` is not `null`):
```suggestion
  recovery_dtbo_offset:
    value: recovery_dtbo.offset
    if: header_version > 0
```

archive/android_img.ksy
```diff
@@ -0,0 +1,116 @@
...
+      - id: size
+        type: u4
+      - id: addr
+        type: u8
+  load_offset:
+    seq:
+      - id: size
+        type: u4
+      - id: offset
+        type: u8
```
generalmimon: I wonder how these types differ, whether they cannot be unified? I guess that `addr` is some absolute position, whereas `offset` is relative one, but I don't know. Is this semantic difference really so significant?

archive/android_img.ksy
evdenis: I tried it initially. I think that the suggested change doesn't work because of endianness. I tried different order for the fields, but I failed to use bitfields to get the proper results

archive/android_img.ksy
evdenis: > I guess that addr is some absolute position, whereas offset is relative one, but I don't know.

Exactly. This offset is the relative position for recovery_dtbo in boot.img, addr is a physical memory address to load the entity too

Here is how google extract recovery_dtbo from boot.img
https://android.googlesource.com/platform/system/core/+/refs/heads/android10-c2f2-release/mkbootimg/unpack_bootimg.py#113

Fixed.

archive/android_img.ksy
evdenis: I've removed recovery_dtbo_offset from instances. Because this is a different kind of offset. recovery_dtbo.offset is offset for recovery_dtbo file in boot.img. Other offsets are memory loading offsets relative to base addr.

archive/android_img.ksy
dgelessus: Correct - for little-endian bit fields/packed integers, Kaitai Struct's bit ordering is counterintuitive, so it's often easier to use classic bit masks, see #597 and #155.

GreyCat:
Thanks for this contribution!

PR #275, opened 2020-04-08
Add IEEE 802.1AD tpid to ethernet frame var2

H4rryK4ne:
Hi,

I want to propose an extension of the ethernet_frame implementing the stacked VLAN tags according to 802.1AD (taken from https://en.wikipedia.org/wiki/IEEE_802.1ad). 

I have another variant of implementation, which I will link here, as soon I have created the pull request. 

Best regards,
Martin


H4rryK4ne:
I just became aware, that this PR is still open. I was running all the time with my forked version :-/

PR #276, opened 2020-04-08
Add IEEE 802.1AD tpid to ethernet frame var1

H4rryK4ne:
Hi,

I want to propose an extension of the ethernet_frame implementing the stacked VLAN tags according to 802.1AD (taken from https://en.wikipedia.org/wiki/IEEE_802.1ad). 

I have another variant of implementation, but I'm not sure, which is the better solution. Please take a look to https://github.com/kaitai-io/kaitai_struct_formats/pull/275

Best regards,
Martin

### DIFF
--- a/network/ethernet_frame.ksy
+++ b/network/ethernet_frame.ksy
@@ -21,17 +21,20 @@ seq:
   - id: src_mac
     size: 6
     doc: Source MAC address
-  - id: ether_type_1
+  - id: ether_type_or_tpid
     type: u2be
     enum: ether_type_enum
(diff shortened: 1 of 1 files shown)


network/ethernet_frame.ksy
```diff
@@ -82,4 +101,4 @@ enums:
     0x0806: arp
     0x8100: ieee_802_1q_tpid
     0x86dd: ipv6
-    #0x88a8: ieee_802_1ad_tpid
+    0x88a8: ieee_802_1ad_tpid
```
generalmimon, line 104: Missing newline.

PR #277, opened 2020-04-14
Autosar SOME/IP and SOME/IP-SD

H4rryK4ne:
Data with this protocol can be transported with UDP or TCP packages. The protocol type is identified when it is send to a specified port of an ECU. 

SOME/IP with Service-Discovery will usually send to a standard port (defined in the car network specifications). With Service-Discovery entries the ECUs exchange additional ports and addresses where SOME/IP data can be subscribed or found.


network/some_ip/some_ip_sd.ksy
```diff
@@ -0,0 +1,48 @@
...
+- id: len_entries
+  type: u4
+- id: entries
+  type: some_ip_sd_entries
+  size: len_entries
+- id: len_options
+  type: u4
+- id: options
+  type: some_ip_sd_options
+  size: len_options
```
GreyCat, line 35: Please indent this block 2 spaces to the right to match [recommended formatting on the style guide](https://doc.kaitai.io/ksy_style_guide.html#general)?

network/some_ip/some_ip_sd.ksy
```diff
@@ -0,0 +1,48 @@
...
+    seq:
+      - id: reboot
+        type: b1
+      - id: unicast
+        type: b1
+      - id: inital_data
+        type: b1
+      - id: reserved
+        type: b5
+    doc-ref: AUTOSAR_PRS_SOMEIPServiceDiscoveryProtocol.pdf - Figure 4.3
```
GreyCat, line 48: Nit: EOF newline missing.

network/some_ip/some_ip_sd.ksy
```diff
@@ -0,0 +1,48 @@
...
+  size: len_options
+  
+types:       
+  sd_flags:
+    seq:
+      - id: reboot
+        type: b1
+      - id: unicast
+        type: b1
+      - id: inital_data
```
GreyCat, line 44: Typo?
```suggestion
      - id: initial_data
```

network/some_ip/some_ip_sd_entries.ksy
```diff
@@ -0,0 +1,80 @@
...
+            type: b4
+          - id: event_group_id
+            type: u2
+            
+    enums:
+      entry_types:
+        0x00 : find
+        0x01 : offer
+        0x06 : subscribe
+        0x07 : subscribe_ack
```
GreyCat, line 80: Nit: missing EOF newline

H4rryK4ne:
Is there something to do for me, to get this merged?

network/some_ip/some_ip.ksy
```diff
@@ -0,0 +1,161 @@
...
+          system (i.e. the vehicle). TheMessage ID is similar to a CAN ID and
+          should be handled via a comparable process.
+          [PRS_SOMEIP_00038] Message IDs of method calls shall be structured in
+          the ID with 2^16 services with 2^15 methods.        
+        doc-ref: AUTOSAR_PRS_SOMEIPProtocol.pdf 4.1.1.1  Message ID
+        
+        instances:
+          value:
+            pos: 0
+            io: _io
```
generalmimon, line 104: This line is redundant.

network/some_ip/some_ip.ksy
```diff
@@ -0,0 +1,161 @@
...
+            type: u2
+        doc: |
+          The Request ID allows a provider and subscriber to differentiate 
+          multiple parallel usesof the same method, event, getter or setter.
+        doc-ref: AUTOSAR_PRS_SOMEIPProtocol.pdf - section 4.1.1.3  Request ID
+        
+        instances:
+          value:
+            pos: 0
+            io: _io
```
generalmimon, line 122:
```suggestion
```

network/some_ip/some_ip.ksy
```diff
@@ -0,0 +1,161 @@
...
+        0x06 : time_out
+        0x07 : wrong_protocol_version
+        0x08 : wrong_interface_version
+        0x09 : malformed_message
+        0x0a : wrong_message_type
+
+  data:
+    seq:
+      - id: body
+        size-eos: true
```
generalmimon, line 161: This type is useless, see my comment above.

network/some_ip/some_ip.ksy
```diff
@@ -0,0 +1,161 @@
...
+seq:
+  - id: header
+    type: header
+  - id: payload
+    size: header.length - 8
+    type:
+      switch-on: header.message_id.value
+      cases:
+        0xffff8100: some_ip_sd
+        _: data
```
generalmimon, line 26: Because the `data` contains just a raw byte array stretching to the end of the stream, it's actually pointless to define it. If you leave out this line from the `cases`, this will be the implicit behavior: when `header.message_id.value != 0xffff8100`, then the `payload` effectively won't have any value of the `type` key. But omitted `type` key means byte array in KS, and this applies here as well. This case is mentioned in the [User Guide](https://doc.kaitai.io/user_guide.html#tlv).

So you can get rid of the whole `data` type completely, and it'll work the same, except that it will be accessed directly as `payload`, not `payload.body`.

```suggestion
```

network/some_ip/some_ip.ksy
```diff
@@ -0,0 +1,159 @@
+meta:
+  id: some_ip
+  title: AUTOSAR SOME/IP
+  license: CC0-1.0
+  ks-version: 0.9
```
generalmimon, line 5: I wonder why you declare `0.9` as the lowest possible compiler version? I don't think that I've seen some 0.9 feature (well, I've not tested it), so I hope it would properly compile even with the KSC 0.8.

But rather don't change it without testing that it really works under version 0.8, I'm just curious if you had encountered some issue with the 0.8 version or you needed some 0.9 feature.

H4rryK4ne: I just used KSC v0.9 only, since the python code compiled with v0.8 didn't work. There was an issue with circular references which had been fixed with v0.9.

GreyCat:
Merging this in. Thanks for @H4rryK4ne for authoring this and for @generalmimon for extensive review, great job!

GreyCat:
Also, please note that we currently fail miserably at automatic compilation of 0.9-related features — e.g. see [gettext_mo](http://formats.kaitai.io/gettext_mo/index.html).

I'll try to fix it soon.

H4rryK4ne:
> The only thing that annoys me a bit is how should be the format specifications of one particular protocol split into multiple files displayed in the [format gallery](https://formats.kaitai.io/). Usually, there is just one entry-point KSY spec, and the rest are just import-only files, not designed to be used directly. But the format gallery generator currently show just the individiual files, without any link to each other, and when the user will see a bunch of `some_ip_container`, `some_ip_sd`, `some_ip_sd_entries`, `some_ip_sd_options`, etc. in the format list, he might be confused which one to use.

@generalmimon:
I agree with you. It is somehow confusing. Especially the some_ip_container file could be merged into the some_ip directly. I will try a bit, to get this pretty.

PR #278, opened 2020-04-25
Use Pygments syntax highlighter

generalmimon:
It seems to me that the quality of the syntax highlighting provided by highlight.js that we use now is quite limited, one can see many highlighting issues across the whole format gallery. We use quite a limited subset of YAML actively for KSY purposes, and yet this library has various problems with that. For example https://formats.kaitai.io/utf8_string/index.html, this is not exactly how it should look like.

So I propose switching to Pygments, as we use for the documentation (https://github.com/kaitai-io/kaitai_struct_doc/pull/34). I think for https://doc.kaitai.io/ it works quite well.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,6 +14,9 @@ jobs:
       - uses: actions/setup-ruby@v1
         with:
           ruby-version: '2.x'
+      - uses: actions/setup-python@v1
+        with:
+          python-version: '3.x'
       - name: Install deps
         run: |
           sudo apt-get update
@@ -41,8 +44,6 @@ jobs:
           echo 'gem: --no-document' | sudo tee /etc/gemrc
           cd _build
           bundle install
-      - name: Install setuptools
-        run: python3 -m pip install --user setuptools
       - name: Install Pygments
         run: python3 -m pip install --user pygments
       - name: Install Graphviz lexer for Pygments
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,8 +46,7 @@ jobs:
       - name: Install Graphviz lexer for Pygments
         run: |
           git clone -b master --depth 1 https://github.com/nikeee/pygments-lexer-graphviz.git
-          cd pygments-lexer-graphviz/
-          sudo python setup.py develop
+          sudo python3 -m pip install --editable pygments-lexer-graphviz/
       - name: Build targets
         working-directory: _build
         run: ./build-target .. target
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,6 +41,11 @@ jobs:
           echo 'gem: --no-document' | sudo tee /etc/gemrc
           cd _build
           bundle install
+          git clone -b 2.5.2 --depth 1 https://github.com/pygments/pygments.git
+          git clone -b master --depth 1 https://github.com/nikeee/pygments-lexer-graphviz.git
+          cd pygments-lexer-graphviz/
+          python setup.py develop
+          cd ../
       - name: Build targets
         working-directory: _build
         run: ./build-target .. target
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,11 +42,9 @@ jobs:
           cd _build
           bundle install
       - name: Install Pygments
-        run: git clone -b 2.5.2 --depth 1 https://github.com/pygments/pygments.git
+        run: python3 -m pip install --user pygments
       - name: Install Graphviz lexer for Pygments
-        run: |
-          git clone -b master --depth 1 https://github.com/nikeee/pygments-lexer-graphviz.git
-          sudo python3 -m pip install --editable pygments-lexer-graphviz/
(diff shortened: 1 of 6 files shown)


_build/header.html.erb
```diff
@@ -47,8 +46,10 @@ section#format-index .row {
 }
 
   </style>
+<!--
   <script src="<%= @base_url %>js/highlight.pack.js"></script>
   <script>hljs.initHighlightingOnLoad();</script>
+-->
```
GreyCat, line 52: Given that we're switching from one to another, it makes sense to remove highlight.js altogether.
```suggestion
```

.github/workflows/main.yml
```diff
@@ -41,6 +41,11 @@ jobs:
           echo 'gem: --no-document' | sudo tee /etc/gemrc
           cd _build
           bundle install
+          git clone -b 2.5.2 --depth 1 https://github.com/pygments/pygments.git
+          git clone -b master --depth 1 https://github.com/nikeee/pygments-lexer-graphviz.git
+          cd pygments-lexer-graphviz/
+          python setup.py develop
+          cd ../
```
GreyCat, line 48: Something here seems to break GH Actions build (may be the very first thing, calling `git`?). I suggest that we'll extract it into separate step to avoid confusion with installation of Ruby dependencies, or at least rename the step.

_build/header.html.erb
generalmimon: Of course, thanks! It's just a forgotten comment.

.github/workflows/main.yml
generalmimon: Yeah, thanks for the catch, it was missing a `sudo`. In https://github.com/kaitai-io/kaitai_struct_formats/pull/278/commits/00963931d294e7331fca9cf2241ceade85ffa11f, it finally works.

.github/workflows/main.yml
```diff
@@ -41,6 +41,13 @@ jobs:
           echo 'gem: --no-document' | sudo tee /etc/gemrc
           cd _build
           bundle install
+      - name: Install Pygments
+        run: git clone -b 2.5.2 --depth 1 https://github.com/pygments/pygments.git
```
dgelessus, line 45: Any particular reason against using Ubuntu's `python3-pygments` package?

.github/workflows/main.yml
```diff
@@ -41,6 +41,13 @@ jobs:
...
+      - name: Install Graphviz lexer for Pygments
+        run: |
+          git clone -b master --depth 1 https://github.com/nikeee/pygments-lexer-graphviz.git
+          cd pygments-lexer-graphviz/
+          sudo python setup.py develop
```
dgelessus, line 50: This is a fairly outdated way of installing a Python package from a local clone. I would recommend using

```suggestion
          git clone -b master --depth 1 https://github.com/nikeee/pygments-lexer-graphviz.git
          sudo pip install --editable pygments-lexer-graphviz/
```

If possible I would also leave out the `--editable` flag to `pip` (it shouldn't make a difference in this case).

dgelessus: You can also get rid of the `sudo` if you pass `--user` to the `pip` call (not that it makes much of a difference in a container).

dgelessus: A one-line version without `sudo` would be
```suggestion
          python3 -m pip install --user git+https://github.com/nikeee/pygments-lexer-graphviz.git@master
```

.github/workflows/main.yml
dgelessus: Ah ok, that explains it - yeah, Ubuntu's packages are often quite old... In that case it probably makes more sense to do a regular `python3 -m pip install --user pygments` to install the latest release version of Pygments, rather than cloning a specific version and installing it manually.

dgelessus:
re. that install failure: `python3 -m pip install --user setuptools`. I really don't know how GitHub Actions' Python has `pip` but not `setuptools`, that's very unusual.

dgelessus:
It might be a good idea to explicitly use https://github.com/actions/setup-python to install Python - that will probably give a more reasonable environment with `setuptools` already installed:

```yaml
- uses: actions/setup-python@v1
  with:
    python-version: '3.x'
```

dgelessus:
Ah, it looks like the pygments.rb gem is calling Python 2 and not Python 3. Maybe `actions/setup-python` will help with that (perhaps it will do some `PATH` manipulation or change other environment variables to make Python 3 preferred). If not, you can probably just replace all `python3` calls with `python2` (although that's not a great solution, as Python 2 isn't really supported anymore).

generalmimon:
It looks like it finally works!
@dgelessus Thank you so much for your great real-time support! Things are far easier when you're online 😃

dgelessus:
@generalmimon Really I need to be thanking you for all the work you're doing on KS - happy to help with that where I can 😃

dgelessus:
It looks like pygments.rb doesn't support Python 3 (see tmm1/pygments.rb#152), which is unfortunate, but there's not really anything we can do about it. So it looks like we are indeed stuck with Python 2 here.

GreyCat:
Big thanks for @generalmimon and @dgelessus for getting this into production-ready state!

PR #288, opened 2020-05-19
fix duplicate enums in elf kaitai-io#279

taviso:
- DT_ENCODING is not encoded into files, it's a special value used by tools.

REF: https://docs.oracle.com/cd/E19683-01/816-7529/chapter6-42444/index.html

- DT_LOOS and DT_HIOS
- DT_VALRNGLO and DT_VALRNGHI
- DT_ADDRRNGLO and DT_ADDRRNGHI
- DT_LOPROC and DT_HIPROC

These are not encoded into files, they're just hi/lo region values.

ref: https://docs.oracle.com/cd/E23824_01/html/819-0690/chapter6-42444.html

- DT_SUNW_RTLDINF is for internal use and not encoded into files.

ref: https://docs.oracle.com/cd/E23824_01/html/819-0690/chapter6-42444.html

### DIFF
--- a/executable/elf.ksy
+++ b/executable/elf.ksy
@@ -640,18 +640,14 @@ enums:
     28: fini_arraysz     # Size in bytes of DT_FINI_ARRAY
     29: runpath          # Library search path
     30: flags            # Flags for the object being loaded
-    32: encoding         # Start of encoded range
     32: preinit_array    # Array with addresses of preinit fct
     33: preinit_arraysz  # Size in bytes of DT_PREINIT_ARRAY
     34: maxpostags       # Number used
-    0x6000000d: loos
     0x6000000d: sunw_auxiliary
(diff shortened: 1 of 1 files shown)


executable/elf.ksy
```diff
@@ -676,7 +670,6 @@ enums:
     0x6ffffdfd: posflag_1       # Flags for DT_* entries, effecting the following DT_* entry.
     0x6ffffdfe: syminsz         # Size of syminfo table (in bytes)
     0x6ffffdff: syminent        # Entry size of syminfo
-    0x6ffffdff: valrnghi
     0x6ffffe00: addrrnglo
```
generalmimon, line 673: According to your description, this one should be probably deleted too?

```suggestion
```

generalmimon:
@taviso Thanks for the contribution!

PR #291, opened 2020-05-26
Support TXT, SRV and PTR records

NN---:


network/dns_packet.ksy
```diff
@@ -144,6 +148,29 @@ types:
...
+      - id: target
+        type: domain_name
+  txt:
+    seq:
+      - id: length
+        type: u1
+      - id: text
+        type: str
+        size: length
+        encoding: UTF-8
```
GreyCat, line 168: Are we sure that this comes in UTF-8? My guess would be that DNS protocol longs before predates Unicode, so likely it's treated as raw byte array?

NN---: Good point.
Let me double check.

NN---: https://tools.ietf.org/html/rfc1464

RFC says ASCII.
I’ll check whether there is a update to this RFC.

NN---: I split dns into dns and mdns packet.
It is better to have on dns_packet with UTF-8 if it is possible.

GreyCat: Let's stick with UTF-8 for now then, and we can always make it more strict or more relaxed later.

GreyCat:
Merging this, thanks for your contribution!

PR #296, opened 2020-06-11
Implemented  IEEE 802.11 prism pcap header

KOLANICH:


network/ieee802_11_prism.ksy
```diff
@@ -0,0 +1,66 @@
...
+            doc: usually zero
+          - id: is_supplied
+            type: u2
+          - id: size
+            type: u2
+          - id: value
+            type:
+              switch-on: type
+              cases:
+                _: u4
```
generalmimon, line 54: Please don't use type switching with the else case only, this situation has been broken in some languages until recently (fixed in commits https://github.com/kaitai-io/kaitai_struct_compiler/search?q=else+only+switch&type=Commits). Because your spec doesn't have any other 0.9 only features, I suggest to eliminate this small detail to work nicely even with 0.8.

```suggestion
            type: u4
```

generalmimon:
@KOLANICH Do you have some test files so that I can check if it works? I tried hard to find something on the internet, but found nothing.

KOLANICH:
If you are unlucky to have a broadcom wlan card + linux, you can capture them yourself.  I am not wiling to share my captures, but I have found https://web.archive.org/web/20101116031635if_/http://www.raulsiles.com:80/downloads/80211_headers.pcap in the Internet, which it is claimed to have 2 packets, one of them is claimed to be prism (I haven't checked for that).

Also have found a website with wlan datasets https://www.crawdad.org (and bookmarked it)  and http://pcapr.net/view/fernpromero/2010/3/6/4/prismheader.cap.html , though haven't checked if they really contain the needed files

network/ieee802_11_prism.ksy
```diff
@@ -0,0 +1,63 @@
+meta:
+  id: ieee802_11_prism
+  title: prism II headers for 802.11 capture
+  license: Unlicense
+  endian: le
+  encoding: ascii
+
+doc-ref:
+  - https://www.tcpdump.org/linktypes/LINKTYPE_IEEE802_11_PRISM.html
```
generalmimon, line 9: [Array in `doc-ref`](https://github.com/kaitai-io/kaitai_struct/issues/269) is also a 0.9 feature.

network/ieee802_11_prism.ksy
```diff
@@ -0,0 +1,63 @@
...
+      field:
+        seq:
+          - id: flags
+            type: u1
+          - id: type
+            type: u1
+            enum: type
+          - id: unknown
+            type: u2
+            doc: usually zero
```
generalmimon, line 45: According to Wireshark with the loaded [`80211_headers.pcap`](https://web.archive.org/web/20101116031635if_/http://www.raulsiles.com:80/downloads/80211_headers.pcap) file that you provide and also to the spec at https://www.tcpdump.org/linktypes/LINKTYPE_IEEE802_11_PRISM.html you specified in the `doc-ref`, this is not the correct interpretation.

These 4 bytes that I highlighted form together in both Wireshark and the mentioned specification a single `u4` integer, which is mapped to an enum (with integer values like `0x00020044` = MAC time etc.). In the [spec](https://www.tcpdump.org/linktypes/LINKTYPE_IEEE802_11_PRISM.html), there are two different enums presented, one for packets with the msgcode value 0x00000041 and the other for packets with the msgcode value 0x00000044. 

Of course, you don't have to stick to this approach, it's OK to split the single enum field into multiple attributes, but it has to parse the right values. The first packet in the file `80211_headers.pcap` gets parsed by Wireshark like this:

```yaml
Prism capture header
    Message Code: 0x00000044
    Message Length: 144
    Device Name: ath0
    DID Host Time -23742
        DID: Host Time (0x00010044)
        Status: Supplied (0)
        Length: 4
        Host Time: 4294943554
    DID Mac Time 186645954
        DID: Mac Time (0x00020044)
        Status: Supplied (0)
        Length: 4
        MAC timestamp (lower 32 bits): 186645954
    # ...
```

whereas your KSY reads this (data exported to JSON by the [Web IDE](https://ide.kaitai.io/devel/)):

```js
  "body": {
    "messageCode": 0x44,
    "size": 144,
    "deviceName": "ath0",
    "fields": {
      "fields": [
        {
          "flags": 0x44,
          "type": 0,
          "unknown": 1,
          "isSupplied": 0,
          "size": 4,
          "value": 4294943554
        },
        {
          "flags": 0x44,
          "type": 0,
          "unknown": 2,
          "isSupplied": 0,
          "size": 4,
          "value": 186645954
        },
        // ...
```

Note that the DID field type isn't recognized at all by your spec. 

I can see the motive of splitting this field the way you did it, but it unfortunately parses wrong values when the `message_code` is `0x44`. The reason is that though the enum values in case of `message_code == 0x44` appear to be similar as in case of `message_code == 0x41`, the "changing number" is at a different position. From the [specification](https://www.tcpdump.org/linktypes/LINKTYPE_IEEE802_11_PRISM.html):

```
For packets with the msgcode value 0x00000041, the DID field values are:
0x00001041 - host time;
0x00002041 - MAC time;
0x00003041 - channel;
...

For packets with the msgcode value 0x00000044, the DID field values are:
0x00010044 - host time;
0x00020044 - MAC time;
0x00030044 - channel;
...
```

So if you want to use the multi-attribute approach, one layout isn't enough and you'll probably have to be creative. What comes to my mind:

```yaml
types:
  field:
    seq:
      - id: type
        type: field_type
        size: 4
  field_type:
    seq:
      - id: code
        type: u1
        doc: seems to be equal to `_root.message_code`
      - id: skip
        type: b12
        if: code == 0x44
        doc: should be 0
      - id: type
        type: b4
        enum: type
enums:
  type:
    0x1: host_time
    0x2: mac_time
    0x3: channel
    # ...
```

network/ieee802_11_prism.ksy
```diff
@@ -0,0 +1,63 @@
...
+          - id: is_supplied
+            type: u2
```
generalmimon, line 47: The name seems misleading to me, I expect from all attributes starting with `is_` to be of type boolean, but this is some magic number. The spec says that

> The Status field is 0 if the value of that parameter is supplied and 1 if it is not supplied.

so I would either create an enum with these two states, or a value instance returning boolean. The name of the value instance with `boolean` type can start with `is_`, but in my opinion an attribute of any other type shouldn't. I'd rather go for the enum.

This enhancement is particularly important because the value logic of this field are counter-intuitive, 0 for `true` (**is** supplied) and 1 for `false` (**not** supplied), and the user of this KSY should be able to find out this just by looking in the KSY spec, without having to search in external sources.

network/ieee802_11_prism.ksy
```diff
@@ -0,0 +1,63 @@
...
+          - id: size
+            type: u2
+          - id: value
+            type: u4
```
generalmimon, line 51: This violates the spec as well. It says

> The Length field **indicates the length** of the following **Data** field; in theory, **it could have any value**, but it is almost always, if not always, 4. The Data field contains the value of the parameter.

Despite that it is _almost always 4_, it's better to not rely on it, and behave according to what the spec says with certainty: indicates the length of the following Data field.

If the `field` type was enclosed in a substream with a fixed size, I would still say "well, I can't read the full data of this single `field`, because it's more than 4, but at least I can access the following `field`s", but this `field` type uses the parent substream, so ignoring the `size` can quickly lead to malformed data and almost certainly with an EOF exception at the end of the file.

generalmimon: I know very well that you don't like this kind of stuff, but still please consider renaming `size` to `len_value` (as per the style guide).

network/ieee802_11_prism.ksy
KOLANICH: >These 4 bytes that I highlighted form together in both Wireshark and the mentioned specification a single u4 integer, which is mapped to an enum (with integer values like 0x00020044 = MAC time etc.).

Yes, I have initially done as in wireshark, but when I have read the docs in the net, it is clearly stated in them that the format is reverse engineered, so there is no correct public info on that. So I have splitted the field without any testing of the resulting spec.

I'm sorry for all the errors, I gonna try to fix them somewhen.

generalmimon: It's OK, take your time, it's just that this PR is not yet finished. I'll mark it as a draft if you don't object, you can hopefully mark it as "ready for review" any time.

network/ieee802_11_prism.ksy
KOLANICH: Fixed. Not as a enum, though.

network/ieee802_11_prism.ksy
KOLANICH: Usage of length is fixed.
Change of the name isn't.

>I know very well that you don't like this kind of stuff, but still please consider renaming size to len_value (as per the style guide).

I'm sorry, but IMHO it's the style guide should be fixed instead of just following the badly-designed style guide.

network/ieee802_11_prism.ksy
KOLANICH: Thanks. I have noticed that the second nibble (`1` or `4`) may mean the offset from the first nibble (`4`), so I have tried to interpret it as such.

KOLANICH:
@generalmimon

network/ieee802_11_prism.ksy
generalmimon: @KOLANICH:
> I'm sorry, but IMHO it's the style guide that should be fixed instead of just following the badly-designed style guide.

Why exactly do you think it is badly designed? IMHO introducing a naming convention like this brings consistency and order and improves orientation in the spec. When you see an attribute starting with `len_`, you know immediately that it's a size in bytes of some other attribute, and that attribute is named exactly as what follows after the `len_` prefix.

If you have a problem with the particular `len_` prefix and you'd rather like `size_` for example, we can talk about such change, but I don't think it matters. It's just nice to have some established convention, rather than for everyone to name it according to the current mood.

KOLANICH: Starting an attr with `len_` or any other generic prefix means I cannot meaningfully use editor autocompletion.

network/ieee802_11_prism.ksy
```diff
@@ -0,0 +1,113 @@
...
+              mandatory_part_offset_from_first_byte:
+                value: mandatory_part_offset - 1
+              mandatory_part:
+                pos: mandatory_part_offset_from_first_byte / 2
+                type: mandatory_part(mandatory_part_skip_nibble)
+                io: placeholder._io
+              mandatory_part_skip_nibble:
+                value: mandatory_part_offset_from_first_byte % 2 == 1
+              type:
+                value: mandatory_part.type
```
generalmimon, line 85: Oh my God, this is so much over-engineered! Let me remind you that according to the [specification](https://www.tcpdump.org/linktypes/LINKTYPE_IEEE802_11_PRISM.html) from `doc-ref` (yeah, I know it's bad to rely on a single source, but I don't know any others), your `header_size` attribute is always equal to **4** and the only valid values for `mandatory_part_offset` are 1 and 4. So your spec permits lots of combinations that are likely to be incompatible with any other parser implementations. This flexibility may seem OK and basically in the spirit of the [Postel's law](https://en.wikipedia.org/wiki/Robustness_principle) (_"be liberal in what you accept from others"_), but IMHO here it does more harm than good. Consider that your format doesn't have any safeguards against invalid packets. This doesn't hurt by itself, but it sucks when you can't even check

generalmimon, line 85: Oh my God, this is so much over-engineered! Let me remind you that according to the [specification](https://www.tcpdump.org/linktypes/LINKTYPE_IEEE802_11_PRISM.html) from `doc-ref` (yeah, I know it's bad to rely on a single source, but I don't know any others), the first byte of the DID field type (here split into `header_size` and `mandatory_part_offset` fields) is either **0x41** or **0x44**, nothing more. So your spec permits lots of combinations that will be rejected by any other parser implementation, and that won't ever generated by any reputable serialization implementation. OK, this is not so serious yet. What is worse, if we imagine that this spec may be used for serialization in the future, it would be very easy to start generating invalid packets that don't conform to the specification and aren't recognized by anything. Would it be the programmer's fault that he set some clearly incorrect values? Nope, he just looked into the KSY and saw that `header_size` and `mandatory_part_offset` are just plain `b4` bit-integers, so apparently all values are permitted, right? Wrong.

Of course, adding `valid/eq: 4` to `header_size` and `valid/any-of: [1, 4]` to `mandatory_part_offset` would address this issues, but then the `mandatory_part_*` stuff seems unnecessarily complex and cumbersome, when you have just 2 options to handle.

Please don't devise this **artificial structure** just for fun to make things more complex and unclear. Instead, stick to the specification and make simple things in a simple way. When the specification talks about two possible options, don't try to map random bits to their _potential_ meaning (as it currently fits) in order to invent a general formula or rule that magically handles the full theoretical range of inputs, but rather create an enum with these two permitted values and choose the behavior accordingly. The `.ksy` specification should match the format it describes as closely as possible, not try to outsmart the format authors themselves with some _wild guesses_. It helps not to think of the `.ksy` spec as a tool for generating parsing code, but rather as a format documentation by itself.

network/ieee802_11_prism.ksy
```diff
@@ -0,0 +1,113 @@
...
+              cases:
+                'type::host_time': u4
+                'type::mac_time': u4
+                'type::channel': u4
+                'type::rssi': u4
+                'type::signal_quality': u4
+                'type::noise': u4
+                'type::data_rate': u4
+                'type::is_tx': u4
+                'type::frame_length': u4
```
generalmimon, line 60: This page http://www.martin.cc/linux/prism actually has some interesting information about the representation of these values, could you check it out and possibly update your spec?

network/ieee802_11_prism.ksy
KOLANICH: > Oh my

Thou shalt not take the name in vain.

>according to the specification
>stick to the specification 

**There is no spec**, all the "specs" are reverse-engineered, so they are just wild guesses matching the files seen by the authors. We can say that all the reverse-engineered "specs" are overfitted. Mine is no better, but it just incorporates some domain knowledge about file formats in general. I mean, if we assumme that in all the cases the same file format is used, and we see that the offset varies, it is usually 3 cases: 
* the offset is encoded in the file itself explicitly as a number, relative to a fixed feature
* the offset is fixed, but relative to a feature which offset varies
* the offset is encoded, and the position of the feature is encoded

So, what else we know about this format?
* After this header another stuff starts untill the end.
* in the header itself there is pretty little stuff to be used as a base forz a fixed offset

So, I guesed it is likely that the offset is encoded, and, since sometimes the stuff starts from the second byte, the offset is likely to be encoded in the first one.

>So your spec permits lots of combinations that will be rejected by any other parser implementation

Yes. And it is OK. We just don't know, am I right, or are the "specs" writers are right, or none of us is right.

>What is worse, if we imagine that this spec may be used for serialization in the future, it would be very easy to start generating invalid packets

We don't know if they are valid or not. And what is good, if any official tools parsing these files can be ever found, we can be able to check, if I guessed right or not.

network/ieee802_11_prism.ksy
KOLANICH: Thanks, in fact I have looked into it, but it seemed like it just a duplicate of the info from the source from doc-ref.

I will look into it again more carefully.

KOLANICH: Could you point out what in your opinion should be added?

generalmimon: > Could you point out what in your opinion should be added?

1. 
   ```
    * rssi:     Appears to be a signed dbm value
   ```
   
   ```
    * noise:     (...) - Should be signed dbm value
   ```
   
   These values should be **signed (`s4`)**, not ~~unsigned (`u4`)~~ as in your spec.

2.
   ```
    * rate:     Appears to be in units/multiples of 500Khz
   ```
   
   Maybe it would make sense to declare a value instance to get the value in standard units like `kHz` or `Hz` .
   
   Additionally, using [verbose enums](https://doc.kaitai.io/user_guide.html#verbose-enums) you document the    `type::data_rate` enum value like this:
   
   ```yaml
   enums:
     type:
       # ...
       0x8:
         id: data_rate
         doc: in units/multiples of 500Khz
   ```

   Or maybe reflect the unit in the name of the enum value, like `data_rate_div_500khz` (meaning that the value stored here is calculated like `rate in Hz (basic unit) / 500 kHz`) or something like that.

3.
   ```
    * istx:     Not Supplied  - 0 = rx packet, 1 = tx packet
   ```
   
   From a raw `'type::is_tx': u4` type, this value meaning is not known to the end user. Best practice in terms of clarity and usability would be to create a value instance returning an enum value of something like `enums/packet_direction: {0: rx, 1: tx}`, but merely documenting this value logic in the `type::is_tx` enum value is also OK.

The info about how to interpret the `u4` values in context of individual field types is essential and thus should IMHO definitely appear somewhere in the KSY spec. At the very least, as a YAML comment `# ...` next to the `type/cases` items, a better option would be `doc` key in [verbose enums](https://doc.kaitai.io/user_guide.html#verbose-enums) (because these docblocks could then appear in auto-generated API documentations and autocompletion boxes when typing the application code in a smart IDE), and the best option is the code that _documents itself_, i.e. it is clear from the attribute names what's the meaning of the value in the attribute, in which units it is, etc. However, this is not always possible/convenient, in these cases it should be accompanied with the docblock.

---

My opinion is that stopping parsing on "magic" `u4` integers without any effort to try to document their meaning is insufficient, and it forces all potential users of the KSY spec to do it on their own. Of course, if it is the best we can do with regard to publicly available resources, then it is understandable. But if we can make the KSY spec better and more usable with minimal effort, IMHO we should always do it.

KOLANICH: Gonna do.

BTW: https://github.com/kaitai-io/kaitai_struct/issues/522 .

PR #297, opened 2020-06-17
Support IPv6 (Type AAAA).

pegasusplus:
Add IPv6 address type support, hope it is useful

### DIFF
--- a/network/dns_packet.ksy
+++ b/network/dns_packet.ksy
@@ -79,6 +79,7 @@ types:
           cases:
             "type_type::ptr": domain_name
             "type_type::a": address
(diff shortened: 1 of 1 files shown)


network/dns_packet.ksy
```diff
@@ -121,6 +122,12 @@ types:
         type: u1
         repeat: expr
         repeat-expr: 4
+  address_v6:
+    seq:
+      - id: ip_v6
+        type: u1
+        repeat: expr
+        repeat-expr: 16
```
generalmimon, line 130: It's better and clearer to use a byte array instead:

```suggestion
      - id: ip_v6
        size: 16
```

PR #299, opened 2020-07-03
Added Sound Resources 'snd '

tisoft:
* used by Sound Manager from Mac OS (Classic)
* mostly found in resource forks


macos/snd_.ksy
```diff
@@ -0,0 +1,201 @@
+meta:
+  id: snd_
+  endian: be
+  xref:
+    wikidata: Q7564684
+    reference: "https://developer.apple.com/library/archive/documentation/mac/pdf/Sound/Sound_Manager.pdf"
```
generalmimon, line 6: This should be specified in the [`doc-ref` key](https://doc.kaitai.io/user_guide.html#doc-ref) instead.

```suggestion
doc-ref: "https://developer.apple.com/library/archive/documentation/mac/pdf/Sound/Sound_Manager.pdf"
```

dgelessus:
By the way, I have a separate repo with some specs related to Mac OS resource forks: https://github.com/dgelessus/resource_fork.ksy

At the moment the repo only has a spec for the resource fork format itself (plus some specs related to System 7's resource compression). I was also planning to put KSYs for common resource types into that repo, but I haven't gotten around to that yet. (That's actually the main reason why I made it a separate repo instead of adding the specs to kaitai_struct_formats - I didn't want to spam the KSF repo with lots of small resource type specs.)

Would it be okay if I include this spec in my resource_fork.ksy repo (when I start working on adding resource type specs there)? I'll of course credit you in the commit and include a link back to the original PR here. Or if you prefer, you can also make a PR on my repo, but perhaps after this one is merged, so we don't review the same PR in two places.

macos/snd_.ksy
```diff
@@ -0,0 +1,201 @@
+meta:
+  id: snd_
```
dgelessus, line 2: At the moment, KSY spec identifiers aren't namespaced, so all spec identifiers in kaitai_struct_formats should be descriptive on their own, without the directory structure. A name like `mac_os_snd` or `mac_os_sound_resource` would be better.

Also, I don't know if/how well the compiler supports trailing underscores. It's not obvious how something like `snd_` should be convered to CamelCase (`Snd` or `Snd_`?). Just to be safe, I would choose an identifier that doesn't end with an underscore.

macos/snd_.ksy
```diff
@@ -0,0 +1,201 @@
...
+  endian: be
```
dgelessus, line 3: All specs in kaitai_struct_formats [need to include a `license` in the metadata](https://github.com/kaitai-io/kaitai_struct_formats/blob/master/CONTRIBUTING.md#contributing).

macos/snd_.ksy
```diff
@@ -0,0 +1,201 @@
...
+    81: buffer_cmd
+    82: rate_cmd
+  sound_header_type:
+    0x00: standard
+    0xFF: extended
+    0xFE: compressed
+seq:
+  - id: format
+    type: u2
+  - id: number_of_data_formats
```
dgelessus, line 39: According to the [KSY naming conventions](https://doc.kaitai.io/ksy_style_guide.html#attr-id), this should be called `num_data_formats`.

macos/snd_.ksy
```diff
@@ -0,0 +1,201 @@
...
+    if: format==1
+  - id: data_format
+    type: data_format
+    repeat: expr
+    repeat-expr: number_of_data_formats
+    if: format==1
+  - id: reference_count
+    type: u2
+    if: format==2
+  - id: number_of_sound_commands
```
dgelessus, line 50: As above, should be called `num_sound_commands` by convention.

macos/snd_.ksy
```diff
@@ -0,0 +1,201 @@
...
+    type: u2
+  - id: sound_command
```
dgelessus, line 52: This should be called `sound_commands`, because it's an array and not a single value.

macos/snd_.ksy
```diff
@@ -0,0 +1,201 @@
...
+types:
+  data_format:
+    seq:
+      - id: id
+        type: u2
+        enum: data_type
+      - id: options
+        type: u4
+  sound_command:
+    instances:
```
dgelessus, line 65: According to the style guide, the `instances` section should come after `seq`.

macos/snd_.ksy
```diff
@@ -0,0 +1,201 @@
...
+        type: u2
+      - id: param1
+        type: u2
+      - id: param2
+        type: u4
+  standard_sound_header:
+    seq:
+      - id: sample_ptr
+        type: u4
+      - id: lenght
```
dgelessus, line 97: Typo, should be `length`.

macos/snd_.ksy
```diff
@@ -0,0 +1,201 @@
...
+      - id: encode
+        type: u1
+        enum: sound_header_type
+      - id: base_frequency
+        type: u1
+      - id: num_frames
+        type: u4
+      - id: aiff_sample_rate
+        size: 10
+        #this is a float80
```
dgelessus, line 137: Relevant - #210 adds specs for floating-point types that KS doesn't support natively, including 80-bit floats. That PR hasn't been merged yet though, so it can't be used here at the moment.

macos/snd_.ksy
```diff
@@ -0,0 +1,201 @@
...
+      - id: sample_size
+        type: u2
+      - id: future_use_1
+        type: u2
+      - id: future_use_2
+        type: u4
+      - id: future_use_3
+        type: u4
+      - id: future_use_4
+        type: u4
```
dgelessus, line 153: It might be simpler to declare this as a single byte array - if all of the fields are for future use only, it doesn't matter what their data types are. Also, by convention fields that are reserved for future use should be called `reserved`.

```suggestion
      - id: reserved
        size: 14
```

macos/snd_.ksy
```diff
@@ -0,0 +1,201 @@
...
+        size: 10
+        #this is a float80
+      - id: marker_chunk
+        type: u4
+      - id: format
+        size: 4
+        type: str
+        encoding: ASCII
+      - id: future_use_2
+        type: u4
```
dgelessus, line 187: As above, this should be called `reserved`.

macos/snd_.ksy
```diff
@@ -0,0 +1,201 @@
...
+      - id: aiff_sample_rate
+        size: 10
+        #this is a float80
+      - id: marker_chunk
+        type: u4
+      - id: format
+        size: 4
+        type: str
+        encoding: ASCII
```
dgelessus, line 185: Do we know for sure that this field always contains valid ASCII? To be safe it might be better to declare this as a byte array or `u4`, so that non-ASCII chars in the four-char code don't break parsing.

macos/snd_.ksy
tisoft: Changed to doc-ref

macos/snd_.ksy
tisoft: changed to mac_os_resource_snd

macos/snd_.ksy
tisoft: added MIT License

macos/snd_.ksy
tisoft: added MIT license

macos/snd_.ksy
tisoft: Done.

macos/snd_.ksy
tisoft: Done.

macos/snd_.ksy
tisoft: Done.

macos/snd_.ksy
tisoft: Done.

macos/snd_.ksy
tisoft: Done.

macos/snd_.ksy
tisoft: Done.

macos/snd_.ksy
tisoft: Since its coming from the Mac, it's probably Macroman, but personally I have never seen anything non ASCII in it. Docs mention the following:

- NONE
- ACE2
- ACE8
- MAC3
- MAC6

Additionally I have seen the following in the software, I am using the parser for:

- ALAW
- ULAW
- TWOS
- IMA4

macos/snd_.ksy
```diff
@@ -1,9 +1,9 @@
 meta:
-  id: snd_
+  id: mac_os_resource_snd
```
generalmimon, line 2: @tisoft Please also rename the file from `macos/snd_.ksy` to `macos/mac_os_resource_snd.ksy`, the best practice is to keep the file name same as the top level `/meta/id` to avoid confusion.

macos/snd_.ksy
```diff
@@ -0,0 +1,200 @@
...
+    0xFF: extended
+    0xFE: compressed
+doc-ref: "https://developer.apple.com/library/archive/documentation/mac/pdf/Sound/Sound_Manager.pdf"
+seq:
+  - id: format
+    type: u2
+  - id: num_data_formats
+    -orig-id: number_of_data_formats
+    type: u2
+    if: format==1
```
generalmimon, line 43: More readable formatting:

```suggestion
    if: format == 1
```

macos/snd_.ksy
```diff
@@ -0,0 +1,200 @@
...
+  - id: data_formats
+    type: data_format
+    repeat: expr
+    repeat-expr: num_data_formats
+    if: format==1
```
generalmimon, line 48:
```suggestion
    if: format == 1
```

macos/snd_.ksy
```diff
@@ -0,0 +1,200 @@
...
+  - id: reference_count
+    type: u2
+    if: format==2
```
generalmimon, line 51:
```suggestion
    if: format == 2
```

macos/snd_.ksy
```diff
@@ -0,0 +1,200 @@
...
+    seq:
+      - id: raw_cmd
+        type: u2
+      - id: param1
+        type: u2
+      - id: param2
+        type: u4
+    instances:
+      cmd:
+        value: raw_cmd&0x7FFF
```
generalmimon, line 77:
```suggestion
        value: raw_cmd & 0x7FFF
```

macos/snd_.ksy
```diff
@@ -0,0 +1,200 @@
...
+        enum: cmd_type
+      is_data_offset:
+        value: raw_cmd&0x8000
```
generalmimon, line 80:
```suggestion
        value: raw_cmd & 0x8000
```

macos/snd_.ksy
tisoft: Done.

macos/snd_.ksy
tisoft: Done.

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,200 @@
...
+      - id: encode
+        type: u1
+        enum: sound_header_type
+      - id: base_frequency
+        type: u1
+      - id: num_frames
+        type: u4
+      - id: aiff_sample_rate
+        size: 10
+        #this is a float80
```
generalmimon, line 140: I think this comment should be put into the `doc` key instead, because it's quite important thing to know when you're a programmer using this spec and you're wondering how to interpret this value. The value of `doc` is usually put into the docblock in the generated code, so a smart IDE can show this comment in a autocomplete box, which makes sense in this case.

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,200 @@
+meta:
+  id: mac_os_resource_snd
+  endian: be
+  license: MIT
+  xref:
+    wikidata: Q7564684
+enums:
```
generalmimon, line 7: Cccording to the [KSY Style Guide](https://doc.kaitai.io/ksy_style_guide.html#type), `enums` should be present after `doc-ref` and `seq`. In most cases, it's listed at the very end.

macos/mac_os_resource_snd.ksy
tisoft: I agree. I do want to add some doc keys anyway. What is the policy on taking documentation bits from the original documentation? Is that allowed, or should I better rephrase?

generalmimon: > What is the policy on taking documentation bits from the original documentation? Is that allowed, or should I better rephrase?

Uhh, I'm no legal advisor at all, but I think it's allowed, given that you state the original documentation in the `doc-ref` key, as you already do. Anyway, I don't recommend pasting whole sentences to the `doc` if it's not strictly necessary, not from legal reasons, but for better readability. It's generally better to keep the descriptions as concise as possible, and there are usually a lot of unnecessary words in a full sentence, whereas often a few words are enough.

macos/snd_.ksy
dgelessus: Yeah, most likely all possible codes that occur in practice are pure ASCII, so `encoding: ASCII` will probably be fine.

macos/mac_os_resource_snd.ksy
dgelessus: > What is the policy on taking documentation bits from the original documentation? Is that allowed, or should I better rephrase?

Strictly speaking you're not allowed to use text from Apple's documentation, because of copyright. In practice I don't think anyone will care though, as long as you make it clear where the text comes from (which is a good idea anyway).

> Anyway, I don't recommend pasting whole sentences to the `doc` if it's not strictly necessary, not from legal reasons, but for better readability. It's generally better to keep the descriptions as concise as possible, and there are usually a lot of unnecessary words in a full sentence, whereas often a few words are enough.

The original docs aren't always long sentences - for example Apple's older docs often have short descriptions like this:

```pascal
CONST
   stdSH    = $00;    {standard sound header}
   extSH    = $FF;    {extended sound header}
   cmpSH    = $FE;    {compressed sound header}
```

tisoft: I have added some documentation from the apple docs. Would that be ok? If so I would do the same for the enums, if not, I can remove it again.

macos/mac_os_resource_snd.ksy
tisoft: Moved enums below

tisoft:
> By the way, I have a separate repo with some specs related to Mac OS resource forks: https://github.com/dgelessus/resource_fork.ksy
> 
> At the moment the repo only has a spec for the resource fork format itself (plus some specs related to System 7's resource compression). I was also planning to put KSYs for common resource types into that repo, but I haven't gotten around to that yet. (That's actually the main reason why I made it a separate repo instead of adding the specs to kaitai_struct_formats - I didn't want to spam the KSF repo with lots of small resource type specs.)
> 
> Would it be okay if I include this spec in my resource_fork.ksy repo (when I start working on adding resource type specs there)? I'll of course credit you in the commit and include a link back to the original PR here. Or if you prefer, you can also make a PR on my repo, but perhaps after this one is merged, so we don't review the same PR in two places.

I don't know where the best place for this is, I'm all open for adding it to your repo, if that is a better place. I would just suggest we do the review here, and if it's decided, it should better be in your repo, then I will create a pull request for that. The 'snd ' resource is not only used in Resource forks, though. I am using the parser, to read the sounds out of a Mac game:
* https://www.deltatao.com/darkcastle/dc3demo.sit.bin contains 'snd ' resources in the file 'DC Demo Data'
* Clanlord uses them in the CL_Sounds file: https://www.deltatao.com/downloads/clanlord/data/

Which brings me to the next question: I also have Kaitai files for the format used by the DC Demo Data and Cl_Sounds/CL_Images. Would they be worth adding to the repo, or is that too niche?

dgelessus:
> The 'snd ' resource is not only used in Resource forks, though.

Right, there are a few other resource types where that's the case too. For example, PICT images can be stored in a `'PICT'` resource or as a separate file, and the formats are identical except that PICT files have an extra 512-byte header that PICT resources don't. I think QuickTime movies can also be stored as resources instead of .mov files.

For formats like these, it's probably better to put them into kaitai_struct_formats, and then have my resource_fork.ksy import them from KSF as needed. This isn't a big issue, because KSF is sort of like a standard library for Kaitai Struct - I don't have to explicitly include it somehow, because it comes included with the compiler.

Actually I'm wondering if perhaps my resource fork specs would be better off in KSF after all, instead of keeping them in a separate repo. Really my main worry is that in the future I might add a bunch of small specs that are only relevant to an obsolete/"retro" system (classic Mac OS). If that's not a problem, I can make a PR to add my resource fork-related specs into KSF.

dgelessus:
> * https://www.deltatao.com/darkcastle/dc3demo.sit.bin contains 'snd ' resources in the file 'DC Demo Data'

By the way, this link didn't work for me. I looked at [the listing of the parent directory](http://deltatao.com/downloads/darkcastle/) and found another download that also has the data file you mentioned: http://deltatao.com/downloads/darkcastle/dcdemo.sit.hqx

tisoft:
> > * https://www.deltatao.com/darkcastle/dc3demo.sit.bin contains 'snd ' resources in the file 'DC Demo Data'
> 
> By the way, this link didn't work for me. I looked at [the listing of the parent directory](http://deltatao.com/downloads/darkcastle/) and found another download that also has the data file you mentioned: http://deltatao.com/downloads/darkcastle/dcdemo.sit.hqx

Yeah, that was the correct one, I copied the wrong link from the downloads page 🤷‍♂️

tisoft:
Do we need anything else for this to get in?

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,277 @@
...
+        value: raw_cmd & 0x8000
+      sound_header_type:
+        pos: param2 + 20
+        type: u1
+        enum: sound_header_type
+        if: is_data_offset > 0 and cmd == cmd_type::buffer_cmd
+      sound_header:
+        pos: param2
+        size-eos: true
+        if: is_data_offset > 0 and cmd == cmd_type::buffer_cmd
```
generalmimon, line 62: I recommend using [bit-sized integers](https://doc.kaitai.io/user_guide.html#_bit_sized_integers) here, instead of reading a `u2` integer `raw_cmd` and then extracting `cmd` and `is_data_offset` using bitwise AND operators:

```suggestion
    seq:
      - id: is_data_offset
        type: b1
      - id: cmd
        type: b15
        enum: cmd_type
      - id: raw_cmd
        type: u2
      - id: param1
        type: u2
      - id: param2
        type: u4
    instances:
      sound_header_type:
        pos: param2 + 20
        type: u1
        enum: sound_header_type
        if: is_data_offset and cmd == cmd_type::buffer_cmd
      sound_header:
        pos: param2
        size-eos: true
        if: is_data_offset and cmd == cmd_type::buffer_cmd
```

Note that the `b1` type gets implicitly converted to boolean type (`true`/`false`), so you can omit the `is_data_offset > 0` check and use `is_data_offset` as a normal boolean. See the [User Guide > Basic data types](https://doc.kaitai.io/user_guide.html#_basic_data_types):

> **Booleans** can be specified as literal `true` and `false` values as in most languages, but also can be derived by using `type: b1`. This method parses a single bit from a stream and represents it as a boolean value: 0 becomes false, 1 becomes true. This is very useful to parse flag bitfields, as you can omit `flag_foo != 0` syntax and just use something more concise, such as `is_foo`.

tisoft:
Added your changes and verified it still works :)

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,275 @@
...
+        type: u4
+        doc: number of samples in array
+      - id: sample_rate
+        -orig-id: sampleRate
+        type: u2
+        doc: sample rate (Fixed Point, Integer part)
+      - id: sample_rate_frac
+        -orig-id: sampleRate
+        type: u2
+        doc: sample rate (Fixed Point, Fraction part)
```
generalmimon, line 83: I see that you're using this couple of fields 3 times in the spec. Can you extract them to a new user type, which will additionally calculate the final value of _sample rate_? I mean putting the _Integer part_ and _Fraction part_ of the fixed point number together using a [value instance](https://doc.kaitai.io/user_guide.html#_calculated_value_instances).

macos/mac_os_resource_snd.ksy
```diff
@@ -75,12 +75,8 @@ types:
...
       - id: sample_rate
         -orig-id: sampleRate
-        type: u2
-        doc: sample rate (Fixed Point, Integer part)
-      - id: sample_rate_frac
-        -orig-id: sampleRate
-        type: u2
-        doc: sample rate (Fixed Point, Fraction part)
+        type: unsigned_fixed_point
+        doc: sample rate
```
generalmimon, line 79: The `doc` key should not be included if it's only a trivial transcription of the `id`, see [KSY Style Guide > Sequence attributes](https://doc.kaitai.io/ksy_style_guide.html#seq-attr):

> Every key is optional. Attributes SHOULD have at least `id` and `doc` — however, see below for notes about omitting `id`, and `doc` SHOULD NOT be included if it’s trivial (i.e. if it is a copy of `id`, and there is really nothing more to say about that attribute).

```suggestion
```

macos/mac_os_resource_snd.ksy
```diff
@@ -113,12 +109,8 @@ types:
...
```
generalmimon, line 113:
```suggestion
```

See https://github.com/kaitai-io/kaitai_struct_formats/pull/299#discussion_r470956521.

macos/mac_os_resource_snd.ksy
```diff
@@ -176,12 +168,8 @@ types:
...
```
generalmimon, line 172:
```suggestion
```

See https://github.com/kaitai-io/kaitai_struct_formats/pull/299#discussion_r470956521.

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,275 @@
+meta:
+  id: mac_os_resource_snd
+  title: Classic MacOS Sound Resource
+  application: Sound Manager
+  endian: be
+  license: MIT
+  xref:
+    wikidata: Q7564684
+doc-ref: "https://developer.apple.com/library/archive/documentation/mac/pdf/Sound/Sound_Manager.pdf"
```
generalmimon, line 9: Can you please add a few sentences describing the format here in the `doc` key, i.e. some general information about the format? What it is used for? What kind of data it stores? What is the main usage field?

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,278 @@
...
+doc: Sound resources were introduced in Classic MacOS with the Sound Manager program.
+  They can contain sound commands to generate sounds with given frequencies as well as sampled sound data.
+  They are mostly found in resource forks, but can occasionally appear standalone or embedded in other files.
```
generalmimon, line 12: Thanks! However, please use [YAML multiline syntax](https://yaml-multiline.info/) instead:

```suggestion
doc: |
  Sound resources were introduced in Classic MacOS with the Sound Manager program.
  They can contain sound commands to generate sounds with given frequencies as well as sampled sound data.
  They are mostly found in resource forks, but can occasionally appear standalone or embedded in other files.
```

The reason is that when you use any of the [_flow scalars_](https://yaml-multiline.info/#flow-scalars) (plain, single-quoted or double-quoted), single newlines are converted to a single space (at least they _should be_ in a YAML-compliant parser, a quick test at https://ide.kaitai.io/devel/ shows that the [yaml.js](https://github.com/jeremyfa/yaml.js) library that we use is not among them, because the single newlines are preserved instead). This makes the generated docblocks hard to read, because the lines tend to be pretty long. This was generated by doing ` ksc -t java mac_os_resource_snd.ksy`:

```java
/**
 * Sound resources were introduced in Classic MacOS with the Sound Manager program. They can contain sound commands to generate sounds with given frequencies as well as sampled sound data. They are mostly found in resource forks, but can occasionally appear standalone or embedded in other files.
 * @see <a href="https://developer.apple.com/library/archive/documentation/mac/pdf/Sound/Sound_Manager.pdf">Source</a>
 */
public class MacOsResourceSnd extends KaitaiStruct {
```

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,278 @@
...
+
+enums:
+  data_type:
+    0x01: square_wave_synth
+    0x03: wave_table_synth
+    0x05: sampled_synth
+  cmd_type:
+    00: null_cmd
+    03: quiet_cmd
+    04: flush_cmd
```
generalmimon, line 258: Be really careful with prefixing the decimal values with padding zeros. This actually **changes the** number's **meaning**: the number literal is interpreted in **octal** notation (base 8) instead of **decimal** (base 10) if it starts with additional 0s.

For example, try the following YAML snippet in a [YAML parser](https://yaml-online-parser.appspot.com/):

```yaml
012: foo
027: bar
071: baz
```

You'll get the following output:

```js
{
  "10": "foo", // 8 * `1` + `2`
  "23": "bar", // 8 * `2` + `7`
  "57": "baz" //  8 * `7` + `1`
}
```

Here you're only lucky that the numbers `00`, `03` and `04` are written in octal base the same as in decimal. However, as you can see above, this can really cause unexpected results. Please rather eliminate the padding zeros to avoid confusion:

```suggestion
    0: null_cmd
    3: quiet_cmd
    4: flush_cmd
```

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,278 @@
...
+        doc: pointer to LeftOverBlock
+      - id: compression_id
+        -orig-id: compressionID
+        type: u2
+        doc: ID of compression algorithm
+      - id: packet_size
+        -orig-id: packetSize
+        type: u2
+        doc: number of bits per packet
+      - id: snth_id
```
generalmimon, line 226: What is _snth_? You stated the original identifier using the `-orig-id` below, so you can calmly use the full expansion of that word if it'll make more sense.

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,278 @@
...
+        type: u4
+        doc: loop point beginning
+      - id: loop_end
+        -orig-id: loopEnd
+        type: u4
+        doc: loop point ending
+      - id: encode
+        type: u1
+        enum: sound_header_type
+        doc: sample’s encoding option
```
generalmimon, line 94: I think there's a little point in using a Unicode symbol [_Right Single Quotation Mark_](https://unicode-table.com/en/2019/) `’` with code point `U+2019` instead of the normal [`U+0027` Apostrophe](https://unicode-table.com/en/0027/) `'` from the ASCII table.

It might look a bit nicer, but as the whole file is encoded in pure ASCII, sticking to ASCII chars will help with portability:

```suggestion
        doc: sample's encoding option
```

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,278 @@
...
```
generalmimon, line 128:
```suggestion
        doc: sample's encoding option
```

See https://github.com/kaitai-io/kaitai_struct_formats/pull/299#discussion_r471788868.

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,278 @@
...
```
generalmimon, line 187:
```suggestion
        doc: sample's encoding option
```

See https://github.com/kaitai-io/kaitai_struct_formats/pull/299#discussion_r471788868.

dgelessus:
I'm wondering if it would make sense to list the resource type FourCC under `meta/xref`, similar to how file extensions and MIME types are also listed there, in case a tool wants to read it. In a PR for another Mac file format (#312) I started using the key `mac-os-resource-type` for the resource type FourCC (and `mac-os-file-type` for the file type FourCC), so maybe you want to add `mac-os-resource-type: 'snd '` under `xref`.

(As far as I know, `mac-os-resource-type` and `mac-os-file-type` haven't been used before, outside of my PR. But `meta/xref` can contain arbitrary key/value pairs, so it's possible to add new metadata there without modifying the compiler or other tools.)

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,279 @@
...
+          cases:
+            sound_header_type::standard: standard_sound_header
+            sound_header_type::extended: extended_sound_header
+            sound_header_type::compressed: compressed_sound_header
+  standard_sound_header:
+    seq:
+      - id: sample_ptr
+        -orig-id: samplePtr
+        type: u4
+        doc: pointer to samples (or NIL if samples follow data structure)
```
generalmimon, line 76: I'd prefer _0_ instead of _NIL_, if this was the intent.

tisoft:
I think I responded to all open comments. If not, please remind me of the open TODOs. 😄

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,280 @@
...
+        size-eos: true
+        doc: compressed sound data
+  unsigned_fixed_point:
+    seq:
+      - id: integer_part
+        type: u2
+        doc: Integer Part of a fixed point value
+      - id: fraction_part
+        type: u2
+        doc: Fraction Part of a fixed point value
```
generalmimon, line 247: I'd say that these `doc`s are just trivial transcriptions of `id`s, and they bring no new information. As such, they [should be removed](https://doc.kaitai.io/ksy_style_guide.html#seq-attr):

> Attributes SHOULD have at least `id` and `doc` — however, see below for notes about omitting `id`, and **`doc` SHOULD NOT be included if it’s trivial** (i.e. if it is a copy of `id`, and there is really nothing more to say about that attribute).

```suggestion
      - id: fraction_part
        type: u2
```

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,280 @@
...
+        doc: reserved
+      - id: format
+        size: 4
+        type: str
+        encoding: ASCII
+        doc: data format type
+      - id: reserved
+        -orig-id: futureUse2
+        size: 4
+        doc: reserved
```
generalmimon, line 211:
```suggestion
      - id: reserved
        -orig-id: futureUse2
        size: 4
```

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,280 @@
...
+        type: u4
+        doc: pointer to instrument info
+      - id: aes_recording
+        -orig-id: AESRecording
+        type: u4
+        doc: pointer to audio info
+      - id: sample_size
+        -orig-id: sampleSize
+        type: u2
+        doc: number of bits per sample
```
generalmimon, line 155: What about `bits_per_sample` _(or `num_bits_per_sample`, but this one is probably overly verbose)_? `size` almost always refers to **byte** size, so using this word as a _number of bits_ is confusing.

`sample_bitsize` would be also a solution, but looks weird to me.

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,280 @@
...
+        type: u4
+        doc: reserved
+      - id: instrument_chunks
+        -orig-id: instrumentChunks
+        type: u4
+        doc: pointer to instrument info
+      - id: aes_recording
+        -orig-id: AESRecording
+        type: u4
+        doc: pointer to audio info
```
generalmimon, line 151: What about putting the `_ptr` suffix after these two field names? Now, I'd expect `instrument_chunks` to be of type `instrument_chunk[]` (an array `instrument_chunk` as base type) and `aes_recording` to be a single user type, so I find this naming seems misleading when both fields are actually just pointers.

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,280 @@
...
+        type: u4
+        doc: loop point ending
+      - id: encode
+        type: u1
+        enum: sound_header_type
+        doc: sample's encoding option
+      - id: base_frequency
+        type: u1
+        doc: base frequency of sample
+        -orig-id: baseFrequency
```
generalmimon, line 100: It should be stated here in what units is the _frequency_ encoded here, or preferably directly create a value instance with a formula that converts this value to hertz (Hz), if it's possible.

Also if the value stored in this attribute is not directly the _frequency in Hertz_, it should not be called `base_frequency`.

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,280 @@
...
+    repeat: expr
+    repeat-expr: num_sound_commands
+types:
+  data_format:
+    seq:
+      - id: id
+        type: u2
+        enum: data_type
+      - id: options
+        type: u4
```
generalmimon, line 44: It's strange that in the rest of the spec you've described more or less obvious things like `num_channels` &ndash; _number of channels in sample_ (which is not wrong, just a bit redundant), while this broad name `options` doesn't have any description. I think that this field is the one that would deserve an explanation the most.

generalmimon (changes requested):
Overall looks good, but there are a few little things that can be improved.

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,280 @@
...
+        type: u2
+        doc: ID of compression algorithm
+      - id: packet_size
+        -orig-id: packetSize
+        type: u2
+        doc: number of bits per packet
+      - id: synthesizer_id
+        -orig-id: snthID
+        type: u2
+        doc: resource ID of Sound Manager Synthesizer (snth)
```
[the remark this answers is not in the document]
dgelessus: I've found an explanation of this field in an older version of Inside Macintosh (Volume VI, dated 1991). A PDF copy of this version can be found [here](https://vintageapple.org/inside_o/pdf/Inside_Macintosh_Volume_VI_1991.pdf) (warning: it's 400 MB large!). The field is documented on page "22-49" (absolute page number 1169 in the PDF).

> <table><tbody><tr><td>snthID</td><td>Indicates the resource ID number of the 'snth' resource that was used to compress the packets contained in the compressed sound header. A 3:1 'snd ' resource would have a snthID of 11, and a 6:1 'snd ' would have a snthID of 13. If a compressed sound header contains samples that aren't compressed, you should set the snthID field to 0.</td></tr></tbody></table>

It seems that Apple stopped using this field, so later versions of the Sound Manager documentation just say that it's unused and should be set to 0. I think it would be good to point out that it's no longer used, but the previous meaning should still be documented.

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,280 @@
...
+        type: u4
+        doc: pointer to StateBlock
+      - id: left_over_samples_ptr
+        -orig-id: leftOverSamples
+        type: u4
+        doc: pointer to LeftOverBlock
+      - id: compression_id
+        -orig-id: compressionID
+        type: u2
+        doc: ID of compression algorithm
```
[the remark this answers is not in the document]
tisoft: I agree, that it needs to be a signed value. There are some reserved IDs, but in principle every value is valid, so I don't think an enum would be feasible.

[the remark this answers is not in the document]
generalmimon: > There are some reserved IDs, but in principle every value is valid, so I don't think an enum would be feasible.

Well, if you think so, you can at least define both integer and enum variant of this attribute:

```ksy
seq:
  - id: compression_id
    type: u2
instances:
   compression_type:
    value: compression_id
    enum: compression_type_enum
enums:
  compression_type_enum:
    -2: variableCompression
    -1: fixedCompression
    0: notCompressed 
    3: threeToOne
    4: threeToOne
```

It's definitely better than just an opaque integer.

A user of this can first switch on the enum values of `compression_type` and if it has an unknown enum value, then access the raw integer `compression_id` and do something with it.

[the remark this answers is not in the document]
tisoft: Thats a good idea! Will do that.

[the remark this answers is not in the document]
tisoft: added the enum

macos/mac_os_resource_snd.ksy
tisoft: the frequency is represented as midi note. (60 being "middle C")

https://en.wikipedia.org/wiki/MIDI_tuning_standard

It can be converted to a frequency in hertz, but that needs power function support for floats and I don't think that is currently supported in Kaitai, or is it?

MIDI to freqeuncy formula

values from 0-127 are supported, so I could add a 127 entry lookup switch, but don't know if that is really wanted. 😄

macos/mac_os_resource_snd.ksy
tisoft: The options are passed to the SndNewChannel as the "init" parameter.

There are some documented values for this field (Page 2-91 in https://developer.apple.com/library/archive/documentation/mac/pdf/Sound/Sound_Manager.pdf), but I don't know if it is feasible to extract them as own instances.

What do you think?

macos/mac_os_resource_snd.ksy
generalmimon: > values from 0-127 are supported, so I could add a 127 entry lookup switch, but don't know if that is really wanted. 😄

OK, you don't have to do that if you don't want to, but it actually wouldn't be so bad idea. Lookup tables can be defined in KS, and there's nothing wrong about that: in fact, it will be naturally faster than evaluating lots of expressions from scratch. And a one-time 1k bytes of memory (`8 * 128` for `double`) are nowadays a ridiculously small amount, so that isn't an issue either.

Neither does it take up much space in the spec: I wrote a JavaScript snippet putting 8 numbers per line with 2 decimals, and it would be 16 lines with max 70 chars per line, which is acceptable _(open DevTools with `F12` or `Ctrl+Shift+I` and paste, or maybe https://jsbin.com/ if you're on mobile)_:

```javascript
const lookupTable = [];
for (let d = 0; d < 127; d++) {
  lookupTable[d] = (2 ** ((d - 69) / 12)) * 440;
}

function printTable(table, {decimals, numsPerLine}) {
  return table
    .map(f => f.toFixed(decimals))
    .reduce((acc, val) => {
        let row = acc[acc.length - 1];
        if (row.length >= numsPerLine) {
          row = [];
          acc.push(row);
        }
        row.push(val);
        return acc;
    }, [[]])
    .map(a => a.join(", "))
    .join("\n");
}

console.log(printTable(lookupTable, {decimals: 2, numsPerLine: 8}));
```

Check out the MacOS dcmp_2 spec by @dgelessus for comparison 😁:

https://github.com/kaitai-io/kaitai_struct_formats/blob/d1e58e36f26ecff1375c6022dc040342914ee4d1/macos/resource_compression/dcmp_2.ksy#L89-L156

macos/mac_os_resource_snd.ksy
```diff
@@ -98,7 +98,6 @@ types:
       - id: midi_note
         type: u1
         doc: base frequency of sample, expressed as MIDI note values, 60 is middleC
```
generalmimon:
```suggestion
      - id: midi_note
        type: u1
        doc: base frequency of sample, expressed as MIDI note values, 60 is middle C
```

macos/mac_os_resource_snd.ksy
generalmimon: @tisoft If you think that the lookup table is a bit overkill, then please at least put the formula `(2 ** ((d - 69) / 12)) * 440` in the `doc` key, that might be useful for the end user.

tisoft: Implemented the lookup table for the standard header, will unify the headers, so that it is available in all.

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,262 @@
...
+        if: sound_header_type == sound_header_type::compressed
+      - id: reserved
+        -orig-id: futureUse2
+        size: 4
+        if: sound_header_type == sound_header_type::compressed
+      - id: state_vars_ptr
+        -orig-id: stateVars
+        type: u4
+        doc: pointer to StateBlock
+        if: sound_header_type == sound_header_type::compressed
```
generalmimon, line 159: You got the point about reusing code, but I'd prefer extracting these fields from the `compressed_sound_header` to a special subtype. There are a lot of consecutive `if: sound_header_type == sound_header_type::compressed` to my taste.

Again, I refer to the [`bmp.ksy` spec](https://github.com/kaitai-io/kaitai_struct_formats/blob/d1e58e36f26ecff1375c6022dc040342914ee4d1/image/bmp.ksy#L350-L372), where I solved this by splitting the `bitmap_header` to multiple extensions _(subtypes), which are enabled/disabled as a whole using only one `if` per extension. IMHO this makes the spec clearer and more understandable.

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,262 @@
...
+        type: u1
+        enum: sound_header_type
+        doc: sample's encoding option
+      - id: midi_note
+        type: u1
+        doc: base frequency of sample, expressed as MIDI note values, 60 is middle C
+        -orig-id: baseFrequency
+      - id: num_frames
+        type: u4
+        if: sound_header_type != sound_header_type::standard
```
generalmimon, line 125: I'd prefer positive matches than negative:

```suggestion
        if: |
          sound_header_type == sound_header_type::extended
          or sound_header_type == sound_header_type::compressed
```

Repeating this for multiple fields is inconvenient, so it might be a good idea to extract it to a value instance `is_extended_or_compressed` (or whatever), or move the fields to a new "extension"

The reason behind this is that it just _explicitly_ says what we know and nothing more: we shouldn't assume anything about the headers we are not aware of. If Apple released a new `sound_header` version and someone added support for it to this KSY spec, it would be _implicitly_ matched by the `if: sound_header_type != sound_header_type::standard` condition, which might be wrong.

macos/mac_os_resource_snd.ksy
tisoft: I'll have a look 👀

macos/mac_os_resource_snd.ksy
tisoft: I have added some instances and enums to better describe the contents of the options field

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,345 @@
...
+      - id: midi_note
+        type: u1
+        doc: base frequency of sample, expressed as MIDI note values, 60 is middle C
+        -orig-id: baseFrequency
+      - id: extended_or_compressed
+        type: extended_or_compressed
+        if: sound_header_type == sound_header_type::extended or sound_header_type == sound_header_type::compressed
+      - id: sample_area
+        -orig-id: sampleArea
+        size-eos: true
```
generalmimon, line 158: I would prefer the variant when the size of `sample_area` is calculated, as in your previous versions of the spec:

https://github.com/kaitai-io/kaitai_struct_formats/blob/9adbc8f49d6000f0023dd8c1b8ed2d3faae37d16/macos/mac_os_resource_snd.ksy#L97

https://github.com/kaitai-io/kaitai_struct_formats/blob/9adbc8f49d6000f0023dd8c1b8ed2d3faae37d16/macos/mac_os_resource_snd.ksy#L137-L140

---

https://github.com/kaitai-io/kaitai_struct_formats/blob/9adbc8f49d6000f0023dd8c1b8ed2d3faae37d16/macos/mac_os_resource_snd.ksy#L141

https://github.com/kaitai-io/kaitai_struct_formats/blob/9adbc8f49d6000f0023dd8c1b8ed2d3faae37d16/macos/mac_os_resource_snd.ksy#L196-L199

You can go with multiple `sample_area`s switched on/off using `if`s, or use the conditional operator _(aka ternary op)_ for joining the expressions:

```suggestion
        size: |
          sound_header_type == sound_header_type::standard ? num_samples : 
          sound_header_type == sound_header_type::extended ? num_frames * num_channels * bits_per_sample / 8 :
          _io.size - _io.pos
```

The last fallback expression (`size: _io.size - _io.pos`) is actually an equivalent of `size-eos: true`, but if think that it's better to have a second field switched using `if` with explicit `size-eos: true` instead, it's OK too.

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,348 @@
...
+    0x0300:
+      id: mace3
+      -orig-id: initMACE3
+      doc: MACE 3:1
+    0x0400:
+      id: mace6
+      -orig-id: initMACE6
+      doc: MACE 6:1
+
+  cmd_type:
```
generalmimon, line 317: Maybe these enum values deserve adding `doc` descriptions from page 2-12 [in the PDF](https://developer.apple.com/library/archive/documentation/mac/pdf/Sound/Sound_Manager.pdf)?

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,348 @@
...
+        523.25, 554.37, 587.33, 622.25, 659.26, 698.46, 739.99, 783.99,
+        830.61, 880.00, 932.33, 987.77, 1046.50, 1108.73, 1174.66, 1244.51,
+        1318.51, 1396.91, 1479.98, 1567.98, 1661.22, 1760.00, 1864.66, 1975.53,
+        2093.00, 2217.46, 2349.32, 2489.02, 2637.02, 2793.83, 2959.96, 3135.96,
+        3322.44, 3520.00, 3729.31, 3951.07, 4186.01, 4434.92, 4698.64, 4978.03,
+        5274.04, 5587.65, 5919.91, 6271.93, 6644.88, 7040.00, 7458.62, 7902.13,
+        8372.02, 8869.84, 9397.27, 9956.06, 10548.08, 11175.30, 11839.82, 12543.85
+      ]
+    doc: |
+      Lookup table to convert a MIDI note into a frequency in hz
```
generalmimon, line 59:
```suggestion
      Lookup table to convert a MIDI note into a frequency in Hz
```

Sorry for my nitpicking 🙂

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,348 @@
...
+          sound_header_type == sound_header_type::extended ? extended_or_compressed.num_frames * num_channels * extended_or_compressed.bits_per_sample / 8 :
+          _io.size - _io.pos
+        doc: sampled-sound data
+    instances:
+      base_freqeuncy:
+        value: _root.midi_note_to_frequency[midi_note]
+        #TODO: If https://github.com/kaitai-io/kaitai_struct/issues/216 is implemented:
+        #TODO: value: (2 ** ((midi_note - 69) / 12)) * 440
+        doc: |
+          base frequency of sample in hz
```
generalmimon, line 169:
```suggestion
          base frequency of sample in Hz
```

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,445 @@
...
+      - id: extended_or_compressed
+        type: extended_or_compressed
+        if: sound_header_type == sound_header_type::extended or sound_header_type == sound_header_type::compressed
+      - id: sample_area
+        -orig-id: sampleArea
+        size: |
+          sound_header_type == sound_header_type::standard ? num_samples :
+          sound_header_type == sound_header_type::extended ? extended_or_compressed.num_frames * num_channels * extended_or_compressed.bits_per_sample / 8 :
+          _io.size - _io.pos
+        doc: sampled-sound data
```
generalmimon, line 162: `sample_area` is present here only if `sample_ptr == 0`, right?

```suggestion
        if: sample_ptr == 0
        doc: sampled-sound data
```

Also please create an `instance` with `pos: sample_ptr` parsing the samples area if `sample_ptr != 0`.

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,445 @@
...
+        enum: cmd_type
+      - id: param1
+        type: u2
+      - id: param2
+        type: u4
+    instances:
+      sound_header:
+        pos: param2
+        type: sound_header
+        size-eos: true
```
generalmimon, line 115: I suppose that you don't know here if the `sound_header` ` structure will stretch to the EOF, do you?

```suggestion
```

macos/mac_os_resource_snd.ksy
tisoft: I think samplePtr points to a memory location, not a file location, so I don't think I can just take the value in an instance pos. Additionally in the case of a compressed header, I don't know the compressed size:

> A pointer to the sampled-sound data. If the sampled sound is located in memory immediately after the futureUse4 field, then this field should be set to NIL. Otherwise, this field is a pointer to the memory location of the sampled-sound data.

https://developer.apple.com/library/archive/documentation/mac/pdf/Sound/Sound_Manager.pdf 2-107

generalmimon: @tisoft 

> I think samplePtr points to a memory location, not a file location, so I don't think I can just take the value in an instance pos. Additionally in the case of a compressed header, I don't know the compressed size

OK, fine, the Mac OS sound resource is really a weird format 😃 Let it be then.

macos/mac_os_resource_snd.ksy
```diff
@@ -0,0 +1,445 @@
...
+        #TODO: value: (2 ** ((midi_note - 69) / 12)) * 440
+        doc: |
+          base frequency of sample in Hz
+          Calculated with the formula (2 ** ((midi_note - 69) / 12)) * 440
+        doc-ref: https://en.wikipedia.org/wiki/MIDI_tuning_standard
+        if: midi_note >= 0 and midi_note < 128
+      sound_header_type:
+        pos: 20
+        type: u1
+        enum: sound_header_type
```
generalmimon, line 176: Oh, sorry, I forgot to add to https://github.com/kaitai-io/kaitai_struct_formats/pull/299#discussion_r506129877 that this won't work anymore. You need to offset the `pos: 20` relatively to the start of the `sound_header`. A common solution is to capture and save the `_io.pos` in the beginning of the `sound_header`, like this:

```ksy
types:
  sound_header:
    seq:
      - size: 0
        if: start_ofs < 0 # invoking the `start_ofs` value instance to save the current `_io.pos`
      - id: sample_ptr
        -orig-id: samplePtr
        type: u4
      # ...
    instances:
      start_ofs:
        value: _io.pos
      sound_header_type:
        pos: start_ofs + 20
        type: u1
        enum: sound_header_type
```

Or, you can also move the `sound_header_type` outside of the `sound_header` like you had before:

https://github.com/kaitai-io/kaitai_struct_formats/blob/9adbc8f49d6000f0023dd8c1b8ed2d3faae37d16/macos/mac_os_resource_snd.ksy#L70

https://github.com/kaitai-io/kaitai_struct_formats/blob/9adbc8f49d6000f0023dd8c1b8ed2d3faae37d16/macos/mac_os_resource_snd.ksy#L81-L86

However, maybe I prefer the version with capturing `start_ofs`, because that's basically guaranteed to work under all circumstances, regardless of whether the `sound_header` is wrapped in a substream or not. With the `pos: param2 + 20` version, there's a room for errors like `param1 + 20` etc.

generalmimon:
The spec looks good to me, I'll only test some snippets from the PDF.

FWIW, here's my JavaScript snippet for converting the listings in the spec to hex dump, which can be consumed by an hex editor:

```javascript
const def = `data 'snd ' (1, "Simple Beep", purgeable) {
    /*the sound resource header*/
    $"0001" /*format type*/
    $"0001" /*number of data types*/
    $"0001" /*square-wave data*/
    $"00000000" /*initialization option*/
    /*the sound commands*/
    $"001B" /*number of sound commands (27)*/
    $"002C" /*command 1--timbreCmd 090 000*/
    $"005A00000000"
    $"002B" /*command 2--ampCmd 224 000*/
    $"00E000000000"
    $"002A" /*command 3--freqCmd 000 069*/
    $"000000000045"
    $"000A" /*command 4--waitCmd 040 000*/
    $"002800000000"
    $"002B" /*command 5--ampCmd 200 000*/
    $"00C800000000"
    /*commands 6 through 26 are omitted; they are */
    /* alternating pairs of waitCmd and ampCmd commands */
    /* where the first parameter of ampCmd has the */
    /* values 192, 184, 176, 168, 160, 144, 128, 96, */
    /* 64, and 32*/
    $"002B" /*command 27--ampCmd 000 000*/
    $"000000000000"
};`;
console.log(
  Array.from(
    def.matchAll(/\$"([0-9A-F ]+)"/g)
  )
  .map(match => match[1].replace(/\s/g, ''))
  .join('')
);
```

generalmimon (approved):
Looks good to me, thank you @tisoft for contributing such a big chunk of work!

I'm also astonished at how quickly you were incorporating my suggestions 😉

tisoft:
Anything else needed from me before this can be merged?

PR #303, opened 2020-07-13
Add specs for Mac OS resource fork and resource compression formats

dgelessus:
These specs come from my [resource_fork.ksy repo](https://github.com/dgelessus/resource_fork.ksy) (as of commit dgelessus/resource_fork.ksy@f6225e07a1a66aab6f9b2f1f7a1ef97810e3b4c5), with some small modifications and renaming to fit the kaitai_struct_formats directory structure better.

This PR only adds specs for the resource fork format itself and its compression formats. My resource_fork.ksy repo also has a few specs for [individual resource types](https://github.com/dgelessus/resource_fork.ksy/tree/master/src/resources), which I might add in a separate PR in the future.

This is sort of a follow-up to the discussion in #299.


common/data_with_io.ksy
```diff
@@ -0,0 +1,11 @@
+meta:
+  id: data_with_io
```
KOLANICH, line 2: In my specs I usually call this stuff a `placeholder`.

macos/compressed_resource.ksy
```diff
@@ -0,0 +1,182 @@
...
+        doc: |
+          The type-specific part of the header,
+          as a raw byte array.
+      type_specific_part:
+        io: type_specific_part_raw_with_io._io
+        pos: 0
+        type:
+          switch-on: common_part.header_type
+          cases:
+            8: type_specific_part_type_8
```
KOLANICH, line 75: Could the names be more meaningful?

macos/resource_compression/dcmp_0.ksy
```diff
@@ -0,0 +1,525 @@
...
+    repeat-until: _.tag == 0xff
+    doc: |
+      The sequence of chunks that make up the compressed data.
+types:
+  chunk:
+    seq:
+      - id: tag
+        type: u1
+        doc: |
+          The chunk's tag byte.
```
KOLANICH, line 51: `u1` is already a byte.

macos/resource_compression/dcmp_0.ksy
```diff
@@ -0,0 +1,525 @@
...
+          This controls the structure of the body and the meaning of the chunk.
+      - id: body
+        type:
+          switch-on: |
+            tag >= 0x00 and tag <= 0x1f ? "literal"
```
KOLANICH, line 56: I wonder if it is more efficient to encode as numbers. I mean enum members.

common/data_with_io.ksy
dgelessus: I'm open to suggestions for better names here :) But I don't think `placeholder` is a good term for this... To me, "placeholder" sounds like a part that hasn't been specced/reversed yet, or a field that is not yet used.

macos/compressed_resource.ksy
dgelessus: This part of the format is practically undocumented - there are no official docs, and the unofficial docs that exist are short or incomplete (which I'm trying to fix 🙂), so there aren't any good descriptive names for this. I used more descriptive names before, but it turned out that they were inaccurate (there were some additional compression schemes that I didn't know about when I made up the names). So I switched to naming them by the number in the preceding header, because that's the only part that I know is accurate 🙂 (see dgelessus/python-rsrcfork/commit/f3b3de496ed607282413fba0ba9bf80a77c695c9)

macos/resource_compression/dcmp_0.ksy
dgelessus: > `u1` is already a byte.

I suppose I could change "tag byte" to just "tag" everywhere, as it's not very relevant that it's a byte.

> Also, can it be a enum or a bit-sized type?

An enum would be overly complicated, because there are a few long ranges of tag values with almost identical meaning. Bit fields don't work here, because the boundaries between the ranges are not always nicely aligned (for example `0x4b`, which is `0b1001011` in binary), and because the different ranges use different bit structures.

macos/resource_compression/dcmp_0.ksy
dgelessus: An enum works quite well here, thanks for the suggestion!

It could be implemented even more nicely with kaitai-io/kaitai_struct#489, but we don't have any feature like that yet.

macos/resource_compression/dcmp_0.ksy
dgelessus: > as it's not very relevant that it's a byte

Actually it is relevant for some chunk types. For example `literal_body` and `backreference_body` store small number values in the tag byte to save space. To extract those values from the tag, you need to know that it's specifically a byte. To make the logic behind those calculations clearer I called it "tag byte" everywhere in the `doc`s.

macos/resource_compression/dcmp_0.ksy
```diff
@@ -0,0 +1,535 @@
...
+              Fixed lookup table that maps tag byte numbers to two bytes each.
+
+              The entries in the lookup table are offset -
+              index 0 stands for tag 0x4b, 1 for 0x4c, etc.
+          value:
+            value: lookup_table[tag - 0x4b]
+            doc: |
+              The two bytes that the tag byte expands to,
+              based on the fixed lookup table.
+        doc: |
```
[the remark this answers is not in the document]
dgelessus: OK, apparently I misread that part of the style guide - I applied the convention for fields/instances (`doc` as the last thing) to non-top-level types as well. You're completely right that in terms of readability it makes more sense to put type `doc`s at the top, I'll change that.

common/data_with_io.ksy
```diff
@@ -0,0 +1,11 @@
...
+  title: Byte array with an `_io` member
+  license: MIT
+  ks-version: "0.8"
```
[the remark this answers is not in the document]
dgelessus: Yeah, this spec is very simple, so it's probably compatible with many older ksc versions too and the minimum version can be removed.

I usually put `ks-version: "0.8"` in new specs when I want to say that they specifically don't require 0.9. They might be compatible with older versions too, but I would expect that practically no one uses versions older than 0.8 anymore, so I normally don't bother checking. (If someone is stuck on an older ksc version for whatever reason, they can remove the `ks-version` and see if the spec compiles anyway.)

macos/resource_fork.ksy
```diff
@@ -0,0 +1,593 @@
...
+        doc: Reserved space for a copy of the resource file header.
+      - id: reserved_next_resource_map_handle
+        type: u4
+        doc: Reserved space for a handle to the next loaded resource map in memory.
+      - id: reserved_file_reference_number
+        type: u2
+        doc: Reserved space for the resource file's file reference number.
+      - id: file_attributes
+        type: file_attributes
+        size: sizeof<file_attributes> # Force creation of a substream
```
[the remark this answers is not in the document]
dgelessus: Normally I wouldn't have used an explicit `size` here at all. I only needed to add one so that a new substream is created for this field, so that the `as_int` instance in `file_attributes` works correctly (hence the comment next to it). Otherwise I want the same behavior as if the field had no explicit `size` at all (i. e. let the type read all the data that it needs and continue reading directly after that), so I tried to simulate that using `sizeof<...>`.

I understand what you mean with checking that the fields in the structure add up to the expected size, but I don't think setting a manually hardcoded `size` really helps with that. No matter if you have a manual `size` or not, if you get the number/size of fields in the type wrong, you'll get errors when parsing data. The manual `size` might make the error messages a little more useful if you accidentally make the type too long, but it will silence errors if the type is too short. There's also an extra chance of errors if you miscalculate the manual `size`.

I don't see other specs in KSF do this either - for example `archive/gzip` doesn't have `size`s on its flag integers:

https://github.com/kaitai-io/kaitai_struct_formats/blob/db23b8c0fb214d7e0259409b6fe3bc76bd020dfc/archive/gzip.ksy#L33-L46

It might not be a bad idea to check a type's calculated size against an expected byte count (especially for things like bit flag integer fields, where the size is explicitly given in the documentation), but `size` isn't the right tool for that. Ideally, checks like this would be done at the type level and statically by the compiler, not when the type is used and at runtime. This would of course require new syntax and support by the compiler and is not something we can do in specs right now.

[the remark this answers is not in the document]
generalmimon: > There's also an extra chance of errors if you miscalculate the manual `size`.

The classical approach (i.e. used by at least 95% of manually written parsers I assume) is to read a n-byte integer, which represents the packed bitfield, and then using bitwise operators to extract the individual fields. The specifications usually count on this as well: first, you'll get the information that there's a field `flags` that has 2 bytes in sizes, and _then_ you'll see the internal structure of the `flags` bitfield. I haven't seen a format specification that would just show a list of bit integers in `flags` with just the number of bits the individual fields occupy, without any byte size, saying _dear programmers, summing the bit sizes up to get the `flags` size in bytes is left as an exercise for the reader_.

That said, you usually get the byte size in advance, so the _miscalculation_ IMHO doesn't occur often, if you don't invent your own format and you don't know yet how many bits you'll need in the end.

> The manual `size` might make the error messages a little more useful if you accidentally make the type too long, but it will silence errors if the type is too short.

Of course, but `size: sizeof<..>` does the same. When the bits have a total of 15 bits, `sizeof<..>` will be 2. But there are cases where this behavior is intended. As I say above, when you're developing a `.ksy` spec for some format, you may postpone dissecting the internal structure of the `flags` to the point when you'll need it, so you first just mark they occupy 2 bytes and skip it. Or imagine that you're trying to figure out the structure of a bitfield during reverse-engineering. You can save yourself the trouble of ensuring that the fields have a total size of 16 bits by constantly updating the bit size of the "unknown purpose" field at the end by just wrapping the type with flags in a substream with fixed byte size.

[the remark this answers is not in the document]
generalmimon: > Normally I wouldn't have used an explicit `size` here at all. I only needed to add one so that a new substream is created for this field, so that the `as_int` instance in `file_attributes` works correctly (hence the comment next to it). Otherwise I want the same behavior as if the field had no explicit `size` at all (i. e. let the type read all the data that it needs and continue reading directly after that), so I tried to simulate that using `sizeof<...>`.

I can't recommend using the current substream in a subtype which ends with a bit-sized integer field, because if the subtype `seq` accidentally doesn't end on a byte boundary (8, 16, 24, ...), it can wreak havoc sometimes. This is due to the bug relating to the byte alignment after bit-sized integer followed by a byte aligned field. The current implementation treats all subtypes as byte-aligned and doesn't look what's inside them. So if a subtype does not end on a byte boundary and it's followed by a byte-aligned field (so the `alignToByte()` method should be called, but doesn't happen of course), it will eventually cause weird things first time you parse a bit-sized integer in the same stream (see https://github.com/kaitai-io/kaitai_struct/issues/155#issuecomment-616219006).

Putting the subtype in a substream will ensure that everything after it will be "aligned" automatically, because it doesn't affect the `bits` and `bitsLeft` properties of the parent stream.

macos/resource_compression/dcmp_2.ksy
```diff
@@ -0,0 +1,289 @@
...
+    repeat-expr: header_parameters.num_custom_lookup_table_entries
+    if: header_parameters.flags.has_custom_lookup_table
+    doc: |
+      The custom lookup table to be used instead of the default lookup table.
+  - id: data
+    type:
+      switch-on: header_parameters.flags.tagged
+      cases:
+        false: untagged_data
+        true: tagged_data
```
generalmimon, line 55: I'd personally prefer this variant:

```suggestion
      cases:
        true: tagged_data
        _: untagged_data
```

..., because then the "default byte array" case isn't generated at all and the byte array type also doesn't pollute the resulting combined type in static typed languages (combining two different user types with a byte array type results in generic `Object` in Java, whereas combining only two user types yields `KaitaiStruct`). If any attribute has size and doesn't have the else case denoted by underscore `_`, the compiler logically assumes that the cases do not fill the full range of possible values the expression in `switch-on` can have and thus generates the fallback byte array case.

But it's rather such a side note, if you think that having verbatim values `true` and `false` helps readability, I accept it. The type casting in statically-typed languages will still be needed in most cases because the result type would be `KaitaiStruct` (unless the programmer wants to only access properties available in `KaitaiStruct` directly like `_io` or `_parent`), but it can reduce confusion of why this attribute has type `Object` when it should apparently be `KaitaiStruct`.

dgelessus: Makes sense. It's really annoying that the compiler behaves this way - IMO the default behavior of `switch-on` should be to generate an error if a value is encountered that isn't listed under cases (and there is no explicit `_` branch). Of course this can't be changed easily because of backwards compatibility. Also, there's no explicit name for the byte array type, so the implicit default case is actually necessary if you want to have a byte array case in a `switch-on`...

kaitai-io/kaitai_struct#208 would be useful to have here.

generalmimon:
@dgelessus Please fix the broken links to your repository - https://github.com/kaitai-io/kaitai_struct_formats/search?q=python-rsrcfork.

Referring to the ever-changing `master` branch always seems like a _great idea_, until it isn't 🙂

dgelessus:
Fixed, thanks. Yeah, I originally I made them non-permalinks on purpose, so that they point to the current code/documentation if I update it. Of course this breaks when the repo gets restructured (what idiot does that anyway? 😛), and realistically those files aren't going to change much anymore, so a permalink works just as well.

PR #305, opened 2020-07-17
Added Broadcom patchram and Bluetooth HCI

KOLANICH:
@matthiasseemoo @jiska2342


jiska2342:
Looks plausible :)
The `.hcd` comes in slightly different variants, though, depending on the chip. Especially the Patchram can be at different positions. Some of that is summarized in our firmware files but it's a mess... https://github.com/seemoo-lab/internalblue/tree/master/internalblue/fw

hardware/bluetooth/updateVendorIds.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python3
```
generalmimon, line 1: It's nice that you wrote a script for downloading the vendor IDs and putting them into the `bluetooth_vendors_ids.ksy` spec, but you don't expect to get this Python script merged into the official KS format gallery repo, do you? I suggest placing it into your own repo instead, and you can add a link to that repo into the `doc` key of `bluetooth_vendors_ids.ksy` spec. Let's keep the https://github.com/kaitai-io/kaitai_struct_formats repo reserved for `.ksy` format specifications.

hardware/bluetooth/.gitignore
```diff
@@ -0,0 +1 @@
+/company-identifiers.html
```
generalmimon, line 1: I have no idea what this is about.

hardware/bluetooth/bluetooth_hcd.ksy
```diff
@@ -0,0 +1,25 @@
+meta:
+  id: bluetooth_hcd
+  title: Bluetooth Host Controller Interface Control Protocol
+  license: Unlicense
+  file-extension:
+    - hcd
```
generalmimon, line 6: Quoting from the [KSY Style Guide](https://doc.kaitai.io/ksy_style_guide.html#meta):

> c. **`file-extension`** — if there's **only one** extension, **MUST be a string**; if there are several, (...)

```suggestion
  file-extension: hcd
```

hardware/bluetooth/bluetooth_vendors_ids.ksy
```diff
@@ -0,0 +1,2343 @@
+meta:
+  id: bluetooth_vendors_ids
+  title: Bluetooth Company identifiers
+  endian: le
```
generalmimon, line 4: Can be removed:

```suggestion
```

hardware/bluetooth/control/bluetooth_control_command.ksy
```diff
@@ -0,0 +1,42 @@
+meta:
+  id: bluetooth_control_command
+  title: Bluetooth Host Controller Interface Control Protocol
+  license: Unlicense
+  file-extension:
+    - hcd
```
generalmimon, line 6: See https://github.com/kaitai-io/kaitai_struct_formats/pull/305#discussion_r456796225:

```suggestion
  file-extension: hcd
```

generalmimon: Isn't `bluetooth_control_command` meant to be an import-only spec? Is there some version of a binary file with `.hcd` extension that can be directly parsed using the `bluetooth_control_command`, without using the `bluetooth_hcd` format?

If the `bluetooth_control_command` is not meant to represent the full `.hcd` format, then there IMHO shouldn't be the `file-extension` key at all.

hardware/bluetooth/control/bluetooth_control_command.ksy
```diff
@@ -0,0 +1,42 @@
...
+      - id: packed
+        type: u2
+    instances:
+      group:
+        -orig-id: OpCode Group Field
+        value: packed >> 10
+        enum: group
+      command:
+        -orig-id: OpCode Command Field
+        value: packed & 0x3FF
```
generalmimon, line 39: Is there a reason why you don't use the little-endian bit-sized integers (https://github.com/kaitai-io/kaitai_struct/issues/155) instead, when we already have them implemented? Compatibility with stable 0.8 can be hardly a decent motive why use this old-school approach, because you already use lists in `doc-ref`.

I think relying on version 0.9 is OK, many existing specs also do it, and hopefully the 0.9 version will be released soon.

```suggestion
    meta:
      bit-endian: le
    seq:
      - id: command
        -orig-id: OpCode Command Field
        type: b10
      - id: group
        -orig-id: OpCode Group Field
        type: b6
        enum: group
```

hardware/bluetooth/control/bluetooth_control_command.ksy
KOLANICH: In fact originally there was only one file with a mess inside, then I have copied it into multiple places and removed the irrelevant parts and made other changes. It seems I have forgotten to remove the `file-extension` from it.

hardware/bluetooth/control/bluetooth_control_command.ksy
KOLANICH: Is it implemented for all the target languages, the most importantly python and C++?

And, BTW, why not just `endian` or even not my endianness encoding scheme (the links to the impl in python (it has to be ported to scala to be usable, it makes no sense to move these computations to the runtime) are present in the issue)?

hardware/bluetooth/.gitignore
KOLANICH: No, my script caches the webpage into this file.

hardware/bluetooth/updateVendorIds.py
KOLANICH: >but you don't expect to get this Python script merged into the official KS format gallery repo, do you?

Of course I do. This script makes no sense without the file it is intended to update. And the file to be updated will rot without this script. Of they both should belong to the same repo.

>I suggest placing it into your own repo instead

It may make sense to put the whole `bluetooth` dir into an own repo and link it as a submodule.

hardware/bluetooth/control/bluetooth_control_command.ksy
dgelessus: > Is it implemented for all the target languages, the most importantly python and C++?

Yes, see https://github.com/kaitai-io/kaitai_struct/issues/155#issuecomment-623027660 and specifically kaitai-io/kaitai_struct_python_runtime@c5ff9a1e397aa3307ba2a0e217b58dfe898cbeed and kaitai-io/kaitai_struct_cpp_stl_runtime@37a8d3bc137313b9672fdf367bc942cb8691853b.

> why not just `endian` or even not my endianness encoding scheme

This has been discussed a few months ago in kaitai-io/kaitai_struct#155 and the implementation work has already been done - it's a bit late now for alternative feature suggestions... A separate `bit-endian` attribute is needed because of backwards compatibility - `endian` previously hasn't affected bit fields, so changing that would silently break some existing specs. The two options for `bit-endian` are enough to support all possible bit/byte order combinations (see https://github.com/kaitai-io/kaitai_struct/issues/155#issuecomment-615345304), so a more complex format for specifying endianness shouldn't be necessary.

KOLANICH: >it's a bit late now for alternative feature suggestions...

OK. IMHO it still makes sense to just break the compatibility and just fix the specs in this repo and just force the authors to fix their specs, rather than to accumulate tech debt and mess (I cannot imagine real world scenarios where numbers are LE but bit fields are used as BE and IMHO the name chosen is very containtuitive).

Thanks for the suggestion and the clarification. Should probably apply it to my other specs too

PR #311, opened 2020-08-07
Added ASUS trx

KOLANICH:


KOLANICH:
Thank you.

firmware/asus_trx.ksy
```diff
@@ -0,0 +1,124 @@
...
+  license: GPL-2.0
+  title: Image headers used in ASUS devices official firmware
+  xref:
+    wikidata: Q152864
+  endian: le
+  bit-endian: le
+  encoding: utf-8
+
+-license: |
+  Copyright (C) 2004  Manuel Novoa III  <mjn3@codepoet.org>
```
generalmimon, line 12: @KOLANICH To be frank, I don't understand why you put GPL-2.0 on this trivial spec and declared this innocent guy _Manuel Novoa III_ the copyright holder. As you probably know, [_copyright protects the expression of an idea but it does not protect the underlying idea itself_](https://www.bclplaw.com/en-US/insights/does-copyright-protect-data-file-formats.html#:~:text=copyright%20protects,the%20underlying%20idea%20itself). It also [_does not protect individual words, short phrases, and slogans_](https://www.dmlp.org/legal-guide/works-not-covered-copyright#words).

I compared the reference code you linked in `doc-ref` with your spec and I found that you didn't take any code, just these two short comments if I'm not mistaken:

https://github.com/kaitai-io/kaitai_struct_formats/blob/7d2894cba0363eb9d74edeb434eafbc0ecc7c7ac/firmware/asus_trx.ksy#L98

https://github.com/kaitai-io/kaitai_struct_formats/blob/7d2894cba0363eb9d74edeb434eafbc0ecc7c7ac/firmware/asus_trx.ksy#L108

So these 2 phrases consisting of total 12 words are all that can be considered _copyrightable content_ if we take it too seriously, but I think they are too easy to defend as _short phrases_ in front of a hypothetical court (which of course never even starts unless the entire world has gone nuts 😆).

@KOLANICH, do you still think it's wise to put GPL-2.0 on this spec? I mean - I noticed that you don't like the GPL licenses too much because they are _viral_, and neither I can understand why use the GPL on this simple KSY.

KOLANICH: >To be frank, I don't understand why you put GPL-2.0 on this trivial spec

Because it was derived from the file having GPL-2.0 license.

>you probably know, copyright protects the expression of an idea but it does not protect the underlying idea itself. It also does not protect individual words, short phrases, and slogans.

The exact boundary can be determined only by a court as a part of lawsuit. Sometimes multiple times with appealing the decision of the previous court.

>and declared this innocent guy Manuel Novoa III the copyright holder.

One that wilingly has chosen GPL is not innocent.

Mercurial is not this spec and is not derived from that code, and even ist author is another person, but it is one of the many examples of the mindset of the people that choose GPL.

https://www.mercurial-scm.org/pipermail/mercurial/2011-March/037583.html

>There are basically two camps of people. People like Martin, who think they can circumvent the GPL by looking at the original Python code to see how Mercurial works and then write a new Mercurial (or parts of it) in a different programming language, claiming it can then be under a different license of their choice, in particular a license which lifts all those ugly constraints they don't like. And there are others (like myself) who think such a thing might be a derived work. Ultimately, this can only be decided in court.

TL;DR: Adrian Buehlmann bullied (all the) persons capable to create an alternative impl of Mercurial with a threat of possibility of bringing a lawsuit against them, unless they do clean room/ black-box RE.

Another example, git maintainer, Junio Hamano:
https://public-inbox.org/git/xmqqeexbrffe.fsf@gitster-ct.c.googlers.com/
> It takes me some effort to convince myself that it is worth for us to spend extra brain cycles even thinking about making such statements, when those who would benefit the most from such an effort are the ones who call the software we work on "contamination" and proclaim that they cannot be bothered to spend the [necessary effort](https://public-inbox.org/git/4B11ADEA-8E8F-444F-B4CA-C4A3E9536F8C@mail.ru/) to comply with the license as a waste of time.

TL;DR: J. Hamano considers clean room/black-box reverse engeneering of "free software" as "necessary effort".

And ones choosing GPL choose GPL exactly because this license was designed to be non-permissive. One either comples with GPL in its most strict interpretation, or pays lawers in courts for deciding in a concrete case and then with some probability of paying damages. It is the result of Richard Stallman's ambitions partly. Society is all about negotiations and compromises between actors of power, and Stallman &  Co. were determined to find a compromise between them and some subset of proprietary software creators, and have managed to do this. FSF has created a license, using which some interested copyright bullies can claim they support and use free software, in the same time doing bullying on the basis of GPL inacceptable conditions violation. The cost for the bullies is that they open the sources. What FSF gets is promotion of its name which is written in eny copy of GPL virus. And some law firms formerly closely associated with FSF get money. And of course, there is proliferation of "free software" as a byproduct. The costs for FSF are the costs of the developing of the license. "Negotiations" have succeeded, the parties (FSF and all the copyright holders that otherwise would have used another license, probably a fully proprietary one - at another ) have managed to aggree to act for mutual benefit.

PR #312, opened 2020-08-08
Add spec for the Apple icon image (ICNS) format

### LINKED ISSUE #218: Add Apple ICNS
```yaml
meta:
  id: apple_icns
  title: Apple Icon Image format
  application:
    - Apple MacOS
    - Apple 
  file-extension: icns
  xref:
    justsolve: ICNS
    mime: image/icns
    uti: com.apple.icns
    wikidata: Q2858737

doc: |
  Icons file format used in Apple MacOS.

doc-ref:
  - https://en.wikipedia.org/wiki/Apple_Icon_Image_format
```

dgelessus:
Closes #218.

I'm marking this as a draft because I still want to do some more testing. The spec is basically finished though, so feel free to leave comments.

### DIFF
--- a/image/icns.ksy
+++ b/image/icns.ksy
@@ -1,0 +1,889 @@
+meta:
+  id: icns
+  title: Apple icon image (ICNS)
+  application:
+    # Listing only official Apple software.
+    # Various third-party software also supports reading and writing the format.
+    - macOS
+    - Preview (macOS)
+    - icns Browser
+    - Icon Composer
+    - iconutil
+  file-extension: icns
+  xref:
+    justsolve: ICNS
+    mac-os-file-type: icns
+    mac-os-resource-type: icns
+    mime:
+      - image/icns
+      - image/x-icns
+    pronom: fmt/1185
+    uti: com.apple.icns
+    wikidata: Q2858737
+  license: MIT
+  ks-version: "0.9"
+  imports:
+    - /common/bytes_with_io
+  endian: be
+doc: |
+  Icon image format used by Mac OS 8.5 and later,
+  including all versions of Mac OS X/macOS.
+  This is the Mac equivalent of the Windows ICO format.
+
+  An ICNS file stores an *icon family*:
+  a collection of images with visually the same content,
+  but at different resolutions and color depths.
+  When the system renders an icon on screen,
+  it chooses the most optimal format from the icon family to display to the user,
+  depending on the icon's displayed size and the capabilities of the display hardware.
+
+  An icon family in an ICNS file can also contain other nested icon families.
+  This feature is sometimes used to represent different states or visual variants of the same icon,
+  such as an opened, selected or dark mode version.
+
+  Each icon in an icon family is identified by a four-byte type code,
+  which also indicates the image data format and resolution.
+  Supported formats include PNG and JPEG 2000 (for larger sizes and modern systems)
+  and multiple raw bitmap formats with varying resolution, color depth, and transparency support
+  (for smaller sizes and/or compatibility with older systems).
+
+  ICNS data can be stored either as a standalone file,
+  or as a resource with type code `'icns'` in a resource fork.
+  The latter was especially common on Classic Mac OS and for Carbon applications,
+  and is still used (as of macOS 10.14) by the Finder to store custom file and folder icons set by the user.
+doc-ref:
+  - '<OSServices/IconStorage.h>'
+  - '<OSServices/IconStorage.r>'
+  - '<HIServices/Icons.r>'
+  - 'https://sourceforge.net/p/icns/ libicns SourceForge project'
+  - 'https://github.com/python-pillow/Pillow/blob/master/src/PIL/IcnsImagePlugin.py Python Pillow ICNS plugin code'
+  - "https://www.macdisk.com/maciconen.php#icns Information about 'icns' resources in resource forks"
+seq:
+  - id: root_element
+    type: icon_family_element
+    valid:
+      expr: |
+        _.header.type.as_enum == icon_family_element::header::type::main_family
+    doc: |
+      The element at the root of the file.
+      This element must have type `"icns"`,
+      should be as long as the entire stream,
+      and contains the main icon family.
+types:
+  icon_family_element:
+    doc: |
+      A single element in an icon family.
+
+      It is normally safe to ignore elements with an unrecognized type code or unparseable data when reading -
+      they are most likely new icon types/resolutions, icon family variants, or metadata introduced in a newer system version.
+      However, when modifying and writing an icon family,
+      such elements should be stripped out,
+      to avoid leaving outdated information in the icon family that could be used by newer systems.
+    -webide-representation: '{header}'
+    seq:
+      - id: header
+        type: header
+        doc: The header storing the element's type code and length.
+      - id: data_with_io
+        type: bytes_with_io
+        size: header.len_data
+        doc: |
+          Use `data` instead,
+          unless you need access to this field's `_io`.
+    instances:
+      data:
+        value: data_with_io.data
+        doc: The raw data stored in the element.
+      data_parsed:
+        io: data_with_io._io
+        pos: 0
+        type:
+          switch-on: header.type.as_enum
+          cases:
+            'header::type::main_family': icon_family_data
+            'header::type::tile_variant_family': icon_family_data
+            'header::type::rollover_variant_family': icon_family_data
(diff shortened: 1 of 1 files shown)


image/icns.ksy
```diff
@@ -0,0 +1,889 @@
...
+          - id: height
+            type: u4
+            doc: The height of the icon in pixels.
+        seq:
+          - id: data
+            type:
+              switch-on: _io.size == width * height * 3
+              cases:
+                false: icns_style_packbits
+                # If true, the data is uncompressed and should be read as a raw byte array.
```
generalmimon, line 790: This attribute doesn't have the `size` key, so it won't be read as a byte array. If the condition `_io.size == width * height * 3` is `true`, the value of `data` will be `null` instead (or the equivalent in the particular target language).

image/icns.ksy
```diff
@@ -0,0 +1,889 @@
...
+      data:
+        value: data_with_io.data
+        doc: The raw data stored in the element.
+      data_parsed:
+        io: data_with_io._io
+        pos: 0
+        type:
+          switch-on: header.type.as_enum
+          cases:
+            'header::type::main_family': icon_family_data
```
KOLANICH, line 103: Given that they are text strings of fixed length and have a kind of structure within them .... would it worth not to use enums for them, but to decompose and parse them in KS?

image/icns.ksy
```diff
@@ -0,0 +1,889 @@
...
+        params:
+          - id: width
+            type: u4
+            doc: The width of the icon in pixels.
+          - id: height
+            type: u4
+            doc: The height of the icon in pixels.
+        seq:
+          - id: icon
+            size: width * height / 8
```
KOLANICH, line 717: `repeat-expr` + `b1`?

image/icns.ksy
```diff
@@ -0,0 +1,889 @@
...
+            doc: The width of the icon in pixels.
+          - id: height
+            type: u4
+            doc: The height of the icon in pixels.
+        seq:
+          - id: data
+            type:
+              switch-on: _io.size == width * height * 3
+              cases:
+                false: icns_style_packbits
```
KOLANICH, line 789: IMHO, for uniformity a type having exactly the same interface as `icns_style_packbits` is needed. We also would like to have all the types containing image data having the same interface. Even if we don't have interfaces in KS currently, for dynamic duck-typed languages this fact is a bit less important.

image/icns.ksy
dgelessus: Thanks, I completely missed that.

I think I already mentioned this in my previous PR, but I'm *really* not a fan of how the implicit default case in `switch-on`s works, and that it's *necessary* to use the implicit default case to have a byte array case in a `switch-on`. I've explained my thoughts on this in more detail in a separate issue: kaitai-io/kaitai_struct#799

image/icns.ksy
dgelessus: Yes, the type codes follow a certain pattern, but it's not consistent enough for parsing to make sense or work properly. The most predictable type codes are the ones for the Classic Mac OS-style bitmaps, which all follow the `ic□#`/`ic□4`/`ic□8`/`i□32`/`□8mk` naming scheme - except:

* the `m` ("mini") size doesn't have an RGB version (`im32`/`m8mk` don't exist)
* the `l` ("large") size the 1-bit+mask variant has type code `ICN#` (not `icl#`)
* the `t` ("thumbnail") size *only* has an RGB version (`ict#`/`ict4`/`ict8` don't exist)
* `it32` has a *slightly* different structure (four extra zero bytes at the beginning) than all the other `i□32` types

And it doesn't get much better with the PNG/JP2 types, which are called `ic0□` (regular scale) and `ic1□` (HiDPI scale), except:

* `ic04` (16x16) and `ic05` (32x32) are actually ARGB bitmaps - their PNG/JP2 versions are called `icp4` and `icp5`
* there's also `icp6`, which is 64x64 - a size that doesn't exist in *any* other type (it doesn't even have an ARGB or HiDPI version)
* `ic10` is the lowest-numbered HiDPI icon, but has the highest resolution (512x512) - because older OSes actually considered a non-HiDPI 1024x1024 icon

There's also `icsb`/`icsB`, which are ARGB in another odd size (18x18). And then you have the metadata and icon family types, which follow no naming pattern at all - one of them isn't even readable text (`0xfdd92fa8`)!

KOLANICH: Thanks. It would be nice to write content this summary into the spec itself.

dgelessus: > It would be nice to write content this summary into the spec itself.

This is just a summary of what's already written in the spec - all the available type codes are listed in the enum, their format/structure is documented in the big switch and the `*_data` types, and the types that would fit the naming scheme but don't exist are pointed out in comments in between the real types.

> I also wonder, if they really don't exist or just undocumented. I mean, have anyone done any experiments on feeding the data with "nonexistent" ids into Apple software and observing the results?

They really don't exist as far as I can tell. I modified some icon files to include those icon types (`im32`, `m8mk`, `ict#`, `ict4`, `ict8`, `ic06`) on system versions that understand other similar icon types, and none of them show up in the Preview application. Actually the valid "mini" sizes don't show up under Preview either (even on Mac OS X 10.4) - to view those I had to use an older tool called icns Browser, and that clearly indicates that "mini" RGB and mask don't exist (it doesn't even show empty placeholders for those types, which it does for types that are valid but not in the file).

image/icns.ksy
dgelessus: I thought about it, but decided that it probably wouldn't have any practical use.

* If you're using the spec in a visualizer, having the data parsed into rows/pixels isn't very helpful - if you want to see the bitmap, just look at the hex dump, or maybe the visualizer has a built-in option to render bitmap data.
* If you're using the spec from a programming language, there are probably existing libraries that can read arrays of packed integers much more efficiently than KS can at the moment, or perhaps even image processing libraries that can read raw bitmap data into high-level image objects. In the case of Python that would be `array`, `numpy`, and/or `PIL`/Pillow.

image/icns.ksy
dgelessus: > for uniformity a type having exactly the same interface as `icns_style_packbits` is needed in `true` branch.

How would that be possible? `icns_style_packbits` is compressed data and the byte array case is uncompressed data - they *have* to be handled differently by code that uses the spec.

> We also would like to have all the types containing image data having the same interface.

I thought about this, and I intentionally made the different image data types have similar/compatible param/field/instance names. I didn't make them all completely identical though - for example the Classic-style bitmap data types don't have a `scale` param/instance, because none of them are HiDPI. I wasn't sure if it's a good idea to add instances that only have a constant value - for people just reading the spec it's unnecessary information, and in a visualizer it might be confusing why those values are there.

image/icns.ksy
KOLANICH: Thanks.

image/icns.ksy
KOLANICH: >How would that be possible? icns_style_packbits is compressed data and the byte array case is uncompressed data - they have to be handled differently by code that uses the spec.

I have implemented this kind of stuff once in one of my specs, but I don't remeber which one exactly. I had to use 2 types, one is for compressed and another one for uncompressed. I had to do that because `process` allows no `switch-on`. So for uniformity I had to made an uncompressed one have the same fields as compressed one, switch the wrapped types, and then pass through their content through instances using dirty hacks as `.as<>` to get a bit simpler API. I know that passing through instances duplicates the size, but it is more a bug on compiler side rather than in specs.

dgelessus: Turns out that I was wrong and these bitmaps are always compressed - there's no special case for uncompressed bitmaps. The problem was that I originally tested it with a bitmap that (in uncompressed form) is also valid compressed data, and when decompressed it looks very similar to the original, so I didn't notice that the data was misinterpreted.

dgelessus:
Alright, I think this is ready now - I'm relatively confident that this spec handles basically everything correctly, after doing some testing with [my little ICNS inspector/extractor tool/library](https://github.com/dgelessus/python-icns) that I wrote based on this spec.

No major changes in the last two commits - I just moved around some types/enums to make them more easily accessible from application code.

dgelessus:
> I'm relatively confident that this spec handles basically everything correctly

Forget that, I found another weird corner case... there are icons in some of Apple's own apps (the Lion version of iWork '09 specifically) that store raw bitmap data in icon types which I thought could only contain PNG/JPEG 2000 images. Apple's ICNS implementation handles this case without problems, so I guess that means it's allowed.

PR #314, opened 2020-08-17
Added Qt Installer Framework

KOLANICH:
Since [my PR into QIF repo](https://github.com/qtproject/installer-framework/pull/7) was ignored for years, it seems they are both disinterested in maintaining the spec and disrespectful enough to just ignore PRs. So I post the spec here.


archive/qt_installer_framework.ksy
```diff
@@ -0,0 +1,225 @@
+meta:
+  id: qt_installer_framework
+  title: Qt Installer Framework archive
+  application: Qt Installer Framework
+  file-extension: exe
+  endian: le
+  encoding: utf-8
+  license: GPL-3.0  # With exceptions, see https://github.com/qtproject/installer-framework/blob/master/LICENSE.GPL3-EXCEPT
```
generalmimon, line 8: Looks like this exception is explicitly listed under the SPDX license exceptions, see https://spdx.org/licenses/Qt-GPL-exception-1.0.html.

Therefore I think there's some way how to express the _"`GPL-3.0` with `Qt-GPL-exception-1.0`_ using a SPDX expression, but I don't know how it would look like. Possibly using the `WITH` operator, as the https://spdx.org/licenses/ page suggests?

> - [**License Exceptions**](https://spdx.org/licenses/exceptions-index.html): The list of commonly found exceptions to free and open source licenses, which can be used with the License Expression operator, "WITH" to create a license with an exception.

archive/qt_installer_framework.ksy
```diff
@@ -0,0 +1,225 @@
...
+  xref:
+    wikidata: Q201904
+
+doc: |
+  Qt installer framework is a set of libs to make SFX installers. Installers usually contain 7zip-compressed archives. Obviously, Qt installer itself is built using this framework.
+  
+  Warning 1: KSC has a bug. It makes the computed values be int32_t. Of course their type should be either explicitly specified by a programmer or derived automatically. The workaroind is to just replace all int32_t to int64_t in sources.
+  Warning 2: don't use this spec on Linux against Qt distribution with overcommit enabled unless you have lot of RAM (> 12 GiB). There is a severe memory leak somewhere (currently I have no idea where exactly). The leak is present in both C++ and python-compiled code. In python even if I have patched the generated source to explicitly free all the `bytes`, `BytesIO`s and `KaitaiStream` objects the leak is still present. At least it is neither in `bytes` nor in `BytesIO`. In C++ I have not patched anything but used move semantics to free the stuff since std::unique_ptr is used. The leak is still present. IDK where it is and how to fix it.
```
generalmimon, line 16: Can you please wrap the long lines to 80 characters, as the [KSY Style Guide](https://doc.kaitai.io/ksy_style_guide.html#doc) recommends?

> Lines should be wrapped to be 80 columns long. If it doesn’t fit into single line after wrapping, then it’s a multi-line docstring, so use proper multi-line syntax.

archive/qt_installer_framework.ksy
```diff
@@ -0,0 +1,225 @@
...
+            enum: type
+          - id: signature
+            contents: [0x68, 0xd6, 0x99, 0x1c, 0x0a, 0x63, 0xc2]
+        enums:
+          type:
+            0xf8: binary
+            0xf9: data
+
+      range:
+        -orig-id: https://github.com/qtproject/installer-framework/blob/cf8d5b9a650cc413dde79cf7d72da407c8993ddc/src/libs/installer/fileio.cpp#L52
```
generalmimon, line 108: It's unusual to have URL in `-orig-id`, why don't you use `doc-ref` instead?

archive/qt_installer_framework.ksy
KOLANICH: I am against manual wrapping. Instead wrapping must be done by a text editor depending on window size.

archive/qt_installer_framework.ksy
KOLANICH: Nice catch.

archive/qt_installer_framework.ksy
KOLANICH: Thanks, fixed.

archive/qt_installer_framework.ksy
KOLANICH: Fixed.

PR #320, opened 2020-08-21
Added CRDB

KOLANICH:
Replaces #88 and addresses license issues by using an SPDX expression.

### DIFF
--- a/database/crdb.ksy
+++ b/database/crdb.ksy
@@ -1,0 +1,187 @@
+meta:
+  id: crdb
+  title: GNU/Linux wireless regulation database
+  license: copyleft-next-0.3.0 AND ISC
+  endian: be
+  encoding: ASCII
+
+-license-URIs:
+  - https://github.com/copyleft-next/copyleft-next/blob/master/Releases/copyleft-next-0.3.0
+
+-license-header: |
+  Copyright (c) 2013, Luis R. Rodriguez <mcgrof@do-not-panic.com>
+
+  CRDA is licensed under copyleft-next 0.3.0
+
+  Refer to copyleft-next-0.3.0 for license details, you can also always get the latest updates of copyleft-next from:
+
+  https://gitorious.org/copyleft-next/
+
+  This file incorporates work covered by the following copyright and permission notice:
+
+  Copyright (c) 2008, Luis R. Rodriguez <mcgrof@do-not-panic.com>
+  Copyright (c) 2008, Johannes Berg <johannes@sipsolutions.net>
+  Copyright (c) 2008, Michael Green <Michael.Green@Atheros.com>
+
+  Permission to use, copy, modify, and/or distribute this software for any purpose with or without fee is hereby granted, provided that the above copyright notice and this permission notice appear in all copies.
+
+  THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+doc: |
+  the test file: https://git.kernel.org/pub/scm/linux/kernel/git/sforshee/wireless-regdb.git/plain/regulatory.bin
+doc-ref:
+  - "https://wireless.wiki.kernel.org/en/developers/Regulatory/wireless-regdb"
+  - "https://git.kernel.org/pub/scm/linux/kernel/git/mcgrof/crda.git/tree/regdb.h"
+  - "https://git.kernel.org/pub/scm/linux/kernel/git/mcgrof/crda.git/tree/reglib.c"
+  - "https://git.kernel.org/pub/scm/linux/kernel/git/sforshee/wireless-regdb.git/tree/db2bin.py"
+seq:
+  - id: signature
+    contents: ["RGDB"]
+  - id: version
+    type: u4
+  - id: states_ptr
+    -orig-id: countries_ptr
+    type: u4
+    doc: The list is sorted alphabetically to allow binary searching (should it become really huge).
+  - id: states_count
+    -orig-id: countries_count
+    type: u4
+  - id: authentication_size
+    -orig-id: signature_len
+    type: u4
+    doc: length (in bytes) of the signature at the end of the file
+instances:
+  states:
+    pos: states_ptr
+    type: state
+    repeat: expr
+    repeat-expr: states_count
+  authentication:
+    pos: _io.size - signature_size
+    size: authentication_size
+    doc: digital signature
+types:
+  freq_range:
+    doc: |
+      everything is in in kHz
+      Values of zero mean "not applicable", i.e. the regulatory does not limit a certain value.
+    seq:
+     - id: start
+       type: u4
+     - id: end
+       type: u4
+     - id: max_bandwidth
+       type: u4
+  power_rule:
+    seq:
+      - id: max_antenna_gain
+        type: u4
+        doc: in mBi (100 * dBi)
+      - id: max_eirp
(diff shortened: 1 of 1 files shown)


database/crdb.ksy
```diff
@@ -0,0 +1,187 @@
...
+    type: u4
+    doc: length (in bytes) of the signature at the end of the file
+instances:
+  states:
+    pos: states_ptr
+    type: state
+    repeat: expr
+    repeat-expr: states_count
+  authentication:
+    pos: _io.size - signature_size
```
GreyCat, line 60: Fails to compile now due to

```
crdb:!:!:/instances/authentication/pos: unable to access 'signature_size' in crdb context
```

database/crdb.ksy
```diff
@@ -0,0 +1,187 @@
+meta:
+  id: crdb
```
GreyCat, line 2: "crdb" doesn't seem to be very widely used term. [Linux kernel wiki](https://wireless.wiki.kernel.org/en/developers/Regulatory/wireless-regdb) names it "wireless-regdb", which is available in text and binary forms. "CRDB" would be abbreviation for "Central Regulatory (Domain Agent) DataBase", and it's not used widely. In fact, google search for that rather reveals [CRDB Bank on Wikipedia](https://en.wikipedia.org/wiki/CRDB_Bank), stuff in [Redis](https://docs.redislabs.com/latest/rs/administering/creating-databases/create-active-active/), [Cosmic-Ray DataBase](https://lpsc.in2p3.fr/crdb/), etc.

Suggestion: rename this to "wireless_regdb".

database/crdb.ksy
```diff
@@ -0,0 +1,187 @@
...
+  title: GNU/Linux wireless regulation database
```
GreyCat, line 3: If we'll be super-pendantic, this is not *GNU*-specific, but actually Linux kernel-specific.

KOLANICH: Fixed. Thanks.

database/crdb.ksy
KOLANICH: Renamed.

database/crdb.ksy
KOLANICH: Thanks, fixed.

database/wireless_regdb.ksy
```diff
@@ -0,0 +1,189 @@
+meta:
+  id: wireless_regdb
+  title: Linux wireless regulation database (aka Central Regulatory (Domain Agent) DataBase)
+  license: copyleft-next-0.3.0 AND ISC
+  endian: be
+  encoding: ASCII
+  xref:
+    wikidata: Q11375
```
GreyCat, line 8: `wikidata` is not supposed to point to arbitrary Wikipedia topics loosely associated with this data format (e.g. "wireless network" in this case). It is supposed to point exactly to the the Wikidata entry associated with this format. Otherwise, it's better *not* to specify it.

GreyCat: NB also: 23aca49

database/wireless_regdb.ksy
```diff
@@ -0,0 +1,189 @@
...
+    contents: ["RGDB"]
+  - id: version
+    type: u4
+  - id: states_ptr
+    -orig-id: countries_ptr
+    type: u4
+    doc: The list is sorted alphabetically to allow binary searching (should it become really huge).
+  - id: states_count
+    -orig-id: countries_count
+    type: u4
```
GreyCat, line 50: I understand the original intent behind renaming "countries" to "states", but in this case I would rather not do that:

* "states" might be confused with USA states
* "states" might be confused with different persistent states of some kind of process

Let's stick with "countries"?

database/wireless_regdb.ksy
```diff
@@ -0,0 +1,189 @@
...
+instances:
+  states:
+    pos: states_ptr
+    type: state
+    repeat: expr
+    repeat-expr: states_count
+  authentication:
+    pos: _io.size - authentication_size
+    size: authentication_size
+    doc: digital signature
```
GreyCat, line 64:
```suggestion
    doc: Digital signature: RSA-signed SHA-1 hash of the file.
```

database/wireless_regdb.ksy
```diff
@@ -0,0 +1,189 @@
...
+          type: b1
+          doc: indoor operation not allowed
+        - id: no_cck
+          -orig-id: RRF_NO_CCK
+          type: b1
+          doc: CCK modulation not allowed
+        - id: no_ofdm
+          -orig-id: RRF_NO_OFDM
+          type: b1
+          doc: OFDM modulation not allowed
```
GreyCat, line 152: Lines 110 to 152 have more than 1 indentation.

database/wireless_regdb.ksy
KOLANICH: How about `jurisdictions`?

database/wireless_regdb.ksy
KOLANICH: I have set this because https://en.wikipedia.org/wiki/Central_Regulatory_Domain_Agent redirects there. Can remove.

database/wireless_regdb.ksy
KOLANICH: Also `realm`s?

database/wireless_regdb.ksy
```diff
@@ -0,0 +1,196 @@
...
+        repeat-expr: count
+
+  realm:
+    -orig-id: regdb_file_reg_country
+    seq:
+      - id: two_letter_code
+        -orig-id: alpha2
+        type: str
+        size: 2
+        -doc-ref: https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes
```
generalmimon, line 169:
```suggestion
        doc-ref: https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes
```

The key `doc-ref` shouldn't be dash-prefixed.

database/wireless_regdb.ksy
```diff
@@ -0,0 +1,196 @@
...
+      - id: ptr
+        type: u4
+    instances:
+      rule:
+        pos: ptr
+        type: rule
+  flags:
+    seq:
+      - id: reserved1
+        type: b20
```
generalmimon, line 109: Given that there are bit-sized integers used here, there should be explicit `meta/bit-endian: be` at the top level. See https://doc.kaitai.io/user_guide.html#bit-endian.

database/wireless_regdb.ksy
```diff
@@ -0,0 +1,196 @@
...
+    repeat: expr
+    repeat-expr: realms_count
+  authentication:
+    pos: _io.size - authentication_size
+    size: authentication_size
+    doc: digital signature
+types:
+  freq_range:
+    doc: |
+      everything is in in kHz
```
generalmimon, line 66:
```suggestion
      everything is in kHz
```

database/wireless_regdb.ksy
KOLANICH: I wonder if signatire format may change in future. I feel like it is not a part of this format, and for this format the signature should be treated as an OpenSSL signature of the hash (why don't they feed the blob to crypto libs directly (the sign call hashes the stuff itself, so they double-hash)?).
So I have added that this is a signature of SHA1 hash, but without specifying that it is RSA.

database/wireless_regdb.ksy
KOLANICH: :facepalm: Fixed! Thanks.

database/wireless_regdb.ksy
KOLANICH: Fixed, thanks.

database/wireless_regdb.ksy
KOLANICH: Fixed, thanks.

database/wireless_regdb.ksy
```diff
@@ -0,0 +1,202 @@
+meta:
+  id: wireless_regdb
+  title: Linux wireless regulation database (aka Central Regulatory (Domain Agent) DataBase)
+  license: copyleft-next-0.3.0 AND ISC
+  endian: be
+  bit-endian: be
+  encoding: ASCII
+
+-license-URIs:
+  - https://github.com/copyleft-next/copyleft-next/blob/master/Releases/copyleft-next-0.3.0
```
generalmimon, line 10: This is not necessary. The SPDX expression `copyleft-next-0.3.0 AND ISC` specified in `meta/license` unambiguously identifies the license on its own. If anyone wants to read the full text of the license, it's always available at `https://spdx.org/licenses/{$identifier}.html` (e.g. [https://spdx.org/licenses/`copyleft-next-0.3.0`.html](https://spdx.org/licenses/copyleft-next-0.3.0.html)).

And BTW, the key `-license-URIs` does not really follow any reputable naming convention (sure, no strict rules apply for dash-prefixed keys, but you know: some sort of consistency is always beneficial). But I'd rather remove it completely.

Issue #153, opened 2019-05-21
Zeno IMproved


KOLANICH:
```yaml
meta:
  id: zim
  title: "(Open) Zeno IMproved"
  application: 
    - Kiwix
    - zimlib
  file-extension: zim
  xref:
    wikidata: Q784695
  license: CC-BY-SA-3.0
  encoding: utf-8
  endian: le
doc: |
  A file format to store ecyclopaedias of articles written in MediaWiki markup language. 
  Files for test: https://dumps.wikimedia.org/other/kiwix/zim/wikipedia/

doc-ref:
  - https://www.openzim.org/wiki/ZIM_file_format
  - https://wiki.openzim.org/wiki/OpenZIM
```

```yaml
WiP:
  - https://github.com/KOLANICH/kaitai_struct_formats/blob/OpenZIM/media/openzim.ksy
```

GreyCat:
ecyclopaedias => encyclopaedias?

KOLANICH:
Good catch! Fixed (currently in the template only), thanks.

generalmimon:
> encyclop<ins>**ae**</ins>dias

This is a British spelling. See https://en.wikipedia.org/wiki/Encyclopedia:

> An **encyclopedia** or **encyclopaedia** ([British English](https://en.wikipedia.org/wiki/British_English)) is a [reference work](https://en.wikipedia.org/wiki/Reference_work) or (...)

and https://www.merriam-webster.com/dictionary/encyclopaedia:

> ## encyclopaedia, encyclopaedic
>
> **Definition of _encyclopaedia_**
>
> chiefly British spellings of [ENCYCLOPEDIA](https://www.merriam-webster.com/dictionary/encyclopedia) , [ENCYCLOPEDIC](https://www.merriam-webster.com/dictionary/encyclopedic)

I thought that we prefer American spelling for KS identifiers, don't we? In fact, @KOLANICH suggested this himself in https://github.com/kaitai-io/kaitai_struct/issues/522#issuecomment-468059463:

> One more thing. American spelling: `meter`, not `metre`.

KOLANICH:
>In fact, @KOLANICH suggested this himself in kaitai-io/kaitai_struct#522 (comment)

Please don't rip out of context. That suggestion was in context of units. Units are the things that must be uniform, and most of units libraries use American spelling for a meter (though some support both), so introducing there `metre` instead of `meter` would cause a need of additional remapping of that id too.

[...]

PR #343, opened 2020-09-21
Added OpenZIM

KOLANICH:
Fixes: #153


media/open_zim.ksy
```diff
@@ -0,0 +1,356 @@
...
+  Files for test: https://dumps.wikimedia.org/other/kiwix/zim/wikipedia/
+
+doc-ref:
+  - https://www.openzim.org/wiki/ZIM_file_format
+  - https://wiki.openzim.org/wiki/OpenZIM
+
+seq:
+  - id: signature
+    -orig-id: magicNumber
+    contents: [0x5A, 0x49, 0x4D, 0x04]
```
generalmimon, line 25:
```suggestion
    contents: ['ZIM', 0x04]
```

media/open_zim.ksy
```diff
@@ -0,0 +1,356 @@
...
+    pos: md5_ptr
+    type: md5
+  minimal_xz_lzma_size:
+    value: 32
+types:
+  md5:
+    seq:
+      - id: data
+        size: 16
+  uuid:
```
generalmimon, line 113: Note that the [devel Web IDE](https://ide.kaitai.io/devel/) supports UUID formatting using the `-webide-representation` key, see https://github.com/kaitai-io/kaitai_struct_webide/commit/7f8e6a91bcfa9af3cb84ba2d9f5b452b1f321e1a:

> **Closes kaitai-io/kaitai_struct_webide#55: implement "uuid" formatter for webide-representation**
> 
> Use `-webide-representation: "{field_name:uuid}"` for the standard big-endian format or `{field_name:uuid=ms}` for Microsoft's implementation

I have no idea which UUID format (big-endian or the Microsoft one) the OpenZIM uses, though. The byte order of both formats can be seen on line [`src/utils.ts:84`](https://github.com/kaitai-io/kaitai_struct_webide/blob/7f8e6a91bcfa9af3cb84ba2d9f5b452b1f321e1a/src/utils.ts#L84).

media/open_zim.ksy
```diff
@@ -0,0 +1,356 @@
...
+      The MIME type list always follows directly after the header, so the mimeListPos also defines the end and size of the ZIM file header.
+      The MIME types in this list are zero terminated strings.
+  - id: main_page_idx
+    -orig-id: mainPage
+    type: u4
+    doc: main page or 0xffffffff if no main page
+  - id: layout_page_idx
+    -orig-id: layoutPage
+    type: u4
+    doc: layout page or 0xffffffffff if no layout page
```
generalmimon, line 65: The value `0xff_ffff_ffff` stated in the comment would be the highest value of a 5-byte unsigned int. I suggest putting underscores in such long hex literals, it makes them more clear:

```suggestion
    doc: layout page or 0xffff_ffff if no layout page
```

media/open_zim.ksy
```diff
@@ -0,0 +1,356 @@
...
+    pos: cluster_pointer_list_ptr
+    type: cluster_ptr
+    repeat: expr
+    repeat-expr: cluster_count
+  main_page:
+    value: url_pointer_list[main_page_idx]
+    if: main_page_idx != 0xffffffff
+  layout_page:
+    value: url_pointer_list[layout_page_idx]
+    if: layout_page_idx != 0xffffffff
```
generalmimon, line 102: To play well with https://github.com/kaitai-io/kaitai_struct_formats/pull/343#discussion_r502899630:

```suggestion
  main_page:
    value: url_pointer_list[main_page_idx]
    if: main_page_idx != 0xffff_ffff
  layout_page:
    value: url_pointer_list[layout_page_idx]
    if: layout_page_idx != 0xffff_ffff
```

media/open_zim.ksy
```diff
@@ -0,0 +1,356 @@
...
+        size: 16
+  version:
+    seq:
+      - id: major
+        -orig-id: majorVersion
+        type: u2
+        doc: |
+          Major version of the ZIM file format (5 or 6)
+          Major version is updated when an incompatible change is integrated in the format (a lib made for a version N will probably not be able to read a version N+1)
+          There are currently 2 major versions :
```
generalmimon, line 125: A tiny typography fix:

```suggestion
          There are currently 2 major versions:
```

media/open_zim.ksy
```diff
@@ -0,0 +1,356 @@
...
+        type: u8
+    instances:
+      entry:
+        value: _parent.url_pointer_list[index].entry
+  str_list:
+    seq:
+      - id: items
+        type: strz
+        repeat: until
+        repeat-until: '_ == ""'
```
generalmimon, line 146: I think this would play better without being enclosed with the single quotes:

```suggestion
        repeat-until: _ == ""
```

media/open_zim.ksy
```diff
@@ -0,0 +1,356 @@
...
+              ptr_size:
+                value: (ptr_is_u8?8:4)
+          blob_ptr_u:
+            seq:
+              - id: value
+                type:
+                  switch-on: _parent.as<cluster>.info.ptr_is_u8 # bug in KSC, as is mandatory
+                  cases:
+                    true: u8
+                    false: u4
```
generalmimon, line 346: What is the bug in KSC? Do you mean that the `value` field here is marked as nullable _(in statically-typed languages like Java, it will have boxed type `Long` instead of primitive `long`)_, despite that _(we humans know)_ the values `true` and `false` cover the full range of possible values of the `ptr_is_u8` (i.e. a `b1` field)?

If so, this has been discussed in https://github.com/kaitai-io/kaitai_struct_formats/pull/303#discussion_r460402466, and the solution is this:

```suggestion
                  switch-on: _parent.as<cluster>.info.ptr_is_u8 # bug in KSC, as is mandatory
                  cases:
                    true: u8
                    _: u4
```

media/open_zim.ksy
```diff
@@ -0,0 +1,356 @@
...
+                type: b3
+              - id: ptr_is_u8
+                -orig-id: extended
+                type: b1
+              - id: compression
+                type: b4
+                enum: compression
+            instances:
+              ptr_size:
+                value: (ptr_is_u8?8:4)
```
generalmimon, line 338: You already know me 😉:

```suggestion
                value: ptr_is_u8 ? 8 : 4
```

media/open_zim.ksy
```diff
@@ -0,0 +1,356 @@
...
+              info:
+                value: _parent.info
+          blob:
+            params:
+              - id: idx
+                type: u8
+            instances:
+              data:
+                pos: _parent.ptr + _parent.blobs_ptrs[idx].as<blob_ptr_u>.value
+                size: _parent.blobs_ptrs[idx+1].as<blob_ptr_u>.value - _parent.blobs_ptrs[idx].as<blob_ptr_u>.value
```
generalmimon, line 316:
```suggestion
                size: _parent.blobs_ptrs[idx + 1].as<blob_ptr_u>.value - _parent.blobs_ptrs[idx].as<blob_ptr_u>.value
```

media/open_zim.ksy
```diff
@@ -0,0 +1,356 @@
...
+            type: blob_ptr_u
+            if: info.compression == compression::none or info.compression == compression::default
+          first_blob_ptr_u_v:
+            value: (info.compression == compression::lzma2?first_blob_ptr_compressed:first_blob_ptr_uncompressed).as<blob_ptr_u>.value
+          blob_count_plus_one:
+            value: first_blob_ptr_u_v / info.ptr_size
+          blob_count:
+            value: blob_count_plus_one - 1
+          blobs_ptrs:
+            value: (info.compression == compression::lzma2?blobs_ptrs_compressed:blobs_ptrs_uncompressed).ptrs
```
generalmimon, line 296:
```suggestion
            value: (info.compression == compression::lzma2 ? blobs_ptrs_compressed:blobs_ptrs_uncompressed).ptrs
```

media/open_zim.ksy
```diff
@@ -0,0 +1,356 @@
...
+      cluster:
+        # BUG: JUNK. Seems to be positioned to a wrong offsets. info.reserved is not 0
+        params:
+          - id: ptr
+            type: u8
+        seq:
+          - id: info
+            type: info_t
+          - id: blobs_ptrs_compressed
+            process: kaitai.compress.lzma(2)
```
generalmimon, line 265: You should state in the top-level `doc` that one needs to install https://github.com/kaitai-io/kaitai_compress so that users don't have to figure out themselves the procedure how to make the parser work.

media/open_zim.ksy
```diff
@@ -0,0 +1,356 @@
...
+            size: 8 + _root.minimal_xz_lzma_size # max(8, 4) + minimal_xz_lzma_size
+            type: blob_ptr_u
+            if: info.compression == compression::lzma2
+          first_blob_ptr_uncompressed:
+            pos: first_blob_offset
+            type: blob_ptr_u
+            if: info.compression == compression::none or info.compression == compression::default
+          first_blob_ptr_u_v:
+            value: (info.compression == compression::lzma2?first_blob_ptr_compressed:first_blob_ptr_uncompressed).as<blob_ptr_u>.value
```
generalmimon, line 290: Beware that `first_blob_ptr_uncompressed` you reference in case of `info.compression != compression::lzma2` is an optional value, which has a meaningful value only `if: info.compression == compression::none or info.compression == compression::default`.

However, {A}: `info.compression != compression::lzma2` is not exactly equivalent to {B}: `info.compression == compression::none or info.compression == compression::default`, so doing `.as<blob_ptr_u>.value` will trigger a `NullPointerException` if `{A} == true and {B} == false`.

The solution is to restrict the [attribute _domain_](https://en.wikipedia.org/wiki/Domain_of_a_function) (respectively the conditions under which the instance has a value) simply using `if`:

```suggestion
          first_blob_ptr_u_v:
            value: (info.compression == compression::lzma2?first_blob_ptr_compressed:first_blob_ptr_uncompressed).as<blob_ptr_u>.value
            if: |
              info.compression == compression::lzma2
              or info.compression == compression::none
              or info.compression == compression::default
```

media/open_zim.ksy
```diff
@@ -0,0 +1,356 @@
...
+            #    enum: compression
+            #  - id: ptr_is_u8
+            #    -orig-id: extended
+            #    type: b1
+            #  - id: reserved
+            #    type: b3
+            seq:
+              - id: reserved
+                type: b3
+              - id: ptr_is_u8
```
generalmimon, line 330: I'd rather pick `is_ptr_u8`, because starting with `is_` prefix is customary for boolean getters (accessors) in programming languages with classes, and also the question sounds better as "Is `ptr` `u8`?".

media/open_zim.ksy
```diff
@@ -0,0 +1,380 @@
...
+              doc: categories, text - see Category Handling
+            0x56: #'V'
+              id: category_article_list
+              doc: categories, article list - see Category Handling
+            0x57: #'W'
+              id: category_list_per_article
+              doc: categories per article, category list - see Category Handling
+            0x58: #'X'
+              id: fulltext_index
+              doc: fulltext index - see ZIM Index Format
```
armijnhemel, line 251: Actually there is also 'Z', which is not documented, but is for 'Xapian indexes':

https://github.com/openzim/libzim/blob/master/src/search.cpp#L267

I have no idea what these do though ;-)

KOLANICH: Thanks for the info!

media/open_zim.ksy
```diff
@@ -0,0 +1,380 @@
...
+          info_t:
+            #seq:
+            #  - id: compression
+            #    type: b4
+            #    enum: compression
+            #  - id: is_ptr_u8
+            #    -orig-id: extended
+            #    type: b1
+            #  - id: reserved
+            #    type: b3
```
generalmimon, line 349: You can use this with [`/meta/bit-endian: le`](https://doc.kaitai.io/user_guide.html#bit-ints-le). This would probably make sense, because the whole format is little-endian as well. Also, the fields in the spec are also listed in the order from the least-significant bit (https://wiki.openzim.org/wiki/ZIM_file_format_old_namespace#Clusters):

> Field Name | Type | Offset | Length | Description
> -- | -- | -- | -- | --
> cluster information | integer | 0 | 1 | **Fourth low bits** : 0: default (no compression), 1: none (inherited from Zeno), 4: LZMA2 compressed<br><br>**Firth bits** : 0: normal (OFFSET_SIZE=4) 1: extended (OFFSET_SIZE=8)

generalmimon (changes requested):

generalmimon:
It seems that there have been made some changes to this format at the beginning of 2021, see https://wiki.openzim.org/wiki/ZIM_file_format:

> # ZIM file format
>
> **Beginning 2021, we change the way we handle namespaces in ZIM file format.**
> 
> This is major change in the way we handle entries in libzim and the semantics around but it is not a break in the binary zim format. Old library/readers are compatible with new zim files.
> 
> This page describe the new format. The old format can be found here : [ZIM file format old namespace](https://wiki.openzim.org/wiki/ZIM_file_format_old_namespace).

Could you check it out and possibly adjust your spec to support the new format too?

media/open_zim.ksy
```diff
@@ -0,0 +1,380 @@
...
+    - libzim
+    - zimlib
+  file-extension: zim
+  xref:
+    wikidata: Q784695
+  license: CC-BY-SA-3.0
+  encoding: utf-8
+  endian: le
+  -requires3rdParty:
+    - https://github.com/kaitai-io/kaitai_compress
```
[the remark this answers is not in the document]
KOLANICH: OK, replaced with `-requires-optional`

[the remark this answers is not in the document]
KOLANICH: >And why do you put this information into a special invented key and don't you just write it into the doc?

Because structured is better than non-structured. We can create a convention which keys mean what, and it is easier to read such docs than a free-form text. And it is easier to process with scripts, when it is needed. So more such keys - the better, if we keep tyem consistent and their names human-understandable.

>This key does not have any potential to become machine-readable anyway, and by attempting to make it so (and not just describing the info in a free-form text field) you only tie your hands yourself because you can't express anything else than the URL 

It doesn't tie my hands: URIs can be distinguished from free-form text automatically.

>(you can't say anything like why this KSY spec depends on it, what happens if you don't install it, some pitfalls, etc.).

It xcan be said in commsnts.

media/open_zim.ksy
```diff
@@ -0,0 +1,380 @@
...
+  id: open_zim
+  title: "(Open) Zeno IMproved"
+  application:
+    - Kiwix
+    - libzim
+    - zimlib
+  file-extension: zim
+  xref:
+    wikidata: Q784695
+  license: CC-BY-SA-3.0
```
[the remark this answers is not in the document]
KOLANICH: The original spec has this license with SA, it may be illegal to change it.

media/open_zim.ksy
```diff
@@ -0,0 +1,380 @@
...
+                type: u4
+                doc: pointer to the directory entry of the redirect target
+        enums:
+          namespace:
+            0x2d: #'-'
+              id: layout
+              doc: eg. the LayoutPage, CSS, favicon.png (48x48), JavaScript and images not related to the articles
+            0x41: #'A'
+              id: article
+              doc: articles - see Article Format
```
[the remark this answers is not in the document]
KOLANICH: Fixed.

media/open_zim.ksy
```diff
@@ -0,0 +1,380 @@
...
+                type: b3
+              - id: is_ptr_u8
+                -orig-id: extended
+                type: b1
+              - id: compression
+                type: b4
+                enum: compression
+            instances:
+              ptr_size:
+                value: "is_ptr_u8 ? 8 : 4"
```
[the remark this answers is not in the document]
KOLANICH: Replaced other quotes in this spec with double ones, now they are more common. BTW `more common` is a very weak argument for changing something, even weaker than `I like that this way`.

media/open_zim.ksy
```diff
@@ -0,0 +1,380 @@
...
+      The MIME type list always follows directly after the header, so the mimeListPos also defines the end and size of the ZIM file header.
+      The MIME types in this list are zero terminated strings.
+  - id: main_page_idx
+    -orig-id: mainPage
+    type: u4
+    doc: main page or 0xffff_ffff if no main page
+  - id: layout_page_idx
+    -orig-id: layoutPage
+    type: u4
+    doc: layout page or 0xff_ffff_ffff (sic!) if no layout page
```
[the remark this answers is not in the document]
KOLANICH: Fixed, I have removed the redundant `doc`s.

media/open_zim.ksy
KOLANICH: Added.

media/open_zim.ksy
KOLANICH: Thanks, fixed.

>You might want to extract such condition to a value instance to avoid repeating it over and over again.

not yet, but maybe.

media/open_zim.ksy
KOLANICH: `-requires-optional`

media/open_zim.ksy
KOLANICH: I guess

```yaml
               - id: compression
                type: b4
                enum: compression
                valid:
                  expr: _ != compression::bzip2 and _ != compression::zlib
```

would do the trick better.

media/open_zim.ksy
KOLANICH: >What is the bug in KSC?

If we remove `.as<cluster>`, we get `don't know how to call method 'info' of object type 'CalcKaitaiStructType'`.

About true/false: fixed.

media/open_zim.ksy
```diff
@@ -0,0 +1,409 @@
...
+                  expr: _ != compression::lzma2 and _ != compression::zstd and _ != compression::bzip2 and _ != compression::zlib
+                doc: LZMA and ZStd are not supported because of behavior of lack of support of streaming processors in KS. bzip2 and zlib are not supported because they are not supported in the official lib (though in principle they probably can be supported).
+            instances:
+              ptr_size:
+                value: "is_ptr_u8 ? 8 : 4"
+          blob_ptr_u:
+            seq:
+              - id: value
+                type:
+                  -affected-by: 561
```
generalmimon, line 394: When I see the [diff showing changes](https://github.com/kaitai-io/kaitai_struct_formats/compare/32d136bbb99d2a9be4c28632c2cec0bc3a764e4d..dc857b989220be64efcd9e4e853a4dbe0f13f6d8) done in the last commit https://github.com/kaitai-io/kaitai_struct_formats/commit/dc857b989220be64efcd9e4e853a4dbe0f13f6d8:

```diff
-                   -affected-by: don't know how to call method 'info' of object type 'CalcKaitaiStructType'
+                   -affected-by: 561
```

how is https://github.com/kaitai-io/kaitai_struct/issues/561 related to the error message that you quoted there earlier?

KOLANICH: Seems to be a manifestation of the same issue - defective or unimplemented machinery for properly resolving types relative to `_parent`.

media/open_zim.ksy
KOLANICH: I prefer to keep it listed in `be` order because it is the convention KS uses for everything. I perceive `bit-endian` as a temporary solution, that IMHO was a mistake to implement that feature that way (because it causes issues for switcheable endianness). IMHO what you call `bit endianness` should be related to just endianness and to the fact of presence of bit-sized types crossing byte boundaries. I guess eventually it would be dropped and replaced to something more uniform, rather than make spec devs to code different types for be and le.

media/open_zim.ksy
```diff
@@ -0,0 +1,409 @@
...
+          namespace:
+            0x2D: # "-"
+              id: layout
+              doc: eg. the LayoutPage, CSS, favicon.png (48x48), JavaScript and images not related to the articles
+            0x41: # "A"
+              id: article
+              doc: articles - see Article Format
+            0x42: # "B"
+              id: article_meta_data
+              doc: article meta data - see Article Format
```
armijnhemel, line 232: It seems that the `C` namespace is missing. The documentation says:

```
All content put in a zim archive must be put in the C namespace. 
```

Most new Zim files that I see put all the content in the `C` namespace.

armijnhemel, line 232: There is also a `H` namespace, which I found in a test file (`nomadic-changthang` which can be found on https://wiki.kiwix.org/wiki/Content_in_all_languages ) but I have no idea what that namespace actually means or represents.

media/open_zim.ksy
generalmimon: > If we remove `.as<cluster>`, we get `don't know how to call method 'info' of object type 'CalcKaitaiStructType'`.

This is not a bug, it is working as intended. See https://doc.kaitai.io/user_guide.html#_enforcing_parent_type - type `blob_ptr_u` may have one of multiple parents (`/types/cluster_ptr/types/cluster` and <code>/types/cluster_ptr/types/cluster/types/**blobs_ptrs**</code>), so they only have the `KaitaiStruct` interface in common (<code>media/open_zim.ksy:{[271-272](https://github.com/kaitai-io/kaitai_struct_formats/blob/208439638843d9ca4e3f656e0964deb173259079/media/open_zim.ksy#L271-L272),[295](https://github.com/kaitai-io/kaitai_struct_formats/blob/208439638843d9ca4e3f656e0964deb173259079/media/open_zim.ksy#L295),[316-319](https://github.com/kaitai-io/kaitai_struct_formats/blob/208439638843d9ca4e3f656e0964deb173259079/media/open_zim.ksy#L316-L319),[359-370](https://github.com/kaitai-io/kaitai_struct_formats/blob/208439638843d9ca4e3f656e0964deb173259079/media/open_zim.ksy#L359-L370)}</code>):

```ksy
    types:
      cluster:
        # ...
        instances:
          # ...
          first_blob_ptr_uncompressed:
            pos: first_blob_offset
            type: blob_ptr_u
            if: info.compression == compression::none or info.compression == compression::default
          # ...
        types:
          blobs_ptrs:
            params:
              - id: count
                type: u8
            seq:
              - id: ptrs
                type: blob_ptr_u
                repeat: eos
```

However, we may definitely improve the error message and make it more useful - in this case, list the types that use `blob_ptr_u` and explain that since it it does not have only one possible parent type, specific attributes of `_parent` are not accessible.

PR #344, opened 2020-09-25
Parse Microsot PE resource table

bvanheu:
This is an attempt to parse the `resource_table`. Heavy use of `instances`, maybe there's a better way to parse this section?

### DIFF
--- a/executable/microsoft_pe.ksy
+++ b/executable/microsoft_pe.ksy
@@ -355,6 +355,11 @@ types:
       body:
         pos: pointer_to_raw_data
         size: size_of_raw_data
+      resource_table:
+        pos: pointer_to_raw_data
+        if: virtual_address == _root.pe.optional_hdr.data_dirs.resource_table.virtual_address
+        size: size_of_raw_data
+        type: resource_directory_table(0, pointer_to_raw_data, virtual_address)
   certificate_table:
     seq:
       - id: items
@@ -408,3 +413,144 @@ types:
         size: length - 8
         doc: Contains a certificate, such as an Authenticode signature.
     doc-ref: 'https://docs.microsoft.com/en-us/windows/desktop/debug/pe-format#the-attribute-certificate-table-image-only'
+
+  resource_directory_table:
+    params:
+      - id: depth
+        type: u4
+      - id: rsrc_section_file_offset
+        type: u4
+      - id: rsrc_section_virtual_address
+        type: u4
+    seq:
+      - id: characteristics
+        type: u4
+      - id: time_date_stamp
+        type: u4
+      - id: major_version
+        type: u2
+      - id: minor_version
+        type: u2
+      - id: number_of_named_entries
+        type: u2
+      - id: number_of_id_entries
+        type: u2
+      - id: named_entries
+        type: resource_directory_entry(depth)
+        repeat: expr
+        repeat-expr: number_of_named_entries
+      - id: id_entries
+        type: resource_directory_entry(depth)
+        repeat: expr
+        repeat-expr: number_of_id_entries
+
+  resource_directory_entry:
+    params:
+      - id: depth
+        type: u4
+    seq:
+      - id: resource_type
+        type: u4
+        enum: enum_resource_type
+        if: depth == 0
+      - id: name_offset
+        type: u4
+        if: depth != 0
+      - id: data_entry_offset
+        type: u4
+    instances:
+      child:
+        pos: (data_entry_offset & ~0x80000000) + _parent.rsrc_section_file_offset
+        type: resource_directory_table(depth + 1, _parent.rsrc_section_file_offset, _parent.rsrc_section_virtual_address)
+        if: (data_entry_offset & 0x80000000) >> 31 == 1
+        parent: _parent
+        io: _root._io
+      resource_data_entry:
+        pos: data_entry_offset + _parent.rsrc_section_file_offset
+        type: resource_data_entry
+        if: (data_entry_offset & 0x80000000) >> 31 == 0
+        parent: _parent
+        io: _root._io
+      resource_directory_string:
+        pos: (name_offset & ~0x80000000) + _parent.rsrc_section_file_offset
+        type: resource_directory_string
+        if: (name_offset & 0x80000000) >> 31 == 1
+        parent: _parent
+        io: _root._io
+    enums:
+      enum_resource_type:
+          1: rt_cursor
+          2: rt_bitmap
+          3: rt_icon
+          4: rt_menu
+          5: rt_dialog
+          6: rt_string
+          7: rt_fontdir
+          8: rt_font
+          9: rt_accelerator
+          10: rt_rcdata
+          11: rt_messagetable
+          12: rt_group_cursor
+          14: rt_group_icon
+          16: rt_version
+          17: rt_dlginclude
+          19: rt_plugplay
+          20: rt_vxd
+          21: rt_anicursor
+          22: rt_aniicon
+          23: rt_html
+          24: rt_manifest
+
+  resource_type:
+    seq:
+      - id: name_offset
+        type: u4
+        enum: enum_resource_type
+    enums:
+      enum_resource_type:
+          1: rt_cursor
+          2: rt_bitmap
+          3: rt_icon
+          4: rt_menu
+          5: rt_dialog
+          6: rt_string
+          7: rt_fontdir
+          8: rt_font
+          9: rt_accelerator
+          10: rt_rcdata
+          11: rt_messagetable
+          12: rt_group_cursor
+          14: rt_group_icon
+          16: rt_version
+          17: rt_dlginclude
+          19: rt_plugplay
+          20: rt_vxd
+          21: rt_anicursor
+          22: rt_aniicon
+          23: rt_html
+          24: rt_manifest
+
+  resource_data_entry:
+    seq:
+      - id: data_rva
+        type: u4
+      - id: size
+        type: u4
+      - id: codepage
+        type: u4
+      - id: reserved
+        type: u4
+    instances:
+      body:
+        io: _root._io
+        pos: (data_rva - _parent.rsrc_section_virtual_address) + _parent.rsrc_section_file_offset
+        size: size
+
+  resource_directory_string:
+    seq:
+      - id: length
+        type: u2
+      - id: string
+        type: str
+        encoding: UTF-16LE
+        size: length * 2


executable/microsoft_pe.ksy
```diff
@@ -408,3 +413,144 @@ types:
         size: length - 8
         doc: Contains a certificate, such as an Authenticode signature.
     doc-ref: 'https://docs.microsoft.com/en-us/windows/desktop/debug/pe-format#the-attribute-certificate-table-image-only'
+
+  resource_directory_table:
+    params:
+      - id: depth
+        type: u4
+      - id: rsrc_section_file_offset
```
KOLANICH, line 421: IMHO: `rsrc_` is not needed

executable/microsoft_pe.ksy
```diff
@@ -408,3 +413,144 @@ types:
...
+        type: u4
+      - id: rsrc_section_virtual_address
+        type: u4
+    seq:
+      - id: characteristics
+        type: u4
```
KOLANICH, line 427: aren't they bit flags? Shouldn't they be a bit-sized type?

executable/microsoft_pe.ksy
```diff
@@ -408,3 +413,144 @@ types:
...
+      - id: time_date_stamp
+        type: u4
+      - id: major_version
```
KOLANICH, line 430: I'd like to move version into an own type

executable/microsoft_pe.ksy
```diff
@@ -408,3 +413,144 @@ types:
...
+        type: u2
+      - id: minor_version
+        type: u2
+      - id: number_of_named_entries
```
KOLANICH, line 434: code style prescribes `num_` prefix, but IMHO the word `number` is too generic for the purpose, in own specs I use `count`.

executable/microsoft_pe.ksy
```diff
@@ -408,3 +413,144 @@ types:
...
+        type: u4
+        enum: enum_resource_type
+        if: depth == 0
+      - id: name_offset
+        type: u4
+        if: depth != 0
+      - id: data_entry_offset
+        type: u4
+    instances:
+      child:
```
KOLANICH, line 462: Haven't verified the computations, but :+1: if they are correct!

executable/microsoft_pe.ksy
```diff
@@ -408,3 +413,144 @@ types:
...
+        io: _root._io
+      resource_directory_string:
+        pos: (name_offset & ~0x80000000) + _parent.rsrc_section_file_offset
+        type: resource_directory_string
+        if: (name_offset & 0x80000000) >> 31 == 1
+        parent: _parent
+        io: _root._io
+    enums:
+      enum_resource_type:
+          1: rt_cursor
```
KOLANICH, line 482: IMHO: `rt_` is unneeded.

executable/microsoft_pe.ksy
```diff
@@ -408,3 +413,144 @@ types:
...
+          23: rt_html
+          24: rt_manifest
+
+  resource_type:
+    seq:
+      - id: name_offset
+        type: u4
+        enum: enum_resource_type
+    enums:
+      enum_resource_type:
```
KOLANICH, line 510: Why is it duplicated?

executable/microsoft_pe.ksy
bvanheu: Agreed, I've removed all reference to `rsrc`.

executable/microsoft_pe.ksy
bvanheu: They're not currently used and reserved for future use. I've added a comment to specify this.

executable/microsoft_pe.ksy
bvanheu: I agree with your statement although I think it's preferable to stick to the style guide. I've used `num_` instead of `number_of_`.

executable/microsoft_pe.ksy
bvanheu: Agreed, removed all `rt_` from the enum.

executable/microsoft_pe.ksy
bvanheu: I was testing different things and it slipped through! I've removed the duplicate enum.

executable/microsoft_pe.ksy
bvanheu: Hopefully! Thanks for the review!

executable/microsoft_pe.ksy
bvanheu: I was thinking maybe it would become a bit too verbose? The rest of the file doesn't use specific type for versions in structure. Maybe keeping this as-is to be consistent with the rest of the file?

KOLANICH: IMHO: code self-documentation and semantic structuring is almost always beneficial.

bvanheu: aight, I've created `resource_directory_table_version`, please take a look

executable/microsoft_pe.ksy
```diff
@@ -355,6 +355,11 @@ types:
       body:
         pos: pointer_to_raw_data
         size: size_of_raw_data
+      resource_table:
```
bvanheu, line 358: Maybe it would have been better is to put these fields under `data_dirs.resource_table`, although I wasn't sure how to create a dependency on parsing each sections *first* then adding the fields, e.g. something like:
```
  data_dir:                                                                     
    seq:                                                                        
      - id: virtual_address                                                     
        type: u4                                                                
      - id: size                                                                
        type: u4                                                                
  instances:
    resource_table:
      type: resource_directory_table(0, _root.pe.sections[x].pointer_to_raw_data, _root.pe.sections[x].virtual_address)
      if: virtual_address == _root.pe.sections[x].virtual_address
```
where `x` would be an implicit iterator.

executable/microsoft_pe.ksy
```diff
@@ -429,10 +429,8 @@ types:
       - id: time_date_stamp
         type: u4
         -orig-id: 'TimeDateStamp'
-      - id: major_version
-        type: u2
-      - id: minor_version
-        type: u2
+      - id: version
+        type: resource_directory_table_version
```
KOLANICH, line 433: why not to use just `version_u2` and replace everywhere we find it?

bvanheu: Sure, done. I prefer to do the replacing elsewhere in the file in another PR.

executable/microsoft_pe.ksy
generalmimon: I recommend renaming the field to `reserved` as per the [style guide](https://doc.kaitai.io/ksy_style_guide.html#attr-id). The name `characteristics` really suggests that it's something meaningful.

executable/microsoft_pe.ksy
```diff
@@ -408,3 +413,134 @@ types:
...
+        type: u4
+      - id: section_virtual_address
+        type: u4
+    seq:
+      # This field is reserved for future use. It is currently set to zero.
+      - id: characteristics
+        type: u4
+      - id: time_date_stamp
+        type: u4
+        -orig-id: 'TimeDateStamp'
```
generalmimon, line 431: Nit: if nothing else is enclosed in quotes, why should be this one?

```suggestion
        -orig-id: TimeDateStamp
```

executable/microsoft_pe.ksy
```diff
@@ -408,3 +413,134 @@ types:
...
+        -orig-id: NumberOfIdEntries
+      - id: named_entries
+        type: resource_directory_entry(depth)
+        repeat: expr
+        repeat-expr: num_named_entries
+      - id: id_entries
+        type: resource_directory_entry(depth)
+        repeat: expr
+        repeat-expr: num_id_entries
+    doc-ref: 'https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#resource-directory-table'
```
generalmimon, line 448: Please move this `doc-ref` key on the top of the type definition as per the [style guide](https://doc.kaitai.io/ksy_style_guide.html#type).

Putting the `doc` and `doc-ref` below `seq`  or `instances` makes the KSY confused most of the time, see https://github.com/kaitai-io/kaitai_struct_formats/pull/303#discussion_r454140325 for a more extreme example.

executable/microsoft_pe.ksy
```diff
@@ -408,3 +413,134 @@ types:
...
+        type: u4
+      - id: section_file_offset
+        type: u4
+      - id: section_virtual_address
+        type: u4
+    seq:
+      # This field is reserved for future use. It is currently set to zero.
+      - id: characteristics
+        type: u4
+      - id: time_date_stamp
```
generalmimon, line 429: Maybe just established name `timestamp`? `time_date_stamp` seems bloated and weird to me.

It's not necessary at all to strictly follow the original naming from the original spec.

```suggestion
      - id: timestamp
```

Also, what timestamp format does it use?

executable/microsoft_pe.ksy
bvanheu: Done

executable/microsoft_pe.ksy
bvanheu: Agreed

executable/microsoft_pe.ksy
bvanheu: Done, i've also changed a couple of other places in this PR

executable/microsoft_pe.ksy
KOLANICH: Instances are parsed lazily. So dependencies are tracked: on each access go an insfance it is checked if it js initialized, and if not, it is tried to be initialized.

bvanheu: The issue I'm facing with this solution would be to iterate over the `_root.pe.sections`, find which one has the same virtual-address in order to create the instance.

In imperative pseudo-code:
```
instance = null;
for section in _root.pe.sections:
  if virtual_address == section.virtual_address:
    instance = resource_directory_table(0, section.pointer_to_raw_data, section.virtual_address))
return instance
```

I was thinking of something like this:
```
      type: resource_directory_table(0, _root.pe.sections[x].pointer_to_raw_data, _root.pe.sections[x].virtual_address)
      if: virtual_address == _root.pe.sections[x].virtual_address
```
where `x` would be an implicit iterator.

Otherwise the instance has to be put into `sections` and will be created once parsing each sections will result in the condition only if the virtual address being true.

KOLANICH: Stateful parsing is not supported out of the box, but if you can reformulate the problem as 2 stages, 1 is iterating using `repeat: until`, and the second is taking its last element, it is possible to do something. You may want to look at https://github.com/kaitai-io/kaitai_struct_formats/pull/265/files#diff-450d8347f68fb61616750e47db6a9f4bR315

executable/microsoft_pe.ksy
```diff
@@ -408,3 +413,134 @@ types:
...
+    seq:
+      - id: resource_type
+        type: u4
+        enum: enum_resource_type
+        if: depth == 0
+      - id: ofs_name
+        type: u4
+        if: depth != 0
+      - id: ofs_data_entry
+        type: u4
```
generalmimon, line 465: I think splitting these `u4` fields to `b1/b31` [bit-sized integers](https://doc.kaitai.io/user_guide.html#_bit_sized_integers) pairs would be a better idea than having this `& ~0x80000000` stuff everywhere.

```suggestion
      - id: is_name
        type: b1
        if: depth != 0
      - id: ofs_name
        type: b31
        if: depth != 0
      - id: is_data_entry
        type: b1
      - id: ofs_data_entry
        type: b31
```

bvanheu: Good idea, thank you!

executable/microsoft_pe.ksy
```diff
@@ -408,3 +413,144 @@ types:
...
+        type: b31le
+        if: depth != 0
+      - id: is_name
+        type: b1be
+        if: depth != 0
+        doc: Whether ofs_name is an id or points to a directory-string
+      - id: ofs_data_entry
+        type: b31le
+      - id: has_child
+        type: b1be
```
generalmimon, line 471: LE and BE bit-sized fields cannot follow on an unaligned bit position. Right now, it has **undefined behavior**, and KSC will respond to it with a compile-time error in the future. See https://doc.kaitai.io/user_guide.html#bit-endian, the _Important_ section:

> _Big-endian_ and _little-endian_ bit integers can follow **only on a byte boundary**. They can’t share the same byte. Joining them on an unaligned bit position is _undefined behavior_, and KSC will be throwing a compile-time error in the future if it detects such situation.
>
> For example, this is **illegal**: (...)

```suggestion
      - id: ofs_data_entry
        type: b31le
      - id: has_child
        type: b1be
```

---

I suggest placing 

```ksy
meta:
  bit-endian: le
```

on the very top of the file, probably below `endian` , because that's not gonna change throughout the file. Then simply omit the `le` suffixes after each bit type:

```suggestion
      - id: ofs_data_entry
        type: b31
      - id: has_child
        type: b1
```

etc.

bvanheu: Thanks for the thorough review and explanation.

I admit I was a bit confused how the bit-endianness worked. I somehow expected `b1le` to actually returns `0` on `80` (i.e. take the 4th byte, interpret the bit in little-endian, returns the first bit which is 0 in the least significant bit).

executable/microsoft_pe.ksy
```diff
@@ -16,6 +16,7 @@ meta:
   license: CC0-1.0
   ks-version: 0.7
```
generalmimon, line 17: Also please change this to

```ksy
  ks-version: 0.9
```

... because little-endian bit-sized integers is a 0.9 feature. See https://doc.kaitai.io/user_guide.html#bit-ints-le:

> ### 4.14.2. Little-endian order
> 
> <table>
> <tr>
> <th>IMPORTANT</th>
> <td>Feature available since v0.9.</td>
> </tr>
> </table>

bvanheu: Done

executable/microsoft_pe.ksy
```diff
@@ -408,3 +414,144 @@ types:
...
+    doc-ref: 'https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#resource-directory-entries'
+    params:
+      - id: depth
+        type: u4
+    seq:
+      - id: resource_type
+        type: u4
+        enum: enum_resource_type
+        if: depth == 0
+      - id: ofs_name
```
bvanheu, line 462: How do you suggest naming this field that can be either and `id` or an offset to a string (which is a `name`), depending on the most-significant bit (`is_name`)?

generalmimon: The best thing to do would be to have two separate fields switched using `if`:

```ksy
  - id: ofs_name
    type: b31
    if: ... 
  - id: some_id
    type: not ... 
```

But that's not possible of course, because the `is_name` comes after these fields.

Regarding the name, I'd

PR #349, opened 2020-10-16
Add Android super.img (Dynamic Partitions) format

tchebb:
The metadata stored by Android at the beginning of a "super" partition, which is what it calls a disk partition that holds one or more Dynamic Partitions. Dynamic Partitions do more or less the same thing that LVM does on Linux, allowing Android to map ranges of non-contiguous extents to a single logical device. This metadata holds that mapping.

[sample.zip](https://github.com/kaitai-io/kaitai_struct_formats/files/5390627/sample.zip) contains a sample file (`super_truncated.img`) that I obtained by extracting only the header portion of a "super" partition provided as part of an official Android emulator image.

### DIFF
--- a/filesystem/android_super.ksy
+++ b/filesystem/android_super.ksy
@@ -1,0 +1,207 @@
+meta:
+  id: android_super
+  title: Android Dynamic Partitions metadata
+  application: Android
+  file-extension: img
+  license: CC0-1.0
+  ks-version: 0.9
+  bit-endian: le
+  endian: le
+
+doc: |
+  The metadata stored by Android at the beginning of a "super" partition, which
+  is what it calls a disk partition that holds one or more Dynamic Partitions.
+  Dynamic Partitions do more or less the same thing that LVM does on Linux,
+  allowing Android to map ranges of non-contiguous extents to a single logical
+  device. This metadata holds that mapping.
+
+doc-ref:
+  - https://source.android.com/devices/tech/ota/dynamic_partitions
+  - https://android.googlesource.com/platform/system/core/+/refs/tags/android-11.0.0_r8/fs_mgr/liblp/include/liblp/metadata_format.h
+
+instances:
+  root:
+    pos: 0x1000
+    type: root
+
+types:
+  root:
+    seq:
+      - id: primary_geometry
+        size: 0x1000
+        type: geometry
+
+      - id: backup_geometry
+        type: geometry
+        size: 0x1000
+
+      - id: primary_metadata
+        size: primary_geometry.metadata_max_size
+        type: metadata
+        repeat: expr
+        repeat-expr: primary_geometry.metadata_slot_count
+
+      - id: backup_metadata
+        # These attributes are intentionally taken from primary_geometry, even
+        # for backup_metadata. The first non-corrupt geometry specifier dictates
+        # the layout of both primary and backup metadata.
+        size: primary_geometry.metadata_max_size
+        type: metadata
+        repeat: expr
+        repeat-expr: primary_geometry.metadata_slot_count
+
+  geometry:
+    seq:
+      - id: magic
+        contents: 'gDla'
+      - id: struct_size
+        type: u4
+      - id: checksum
+        size: 32
+        doc: |
+          SHA-256 hash of struct_size bytes from beginning of geometry,
+          calculated as if checksum were zeroed out
+      - id: metadata_max_size
+        type: u4
+      - id: metadata_slot_count
+        type: u4
+      - id: logical_block_size
+        type: u4
+
+  metadata:
+    seq:
+      - id: magic
+        contents: '0PLA'
+      - id: major_version
+        type: u2
+      - id: minor_version
+        type: u2
+      - id: header_size
+        type: u4
+      - id: header_checksum
+        size: 32
+        doc: |
+          SHA-256 hash of header_size bytes from beginning of metadata,
+          calculated as if header_checksum were zeroed out
+      - id: tables_size
+        type: u4
+      - id: tables_checksum
+        size: 32
+        doc: SHA-256 hash of tables_size bytes from end of header
+      - id: partitions
+        type: table_descriptor(table_kind::partitions)
+      - id: extents
+        type: table_descriptor(table_kind::extents)
+      - id: groups
+        type: table_descriptor(table_kind::groups)
+      - id: block_devices
+        type: table_descriptor(table_kind::block_devices)
+
+    enums:
+      table_kind:
+        0: partitions
+        1: extents
+        2: groups
+        3: block_devices
+
+    types:
+      table_descriptor:
+        params:
+          - id: kind
+            type: u1
+            enum: table_kind
+            -affected-by: 135
+
+        seq:
+          - id: offset
+            type: u4
+          - id: num_entries
+            type: u4
+          - id: entry_size
+            type: u4
+
+        instances:
+          table:
+            pos: _parent.header_size + offset
+            size: entry_size
+            type:
+              switch-on: kind
+              cases:
+                'table_kind::partitions': partition
+                'table_kind::extents': extent
+                'table_kind::groups': group
+                'table_kind::block_devices': block_device
+            repeat: expr
+            repeat-expr: num_entries
+
+      partition:
+        seq:
+          - id: name
+            size: 36
+            type: strz
+            encoding: UTF-8
+          - id: attr_readonly
+            type: b1
+          - id: attr_slot_suffixed
+            type: b1
+          - id: attr_updated
+            type: b1
+          - id: attr_disabled
+            type: b1
+          - id: attrs_reserved
+            type: b28
+          - id: first_extent_index
+            type: u4
+          - id: num_extents
+            type: u4
+          - id: group_index
+            type: u4
+
+      extent:
+        seq:
+          - id: num_sectors
+            type: u8
+          - id: target_type
+            type: u4
+            enum: target_type
+          - id: target_data
+            type: u8
+          - id: target_source
+            type: u4
+
+        enums:
+          target_type:
+            0: linear
+            1: zero
+
+      group:
+        seq:
+          - id: name
+            size: 36
+            type: strz
+            encoding: UTF-8
+          - id: flag_slot_suffixed
+            type: b1
+          - id: flags_reserved
+            type: b31
+          - id: maximum_size
+            type: u8
+
+      block_device:
+        seq:
+          - id: first_logical_sector
+            type: u8
+          - id: alignment
+            type: u4
+          - id: alignment_offset
+            type: u4
+          - id: size
+            type: u8
+          - id: partition_name
+            size: 36
+            type: strz
+            encoding: UTF-8
+          - id: flag_slot_suffixed
+            type: b1
+          - id: flags_reserved
+            type: b31


filesystem/android_super.ksy
```diff
@@ -0,0 +1,174 @@
...
+
+doc: |
+  The metadata stored by Android at the beginning of a "super" partition, which
+  is what it calls a disk partition that holds one or more Dynamic Partitions.
+  Dynamic Partitions do more or less the same thing that LVM does on Linux,
+  allowing Android to map ranges of non-contiguous extents to a single logical
+  device. This metadata holds that mapping.
+
+doc-ref: https://source.android.com/devices/tech/ota/dynamic_partitions/implement
+doc-ref: https://android.googlesource.com/platform/system/core/+/master/fs_mgr/liblp/include/liblp/metadata_format.h
```
generalmimon, line 18: This is actually invalid YAML, duplicate keys aren't allowed. Some parsers allow them, but the latter `doc-ref` will silently overwrite the first one in that case, it won't become an array.

The proper solution is to put these refs to a sequence, which is allowed since 0.9 (see https://github.com/kaitai-io/kaitai_struct/issues/269): 

```suggestion
doc-ref
 - https://source.android.com/devices/tech/ota/dynamic_partitions/implement
 - https://android.googlesource.com/platform/system/core/+/master/fs_mgr/liblp/include/liblp/metadata_format.h
```

And please also increment the top-level `ks-version` to 0.9.

filesystem/android_super.ksy
```diff
@@ -0,0 +1,174 @@
+meta:
+  id: android_super
+  title: Android Dynamic Partitions metadata
+  application: Android
+  file-extension: img
+  license: CC0-1.0
+  ks-version: 0.8
```
generalmimon, line 7:
```suggestion
  ks-version: 0.9
```

After you convert the duplicate `doc-ref` keys to array, see above.

generalmimon (changes requested):

filesystem/android_super.ksy
```diff
@@ -0,0 +1,174 @@
...
+      - id: groups
+        type: table_descriptor(2)
+      - id: block_devices
+        type: table_descriptor(3)
+
+    types:
+      table_descriptor:
+        params:
+          - id: entry_type
+            type: u8
```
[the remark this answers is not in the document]
tchebb: So these values aren't actually part of the specification at all—they never appear in the serialized data anywhere. I've just created them so that I can parameterize the type that's held inside each `table_descriptor` instance. Given that, do you still think making an enum makes sense? Or is there a way to parameterize `table_descriptor` directly by a type so that I don't need to use numbers at all?

tchebb:
Thanks for the review, and congrats on the 0.9 release! I've taken most of your suggestions and also added some shiny new little-endian bit fields to parse various flag fields.

filesystem/android_super.ksy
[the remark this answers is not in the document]
generalmimon: > I've just created them so that I can parameterize the type that's held inside each `table_descriptor` instance. Given that, do you still think making an enum makes sense?

Yes, I think so. Defining enums purely for the internal purpose of the KSY implementation is totally OK, and it's commonly used in specs. For example:

https://github.com/kaitai-io/kaitai_struct_formats/blob/ad875ff48d3ec43ae4ddbcf77b98cefe6d3edb7c/macos/resource_compression/dcmp_0.ksy#L85-L94

It's definitely better than repeating plain magic integers everywhere without any obvious meaning.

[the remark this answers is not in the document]
generalmimon: > Or is there a way to parameterize `table_descriptor` directly by a type so that I don't need to use numbers at all?

Unfortunately not yet. This feature idea is tracked in issue https://github.com/kaitai-io/kaitai_struct/issues/135.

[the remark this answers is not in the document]
tchebb: Cool, makes sense. Should be all set now :)

filesystem/android_super.ksy
```diff
@@ -0,0 +1,196 @@
...
+  dummy: {}
+
+  geometry:
+    seq:
+      - id: magic
+        contents: 'gDla'
+      - id: struct_size
+        type: u4
+      - id: checksum
+        size: 32
```
generalmimon, line 57: Is the checksum format known? Is it an output from [known hash function](https://en.wikipedia.org/wiki/List_of_hash_functions), or something custom?

tchebb: It's SHA-256. The header checksum is computed with a checksum field of all zeroes. Want me to document that inline? The second top-level `doc-ref` URL mentions it.

filesystem/android_super.ksy
```diff
@@ -0,0 +1,196 @@
...
+          table:
+            pos: _parent.header_size + offset
+            size: entry_size
+            type:
+              switch-on: kind
+              cases:
+                'table_kind::partitions': partition
+                'table_kind::extents': extent
+                'table_kind::groups': group
+                'table_kind::blocks': block_device
```
generalmimon, line 122: Phew, it took me a while to debug this problem in the [Web IDE](https://ide.kaitai.io/), but apparently the compiler doesn't verify if the enum key exists. And `table_kind::blocks` enum value doesn't exist, it's named `block_devices` instead:

```suggestion
                'table_kind::block_devices': block_device
```

filesystem/android_super.ksy
generalmimon: > It's SHA-256. The header checksum is computed as if the checksum field were all zeroes. Want me to document that inline? The second top-level `doc-ref` URL mentions it.

Definitely a good idea, put this description to the `doc` key of this attribute, please.

filesystem/android_super.ksy
tchebb: Oops, my bad. Good catch!

filesystem/android_super.ksy
tchebb: Updated.

filesystem/android_super.ksy
generalmimon: > Oops, my bad. Good catch!

In fact, not your bad at all 😐 The point is that the compiler should to catch these things, in the same way as when you reference a non-existing type or field.

filesystem/android_super.ksy
```diff
@@ -0,0 +1,200 @@
...
+    enums:
+      table_kind:
+        0: partitions
+        1: extents
+        2: groups
+        3: block_devices
+
+    types:
+      table_descriptor:
+        params:
```
KOLANICH, line 103:
```yaml
-affected-by: 135
```

KOLANICH (approved):

filesystem/android_super.ksy
```diff
@@ -0,0 +1,200 @@
...
+  device. This metadata holds that mapping.
+
+doc-ref:
+  - https://source.android.com/devices/tech/ota/dynamic_partitions/implement
+  - https://android.googlesource.com/platform/system/core/+/master/fs_mgr/liblp/include/liblp/metadata_format.h
+
+seq:
+  - id: padding
+    size: 0x1000
+    type: dummy
```
KOLANICH, line 25: Is it needed?

tchebb: As I understand it, using a dummy type here creates a substream that's never consumed, so we skip over the bytes and never read them. If I omitted a type and just made it a byte array, we'd be reading and loading into memory the entire 4KB. It also adds a highlight in the Web IDE for all those 4KB, which I don't like since they're totally irrelevant and ignored.

KOLANICH: No. Substreams currently are read as bytes, then for each substream a KaitaiStream object created from bytes. So you should probably move everything except the padding into an instance.

tchebb: Updated to use instance.

tchebb:
@generalmimon just in case you missed my last push, this should be ready to review again now with all comments addressed. No rush, though!

filesystem/android_super.ksy
```diff
@@ -0,0 +1,200 @@
...
+
+doc: |
+  The metadata stored by Android at the beginning of a "super" partition, which
+  is what it calls a disk partition that holds one or more Dynamic Partitions.
+  Dynamic Partitions do more or less the same thing that LVM does on Linux,
+  allowing Android to map ranges of non-contiguous extents to a single logical
+  device. This metadata holds that mapping.
+
+doc-ref:
+  - https://source.android.com/devices/tech/ota/dynamic_partitions/implement
```
generalmimon, line 19: I'm missing here how this link relates to the format you're specifying?

filesystem/android_super.ksy
generalmimon: I don't see how you resolved this, @tchebb? The point is to provide a reference to an issue in the https://github.com/kaitai-io/kaitai_struct repository from the KSY that's blocking or impairing some feature of the particular KSY spec implementation. It's something like `TODO` or `FIXME when feature ... is available in KS language` comments in the programming world.

It's definitely a good idea, because it's true that https://github.com/kaitai-io/kaitai_struct/issues/135 would help here. I'm not sure though how it should be done: maybe a simple `doc` key would be better than this not-much-established `-affected-by` key _(I searched GitHub with term [`extension:ksy affected-by OR -affected-by`](https://github.com/search?q=extension%3Aksy+affected-by+OR+-affected-by) and found nothing, not sure, but e.g. [`extension:ksy -orig-id OR mac-os-resource-type`](https://github.com/search?q=extension%3Aksy+-orig-id+OR+mac-os-resource-type) works, so the former probably works as well)_. It's likely that readers of the spec wouldn't understand this key, and it's naturally not documented anywhere.

Instead, I would suggest something like this, which basically says `-affected-by: 188`:

https://github.com/kaitai-io/kaitai_struct_formats/blob/53795d043f170e2f2654b5a5774a48fadb1d349e/image/bmp.ksy#L104-L109

generalmimon (changes requested):

filesystem/android_super.ksy
KOLANICH: >It's likely that readers of the spec wouldn't understand this key, and it's naturally not documented anywhere.

It is likely that it is currently only used in my specs, and my specs using it are not currently merged. Using `doc` for it is IMHO consumes too much space. The original purpose of `-affected-by` was for search and automation. I mean some bot that scans the repo on issues being closed and mentioning all the specs that should be fixed.

generalmimon: > The original purpose of `-affected-by` was for search and automation. I mean some bot that scans the repo on issues being closed and mentioning all the specs that should be fixed.

That's a nice idea, thanks!

KOLANICH: IMHO since issues in kaitai-io/kaitai_struct are usually meant, for that repo it makes sense to use just an issue number.

tchebb: I'm not sure where you want me to put this line. Isn't the compiler going to complain about an unknown key if I add it right here?

filesystem/android_super.ksy
tchebb: This is intended to just let people know what the format is used for. I'll change it to the overview page for Dynamic Partitions rather than this specific implementation page.

filesystem/android_super.ksy
```diff
@@ -0,0 +1,200 @@
...
+        size: primary_geometry.metadata_max_size
+        type: metadata
+        repeat: expr
+        repeat-expr: primary_geometry.metadata_slot_count
+
+      - id: backup_metadata
+        size: primary_geometry.metadata_max_size
+        type: metadata
+        repeat: expr
+        repeat-expr: primary_geometry.metadata_slot_count
```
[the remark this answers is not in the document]
tchebb: No. As I understand it, the primary geometry is used to define the entire layout of everything following unless it is corrupt, in which case the backup geometry is used for everything. I'm assuming that nothing is corrupt in the file we're parsing, which means we ignore the values in the backup geometry.

filesystem/android_super.ksy
```diff
@@ -0,0 +1,200 @@
...
+
+  geometry:
+    seq:
+      - id: magic
+        contents: 'gDla'
+      - id: struct_size
+        type: u4
```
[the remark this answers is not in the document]
tchebb: I'm using the 0x1000 length substream because whatever comes after a geometry struct is always aligned like that. I suppose I could use instances here, but an extra ~4096 bytes twice during the entire parsing doesn't seem worth the additional complexity to me.

I'd prefer not to do too much extra work on this format, since it works well for my purposes as is. Are you willing to merge this in its current state, or do you need this refactoring done first?

filesystem/android_super.ksy
generalmimon: > I'm not sure where you want me to put this line. Isn't the compiler going to complain about an unknown key if I add it right here?

No, see https://doc.kaitai.io/user_guide.html#orig-id:

> Kaitai Struct compiler would just ignore any key that starts with `-`, and silently allow it. These kind of keys can be used to store arbitrary additional information, which can be accessible to external tools (i.e. other than the compiler). Feel free to add more arbitrary keys if you need to store extra structured information for some reason. (...)

tchebb: Ah, I see now: there's not supposed to be a space between the dash and `affected-by`. Added.

filesystem/android_super.ksy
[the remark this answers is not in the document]
generalmimon: > I'm assuming that nothing is corrupt in the file we're parsing, which means we ignore the values in the backup geometry.

OK, a weird format 😃

In that case, can you add some comment like `` intentionally `primary_geometry` `` there, so nobody will "correct" it in the future? The names really suggest that `backup_metadata` and `backup_geometry` belong to each other.

filesystem/android_super.ksy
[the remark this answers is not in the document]
generalmimon: > an extra ~4096 bytes twice during the entire parsing doesn't seem worth the additional complexity to me.

Yeah, you're actually right, the complexity isn't worth. And my argument about reading the 4096 byte array is actually the problem of KS (i.e. absence of https://github.com/kaitai-io/kaitai_struct/issues/44 implementation), not the spec itself. The format is OK in the current state.

I always start with a simple idea and then elaborate it in a more detail, and until the elaboration it's not clear if the idea is good or not 🙂

filesystem/android_super.ksy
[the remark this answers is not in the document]
tchebb: Done

filesystem/android_super.ksy
```diff
@@ -0,0 +1,207 @@
...
+doc: |
+  The metadata stored by Android at the beginning of a "super" partition, which
+  is what it calls a disk partition that holds one or more Dynamic Partitions.
+  Dynamic Partitions do more or less the same thing that LVM does on Linux,
+  allowing Android to map ranges of non-contiguous extents to a single logical
+  device. This metadata holds that mapping.
+
+doc-ref:
+  - https://source.android.com/devices/tech/ota/dynamic_partitions
+  - https://android.googlesource.com/platform/system/core/+/master/fs_mgr/liblp/include/liblp/metadata_format.h
```
generalmimon, line 20: It's generally not a good idea to link to the `master` branch (or any branch that _changes_ over time) when you refer to any source code, because, you know: software evolves and code gets refactored and regrouped, old code is deleted etc.

Please try to link a specific tag or a commit hash.

tchebb: Done

generalmimon (approved):
@tchebb Thanks for the contribution, merging in!

PR #355, opened 2020-11-16
add gen3 Nintendo Mii data file formats

HEYimHeroic:
added the Mii Studio Mii data file format: the Mii Studio (studio.mii.nintendo.com) stores Miis in this format. you can pull the Mii data from your browser's local storage, or by using this JavaScript code written by me and polished by bendevnul: https://github.com/RiiConnect24/mii2studio#importingexporting-to-mii-studio (instructions are on the page)

added the SwitchDB Mii data file format: the Nintendo Switch's internal Mii database uses a different format than used in games. this is the format of the Miis used in the internal database. highly compact, and can fit 10 characters of a Mii name (20 bytes). also includes some currently unknown data :( but it doesn't seem that this data is relevant, as even the Mii Studio format doesn't store it.

added the Switch Mii data file format: can store the exact same as the SwitchDB format, just this format is only used in games like Super Smash Bros. Ultimate and Mario Kart 8 Deluxe, while the SwitchDB is (as explained above) only used in the internal Mii database.

i documented and created the .ksy files for all of these formats. if needed, i can provide default and example files.

### DIFF
--- a/game/MiiStudio_Miidatafile.ksy
+++ b/game/MiiStudio_Miidatafile.ksy
@@ -1,0 +1,142 @@
+meta:
+  id: miidata_ms
+  endian: le
+seq:
+  - id: facial_hair_color
+    type: u1
+    doc: Facial hair color. Ranges from 0 to 99. Not ordered the same as visible in editor.
+  - id: beard_goatee
+    type: u1
+    doc: Beard (goatee) type. Ranges from 0 to 5.
+  - id: body_weight
+    type: u1
+    doc: Body weight. Ranges from 0 to 127, small to large.
+  - id: eye_stretch
+    type: u1
+    doc: Eye stretch. Ranges from 0 to 6, small to large.
+  - id: eye_color
+    type: u1
+    doc: Eye color. Ranges from 0 to 99. Not ordered the same as visible in editor.
+  - id: eye_rotation
+    type: u1
+    doc: Eye rotation. Ranges from 0 to 7, down to up. Note that some eye types have a default rotation.
+  - id: eye_size
+    type: u1
+    doc: Eye size. Ranges from 0 to 7, small to large.
+  - id: eye_type
+    type: u1
+    doc: Eye type. Ranges from 0 to 59. Not ordered the same as visible in editor.
+  - id: eye_horizontal
+    type: u1
+    doc: Eye X (horizontal) distance. Ranges from 0 to 12, close to far.
+  - id: eye_vertical
+    type: u1
+    doc: Eye Y (vertical) position. Ranges from 18 to 0, low to high.
+  - id: eyebrow_stretch
+    type: u1
+    doc: Eyebrow stretch. Ranges from 0 to 6, small to large.
+  - id: eyebrow_color
+    type: u1
+    doc: Eyebrow color. Ranges from 0 to 99. Not ordered the same as visible in editor.
+  - id: eyebrow_rotation
+    type: u1
+    doc: Eyebrow rotation. Ranges from 0 to 11, down to up. Note that some eyebrow types have a default rotation.
+  - id: eyebrow_size
+    type: u1
+    doc: Eyebrow size. Ranges from 0 to 8, small to large.
+  - id: eyebrow_type
+    type: u1
+    doc: Eyebrow type. Ranges from 0 to 23. Not ordered the same as visible in editor.
+  - id: eyebrow_horizontal
+    type: u1
+    doc: Eyebrow X (horizontal) distance. Ranges from 0 to 12, close to far.
+  - id: eyebrow_vertical
+    type: u1
+    doc: Eyebrow Y (vertical) distance. Ranges from 18 to 3, low to high.
+  - id: face_color
+    type: u1
+    doc: Skin color. Ranges from 0 to 9. Not ordered the same as visible in editor.
+  - id: face_makeup
+    type: u1
+    doc: Face makeup. Ranges from 0 to 11.
+  - id: face_type
+    type: u1
+    doc: Face shape. Ranges from 0 to 11. Not ordered the same as visible in editor.
+  - id: face_wrinkles
+    type: u1
+    doc: Face wrinkles. Ranges from 0 to 11.
+  - id: favorite_color
+    type: u1
+    doc: Favorite color. Ranges from 0 to 11.
+  - id: gender
+    type: u1
+    doc: Mii gender. 0 = male, 1 = female.
+  - id: glasses_color
+    type: u1
+    doc: Glasses color. Ranges from 0 to 99. Not ordered the same as visible in editor.
+  - id: glasses_size
+    type: u1
+    doc: Glasses size. Ranges from 0 to 7, small to large.
+  - id: glasses_type
+    type: u1
+    doc: Glasses type. Ranges from 0 to 19. Not ordered the same as visible in editor.
+  - id: glasses_vertical
+    type: u1
+    doc: Glasses Y (vertical) position. Ranges from 20 to 0, low to high.
+  - id: hair_color
+    type: u1
+    doc: Hair color. Ranges from 0 to 99. Not ordered the same as visible in editor.
+  - id: hair_flip
+    type: u1
+    doc: Flip hair. 0 = no, 1 = yes.
+  - id: hair_type
+    type: u1
+    doc: Hair type. Ranges from 0 to 131. Not ordered the same as visible in editor.
+  - id: body_height
+    type: u1
+    doc: Body height. Ranges from 0 to 127, short to tall.
+  - id: mole_size
+    type: u1
+    doc: Beauty mark size. Ranges from 0 to 8, small to large.
+  - id: mole_enable
+    type: u1
+    doc: Enable beauty mark. 0 = no, 1 = yes.
+  - id: mole_horizontal
+    type: u1
+    doc: Beauty mark X (horizontal) position. Ranges from 0 to 16, left to right.
+  - id: mole_vertical
+    type: u1
+    doc: Beauty mark Y (vertical) position. Ranges from 30 to 0, low to high.
+  - id: mouth_stretch
+    type: u1
+    doc: Mouth stretch. Ranges from 0 to 6, small to large.
+  - id: mouth_color
+    type: u1
+    doc: Mouth color. The default colors are ordered the same as visible in editor, ranging from 19 to 23. The custom colors are not and range from 0 to 99.
+  - id: mouth_size
+    type: u1
+    doc: Mouth size. Ranges from 0 to 8, small to large.
+  - id: mouth_type
+    type: u1
+    doc: Mouth type. Ranges from 0 to 35. Not ordered the same as visible in editor.
+  - id: mouth_vertical
+    type: u1
+    doc: Mouth Y (vertical) position. Ranges from 18 to 0, low to high.
+  - id: beard_size
+    type: u1
+    doc: Mustache size. Ranges from 0 to 8, small to large.
+  - id: beard_mustache
+    type: u1
+    doc: Mustache type. Ranges from 0 to 5.
+  - id: beard_vertical
+    type: u1
+    doc: Mustache Y (vertical) position. Ranges from 16 to 0, low to high.
+  - id: nose_size
+    type: u1
+    doc: Nose size. Ranges from 0 to 8, small to large.
+  - id: nose_type
+    type: u1
+    doc: Nose type. Ranges from 0 to 17. Not ordered the same as visible in editor.
+  - id: nose_vertical
+    type: u1
+    doc: Nose Y (vertical) position. Ranges from 18 to 0, low to high.
--- a/game/SwitchDB_Miidatafile.ksy
+++ b/game/SwitchDB_Miidatafile.ksy
@@ -1,0 +1,157 @@
+meta:
+  id: miidata_sdb
+  endian: le
+seq:
+  - id: hair_type
+    type: u1
+    doc: Hair type. Ranges from 0 to 131. Not ordered the same as visible in editor.
+  - id: mole_enable
+    type: b1
+    doc: Enable mole. 0 = no, 1 = yes.
+  - id: body_height
+    type: b7
+    doc: Body height. Ranges from 0 to 127, short to tall. In the Mii editor, pressing right will add 1 to this value, and pressing left will subtract 1, allowing for precise editing.
+  - id: hair_flip
+    type: b1
+    doc: Flip hair. 0 = no, 1 = yes.
+  - id: body_weight
+    type: b7
+    doc: Body weight. Ranges from 0 to 127, small to large. In the Mii editor, pressing right will add 1 to this value, and pressing left will subtract 1, allowing for precise editing.
+  - id: hair_color
+    type: u1
+    doc: Hair color. Ranges from 0 to 99. Not ordered the same as visible in editor.
+  - id: gender
+    type: b1
+    doc: Mii gender. 0 = male, 1 = female.
+  - id: eye_color
+    type: b7
+    doc: Eye color. Ranges from 0 to 99. Not ordered the same as visible in editor.
+  - id: eyebrow_color
+    type: u1
+    doc: Eyebrow color. Ranges from 0 to 99. Not ordered the same as visible in editor.
+  - id: mouth_color
+    type: u1
+    doc: Mouth color. The default colors are ordered the same as visible in editor, ranging from 19 to 23. The custom colors are not and range from 0 to 99.
+  - id: facial_hair_color
+    type: u1
+    doc: Facial hair color. Ranges from 0 to 99. Not ordered the same as visible in editor.
+  - id: glasses_color
+    type: u1
+    doc: Glasses color. Ranges from 0 to 99. Not ordered the same as visible in editor.
+  - id: eye_type
+    type: u1
+    doc: Eye type. Ranges from 0 to 59. Not ordered the same as visible in editor.
+  - id: mouth_type
+    type: u1
+    doc: Mouth type. Ranges from 0 to 35. Not ordered the same as visible in editor.
+  - id: glasses_size
+    type: b3
+    doc: Glasses size. Ranges from 0 to 7, small to big.
+  - id: eye_vertical
+    type: b5
+    doc: Eye Y (vertical) position. Ranges from 18 to 0, low to high.
+  - id: facial_hair_mustache
+    type: b3
+    doc: Mustache type. Ranges from 0 to 5.
+  - id: eyebrow_type
+    type: b5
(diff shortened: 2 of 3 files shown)


game/MiiStudio_Miidatafile.ksy
```diff
@@ -0,0 +1,142 @@
...
+    doc: Face shape. Ranges from 0 to 11. Not ordered the same as visible in editor.
+  - id: face_wrinkles
+    type: u1
+    doc: Face wrinkles. Ranges from 0 to 11.
+  - id: favorite_color
+    type: u1
+    doc: Favorite color. Ranges from 0 to 11.
+  - id: gender
+    type: u1
+    doc: Mii gender. 0 = male, 1 = female.
```
dgelessus, line 73: It probably makes sense to define an [enum](https://doc.kaitai.io/user_guide.html#enums) for this field.

Actually, there are a few other fields in the format that could also use enums (colors, beard/eye/eyebrow/etc. types), but those fields have so many possible values and are probably difficult to describe in text sometimes, so it might not be worth the effort for those fields.

(Same for the `gender` fields in the other files.)

game/MiiStudio_Miidatafile.ksy
```diff
@@ -0,0 +1,142 @@
...
+    doc: Flip hair. 0 = no, 1 = yes.
+  - id: hair_type
+    type: u1
+    doc: Hair type. Ranges from 0 to 131. Not ordered the same as visible in editor.
+  - id: body_height
+    type: u1
+    doc: Body height. Ranges from 0 to 127, short to tall.
+  - id: mole_size
+    type: u1
+    doc: Beauty mark size. Ranges from 0 to 8, small to large.
```
dgelessus, line 100: The field's `id` is `mole_size`, but the `doc` calls it a "beauty mark". For consistency it would be better to use the same term in both the `id` and the `doc`.

game/MiiStudio_Miidatafile.ksy
```diff
@@ -0,0 +1,142 @@
...
+    doc: Mustache Y (vertical) position. Ranges from 16 to 0, low to high.
+  - id: nose_size
+    type: u1
+    doc: Nose size. Ranges from 0 to 8, small to large.
+  - id: nose_type
+    type: u1
+    doc: Nose type. Ranges from 0 to 17. Not ordered the same as visible in editor.
+  - id: nose_vertical
+    type: u1
+    doc: Nose Y (vertical) position. Ranges from 18 to 0, low to high.
```
dgelessus, line 142: Minor thing: there should be a newline at the end of the file.

(Same for the other two files.)

game/SwitchDB_Miidatafile.ksy
```diff
@@ -0,0 +1,157 @@
+meta:
+  id: miidata_sdb
+  endian: le
```
dgelessus, line 3: Like with the previous spec, there's no need to set an endianness, because none of the integers are multiple bytes large.

Though because the spec uses bit-sized integers, you should set the bit order using `bit-endian: be`. If you don't set `bit-endian`, the compiler also defaults to `be`, but to avoid confusion it's recommended to always set `bit-endian` explicitly if possible. (In the [`bit-endian` docs](https://doc.kaitai.io/user_guide.html#bit-endian) there's a detailed explanation why, if you're interested.)
```suggestion
  bit-endian: be
```

game/SwitchDB_Miidatafile.ksy
```diff
@@ -0,0 +1,157 @@
...
+    type: b4
+    doc: Mole size. Ranges from 0 to 8, small to big.
+  - id: facial_hair_size
+    type: b4
+    doc: Mustache size. Ranges from 0 to 8, small to big.
+  - id: mii_name
+    type: str
+    size: 20
+    encoding: utf-16le
+    doc: Mii name. Can be up to 10 characters long.
```
dgelessus, line 147: Could you mention in the `doc` how the length of the name is determined, i. e. if you enter a name shorter than 10 characters, how do you find out where the name ends?

If it's zero-terminated (as is commonly done), you're supposed to be able to use `type: strz` here. Unfortunately this is currently broken for strings in UTF-16 and UTF-32 encoding - see kaitai-io/kaitai_struct#187.

game/SwitchDB_Miidatafile.ksy
```diff
@@ -0,0 +1,157 @@
...
+  - id: unknown
+    type: u1
+    repeat: expr
+    repeat-expr: 16
```
dgelessus, line 151: Instead of an array of `u1` values, it's almost always better to use `size` instead, to create a simple byte array.
```suggestion
    size: 16
```

game/SwitchDB_Miidatafile.ksy
```diff
@@ -0,0 +1,157 @@
...
+    doc: Currently unknown data.
+  - id: mii_id
+    type: u1
+    repeat: expr
+    repeat-expr: 4
```
dgelessus, line 156:
```suggestion
    size: 4
```

game/Switch_Miidatafile.ksy
```diff
@@ -0,0 +1,167 @@
+meta:
+  id: miidata_swi
+  endian: le
+seq:
+  - id: unknown_data
+    type: u1
+    repeat: expr
+    repeat-expr: 16
```
dgelessus, line 8:
```suggestion
    size: 16
```

game/Switch_Miidatafile.ksy
```diff
@@ -0,0 +1,167 @@
...
+    doc: Currently unknown data.
+  - id: mii_name
+    type: str
+    size: 20
+    encoding: utf-16le
+    doc: Mii name. Can be up to 10 characters long. Different from the Mii name that appears in Super Smash Bros. Ultimate - in game, this is never seen.
+  - id: unknown_buffer
+    type: u1
+    repeat: expr
+    repeat-expr: 3
```
dgelessus, line 18: If you're relatively sure that these bytes should always be zero, you can use `contents` to check that:

```suggestion
    contents: [0, 0, 0]
```

game/Switch_Miidatafile.ksy
```diff
@@ -0,0 +1,167 @@
...
+  - id: body_height
+    type: u1
+    doc: Body height. Ranges from 0 to 127, short to tall.
+  - id: body_weight
+    type: u1
+    doc: Body weight. Ranges from 0 to 127, small to large.
+  - id: unknown_buffer2
+    type: u1
+    repeat: expr
+    repeat-expr: 2
```
dgelessus, line 35:
```suggestion
    contents: [0, 0]
```

game/Switch_Miidatafile.ksy
```diff
@@ -0,0 +1,167 @@
...
+  - id: mole_horizontal
+    type: u1
+    doc: Mole X (horizontal) position. Ranges from 0 to 16, left to right.
+  - id: mole_vertical
+    type: u1
+    doc: Mole Y (vertical) position. Ranges from 30 to 0, low to high.
+  - id: unknown_buffer3
+    type: u1
+    repeat: expr
+    repeat-expr: 1
```
dgelessus, line 166:
```suggestion
    contents: [0]
```

dgelessus (changes requested):
Thank you for the PR! This looks good overall. I've left a few comments about style/formatting, and some suggestions for using Kaitai Struct's validation features.

game/SwitchDB_Miidatafile.ksy
larsenv: It's zero-terminated.

HEYimHeroic:
wow, what an incredibly detailed list of suggestions! i'll look more closely at your suggested changes soon, but in the meantime, here's some notes i felt like i should mention:

> Actually, there are a few other fields in the format that could also use enums (colors, beard/eye/eyebrow/etc. types), but those fields have so many possible values and are probably difficult to describe in text sometimes, so it might not be worth the effort for those fields.

i actually do have a few separate files available that properly line up everything that isn't the same order as in the editors. it's simply a text file named "maps.txt" which list every selection's ID in the same order they appear in the editor, but it's so large i left it out since i couldn't find a way to properly include it. same with the rotation values, but that might be easier. i've attached the maps.txt and rotation.txt files to this post - i'm fairly new to creating Kaitais, but you've laid out such detail, it seems a lot more possible now :)

> The field's `id` is `mole_size`, but the `doc` calls it a "beauty mark". For consistency it would be better to use the same term in both the `id` and the `doc`.

this was done so the multiple files would have no naming issues when carrying over values. for some reason, the Mii Studio calls some things different names (such as moles "beauty marks" and facial hair all being under "beard"), despite them being the same exact thing as other editors have called it previously. as a standalone file, however, it does make sense to match up the terms. thanks!

> Could you mention in the `doc` how the length of the name is determined, i. e. if you enter a name shorter than 10 characters, how do you find out where the name ends?
> 
> If it's zero-terminated (as is commonly done), you're supposed to be able to use `type: strz` here. Unfortunately this is currently broken for strings in UTF-16 and UTF-32 encoding - see [kaitai-io/kaitai_struct#187](https://github.com/kaitai-io/kaitai_struct/issues/187). In that case as a workaround you can leave it as `type: str` and note in the `doc` that it's zero-terminated.

i see... so, for now, i would use the `str` workaround? just making sure ^^

> If you're relatively sure that these bytes should always be zero, you can use `contents` to check that:

relatively sure, yes ^^ having 00 buffers like this isn't new to the Mii data formats, as i believe all previous versions of the formats (Wii, Wii U/3DS, etc.) all also had 00 buffers. of course, there aren't many Miis found in Switch games, so the data is limited... but from all cases i've seen, included user-created Miis extracted from save files, i've found these values to always be 00. it's helpful to know that many of these values can have caps and limits. i had no idea Kaitai was capable of throwing errors when the value is outside an expected range! (again, very new to these)

there was also one more thing i wanted to ask. having 16 different values for the SwitchDB and Switch's `unknown_data` seems tedious, but representing them using decimal numbers is even worse, so i separated them. is there a way i'd be able to represent those 16 bytes as one hexidecimal value? thanks.

also, i'm glad the fact there even _is_ unknown data is okay. i could elaborate more about the unknown data and things like that, so i plan to do so. i was slightly worried only the Mii Studio Kaitai would be allowed, since it's the only one of the three that doesn't have any byte left unknown... but i'm glad to see they're all okay. ^^

here are the files i mentioned above, with many notes, heh:
[rotation.txt](https://github.com/kaitai-io/kaitai_struct_formats/files/5551038/rotation.txt)
[maps_Studio.txt](https://github.com/kaitai-io/kaitai_struct_formats/files/5551042/maps_Studio.txt)
[maps_Switch.txt](https://github.com/kaitai-io/kaitai_struct_formats/files/5551045/maps_Switch.txt)

larsenv:
@HEYimHeroic You know you can reply to the code comments themselves, right?

generalmimon:
@HEYimHeroic For your own convenience, remember a rule of thumb when making pull requests from your fork repository: **never commit directly on `master` of your fork**. Create a topic branch for every PR you create. See https://contribute.jquery.org/commits-and-pull-requests/#never-commit-on-master:

game/Switch_Miidatafile.ksy
generalmimon: > If you're relatively sure that these bytes should always be zero, you can use `contents` to check that

Yes, _"if you're relatively sure"_, but I don't think we can assume anything about _unknown data_.

dgelessus:
> i actually do have a few separate files available that properly line up everything that isn't the same order as in the editors. it's simply a text file named "maps.txt" which list every selection's ID in the same order they appear in the editor, but it's so large i left it out since i couldn't find a way to properly include it. same with the rotation values, but that might be easier. i've attached the maps.txt and rotation.txt files to this post - i'm fairly new to creating Kaitais, but you've laid out such detail, it seems a lot more possible now :)

It's possible to include and use simple lookup tables in a KSY, using [value instances](https://doc.kaitai.io/user_guide.html#_value_instances):

```yaml
instances:
  face_type_editor_order:
    value: [
      0x00, 0x01, 0x08, 0x02, 0x03, 0x09,
      0x04, 0x05, 0x0a, 0x06, 0x07, 0x0b,
    ]
  face_type_editor_position:
    value: face_type_editor_order[face_type]
```

Though you might have to "invert" the lookup table first - if I'm reading the text files right, your lookup tables translate the position in the editor to the byte stored in the data, which is the opposite of what you need to do when reading the data (translating the stored byte to an editor position).

> this was done so the multiple files would have no naming issues when carrying over values. for some reason, the Mii Studio calls some things different names (such as moles "beauty marks" and facial hair all being under "beard"), despite them being the same exact thing as other editors have called it previously. as a standalone file, however, it does make sense to match up the terms. thanks!

Ah I see, that makes sense. Might still be a good idea to point out in the `doc` when one system/software uses a different name than others when talking about the same thing.

> i see... so, for now, i would use the `str` workaround? just making sure ^^

Yes, exactly. (I just wanted to mention the `strz` bug with UTF-16 so you don't run into it by accident.)

> there was also one more thing i wanted to ask. having 16 different values for the SwitchDB and Switch's `unknown_data` seems tedious, but representing them using decimal numbers is even worse, so i separated them. is there a way i'd be able to represent those 16 bytes as one hexidecimal value? thanks.

I'm not sure what you mean here exactly, sorry... The way you've written `unknown_data` currently (using `type: u1` and `repeat-expr: 16`), it will be parsed as an array of 16 integers (one for each byte). If you use `size: 16` (with no `type`), it will instead be parsed as a raw byte array and stored as your language's type for that (e. g. `bytes` in Python). Kaitai Struct doesn't support integers larger than 64 bits (8 bytes), so you can't parse all 16 bytes into a single integer.

Regarding decimal vs. hexadecimal, that only depends on your code and how it displays the values - the parser doesn't affect how your language prints integers and byte arrays by default. If you're working in the [Web IDE](https://ide.kaitai.io/), you can control how it displays your data by adding [`-webide-representation`](https://github.com/kaitai-io/kaitai_struct_webide/wiki/Features#user-content-webide-representation) attributes in your KSY.

> also, i'm glad the fact there even _is_ unknown data is okay. i could elaborate more about the unknown data and things like that, so i plan to do so. i was slightly worried only the Mii Studio Kaitai would be allowed, since it's the only one of the three that doesn't have any byte left unknown... but i'm glad to see they're all okay. ^^

It's not a problem if your spec skips some parts of the data where the exact meaning isn't known yet. We don't require anyone to 100 % reverse-engineer a format before submitting it to kaitai_struct_formats 🙂

HEYimHeroic:
> @HEYimHeroic You know you can reply to the code comments themselves, right?

yes, i know... but leaving like 5 separate comments seems tedious and would spam notifications, heh

> @HEYimHeroic For your own convenience, remember a rule of thumb when making pull requests from your fork repository: **never commit directly on `master` of your fork**. Create a topic branch for every PR you create.

ah, i see... if my contribution history tells you anything, i'm *also* new to doing things like this on GitHub. sorry ^^ if there's a way i can revert it or something, let me know.

> Though you might have to "invert" the lookup table first - if I'm reading the text files right, your lookup tables translate the position in the editor to the byte stored in the data, which is the opposite of what you need to do when reading the data (translating the stored byte to an editor position).

so, the editor may show:
```
0x00, 0x01, 0x02,
0x03, 0x04, 0x05
```
but i should put the lookup table as:
```
0x05, 0x04, 0x03,
0x02, 0x01, 0x00
```
is that right?

> I'm not sure what you mean here exactly, sorry...

the proper way to store information like this (previous iterations of the format point to this) is to store them as a HEX string. for example, a Mii ID would be properly, accurately stored as something like `097A2DF5`, not 159,002,101. i believe at least part of this `unknown_data` is something similar to the Mii ID used in previous formats, and storing at least 16 HEX values would be nicer and more appropriate than 16 decimal numbers. ^^

dgelessus:
re. "inverting" the lookup table: I meant that in [the mathematical sense](https://en.wikipedia.org/wiki/Inverse_function) of "make a function go the other way" - I don't know any better way to describe it, sorry 😄 As an example, the `face` array from your maps_Studio.txt looks like this:

```
face: [
    0x00,0x01,0x08,0x02,0x03,0x09,
    0x04,0x05,0x0a,0x06,0x07,0x0b
]
```

and it describes this mapping between the position of each face type in the editor and the byte value stored in the data:

| Position in editor | Byte value in data |
|---|---|
| 0 | 0 |
| 1 | 1 |
| 2 | 8 |
| 3 | 2 |
| 4 | 3 |
| 5 | 9 |
| 6 | 4 |
| 7 | 5 |
| 8 | 10 (`0x0a`) |
| 9 | 6 |
| 10 | 7 |
| 11 | 11 (`0x0b`) |

The `face` array lets you easily translate an editor position to a byte value using `face[editor_pos]`, e. g. `face[5] == 0x09`. But you can't easily do a lookup in the other direction, i. e. translate a byte value to an editor position. To do that you would need another lookup table that looks something like this:

```
face_inverse: [
    0, 1, 3, 4, 6, 7,
    9, 10, 2, 5, 8, 11
]
```

And with that array you can translate a byte value to an editor position using `face_inverse[byte_value]`, e. g. `face_inverse[0x09] == 5`. This could be integrated into the KSY using `instances`, to allow your code to directly read the editor positions, instead of having to do the conversion manually.

Anyway, I don't really know how you're using the KSY in your code, so maybe what I'm suggesting here isn't useful to you after all 😄 If you're not sure, you can just leave this out of the KSY and keep the arrays in your code.

> the proper way to store information like this (previous iterations of the format point to this) is to store them as a HEX string. for example, a Mii ID would be properly, accurately stored as something like `097A2DF5`, not 159,002,101. i believe at least part of this `unknown_data` is something similar to the Mii ID used in previous formats, and storing at least 16 HEX values would be nicer and more appropriate than 16 decimal numbers. ^^

That's just a question of how you print/display/format the value. Kaitai Struct doesn't specifically store the number as decimal or hexadecimal - it just stores it as a normal integer, which is normally by default shown as decimal. I don't know what language you're using the KSY with, but in Python for example you could do `print(f"{mii_id:>08X}")` to show the ID as 8 hex digits like in your example (with `mii_id = 159002101` this outputs `097A2DF5`). For the `unknown_data` it's similar - Kaitai Struct isn't storing 16 *decimal* numbers, just 16 *numbers*, which will be shown as decimal by default, but you can make your code display them differently if it helps you with figuring out what's actually stored in that data.

KOLANICH (changes requested):
IMHO should be moved into an own directory.

HEYimHeroic:
okay, sorry i took so long, i just got finished with moving. i'll approve the current changes, and make some more edits from there. after that, it should be ready to go ^^

generalmimon:
Closing this PR, as it's superseded by https://github.com/kaitai-io/kaitai_struct_formats/pull/400. However, from what I've seen, this thread contains some useful review comments which are not pointed out or addressed in https://github.com/kaitai-io/kaitai_struct_formats/pull/400.

PR #356, opened 2020-11-17
Add Mii formats for other consoles

larsenv:
This is related to #355, I've worked on this Kaitai with @HEYimHeroic.

The Wii U/3DS/Miitomo Kaitai in here is in a bad state because I could never figure out how to grab bits from a little-endian integer.

### DIFF
--- a/game/WiiU3DSMiitomo_miidatafile.ksy
+++ b/game/WiiU3DSMiitomo_miidatafile.ksy
@@ -1,0 +1,174 @@
+meta:
+  id: gen2_wiiu_3ds_miitomo
+  endian: le
+seq:
+  - id: unknown_1
+    type: u1
+    doc: Always 3?
+  - id: character_set
+    type: b2
+    doc: 0=JPN+USA+EUR, 1=CHN, 2=KOR, 3=TWN
+  - id: region_lock
+    type: b2
+    doc: 0=no lock, 1=JPN, 2=USA, 3=EUR
+  - id: profanity_flag
+    type: b1
+  - id: copying
+    type: b1
+  - id: unknown_2
+    type: b2
+  - id: mii_position_slot_index
+    type: b4
+  - id: mii_position_page_index
+    type: b4
+  - id: version
+    type: b4
+  - id: unknown_3
+    type: b4
+  - id: system_id
+    type: u1
+    repeat: expr
+    repeat-expr: 8
+  - id: avatar_id
+    type: u1
+    repeat: expr
+    repeat-expr: 4
+  - id: client_id
+    type: u1
+    repeat: expr
+    repeat-expr: 6
+  - id: padding
+    type: u2
+  - id: data_1
+    type: u2
+  - id: mii_name
+    type: str
+    size: 20
+    encoding: utf-16le
+  - id: body_height
+    type: u1
+  - id: body_weight
+    type: u1
+  - id: face_color
+    type: b3
+  - id: face_type
+    type: b4
+  - id: mingle
+    type: b1
+  - id: face_makeup
+    type: b4
+  - id: face_wrinkles
+    type: b4
+  - id: hair_type
+    type: u1
+  - id: unknown_5
+    type: b4
+  - id: hair_flip
+    type: b1
+  - id: hair_color
+    type: b3
+  - id: eye
+    type: u4le
+  - id: eyebrow
+    type: u4le
+  - id: nose
+    type: u2le
+  - id: mouth
+    type: u2le
+  - id: mouth2
+    type: u2le
+  - id: beard
+    type: u2le
+  - id: glasses
+    type: u2le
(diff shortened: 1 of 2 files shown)


larsenv:
@dgelessus Please review this.

generalmimon:
@larsenv Can I ask you to provide some example files for testing?

> The Wii U/3DS/Miitomo Kaitai in here is in a bad state because I could never figure out how to grab bits from a little-endian integer.

Does the section [Bit-sized integers](https://doc.kaitai.io/user_guide.html#_bit_sized_integers) in the User Guide help?

generalmimon:
And yeah, for your own convenience, remember a rule of thumb when making pull requests from your fork repository: **never commit directly on `master` of your fork**. Create a topic branch for every PR you create. See https://contribute.jquery.org/commits-and-pull-requests/#never-commit-on-master:

dgelessus (changes requested):
Thank you for the PR! You've probably already seen my comments in #355, and many of them also apply to this PR, so I've kept my review comments here a bit short. If you have any questions, let me know.

About extracting bits from little-endian integers: Kaitai Struct's default behavior in this case can be confusing and doesn't work at all with some little-endian formats. The recent 0.9 release has improved support for little-endian bit-sized integers. I would recommend that you set `bit-endian: le` under `meta` in the KSY and then try to split the byte integers into bit integers again.

With `bit-endian: le`, you need to order the bit fields from low to high bits. If you look at the expressions that you've currently written under `instances`, the fields with a low number after `>>` (or no `>>` at all) should come earlier than the ones with a high number after `>>`. For example the fields stored in `data_1` should go in the order `gender`, `birth_month`, `birth_day`, `favorite_color`, `favorite`.

For the full details on how Kaitai Struct's bit-sized integers behave depending on the `bit-endian` setting, see the documentation that @generalmimon linked above.

larsenv:
@generalmimon @dgelessus Thanks.

At the time when I made the Kaitai, the IDE didn't support the `bit-endian` attribute. It does now, so I made the changes and now it looks much better in my opinion.

And about using a branch other than master, I'm aware that making a new branch is better, but when I made the PR I quickly did it using the web client and forgot about it.

game/WiiU3DSMiitomo_miidatafile.ksy
```diff
@@ -0,0 +1,178 @@
+meta:
+  id: WiiU3DSMiitomo_miidatafile
```
dgelessus, line 2: This isn't going to work:

```
io.kaitai.struct.format.YAMLParseException: /meta/id: invalid meta ID: 'WiiU3DSMiitomo_miidatafile', expected /^[a-z][a-z0-9_]*$/
```

Kaitai Struct doesn't accept uppercase letters in `id`s, because it might need to translate the identifier depending on the target language. For example, class names are by convention written in `CamelCase` in many languages, so a KSY named `some_long_name` will be translated to a class named `SomeLongName`. This kind of translation becomes more complicated and unclear if the original name already has uppercase letters in it, so Kaitai Struct just forbids uppercase letters in identifiers.

generalmimon: > Kaitai Struct doesn't accept uppercase letters in `id`s, because it might need to translate the identifier depending on the target language.

Yes, the reason is described in https://github.com/kaitai-io/kaitai_struct/issues/148.

larsenv:
Sorry, it's been a week, this is sort of low priority for me.

Is the commit I made better? What else needs to be done?

KOLANICH (changes requested):
IMHO: should be moved into a dedicated dir with all its siblings, modularized and deduplicated.

game/WiiU3DSMiitomo_miidatafile.ksy
```diff
@@ -0,0 +1,178 @@
+meta:
+  id: wiiu3dsmiitomo_miidatafile
```
dgelessus, line 2: The `id` is okay now, but the file name still has uppercase letters - it should also be all lowercase like the `id`.

If you're on Windows or Mac, it's possible that you've already renamed the file to lowercase and Git simply didn't detect the change (this can happen because Windows and Mac use case-insensitive filesystems by default). In that case you might have to use GitHub's web interface to rename the file to lowercase.

game/Wii_miidatafile.ksy
```diff
@@ -0,0 +1,142 @@
+meta:
+  id: gen1_wii
```
dgelessus, line 2: The `id` and file name here also need to be changed to match.

dgelessus (commented):
> What else needs to be done?

See the code comments in my earlier review: https://github.com/kaitai-io/kaitai_struct_formats/pull/356#pullrequestreview-533550254

larsenv:
Going forward, we should continue with #400.

PR #357, opened 2020-11-27
Fix misprint if SWF xref location

Mingun:
Accidentally detected typo

### DIFF
--- a/executable/swf.ksy
+++ b/executable/swf.ksy
@@ -4,7 +4,10 @@ meta:
   file-extension: swf
   xref:
     justsolve: SWF
-    loc: fmt/507
+    pronom: fmt/507
+    loc:
+      - fdd000130
+      - fdd000248
     mime: application/x-shockwave-flash
     wikidata: Q594447
   tags:


dgelessus (approved):
Thanks for the fix!

executable/swf.ksy
```diff
@@ -4,7 +4,10 @@ meta:
   file-extension: swf
   xref:
     justsolve: SWF
-    loc: fmt/507
+    pronom: fmt/507
```
generalmimon, line 7: Actually, this error gets highlighted when validating the KSY against the [`ksy_schema.json`](https://github.com/kaitai-io/ksy_schema/blob/master/ksy_schema.json), but of course, someone would need to do that for all formats 😃:

image

Perhaps we could enforce all specs in KSF to pass through the https://github.com/kaitai-io/ksylint tool, but we need to update the built-in KSY schema in it and possibly adding additional checks to the tool (for instance, if the resource at URL _`https://www.nationalarchives.gov.uk/pronom/` + value of the `meta/xref/pronom` key_ exists, i.e. checking if ttps://www.nationalarchives.gov.uk/pronom/fmt/507 returns proper HTTP 200 code) etc.

Mingun: Yeah, I found it that way.

PR #376, opened 2021-01-02
First draft of xiaomi depthmap

VincenzoLaSpesa:
First draft of a parser of the depthmaps embedded in photos taken with last models of Xiaomi phones.

A photo with depth image contains 3 sub-images inside:
- the photo with exif metadata and depth effect applied ( i.e. bokeh simulation)
- the original image as jpeg
- a file in a proprietary format with the depthmap

The sub-images are just appended one after another ( using the jpeg EOF marker)

The current parser works on the last image.

The parser is still not completed as I still don't fully understand the format, but it can already extract the depthmap.
It needs to be used with the commandline tool in https://github.com/VincenzoLaSpesa/XiaomiDepthMapExtractor 

I will write a better documentation as soon as possible.


image/xiaomi_depthmap.ksy
```diff
@@ -0,0 +1,65 @@
...
+  - id: depthmap
+    size-eos: true
+    
+types:
+  magic_constant:
+    seq:
+      - id: magic
+        contents: "PMPD"
+      - id: magic2
+        contents: [0x02, 0x00, 0x64, 0x00, 0x01, 0x00, 0x00]
```
KOLANICH, line 26: Is it really a magic? Looks like 3 `u2` and then `u1`

VincenzoLaSpesa: Reasonable. I can't be 100% sure as I have that same string on every file I tried. I'll change it.

image/xiaomi_depthmap.ksy
```diff
@@ -0,0 +1,95 @@
...
+      size: 1022
+    - id: tail
+      contents: [0x80]
+
+
+  type_depthmap_info: 
+    doc: 17+1007 = 1kb, info about the shape of the depthmap
+    seq:
+    - id: head
+      contents: [01,00,0xFF,0xFF]
```
generalmimon, line 70:
```suggestion
      contents: [0x01, 0x00, 0xff, 0xff]
```

BTW, be careful with prefixing the decimal integer literals with zero in YAML (but also in other languages). This changes the number's interpretation from decimal (base 10) to octal (base 8), and that might cause unexpected results (e.g. `027` translates to `23`). See https://github.com/kaitai-io/kaitai_struct_formats/pull/299#discussion_r471781903 for more details.

image/xiaomi_depthmap.ksy
```diff
@@ -0,0 +1,95 @@
...
+    type: sector
+    repeat: until
+    repeat-until: not _.has_next
+  - id: depthmap
+    doc: the depthmap
+    size-eos: true
+    
+types:
+  info:
+    doc: Header of the file. Contains the raw lenght and many fields i still don't undestand
```
generalmimon, line 30: Some typos:

```suggestion
    doc: Header of the file. Contains the raw length and many fields I still don't undestand
```

generalmimon (changes requested):

PR #381, opened 2021-01-02
Added MGL Avionics Enigma Waypoint format

KOLANICH:


log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,134 @@
...
+    15: vor
+    16: rep_pt
+    17: lfr
+    18: uhf_ndb
+    19: m_ndb
+    20: m_ndb_or_dme
+    21: lom
+    22: lmm
+    23: loc_or_sdf
+    24: mls_or_ismls
```
dgelessus, line 128: I assume these are common abbreviations in the context of aviation and don't need to be explained further? Maybe for clarity it would be good to at least say what each abbreviation stands for, so that it's easier to look them up.

log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,134 @@
...
+    type: waypoint
+    repeat: eos
+    doc: There is no demand that waypoints are stored in any particular order in the waypoint file. They might be sorted by short name or waypoint type but this is of little consequence. The order of waypoints in the file matters only for route files. Route waypoints are stored in the file in order of their appearance in the route. A route should have at least one waypoint.
+
+types:
+  waypoint:
+    seq:
+      - id: lat
+        type: sph_coord
+        doc: N is positive
```
dgelessus, line 36: This could be checked with `valid`, right?

log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,134 @@
...
+      - id: lon
+        type: sph_coord
+        doc: E is positive
```
dgelessus, line 39: Same as above, perhaps check using `valid`.

log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,134 @@
...
+        doc: This entry is used as “key” and in aviation terms would contain the short identifier of an airfield or navaid beacon. This key is used to lookup related data in other databases, for example an airport data base containing details of an airport such as frequencies and runway headings.
+      - id: name_long
+        type: dc_str(27)
+        doc: |
+          The long name contains a human readable description of the waypoint.
+          For example: “Los Angeles International”. The long name is arbitrary and not used as search key for any purpose."
+
+    instances:
+      is_data_frequency:
+        value: 25 >= type.to_i and type.to_i >= 9
```
dgelessus, line 56: Maybe this is personal preference, but I think it's more readable this way:

```suggestion
        value: type.to_i >= 9 and type.to_i <= 25
```

Also it would probably be good to explain what exactly determines the format of the data and what it means. As I understand it, for navaids it's a radio frequency and for everything else it's an altitude? Is it correct that for the `unspecified` type and for unknown/not yet defined types the data is known to be an altitude, or could it be something different/an unknown data format?

log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,134 @@
...
+  In order to allow recognition of the format any party adopting this format agrees to refer to this format as the “Enigma waypoint format”. All data using this format should be considered to be in the public domain and the implementer will make reasonable effort to allow any interested party to use the data.
+  Exceptions:
+    MGL Avionics reserves copyright and denies use of this data format for any military or related purpose.
+
+doc: |
+  Waypoint format used by Enigma series of EFIS instruments from MGL Avionics.
+  It is common to create an index file containing a simple list of indexes, one to each entry in the waypoint file, sorted by a particular criteria (for example distance of the waypoint to a given location), rather than sorting the actual waypoint list itself. This is a method commonly employed by database applications.
+
+doc-ref:
+  - https://mglavionics.co.za/Docs/Enigma%20Waypoint%20format.pdf
```
dgelessus, line 23: The list can be collapsed if it's only a single URL:

```suggestion
doc-ref: https://mglavionics.co.za/Docs/Enigma%20Waypoint%20format.pdf
```

log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,134 @@
...
+  title: MGL Avionics Enigma Waypoint format
+  application: Enigma software
+  file-extension:
+    - ewd # WAYPOINT.EWD usually
+    - rte # MYROUTE.RTE usually
+  license: Unlicense
+  endian: le
+  encoding: ascii
+
+-license-header: | # the conditions contradict each other and seem to be unenforceable
```
dgelessus, line 12: I think this would fit better under `meta`, because that's where the normal `license` attribute goes.

dgelessus (changes requested):

log/gps/gps_mgl_enigma_waypoint.ksy
generalmimon: I don't think so, IMHO all possible values are valid. It's probably just that values `lat > 0` denote a parallel in the northern hemisphere of the Earth and values `lat < 0` relate to the southern hemisphere.

The comment is just written ambiguously, that's true.

log/gps/gps_mgl_enigma_waypoint.ksy
generalmimon: @dgelessus 

> Maybe this is personal preference, but I think it's more readable this way

I think exactly the same thing and also wanted to point it out. So at least, it is personal preference of both of us 🙂

log/gps/gps_mgl_enigma_waypoint.ksy
dgelessus: doh, I misread this (and the one below) and for some reason thought that "N" and "E" referred to numerical variables that must be greater than zero. The comment makes much more sense if I actually use my brain 🙂 But yeah, maybe it would be good to reword/expand it a bit so it's harder to misunderstand.

log/gps/gps_mgl_enigma_waypoint.ksy
dgelessus: (My comment is probably nonsense, see above: https://github.com/kaitai-io/kaitai_struct_formats/pull/381#discussion_r551049997)

log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,134 @@
+meta:
+  id: gps_mgl_enigma_waypoint
+  title: MGL Avionics Enigma Waypoint format
+  application: Enigma software
+  file-extension:
+    - ewd # WAYPOINT.EWD usually
+    - rte # MYROUTE.RTE usually
```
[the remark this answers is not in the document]
KOLANICH: The spec says that it is just default file names.

log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,134 @@
...
+  license: Unlicense
```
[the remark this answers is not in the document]
KOLANICH: So do I. But the terms look strange, feel strange and I doubt in their enforceability. I have written them asking if they are ready to remove the conditions and just relicense the PDF under CC0/Unlicense, but they will read my  message not earlier than Jan 6. Also this format is also used in a software under GPL (gpsbabel), from which I got the link to the PDF.

log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,134 @@
...
+        type: u1
+    seq:
+      - id: size
+        type: u1
+      - id: value
+        type: str
+        size: value_size
+      - id: garbage
+        type: str
+        size: max_size - value_size
```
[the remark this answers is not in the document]
KOLANICH: I used `value_size` because just `size` feels more like a full size (that is `max_size` in our case).

log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,134 @@
...
+
+enums:
+  type:  # Contact info@MGLAvionics.co.za if you add any waypoint type, so they can update their PDF
+    0: unspecified
+    1:
+      id: aitport_minor
+      doc: Typical assignment for medium sized airports
+    2:
+      id: aitport_major
+      doc: Typical assignment for large and international airports
```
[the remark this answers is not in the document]
KOLANICH: Thanks, gonna fix it

log/gps/gps_mgl_enigma_waypoint.ksy
KOLANICH: I feel like KS needs the syntax like in python for chained comparisons, so it should look like `25 >= type.to_i >= 9` or `9 <= type.to_i <= 25` and translate to the chained comparisons for python.

>As I understand it, for navaids it's a radio frequency and for everything else it's an altitude?

For `unspecified` it is altitude, and for 1 of the types it is undefined at all, according to the PDF.

log/gps/gps_mgl_enigma_waypoint.ksy
KOLANICH: Nice catch.

log/gps/gps_mgl_enigma_waypoint.ksy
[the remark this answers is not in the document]
dgelessus: Ah, I didn't read properly again - I only saw the `size` field above and missed the `value_size` instance definition below.

log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,134 @@
...
+        type: u1
+        enum: type
+      - id: name_short
+        type: dc_str(6)
+        doc: This entry is used as “key” and in aviation terms would contain the short identifier of an airfield or navaid beacon. This key is used to lookup related data in other databases, for example an airport data base containing details of an airport such as frequencies and runway headings.
+      - id: name_long
+        type: dc_str(27)
+        doc: |
+          The long name contains a human readable description of the waypoint.
+          For example: “Los Angeles International”. The long name is arbitrary and not used as search key for any purpose."
```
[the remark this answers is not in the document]
KOLANICH: Nice catch.

dgelessus (commented):
I just noticed that you put this spec in a new directory log/gps - wouldn't it fit better under the existing geospatial directory?

Also I'm really not sure about the license thing. It has some conditions that are definitely not normal for a copyleft/free software/open source license (specifically that you're asked to release any data that uses this format into the public domain). You can speculate that it may be unenforcable, but it's difficult to know for sure, and a spec under such an unclear license shouldn't go into the kaitai_struct_formats repo.

Or more simply: there's no way to describe this custom license using an SPDX expression (`Proprietary` is not a valid SPDX license identifier), and all specs in KSF are required to have one. So unless the documentation is released under a standard license, or the spec is rewritten based on documentation that doesn't have restictions like this, this can't be merged.

log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,148 @@
...
+      value_size:
+        value: max_size > size ? size : max_size
+
+  sph_coord:
+    seq:
+      - id: raw
+        type: s4
+    instances:
+      value:
+        value: raw / 180000
```
generalmimon, line 89: This looks like it should result in a floating-point number instead:

```suggestion
        value: raw / 180000.0
```

log/gps/gps_mgl_enigma_waypoint.ksy
KOLANICH: Fixed, thanks.

log/gps/gps_mgl_enigma_waypoint.ksy
KOLANICH: Fixed, thanks.

log/gps/gps_mgl_enigma_waypoint.ksy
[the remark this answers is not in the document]
KOLANICH: Fixed, thanks.

log/gps/gps_mgl_enigma_waypoint.ksy
KOLANICH: Added some `doc-ref`s.

log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,134 @@
...
+        type: dc_str(27)
+        doc: |
+          The long name contains a human readable description of the waypoint.
+          For example: “Los Angeles International”. The long name is arbitrary and not used as search key for any purpose."
+
+    instances:
+      is_data_frequency:
+        value: 25 >= type.to_i and type.to_i >= 9
+      altitude:
+        value: "((data_raw < 0x8FFFFFFF)?data_raw:-0x8FFFFFFF - data_raw)"
```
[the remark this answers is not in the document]
KOLANICH: Fixed, thanks. And not only this.

log/gps/gps_mgl_enigma_waypoint.ksy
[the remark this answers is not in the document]
KOLANICH: Fixed, thanks.

log/gps/gps_mgl_enigma_waypoint.ksy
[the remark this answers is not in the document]
KOLANICH: I haven't got any response yet, so just changed to `proprietary`

log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,148 @@
...
+        type: dc_str(27)
+        doc: |
+          The long name contains a human readable description of the waypoint.
+          For example: “Los Angeles International”. The long name is arbitrary and not used as search key for any purpose.
+
+    instances:
+      is_data_frequency:
+        value: 9 <= type.to_i and type.to_i <= 25
+      altitude:
+        value: "( ( data_raw < 0x80000000 ) ? data_raw : ( ( data_raw & 0x7FFFFFFF ) - 0x80000000 ) )"
```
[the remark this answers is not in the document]
KOLANICH: The problem is the field on which the interpretation of this field depends (`type`) follows this field. It can be done using a `pos` `instance`, but it would be heavier.

log/gps/gps_mgl_enigma_waypoint.ksy
KOLANICH: Yeah, I haven't tested it at all. Only checked that it compiles.

log/gps/gps_mgl_enigma_waypoint.ksy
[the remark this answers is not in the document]
generalmimon: @KOLANICH:

> The problem is the field on which the interpretation of this field depends (`is_data_frequency` is derived from `type`) follows this field. It can be done using a `pos` `instance`, but it would be heavier.

```ksy
    seq:
      # ...
      - id: data_slot
        size: 4
        type: slot
      - id: type
        type: u1
        enum: type
    instances:
      altitude:
        io: data_slot._io
        pos: 0
        type: s4
        if: not is_data_frequency
        doc: in feet
      frequency:
        io: data_slot._io
        pos: 0
        type: u4
        if: is_data_frequency
        doc: in kHz
    types:
      slot: {}
```

generalmimon (changes requested):

log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,148 @@
...
+    type: waypoint
+    repeat: eos
+    doc: There is no demand that waypoints are stored in any particular order in the waypoint file. They might be sorted by short name or waypoint type but this is of little consequence. The order of waypoints in the file matters only for route files. Route waypoints are stored in the file in order of their appearance in the route. A route should have at least one waypoint.
+
+types:
+  waypoint:
+    seq:
+      - id: lat
+        type: sph_coord
+        doc: Positive values are in Nothern Hemisphere
```
generalmimon, line 36:
```suggestion
        doc: Positive values are in the Nothern Hemisphere
```

log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,148 @@
...
+      - id: lon
+        type: sph_coord
+        doc: Positive values are in Eastern Hemisphere
```
generalmimon, line 39: The same as https://github.com/kaitai-io/kaitai_struct_formats/pull/381#discussion_r554379906:

```suggestion
        doc: Positive values are in the Eastern Hemisphere
```

log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,148 @@
...
+        type: u1
+        enum: type
+      - id: name_short
+        type: dc_str(6)
+        doc: This entry is used as “key” and in aviation terms would contain the short identifier of an airfield or navaid beacon. This key is used to lookup related data in other databases, for example an airport data base containing details of an airport such as frequencies and runway headings.
+      - id: name_long
+        type: dc_str(27)
+        doc: |
+          The long name contains a human readable description of the waypoint.
+          For example: “Los Angeles International”. The long name is arbitrary and not used as search key for any purpose.
```
generalmimon, line 52:
```suggestion
        doc: human-readable description of the waypoint (e.g. "Los Angeles International"), not used as a search key
```

log/gps/gps_mgl_enigma_waypoint.ksy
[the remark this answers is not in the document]
KOLANICH: Yeah, a KaitaiStream object + a buffer object to store 4 bytes (that would likely consume not less than 8 bytes of overhead) + some reuses of tuat stream object .... Feels like much (and unnecessarily) heavier than just manually converting unsigned into signed.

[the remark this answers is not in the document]
generalmimon: > Yeah, a KaitaiStream object + a buffer object to store 4 bytes (that would likely consume not less than 8 bytes of overhead) + some reuses of tuat stream object

The overhead will be insignificant when we implement data views / substreams (https://github.com/kaitai-io/kaitai_struct/issues/44).

log/gps/gps_mgl_enigma_waypoint.ksy
```diff
@@ -0,0 +1,148 @@
...
+        type: u1
+      - id: value
+        type: str
+        size: value_size
+      - id: garbage
+        type: str
+        size: max_size - value_size
+    instances:
+      value_size:
+        value: max_size > size ? size : max_size
```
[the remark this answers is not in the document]
KOLANICH: >“Don't care” values means that they should not be even parsed

They can contain garbage from address space of the software, that may potentially be used to for fun and profit unless properly wiped. That's why IMHO all the formats allowing such abuse should be specified the way highlighting it is possible and allowing the one parsing to check if the implementation is vulnerable and (when serialization is supported) wipe the garbage.

>Also, there isn't any reason why you should ignore invalid len_value greater than the maximum allowed size - that's just an unwanted error suppression. The PDF doc does not suggest doing this at all.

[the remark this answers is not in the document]
generalmimon: > They can contain garbage from address space of the software, that may potentially be used to for fun and profit unless properly wiped.

😆 You're really contradicting yourself, don't you see it? In one place (https://github.com/kaitai-io/kaitai_struct_formats/pull/381#discussion_r554417176) you care about parsing additional 4 bytes, and here you don't mind that you're unnecessarily wasting much more CPU time and resources (not only parsing a few dozens of bytes one-by-one but also interpreting the byte array as a string per character encoding) just _for fun_.

log/gps/gps_mgl_enigma_waypoint.ksy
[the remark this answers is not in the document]
KOLANICH: I got the response. It was very brief.

>I am sorry but we are not interested
>Best regards
>Rainier

So it seems they won't removd the bullshit from the license.

log/gps/gps_mgl_enigma_waypoint.ksy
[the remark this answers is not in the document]
KOLANICH: >You're really contradicting yourself, don't you see it? In one place (#381 (comment)) you care about parsing additional 4 bytes, and here you don't mind that you're unnecessarily wasting much more CPU time and resources (not only parsing a few dozens of bytes one-by-one but also interpreting the byte array as a string per character encoding) just for fun.

No.

* There we get an equivalent result more efficiently.
* Here we get different results.

>I can't imagine any profit from the analysis of junk.

Getting leaked parts of secrets maybe? Or maybe intentionally storing other info in the garbage area, this way the format is somehow forward-compatible.

PR #389, opened 2021-01-07
Added DeLorme gpl

KOLANICH:


log/gps/delorme_track_gpl.ksy
```diff
@@ -0,0 +1,67 @@
...
+        type: u4
+      - id: reserved1
+        type: u4
+        doc: Zero in all observed files
+
+enums:
+  fix:
+    0:
+      id: dupe
+      doc: duplicate of the previous record
```
dgelessus, line 63: How is this interpreted exactly? If a record has this type, are its contents ignored and taken from the previous record instead?

log/gps/delorme_track_gpl.ksy
```diff
@@ -0,0 +1,67 @@
...
+    1: none
+    2: two_d
+    3: three_d
+    4: dgps
```
dgelessus, line 67: This doesn't match the source code you linked, which maps 5 (not 4) to `fix_dgps`:

https://github.com/gpsbabel/gpsbabel/blob/e1e195beca8e0f24ca0e41c02d1986af13865cd0/delgpl.cc#L80-L82

```cpp
    case 5:
      wpt_tmp->fix = fix_dgps;
      break;
```

log/gps/delorme_track_gpl.ksy
```diff
@@ -0,0 +1,67 @@
...
+        type: f8
+        -unit: feet
+      - id: heading
+        type: f8
+        -unit: °
+      - id: speed
+        type: f8
+        -unit: ml/h
+      - id: timestamp
+        type: u4
```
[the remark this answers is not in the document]
generalmimon: I agree, IMHO the best would be to rename the `id` to `unix_timestamp` so that the meaning is clear even from the application code using the generated parser.

dgelessus:
PS: Like with #381, I think this spec would fit better under geospatial rather than log/gps.

I suppose this format could count as a log, because it's a sequence of timestamped positions/movement vectors. But still, it has more in common with other waypoint/route formats than with OS event logs.

generalmimon:
> I suppose this format could count as a log, because it's a sequence of timestamped positions/movement vectors.

It can use `/meta/tags` to denote that (and it will show also in the _Logs_ category at https://formats.kaitai.io/ then):

```ksy
meta:
  tags:
    - log
```

log/gps/delorme_track_gpl.ksy
KOLANICH: IDK in fact. GPSBabel verifies they are equal, but it feels like completely pointless. I feel like these may be angles of a rectangle.

KOLANICH:
I feel like `geospatial` is more for maps, but GPS tracks are more logs than maps, so go to `log/gps`.

PR #401, opened 2021-01-24
Added GBTD GBR0 ksy format

ZetaTwo:
As per: https://github.com/ZetaTwo/gameboy-project/issues/1 I'm making this PR.
It is a very niche file format but it is a fully working KSY implementation of it so at minimum it can serve as an example.

PR #402, opened 2021-01-24
Added GBMB GBR1 (broken) ksy format

ZetaTwo:
As per: https://github.com/ZetaTwo/gameboy-project/issues/1 I am opening this PR.
However, this KSY does not actually properly parse the file format since I have not found a good way to implement it.
The problem is described in: https://github.com/kaitai-io/kaitai_struct/issues/172

The problem is as follows. The file format consists of a sequence of "Objects". Each such object has (among other things) a Type, an ObjectID and a MasterID. If the MasterID is 0, it is a regular object, otherwise it is a "Sub-object" referencing its parent. In particular, there are two type: Map (type=2) and Map Tile Data (type=3) where the Map has two attributes: width and height and the size of the Map Tile Data object is determined by looking up its master object and repeating a record "width*height" times.

Basically, it would be required to write something like:
```
object_map_tile_data:
    seq:
      - id: master_id
        type: u2
      - id: data
        type: object_map_tile_data_record
        repeat: expr
        #repeat-expr: master.tile_count
        repeat-expr: 0
    instances:
      master:
        value: find_master(master_id)
```

How would you suggest I approach this with Kaitai currently?

### DIFF
--- a/game/gbtd_gbr1.ksy
+++ b/game/gbtd_gbr1.ksy
@@ -1,0 +1,156 @@
+meta:
+  id: gbr1
+  endian: le
+  encoding: ascii
+  title: GBM file
+  license: MIT
+  application: Gameboy Map Builder, GBMB
+  file-extension:
+    - gbm
+  tags:
+    - game
+
+doc: |
+  GBR1 files are used by Gameboy Map Builder, GBMB
+  to store maps created with the program.
+doc-ref: http://www.devrs.com/gb/hmgd/gbmb.html
+
+seq:
+  - id: magic
+    contents: [0x47, 0x42, 0x4F, 0x31] #GBO1
+  - id: objects
+    type: gbr1_object
+    repeat: eos
+
+types:
+  gbr1_object:
+    seq:
+      - id: marker
+        contents: [0x48, 0x50, 0x4a, 0x4d, 0x54, 0x4c] # HPJMTL
+      - id: object_type
+        type: u2
+      - id: object_id
+        type: u2
+      - id: master_id
+        type: u2
+      - id: crc
+        type: u4
+      - id: object_length
+        type: u4
+      - id: body
+        size: object_length
+        type:
+          switch-on: object_type
+          cases:
+            0x0001: object_producer
+            0x0002: object_map
+            0x0003: object_map_tile_data
+            0x0004: object_map_properties
+            0x0005: object_map_property_data
+            0x0006: object_map_default_property_value
+            0x0007: object_map_settings
+            0x0008: object_map_property_colors
+            0x0009: object_map_export_settings
+            0x000A: object_map_export_properties
+            0xFFFF: object_deleted
+    
+  object_producer:
+    seq:
+      - id: name
+        type: strz
+        size: 128
+      - id: version
+        type: strz
+        size: 10
+      - id: info
+        type: strz
+        size: 128
+
+  object_map:
+    seq:
+      - id: name
+        type: strz
+        size: 128
+      - id: width
+        type: u4
+      - id: height
+        type: u4
+      - id: prop_count
+        type: u4
+      - id: tile_file
+        type: strz
+        size: 256
+      - id: tile_count
+        type: u4
+      - id: prop_color_count
+        type: u4
+
+  object_map_tile_data:
+    seq:
+      - id: master_id
+        type: u2
+      - id: data
+        type: object_map_tile_data_record
+        repeat: expr
+        #repeat-expr: master.tile_count
+        repeat-expr: 0
+    instances:
+      master:
+      #  value: find_master(master_id) # <--- Doesn't work (https://github.com/kaitai-io/kaitai_struct/issues/172)
+        value: 0
+         
+
+  object_map_tile_data_record:
+    seq:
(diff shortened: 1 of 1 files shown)


game/gbtd_gbr1.ksy
```diff
@@ -0,0 +1,156 @@
...
+    seq:
+      - id: master_id
+        type: u2
+      - id: data
+        type: object_map_tile_data_record
+        repeat: expr
+        #repeat-expr: master.tile_count
+        repeat-expr: 0
+    instances:
+      master:
```
KOLANICH, line 98:
```yaml
-affected-by: 172
```

KOLANICH (commented):
The same issues as in #401 are also present here.

ZetaTwo:
Updated PR based on comments in #401

game/gbtd_gbr1.ksy
```diff
@@ -0,0 +1,171 @@
...
+    contents: GBO1
+  - id: objects
+    type: gbr1_object
+    repeat: eos
+
+types:
+  gbr1_object:
+    seq:
+      - id: marker
+        contents: [0x48, 0x50, 0x4a, 0x4d, 0x54, 0x4c] # HPJMTL
```
KOLANICH, line 29: `contents: HPJMTL`

ZetaTwo: Fixed

ZetaTwo:
Updated PR based on your comments.

game/gbtd_gbr1.ksy
```diff
@@ -69,103 +69,104 @@ types:
...
+          - id: master_id
+            type: u2
+          - id: data
+            type: map_tile_data_record
+            repeat: expr
+            #repeat-expr: master.tile_count
+            repeat-expr: 0
+        instances:
+          master:
+          #  value: find_master(master_id) # <--- Doesn't work (https://github.com/kaitai-io/kaitai_struct/issues/172)
```
KOLANICH, line 115: `-affected-by: 172`

ZetaTwo:
Made an implementation using opaque types in the meanwhile. It's not fantastic but it works.

game/gbtd_gbr1.ksy
```diff
@@ -4,6 +4,7 @@ meta:
   encoding: ascii
   title: GBM file
   license: MIT
+  ks-opaque-types: true
```
KOLANICH, line 7: It'd be nice to insert a link to `lookup_object` source code.

game/gbtd_gbr1.ksy
```diff
@@ -0,0 +1,297 @@
...
+              - id: reserved1
+                type: b3
+              - id: sgb_palette
+                type: b3
+              - id: reserved2
+                type: b1
+              - id: gbc_palette
+                type: b5
+              - id: tile_number
+                type: b10
```
generalmimon, line 120: Does this format really use **little-endian** byte-sized integers and [**big-endian** bit-sized integers](https://doc.kaitai.io/user_guide.html#bit-ints-be)? That would be something quite rare and uncommon. The ["bit endianness"](https://doc.kaitai.io/user_guide.html#bit-endian) is usually the same as byte endianness in the entire format. That's because the bitfields are mostly designed to be packed into a byte-aligned integer, which is first parsed in the same endianness as other byte-aligned integers use, and then the value of such integer is unpacked into the individual bit field members (bit-sized integers) using bitwise AND `&` and SHR `>>` operations.

And reading multi-byte integers in a non-native endianness is mostly less efficient than in the native endianness (since [byte swapping](https://en.wikipedia.org/wiki/Endianness#Byte_swapping) has to be done), so that's something file format authors keep in mind and try to avoid that. Therefore mixing BE and LE integers in the same format does not make sense.

KOLANICH: I think it was a big mistake to introduce byte-swapping for it instead of relying on a bit ranges remaper.

game/gbtd_gbr1.ksy
ZetaTwo: Added link to my implementation in the doc attribute.

game/gbtd_gbr1.ksy
ZetaTwo: I'm not sure I completely follow here but the current implementation does produce the correct result. Running this test script: https://github.com/ZetaTwo/gameboy-project/blob/master/gbm-parser/tester.py with the test file in that repo gives the correct tile_number values:

```
...

Type: Type.map_tile_data, ID: 2
Size: 429
MapTileData
	Num records: 143
	MapTileData[0].tile_number = 0
	MapTileData[1].tile_number = 1
	MapTileData[2].tile_number = 2
	MapTileData[3].tile_number = 3
	MapTileData[4].tile_number = 4
	...
```

If they were the wrong endianness I would expect it to be something like: "0, 512, 256, 768, ...", right?

ZetaTwo:
Addressed both comments.

ZetaTwo:
Can this be merged?

KOLANICH:
Have you used the wrong order of arguments of `git rebase` command?

ZetaTwo:
I did something wrong indeed but I believe I hade sorted it out now.

PR #403, opened 2021-01-24
Add yodesk.ksy

ghost:
As suggested in https://github.com/cyco/WebFun/issues/3 this PR adds a spec for the file format used by [Yoda Stories](https://en.wikipedia.org/wiki/Star_Wars:_Yoda_Stories). The definition contains a few fields whose purpose is unclear to me (e.g. https://github.com/cyco/kaitai_struct_formats/blob/add-yodesk/game/yodesk.ksy#L203) and there's some unusual error handling that could probably be improved but I don't know how (https://github.com/cyco/kaitai_struct_formats/blob/add-yodesk/game/yodesk.ksy#L36). Apart from that, I tried to follow the [style guide](https://doc.kaitai.io/ksy_style_guide.html).

The game [Indiana Jones and His Desktop Adventures](https://en.wikipedia.org/wiki/Indiana_Jones_and_His_Desktop_Adventures) uses a very similar file format (a predecessor) to this one. As far as I can tell it mostly differs in the size of some integers. When this PR goes through I can open another one for that game.

I'm looking forward to your feedback on this one.


game/yodesk.ksy
```diff
@@ -0,0 +1,414 @@
...
+        size: 0x18
+        if: tile_id != -1
+  tiles:
+    seq:
+      - id: size
+        type: u4
+      - id: tiles
+        type: tile
+        repeat: expr
+        repeat-expr: size / ( 32 * 32 + 4)
```
KOLANICH, line 83: ```sizeof<tile>```, but not the best way. A better way is `size: size` + `repeat: eos`

game/yodesk.ksy
```diff
@@ -0,0 +1,414 @@
...
+      - id: type
+        type: u4
+      - id: shared_counter
+        type: u2
+        # contents: 0xFF FF
+      - id: planet_again
+        type: u2
+      - id: tile_ids
+        type: u2
+        repeat: expr
```
KOLANICH, line 163: What is it?

game/yodesk.ksy
```diff
@@ -0,0 +1,414 @@
...
+      - id: weapons
+        type: character_weapon
+        repeat: until
+        repeat-until: _.index == -1
+  character_weapon:
+    seq:
+      - id: index
+        type: s2
+      - id: reference
+        # if character is monster this is a reference to the their weapon (char idx)
```
KOLANICH, line 386: `doc:`

game/yodesk.ksy
generalmimon: @KOLANICH This comment seems to be misplaced, it probably relates to the line below, right?

https://github.com/kaitai-io/kaitai_struct_formats/blob/e18654a2f42b9177449738f9f908708b927f8e62/game/yodesk.ksy#L164

KOLANICH: Yes. IDK why. This is not the first time this happens.

game/yodesk.ksy
```diff
@@ -0,0 +1,414 @@
...
+        type: u4
+      - id: shared_counter
+        type: u2
+        # contents: 0xFF FF
+      - id: planet_again
+        type: u2
+      - id: tile_ids
+        type: u2
+        repeat: expr
+        repeat-expr: 3 * width * height
```
KOLANICH, line 164: Does 3 mean count of channels? If so, what is the order?

game/yodesk.ksy
cyco: I'm now using `repeat: eos`, thanks for the tip. Adding an additional layer to get the sub-stream right took me a little while (even though it's properly document). Accessing a specific tile from code is a little ridiculous with all the nested structures:

Screenshot 2021-01-24 at 17 22 31

```javascript
const tileCategory = catalog.entries.find(entry => entry.type === "TILE");

tileCategory.content.tiles.tiles.tiles[xx];
```

game/yodesk.ksy
cyco: I added a `doc:` comment to the field, explaining that each map has 3 tile layers. Hope that clears things up.

game/yodesk.ksy
cyco: Fixed

cyco:
Thanks for your feedback so far. I added a bunch of enums for `type` fields. What do you think?

game/yodesk.ksy
KOLANICH: It may make sense to encode it in KS.

game/yodesk.ksy
```diff
@@ -0,0 +1,479 @@
...
+    14: lose
+    15: trade
+    16: use
+    17: find
+    18: find_unique_weapon
+  character_type:
+    1: hero
+    2: enemy
+    4: weapon
+  hotspot_type:
```
KOLANICH, line 457: If a enum is used only within a certain class, it can be moved there.

game/yodesk.ksy
```diff
@@ -0,0 +1,479 @@
...
+      - id: instruction_count
+        type: u2
+      - id: instructions
+        type: action_item
+        repeat: expr
+        repeat-expr: instruction_count
+  action_item:
+    seq:
+      - id: opcode
+        type: u2
```
KOLANICH, line 116: Are opcodes known? Should it be a enum?

KOLANICH (approved):

game/yodesk.ksy
cyco: In bb5260b26a90b0d57ce457cf4ded244626183f7e I moved all enums to the classes where they are used.

game/yodesk.ksy
cyco: I split `action_item` into `instruction` and `condition` and added enums for the respective opcodes.

game/yodesk.ksy
cyco: Can you take a look at cc51e05a53dd16ebcfbeff193aafc1436ad4d789 and let me know if that is what you had in mind?

KOLANICH: Reviewed.

game/yodesk.ksy
```diff
@@ -0,0 +1,582 @@
...
+      - id: width
+        type: u2
+      - id: height
+        type: u2
+      - id: type
+        enum: zone_type
+        type: u4
+      - id: shared_counter
+        type: u2
+        # contents: 0xFF FF
```
KOLANICH, line 257: BTW, we have `valid: `

KOLANICH (commented):
>Thanks for your feedback so far. I added a bunch of enums for type fields. What do you think?

:+1:

cyco:
Alright then. I think this is ready to be merged.

game/yodesk.ksy
```diff
@@ -254,7 +254,7 @@ types:
         type: u4
       - id: shared_counter
         type: u2
-        # contents: 0xFF FF
+        valid: 0xFF FF
```
KOLANICH, line 257: Is it really recognized as a number?

generalmimon: > Is it really recognized as a number?

Two answers:

1. in the badly-written YAML parser (i.e. https://github.com/jeremyfa/yaml.js) used in the [Web IDE](https://ide.kaitai.io/devel/): **yes**,
2. using a specification-compliant parser like https://yaml-online-parser.appspot.com/: **no** (`valid: 0xFF FF` will be interpreted as `{ "valid": "0xFF FF" }`, and this will fail in the expression parser with an exception `parsing expression '0xFF FF' failed on 1:6`).

And yeah, we should really at least switch to a better parser in the Web IDE.

generalmimon: The fix would be

```suggestion
        valid: 0xFF_FF
```

cyco: Good catch! I was indeed using the web IDE to test these changes (and loved it btw). Too bad its parser is off.

game/yodesk.ksy
```diff
@@ -0,0 +1,581 @@
...
+      - id: size
+        type: u4
+      - id: characters
+        type: character
+        repeat: until
+        repeat-until: _.index == -1
+  character:
+    seq:
+      - id: index
+        type: s2
```
KOLANICH, line 474: Do other "negative" indices make sense? Or are they positive and just 0xFFFF is special?

cyco: Throughout the whole file signed integers could be considered unsigned with 0xFF FF indicating an empty value. The original game treats alls values expect u4 as signed integers, often checking `x < 0` to determine if an id should be resolvable or if an index marks the end of an array. 

So I think one could argue that the file format contains unsigned values treating 0xFFFF as a special case and the original game has a _bug_ in its parsing code.

game/yodesk.ksy
```diff
@@ -0,0 +1,565 @@
...
+      - id: content
+        type: strz
+        size: length
+  waypoint:
+    seq:
+    - id: x
+      type: u4
+    - id: y
+      type: u4
+
```
generalmimon, line 565: Unnecessary double final newline.

```suggestion
```

game/yodesk.ksy
```diff
@@ -0,0 +1,564 @@
...
+          type: u4
+        - id: pixels
+          size: 32 * 32
+  action:
+    seq:
+      - id: marker
+        contents: "IACT"
+      - id: size
+        type: u4
+      - id: condition_count
```
generalmimon, line 95: `num_conditions` (https://doc.kaitai.io/ksy_style_guide.html#attr-id):

> For fields that designate **number / count** of something (in particular, number of repetitions of some other structure), use `num_` prefix and plural form — i.e. `num_questions`, `num_blocks`, `num_nodes`

game/yodesk.ksy
```diff
@@ -0,0 +1,564 @@
...
+        enum: condition_opcode
+      - id: arguments
+        type: s2
+        repeat: expr
+        repeat-expr: 5
+      - id: text_length
+        type: u2
+      - id: text
+        type: str
+        size: text_length
```
generalmimon, line 120:
```suggestion
      - id: len_text
        type: u2
      - id: text
        type: str
        size: len_text
```

See https://doc.kaitai.io/ksy_style_guide.html#attr-id:

> For fields that designate **size** of some particular data structure (in bytes or some other fixed units), use `len_` prefix and name of that data structure — i.e. `len_block` (length of a single `block` entry), `len_blocks` (total length of whole `blocks` array, made of `block` entries).

game/yodesk.ksy
cyco: Thanks again for your feedback. I replaced all `_count` suffixes with `num_` prefixes.

game/yodesk.ksy
```diff
@@ -0,0 +1,564 @@
+meta:
+  id: yodesk
+  application: "Star Wars: Yoda Stories"
+  file-extension: dta
+  license: CC0-1.0
+  ks-version: 0.9
+  endian: le
+  encoding: ASCII
+seq:
```
generalmimon, line 9: Would you add at least a few sentences about the format in [the `/doc` key](https://doc.kaitai.io/ksy_style_guide.html#doc), i.e. some general information about the format? What it is used for? What kind of data it stores? What is the main usage field?

cyco: It took me a little while, but today I finally got around to adding the comments you requested.

game/yodesk.ksy
```diff
@@ -1,564 +1,923 @@
...
+    tiles_entries:
+      seq:
+        - id: tiles
+          type: tile
+          repeat: eos
+    tile:
+        seq:
+          - id: attributes
+            doc: |
+              Bit mask of various tile attributes. Most bits are purely informational and do not have any effect on game play.
```
KOLANICH: Why not to specify it in KS using bit-sized types instead of putting that info into `doc`?

game/yodesk.ksy
```diff
@@ -0,0 +1,1034 @@
...
+    application: "Star Wars: Yoda Stories"
+    file-extension: dta
+    license: CC0-1.0
+    ks-version: 0.9
+    endian: le
+    encoding: ASCII
+  doc: |
+    [Star Wars: Yoda Stories](https://en.wikipedia.org/wiki/Star_Wars:_Yoda_Stories) is a unique tile based game with procedurally
+    generated worlds.
+    This spec describes the file format used for all assets of the windows
```
generalmimon, line 12: Should it be "**Windows**"?


```suggestion
    This spec describes the file format used for all assets of the Windows
```

game/yodesk.ksy
generalmimon: I agree with @KOLANICH. This is a part of the format structure and should be described by KSY. 

@cyco The best way to do that is to use little-endian bit-sized integers (https://doc.kaitai.io/user_guide.html#bit-ints-le). In short, it's about adding this into the top-level `meta` section

```diff
   ks-version: 0.9
+  bit-endian: le
   endian: le
   encoding: ASCII
```

and then you just start writing the bit-sized members in the order from the least significant bit in the bitfield ("bit 0"):

```ksy
types:
  flags:
    seq:
      - id: is_opaque # Bit 0
        type: b1
      - id: is_floor # Bit 1
        type: b1
      - id: is_object # Bit 2
        type: b1
      # etc. 
```

cyco: Hey, I've updated the PR. Tile attributes are now encoded using bit sized integers as you suggested. Please take a look and let me know what you think.

My editor somehow messed up the indentation of the file, I hope the large diff is not a problem.

cyco:
@generalmimon Just wanted to check in to see if this is still on your radar. We've all spent a lot of time on this PR and it would be a shame not to merge it.

I also have another one coming up https://github.com/cyco/kaitai_struct_formats/blob/add-deskadv/game/deskadv.ksy

game/yodesk.ksy
```diff
@@ -0,0 +1,1190 @@
...
+        type: u2
+        if: index != 0xFF_FF
+  # Utilities
+  prefixed_str:
+    seq:
+      - id: len_content
+        type: u2
+      - id: content
+        type: str
+        size: len_content
```
generalmimon, line 1177: Please put `size` above `type` (https://doc.kaitai.io/ksy_style_guide.html#attr-id):

```suggestion
      - id: content
        size: len_content
        type: str
```

The same for all such `size:` occurrences in the file.

game/yodesk.ksy
```diff
@@ -0,0 +1,1190 @@
...
+
+seq:
+  - id: catalog
+    type: catalog_entry
+    repeat: until
+    repeat-until: _.type == "ENDF"
+
+types:
+  catalog_entry:
+    seq:
```
generalmimon, line 40: This helps debugging in the [Web IDE](https://ide.kaitai.io/devel/):

```suggestion
  catalog_entry:
    -webide-representation: '{type}'
    seq:
```

generalmimon (changes requested):
@cyco

> @generalmimon Just wanted to check in to see if this is still on your radar. We've all spent a lot of time on this PR and it would be a shame not to merge it.

Actually I moved to other things, I'm glad that you remind this. Let's try to finish it...

1. Do you have sample files for testing? Could you put links to them in the `/doc` block?

2. What are the sources that you used as a reference? Are there any official/unofficial descriptions (specifications), forum threads / blog posts discussing the structure or existing open source parsers/generators that you could link in `doc-ref`?

3. Can we choose a better name than `yodesk`? It doesn't matter if the identifier is a bit longer (I'd say that for example [`battery_management_system_protocol`](https://github.com/kaitai-io/kaitai_struct_formats/pull/410/files#diff-e1504689828f54de614eed5ff6e83418386ce288fcdcebce3fcfb13796d10810R2) is still OK, although expanding established abbreviations does not necessarily make sense), but it should be descriptive enough so that people with no background have some idea what format it is just from the name.

   Something like `starwars_yodesk` maybe? It is OK to use short names for a `.ksy` in your own repo if you know what it is, but in a general-purpose repo like the format gallery, we need to be more specific and provide some context. The problem is that `yodesk` is not much widespread name, so people would usually have no idea.

game/yodesk.ksy
```diff
@@ -0,0 +1,1190 @@
...
+          doc: |
+            Tile forms part of a character
+        - id: unused
+          type: b7
+        # Floor flags
+        - id: is_doorway
+          type: b1
+          doc: |
+            These tiles are doorways, monsters can't go
+          if: is_floor
```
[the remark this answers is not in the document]
cyco: I think it's not that clear cut. Bits `0` and `3` can be set independently so we'd have to logical and the value with `0x1f6` before looking up the enum field.

game/yodesk.ksy
```diff
@@ -0,0 +1,1190 @@
...
+        repeat: until
+        repeat-until:  _.tile_id == 0xFF_FF
+  tile_name:
+    seq:
+      - id: tile_id
+        type: u2
+      - id: name
+        type: strz
+        size: 0x18
+        if: tile_id != 0xFF_FF
```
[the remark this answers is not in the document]
cyco: I replaced some occurrences of the pattern with an instance in a6119768d6b45a3f5b3da65f78cc2d295ccb9879. Is there a way to globally define the instance? I don't really think the additional abstraction is justified if only on field uses the check.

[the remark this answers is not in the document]
generalmimon: > I don't really think the additional abstraction is justified if only on field uses the check.

Right, this is focused mainly on places where the check is repeated multiple times, as for example [in the `character` type](https://github.com/kaitai-io/kaitai_struct_formats/blob/2e2b4ea332c84c6ec5c61165cb0218518f867675/game/yodesk.ksy#L1083-L1119). If it only used once (maybe twice), I'm happy with leaving it in `if`.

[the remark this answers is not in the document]
generalmimon: Hmm, I've just realized that this comment of mine is pretty much useless because I suggest to pull the adjacent fields with `if: index != 0xFF_FF` in the [very next comment](https://github.com/kaitai-io/kaitai_struct_formats/pull/403#discussion_r662628427). You've done the `character_body` change too (which is a better way than using the value instance `is_end`) so I'd say feel free to ignore this.

game/yodesk.ksy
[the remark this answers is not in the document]
generalmimon: > Bits `0` and `3` can be set independently so we'd have to logical and the value with `0x1f6` before looking up the enum field.

OK, thanks for clarifying. You can do that like this:

```ksy
seq:
  # ...
  - id: type_flags
    type: b8
  # ...
instances:
  tile_type:
    value: type_flags & 0b1111_0110
    enum: tile_types
  is_floor:
    value: (type_flags & 0b0000_0001) != 0
  is_roof:
    value: (type_flags & 0b0000_1000) != 0
```

game/yodesk.ksy
```diff
@@ -0,0 +1,1190 @@
...
+        0x8:
+          id:  random_is_less_than
+          doc: Current zone's `random` value is less than `args[0]`
+        0x9:
+          id:  enter_by_plane
+        0xa:
+          id:  tile_at_is
+          doc: |
+            Check if tile at `args[0]`x`args[1]`x`args[2]` is equal to
+            `args[3]`
```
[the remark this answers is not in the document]
cyco: I absolutely love this suggestion! It seems so obvious in hindsight.

game/yodesk.ksy
```diff
@@ -0,0 +1,1190 @@
...
+        0x20:
+          id:  random_is_not
+          doc: Current zone's `random` value is not equal to `args[0]`
+        0x21:
+          id:  shared_counter_is_not
+          doc: Current zone's `shared_counter` value is not equal to `args[0]`
+        0x22:
+          id:  is_variable
+          doc: |
+            Check if variable identified by `args[0]`⊕`args[1]`⊕`args[2]` is
```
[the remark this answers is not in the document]
cyco: This was actually supposed to be a concatenation. I've replaced it with tuple notation instead.

cyco:
Here we go again :) I hope I didn't miss anything.

> 1. Do you have sample files for testing? Could you put links to them in the `/doc` block?

Sadly no. Some versions of the game are available on archive.org and various abandonware sites, but nothing I'd feel comfortable linking to.

> 2. What are the sources that you used as a reference? Are there any official/unofficial descriptions (specifications), forum threads / blog posts discussing the structure or existing open source parsers/generators that you could link in `doc-ref`?

This file is solely based on my own reverse engineering efforts for https://github.com/cyco/webfun. Aside from the parser implementation at https://github.com/cyco/WebFun/tree/master/src/engine/file-format there's no documentation on the file format in the project yet. As far as I can tell there's also no official documentation available. Zach Barth wrote an excellent piece on reverse engineering the format back in 2014 (phew, long time ago) here https://www.trashworldnews.com/yoda-stories/ but I'm hesitant to use that article as `doc-ref` as it is incomplete and inaccurate at times.

> 3. Can we choose a better name than `yodesk`? It doesn't matter if the identifier is a bit longer (I'd say that for example [`battery_management_system_protocol`](https://github.com/kaitai-io/kaitai_struct_formats/pull/410/files#diff-e1504689828f54de614eed5ff6e83418386ce288fcdcebce3fcfb13796d10810R2) is still OK, although expanding established abbreviations does not necessarily make sense), but it should be descriptive enough so that people with no background have some idea what format it is just from the name.
>    Something like `starwars_yodesk` maybe? It is OK to use short names for a `.ksy` in your own repo if you know what it is, but in a general-purpose repo like the format gallery, we need to be more specific and provide some context. The problem is that `yodesk` is not much widespread name, so people would usually have no idea. 

 I felt that `starwars` was too generic, the game is called _Star Wars: Yoda Stories_ or just _Yoda Stories_ – so I've renamed the ksy to `yoda_stories_dta` following the pattern used for other files here (e.g. `gran_turismo_vol.ksy`).

generalmimon (changes requested):
These comments remained unpublished in the _Pending_ state for way too long (because I was probably hoping to add something but then got distracted by something else and completely forgot about them), so I'll at least publish them now 😐

PR #404, opened 2021-01-24
Add Minecraft NBT format

generalmimon:
Closes https://github.com/kaitai-io/kaitai_struct_formats/issues/360


game/minecraft_nbt.ksy
```diff
@@ -0,0 +1,221 @@
...
+      - id: tags
+        type: s8
+        repeat: expr
+        repeat-expr: num_tags
+    instances:
+      tags_type:
+        value: 'tag::tag_long'
+enums:
+  tag:
+    0: tag_end
```
KOLANICH, line 209: How about removing the prefix?

KOLANICH:
I feel like it is more `serialization` than `game`

game/minecraft_nbt.ksy
generalmimon: > How about removing the prefix?

Sure, a good idea. I was afraid that using enum keys like `byte`, `int`, `long`, etc. would cause clashes with statically-typed target language keywords, but I've just checked all of them and surprisingly, it only caused a problem in Lua with keyword `end` (https://github.com/kaitai-io/kaitai_struct/issues/90#issuecomment-766440975). But it's a bug in KSC, not this format, and it can be fixed without having to mangle identifiers (you just replace `end` to `["end"]`). So let's go with it.

generalmimon:
> I feel like it is more `serialization` than `game`.

In terms of format structure, you're totally right. However, I think it is more important what the format is used for. The NBT format is specially designed for Minecraft and meant to exclusively store and transfer only Minecraft-related data. I don't think that this format is used in practice as a container for another type of data, although technically it is of course possible.

Moreover, you should admit that the internal structure doesn't matter, as far as only the visible and practical result is concerned. Sure, Minecraft developers chose to use this generic flexible serialization format due to a large variety of data layouts they need to store. But if they had only chosen to go with a bunch of fixed-size chunk types which are not flexible at all, nothing would have changed from the practical perspective (i.e. it would still be a format exclusively used by Minecraft), but it would have no longer been a `serialization` format.

For these reasons, `serialization` is not the first compartment that comes in your mind when searching for a NBT parser. You know why you want to parse NBT tags (probably to make some use of the data created by Minecraft - so you naturally have it associated with the `game`), but you have no idea what the format looks like from inside - only when you do, perhaps you start to think that it has a lot in common with `serialization` formats.

KOLANICH:
>You know why you want to parse NBT tags (probably to make some use of the data created by Minecraft - so you naturally have it associated with the game), but you have no idea what the format looks like from inside - only when you do, perhaps you start to think that it has a lot in common with serialization formats.

And here we take into account the effect KS would do on formats once seralization is implemented. KS gonna make all the formats specified using it easily accessible to everyone using one of the supported languages.  So noone gonna care if this format was for originally minecraft, if devs find it superrior ot BSON/MSGPack/CBOR/anything else for their application, they would use it.

game/minecraft_nbt.ksy
generalmimon: @KOLANICH Solved in https://github.com/kaitai-io/kaitai_struct_formats/pull/404/commits/3f45153a0c9698f03e3ec92f8e284744ff03ab93.

game/minecraft_nbt.ksy
```diff
@@ -0,0 +1,221 @@
...
+        if: not is_root
+        doc: do not read this property, access `type` instead
+      - id: name
+        type: tag_string
+        if: not is_tag_end
+      - id: payload
+        type:
+          switch-on: type
+          cases:
+            'tag::tag_byte': s1
```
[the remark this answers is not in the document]
generalmimon: > BTW, `'` seem to be non-mandatory

Perhaps. It's just a personal preference.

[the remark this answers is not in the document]
generalmimon: > BTW, `'` seem to be non-mandatory

Resolved in https://github.com/kaitai-io/kaitai_struct_formats/pull/404/commits/f9750e94b85a23bde1d0e9f21346720f48635081

game/minecraft_nbt.ksy
```diff
@@ -0,0 +1,229 @@
...
+  - id: root
+    type: named_tag(true)
+types:
+  named_tag:
+    -webide-representation: 'TAG_{type}("{name.data}"): {payload:dec}'
+    params:
+      - id: is_root
+        type: bool
+    seq:
+      - id: root_type
```
KOLANICH, line 98: Feels very incorrect to introduce a param that will always be false. It would eat memory for each instance, but will be really useful only for one. IMHO it may be better to validate anyhow in other way.

Proposal: create an `instance` in the root called valid and not to validate it, since validation of instances is not yet implemented.

generalmimon: > validation of instances is not yet implemented

It _is_ implemented for more than a year (according to this comment https://github.com/kaitai-io/kaitai_struct/issues/435#issuecomment-557963004), see the test `ValidFailInst` at our [CI dashboard](https://ci.kaitai.io/) (the corresponding format spec is [`valid_fail_inst`](https://github.com/kaitai-io/kaitai_struct_tests/blob/01e443a844f86ae192e00eeb947f808d93a7ee05/formats/valid_fail_inst.ksy)).

I'll use an instance then, that really seems to be the best solution.

KOLANICH: >It is implemented for more than a year

Nice then. The last time I have specified it it was broken, but I remember it was closer than year ago.

generalmimon: Resolved in https://github.com/kaitai-io/kaitai_struct_formats/pull/404/commits/4c7af9b056524055710cdfed570b712e500bdfac

game/minecraft_nbt.ksy
```diff
@@ -0,0 +1,225 @@
...
+  - https://web.archive.org/web/20110723210920/https://www.minecraft.net/docs/NBT.txt
+  - https://minecraft.gamepedia.com/NBT_format
+seq:
+  - size: 0
+    if: root_type == tag::end # always `false`, but it doesn't matter
+    doc: Invoke `valid` check before we start parsing
+  - id: root
+    type: named_tag
+instances:
+  root_type:
```
KOLANICH, line 95: I meant

```yaml
instances:
  is_valid:
    value: root.type == tag::compound
    valid: true
```

and it is likely it still won't be compiled.

generalmimon: @KOLANICH:

> I meant
> 
> ```yaml
> instances:
>   is_valid:
>     value: root.type == tag::compound
>     valid: true
> ```
> 
> and it is likely it still won't be compiled.

Yes, it does not work on `value` instances.

game/minecraft_nbt.ksy
```diff
@@ -0,0 +1,229 @@
...
+  - https://minecraft.gamepedia.com/NBT_format
+seq:
+  - id: root
+    type: named_tag(true)
+types:
+  named_tag:
+    -webide-representation: 'TAG_{type}("{name.data}"): {payload:dec}'
+    params:
+      - id: is_root
+        type: bool
```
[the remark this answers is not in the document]
dgelessus: (This is outdated and only got submitted because I already wrote the comment before the code was changed. My next review comment is relevant to the current state of this check.)

game/minecraft_nbt.ksy
KOLANICH: Would be nice, if it is possible to do it.

game/minecraft_nbt.ksy
```diff
@@ -0,0 +1,229 @@
...
+  languages, and thus one must use external libraries to achieve a fully-compliant
+  decoder. But decoding in standard UTF-8 is still better than nothing, and
+  it usually works fine.
+
+  All Unicode codepoints with incompatible representations in MUTF-8 and UTF-8 are
+  U+0000 (_NULL_), U+D800-U+DFFF (_High_ and _Low Surrogates_) and U+10000-U+10FFFF
+  (_Supplementary_ Planes; includes e.g. emoticons, pictograms). When a _MUTF-8_-encoded
+  string is decoded as UTF-8, every such code point will be replaced with
+  2 (for _NULL_ character), 4 (for _Surrogates_) or 6 (for _Supplementary_ characters)
+  "�" symbols (i.e. U+FFFD _Replacement Character_).
```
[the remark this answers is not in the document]
KOLANICH: BTW, in kaitaiStructCompile.py I had to implement a pastprocess pass, to patch some decoding to tolerate errors.

game/minecraft_nbt.ksy
```diff
@@ -0,0 +1,225 @@
...
+            tag::int: s4
+            tag::long: s8
+            tag::float: f4
+            tag::double: f8
+            tag::byte_array: tag_byte_array
+            tag::string: tag_string
+            tag::list: tag_list
+            tag::compound: tag_compound
+            tag::int_array: tag_int_array
+            tag::long_array: tag_long_array
```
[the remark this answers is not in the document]
KOLANICH: BTW, how about introducing a special reserved type that will never be generated, but that can be used for code self-documentation to explicitly mark that this `case` is taken in account (and it may be possible to emit a warning on switches on enums if not all cases are covered)?

game/minecraft_nbt.ksy
generalmimon: > Why not just put the `valid` attribute directly on the `root` field like this?

This is the first thing which occurred to me, but the problem is that it makes the check execute only when the entire file is parsed, and that is pretty late to realize that all the parsing was unnecessary, because it's apparently a completely different format.

game/minecraft_nbt.ksy
```diff
@@ -0,0 +1,225 @@
...
+  tag_compound:
+    -webide-representation: '{dump_num_tags:dec} entries'
+    seq:
+      - id: tags
+        type: named_tag
+        repeat: until
+        repeat-until: _.is_tag_end
+    instances:
+      dump_num_tags:
+        value: 'tags.size - (tags.last.is_tag_end ? 1 : 0)'
```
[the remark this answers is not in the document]
generalmimon: > Is it even allowed to have a compound that doesn't end with an end tag?

No, I'm pretty sure it isn't. There is no way to achieve that: `repeat-until: _.is_tag_end` is correct, meaning that encountering a `TAG_End` is the only valid way to stop parsing `tags`.

The [spec](https://web.archive.org/web/20110723210920/https://www.minecraft.net/docs/NBT.txt) makes it pretty clear:

```
    TYPE: 10 NAME: TAG_Compound
    Payload: A sequential list of Named Tags. This array keeps going until a TAG_End is found.
             TAG_End end
```

The `(tags.last.is_tag_end ? 1 : 0)` thing is just a kind of defensive programming: after spending once several hours at work trying to figure out that some badly-written Wordpress plugin was trying to strip a `https://...` prefix from the beginning of a `http://...` URL using `substring` (but the code didn't check whether the URL really begins with that prefix and it didn't, resulting in that the `substring` ate up 1 extra character), I realized that I don't ever want to make bugs like that. If the assumptions are always correct, then great, it works. But doing a check just to be sure never harms anything (well, probably takes around 1-4 ticks of CPU time) and it's quicker to add it than examine all possible edge cases to find some 100% working assumptions that _might_ lead to code simplification. Moreover, only that an assumption is true for the current state of code doesn't mean that it will remain 100% working in the future when the code is changed.

> Even if a missing end tag is allowed, this check won't work if the compound has no contents at all - it would try to access the last element of an empty array.

Good point, but that does not apply here I believe. The `tags` array cannot end up empty after proper parsing, because it is delimited by `repeat-until` which is a 1+ repetition (the `do { ... } while (...)` statement is usually used; it unconditionally parses the 1st element and then decides whether to parse the 2nd or not).

But yeah, true, I probably should take this case into account too - when we get serialization, it _will_ be possible to make any array empty, and then it could fail.

game/minecraft_nbt.ksy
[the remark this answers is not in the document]
generalmimon: > In the case of `tag::end` there is also no payload, right? I would add an explicit `if: not is_tag_end` to this field as well, to make that explicit

Good idea. Added in https://github.com/kaitai-io/kaitai_struct_formats/pull/404/commits/3a3d3cca060b9b0ac0d0646a6f8d4f33eef749d2.

game/minecraft_nbt.ksy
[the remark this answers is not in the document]
generalmimon: @dgelessus:

> I would shorten this and only say something like:

Much better, thanks!

game/minecraft_nbt.ksy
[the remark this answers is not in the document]
generalmimon: And I believe there is a yet another advantage of `value: 'tags.size - (tags.last.is_tag_end ? 1 : 0)'` over `value: tags.size - 1`: the former makes it explicit what we are removing by doing `- 1`.

game/minecraft_nbt.ksy
```diff
@@ -0,0 +1,229 @@
...
+  **Implementation note:** strings in `TAG_String` are incorrectly decoded with
+  standard UTF-8, while they are encoded in [**Modified UTF-8**](
+    https://docs.oracle.com/javase/8/docs/api/java/io/DataInput.html#modified-utf-8
+  ) (MUTF-8). That's because MUTF-8 is usually not supported natively by most target
+  languages, and thus one must use external libraries to achieve a fully-compliant
+  decoder. But decoding in standard UTF-8 is still better than nothing, and
+  it usually works fine.
+
+  All Unicode codepoints with incompatible representations in MUTF-8 and UTF-8 are
+  U+0000 (_NULL_), U+D800-U+DFFF (_High_ and _Low Surrogates_) and U+10000-U+10FFFF
```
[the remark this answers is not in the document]
generalmimon: Resolved in https://github.com/kaitai-io/kaitai_struct_formats/pull/404/commits/852451a3b1c473bd155783db5dbc7cdabed97970

KOLANICH (approved):

game/minecraft_nbt.ksy
```diff
@@ -0,0 +1,227 @@
...
+    * <https://github.com/twoolie/NBT/blob/f9e892e/tests/world_test/data/scoreboard.dat>
+    * <https://github.com/chmod222/cNBT/tree/3f74b69/testdata>
+    * <https://github.com/PistonDevelopers/hematite_nbt/tree/0b85f89/tests>
+doc-ref:
+  - https://wiki.vg/NBT
+  - https://web.archive.org/web/20110723210920/https://www.minecraft.net/docs/NBT.txt
+  - https://minecraft.gamepedia.com/NBT_format
+seq:
+  - size: 0
+    if: root_type == tag::end # always `false`, but it doesn't matter
```
dgelessus, line 91: Thinking about this again, would it make sense to move the `valid` check to here:

```suggestion
    valid:
      expr: root_type == tag::compound
```

and remove the `valid: tag::compound` below? This makes the intent (checking the root type at exactly this point in parsing) more obvious and avoids having to use a nonsense condition like this to force evaluation of an instance.

Or if you want to keep it this way, please change the condition to

```suggestion
    if: root_type == tag::end or true # Force evaluation of root_type's valid check
```

so that it doesn't matter if the nonsense condition is true or false.

generalmimon: > Thinking about this again, would it make sense to move the `valid` check to here [...] and remove the `valid: tag::compound` below?

I'm OK with that. It does the job, and yeah, you're right that it makes the weird `size: 0` member less cryptic and hacky.

generalmimon: Addressed in https://github.com/kaitai-io/kaitai_struct_formats/pull/404/commits/e37fbbbd100a1279151fe1bc83606c110e32dfb5 and it really simplifies things a lot! Thanks for this good suggestion.

game/minecraft_nbt.ksy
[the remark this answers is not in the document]
generalmimon: > But yeah, true, I probably should take this case into account too - when we get serialization, it _will_ be possible to make any array empty, and then it could fail.

Solved in https://github.com/kaitai-io/kaitai_struct_formats/pull/404/commits/1bb30188421aab44d49fe2d2934612f7a8c72630.

@dgelessus Are you OK with this approach or do you find these checks confusing?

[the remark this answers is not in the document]
dgelessus: Now that I know the reason behind it, it makes sense 🙂 Normally I would put this sort of check in an `assert` (to make it clear that it should really always be true), but that doesn't exist in KS, so this is a reasonable workaround.

dgelessus (approved):
LGTM, thanks!

game/minecraft_nbt.ksy
generalmimon: Well, I'm not sure now. I have some new findings about this snippet:

```ksy
  - size: 0
    valid:
      expr: root_type == tag::compound
```

1. `valid/expr` in unnamed fields (i.e. without `id`) will apparently not work properly in any target language. For some reason, the compiler generates this kind of code everywhere:

   ```javascript
     MinecraftNbt.prototype._read = function() {
       this._unnamed0 = this._io.readBytes(0);
       var _ = this.unnamed0;
   ```

   Notice that the property is named <code><b>_</b>unnamed0</code>, but the assignment to the underscore (`_`) variable references `unnamed0` without the underscore prefix. So I guess the code won't even compile in static-typed languages, and in dynamic-typed it'll work only as long as the `_` alias is not used in the expression.

2. The spec using `valid/expr` won't even compile with KSC for Go and Lua, whereas if it's only using other `valid` subkeys (`eq`, `min`, `any-of`, ...), it will.

---

The first thing is pretty significant, the second not so much, but still perhaps we could wait with `valid/expr` until the next release when it's fixed for Go and Lua. What do you think, @dgelessus?

dgelessus: oof, I should have checked this better before suggesting it, heh. The name problem at least could be fixed by giving the zero-size field an `id`, right? Arguably that would be the right thing to do anyway, because `id`-less fields are supposed to be for data with unknown meaning, which this isn't.

There's no good way to avoid the second issue though, other than just not using `valid/expr`. I suppose for now we could revert that change - not sure if that's worth it or not.

generalmimon: > The name problem at least could be fixed by giving the zero-size field an `id`, right? Arguably that would be the right thing to do anyway, because `id`-less fields are supposed to be for data with unknown meaning, which this isn't.

Yeah, let's do that.

PR #408, opened 2021-01-26
Add OpenSSH Certificate format

N6UDP:
This adds support for the OpenSSH certificate format described [here](https://cvsweb.openbsd.org/src/usr.bin/ssh/PROTOCOL.certkeys?annotate=HEAD).

### DIFF
--- a/security/ssh_certificate.ksy
+++ b/security/ssh_certificate.ksy
@@ -1,0 +1,118 @@
+meta:
+  id: ssh_certificate
+  title: OpenSSH Certificate
+  license: CC0-1.0
+  endian: be
+  imports:
+    - /security/ssh_public_key
+doc: |
+  OpenSSH Certificates are simple certs used by OpenSSH.
+doc-ref: 'https://cvsweb.openbsd.org/src/usr.bin/ssh/PROTOCOL.certkeys?annotate=HEAD'
+seq:
+  - id: cert_name
+    type: ssh_public_key::cstring
+  - id: body
+    type:
+      switch-on: cert_name.value
+      cases:
+        '"ssh-rsa-cert-v01@openssh.com"': cert_rsa
+        '"ssh-dss-cert-v01@openssh.com"': cert_dss
+        '"ecdsa-sha2-nistp256-cert-v01@openssh.com"': cert_ecdsa
+        '"ecdsa-sha2-nistp384-cert-v01@openssh.com"': cert_ecdsa
+        '"ecdsa-sha2-nistp521-cert-v01@openssh.com"': cert_ecdsa
+        '"ssh-ed25519-cert-v01@openssh.com"': cert_ed25519
+types:
+  cert_rsa:
+    doc-ref: 'https://tools.ietf.org/html/rfc4253#section-6.6'
+    seq:
+      - id: nonce
+        type: ssh_public_key::cstring
+      - id: rsa
+        type: ssh_public_key::key_rsa
+      - id: footer
+        type: cert_footer
+  cert_dss:
+    seq:
+      - id: nonce
+        type: ssh_public_key::cstring
+      - id: dsa
+        type: ssh_public_key::key_dsa
+      - id: footer
+        type: cert_footer
+  cert_ecdsa:
+    seq:
+      - id: nonce
+        type: ssh_public_key::cstring
+      - id: ecdsa
+        type: ssh_public_key::key_ecdsa
+      - id: footer
+        type: cert_footer
+  cert_ed25519:
+    seq:
+      - id: nonce
+        type: ssh_public_key::cstring
+      - id: ed25519
+        type: ssh_public_key::key_ed25519
+      - id: footer
+        type: cert_footer
+  cert_footer:
+    seq:
+      - id: serial
+        type: u8
+      - id: type
+        type: u4
+      - id: key_id
+        type: ssh_public_key::cstring
+      - id: validprincipals
+        type: packed_cstring
+      - id: valid_after
+        type: u8
+      - id: valid_before
+        type: u8
+      - id: critical_options
+        type: packed_cstring_tuple
+      - id: extensions
+        type: packed_cstring_tuple
+      - id: reserved
+        type: ssh_public_key::cstring
+      - id: signature_key
+        type: cstring_sshkey
+      - id: signature
+        type: packed_cstring_tuple
+  packed_cstring:
+    seq:
+      - id: len
+        type: u4
+      - id: packed_strings
+        type: cstrings
+        size: len
+  packed_cstring_tuple:
+    seq:
+      - id: len
+        type: u4
+      - id: packed_strings
+        type: cstring_tuples
+        size: len
+  cstrings:
+    seq:
+      - id: strings
+        type: ssh_public_key::cstring
+        repeat: eos
+  cstring_tuples:
+    seq:
+      - id: strings
+        type: cstring_tuple
+        repeat: eos
+  cstring_tuple:
+    seq:
+      - id: name
+        type: ssh_public_key::cstring
+      - id: data
+        type: ssh_public_key::cstring
+  cstring_sshkey:
+    seq:
+      - id: len
+        type: u4
+      - id: value
+        type: ssh_public_key
+


security/ssh_certificate.ksy
```diff
@@ -0,0 +1,118 @@
...
+        type: cert_footer
+  cert_footer:
+    seq:
+      - id: serial
+        type: u8
+      - id: type
+        type: u4
+      - id: key_id
+        type: ssh_public_key::cstring
+      - id: validprincipals
```
generalmimon, line 66:
```suggestion
      - id: valid_principals
```

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,118 @@
...
+        type: packed_cstring
+      - id: valid_after
+        type: u8
+      - id: valid_before
+        type: u8
```
generalmimon, line 71: Are these UNIX timestamps or what datetime format do they use?

Also, it seems that if it's valid _forever_, then the `valid_before` is set to `0xffff_ffff_ffff_ffff`, so you could mention this special value here.

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,118 @@
...
+        '"ecdsa-sha2-nistp256-cert-v01@openssh.com"': cert_ecdsa
+        '"ecdsa-sha2-nistp384-cert-v01@openssh.com"': cert_ecdsa
+        '"ecdsa-sha2-nistp521-cert-v01@openssh.com"': cert_ecdsa
+        '"ssh-ed25519-cert-v01@openssh.com"': cert_ed25519
+types:
+  cert_rsa:
+    doc-ref: 'https://tools.ietf.org/html/rfc4253#section-6.6'
+    seq:
+      - id: nonce
+        type: ssh_public_key::cstring
```
generalmimon, line 29: I don't think that `nonce` is an ASCII string (see [`security/ssh_public_key.ksy:84`](https://github.com/kaitai-io/kaitai_struct_formats/blob/7f3ecee038ae4aa14da7d9a80d8abfd4080c9b59/security/ssh_public_key.ksy#L84)).

When I generate a test file with these commands:

```shell
ssh-keygen -t rsa -f ./id_rsa -N ""
ssh-keygen -s ./id_rsa -I "AcmeCorporation" ./id_rsa.pub
cat ./id_rsa-cert.pub | cut -f2 -d ' ' | openssl base64 -d -A > id_rsa-cert_base64-decoded.bin
```

it seems that this `nonce` is just an array of random bytes, so it should be treated that way. Interpreting it as a string with encoding is useless. Note that the `ASCII` encoding used isn't even guaranteed to properly handle byte values 128-255.

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,118 @@
...
+    seq:
+      - id: strings
+        type: cstring_tuple
+        repeat: eos
+  cstring_tuple:
+    seq:
+      - id: name
+        type: ssh_public_key::cstring
+      - id: data
+        type: ssh_public_key::cstring
```
generalmimon, line 111: Again, this doesn't look like an `ASCII` string. Especially in `signature`:

https://github.com/kaitai-io/kaitai_struct_formats/blob/676b472237de2d421671cbde6ac9226d4dd0ecd0/security/ssh_certificate.ksy#L80-L81


for my test file generated by the commands above, there is a `cstring_tuple` with name "rsa-sha2-512", which contains 384 random bytes in `data` (I don't quite understand why it has a byte length of 384 and not 64 when it claims to be SHA-512, but that doesn't matter - it's clearly not an ASCII string).

generalmimon (changes requested):

generalmimon:
@N6UDP And yeah, for your own convenience, remember a rule of thumb when making pull requests from your fork repository: **never commit directly on `master` of your fork**. Create a topic branch for every PR you create. See https://contribute.jquery.org/commits-and-pull-requests/#never-commit-on-master:

security/ssh_certificate.ksy
N6UDP: Yes, you are correct, I had moved to using the imported type without noticing that.. for reference the RFC text is as follows re: openssh "strings".
```
   string

      Arbitrary length binary string.  Strings are allowed to contain
      arbitrary binary data, including null characters and 8-bit
      characters.  They are stored as a uint32 containing its length
      (number of bytes that follow) and zero (= empty string) or more
      bytes that are the value of the string.  Terminating null
      characters are not used.

      Strings are also used to store text.  In that case, US-ASCII is
      used for internal names, and ISO-10646 UTF-8 for text that might
      be displayed to the user.  The terminating null character SHOULD
      NOT normally be stored in the string.  For example: the US-ASCII
      string "testing" is represented as 00 00 00 07 t e s t i n g.  The
      UTF-8 mapping does not alter the encoding of US-ASCII characters.
```
I will add a non-ASCII type for use here for the fields that don't actually store strings.

security/ssh_certificate.ksy
N6UDP: Yeah the format of the actual signature is in another openssh RFC; it differs by algorithm for both the signature itself and the hash; It would be a nice todo to actually parse it out so folks could verify using their crypto library of choice but for now I'll just make it a byte array.

N6UDP: (btw the byte length is due to length of the RSA key - e.g. 384*8 = 3072 which should be equal to keylength)

N6UDP:
@generalmimon yeah I was being bad/lazy; I plan to either reset + force push the squashed changes to my fork or just delete it when done ;)  Anyway I think I addressed all your comments, let me know if there are other things you see :)

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,196 @@
...
+          critical options is a set of zero or more key options encoded as
+          below. All such options are "critical" in the sense that an implementation
+          must refuse to authorise a key that has an unrecognised option.
+
+          Generally, critical options are used to control features that restrict
+          access where extensions are used to enable features that grant access.
+          This ensures that certificates containing unknown restrictions do not
+          inadvertently grant access while allowing new protocol features to be
+          enabled via extensions without breaking certificates' backwards
+          compatibility.
```
generalmimon, line 98: I find this too verbose and confused. If you found time to shorten it and make it clearer, it would be great.

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,196 @@
+meta:
+  id: ssh_certificate
+  title: OpenSSH Certificate
+  license: CC0-1.0
+  endian: be
+  imports:
+    - /security/ssh_public_key
```
generalmimon, line 7: `imports` should be above `endian` (https://doc.kaitai.io/ksy_style_guide.html#meta), but that's just a nitpick:

```suggestion
  imports:
    - /security/ssh_public_key
  endian: be
```

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,196 @@
...
+      - id: valid_after
+        doc: Unix timestamp (seconds since 1970-01-01 00:00:00)
+        type: u8
+      - id: valid_before
+        doc: | 
+          Unix timestamp (seconds since 1970-01-01 00:00:00)
+          When "forever" is requested this is set to 0xFFFF_FFFF_FFFF_FFFF
+        type: u8
+      - id: critical_options
+        doc: |
```
generalmimon, line 88:
```suggestion
        doc: |
          Unix timestamp (seconds since 1970-01-01 00:00:00)
          When "forever" is requested this is set to 0xFFFF_FFFF_FFFF_FFFF
        type: u8
      - id: critical_options
        doc: |
```

There is some trailing whitespace here (and 4 other similar lines). It's recommended to set up your editor to honor the [project `.editorconfig`](https://github.com/kaitai-io/kaitai_struct_formats/blob/1eae5145325bb992e20dae838dd68d04814dda46/.editorconfig#L8) (for example, there is a plugin [_EditorConfig for VS Code_](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig)).

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,196 @@
...
+        size: len
+  cstring_utf8:
+    doc: variant of cstring_bytes that decodes to UTF-8
+    seq:
+    - id: len
+      type: u4
+    - id: value
+      type: str
+      encoding: UTF-8
+      size: len
```
generalmimon, line 196:
```suggestion
    seq:
      - id: len
        type: u4
      - id: value
        type: str
        encoding: UTF-8
        size: len
```

See [KSY Style Guide > 1. General formatting](https://doc.kaitai.io/ksy_style_guide.html#general):

>
> <table>
> <thead>
> <tr>
> <th align=left>Good</th>
> <th align=left>Bad</th>
> </tr>
> </thead>
> <tbody>
> <tr>
> <td>
> 
> ```yaml
> imports:
>   - foo
>   - bar
>
> ```
>
> </td>
> <td>
> 
> ```yaml
> imports:
> - foo
> - bar
>
> ```
>
> </td>
> </tr>
> </tbody>
> </table>

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,196 @@
...
+      (number of bytes that follow) and zero (= empty string) or more
+      bytes that are the value of the string.  Terminating null
+      characters are not used.
+
+      Strings are also used to store text.  In that case, US-ASCII is
+      used for internal names, and ISO-10646 UTF-8 for text that might
+      be displayed to the user.  The terminating null character SHOULD
+      NOT normally be stored in the string.  For example: the US-ASCII
+      string "testing" is represented as 00 00 00 07 t e s t i n g.  The
+      UTF-8 mapping does not alter the encoding of US-ASCII characters.
```
generalmimon, line 182: Are the double spaces intentional?

```suggestion
    doc: |
      Arbitrary length binary string. Strings are allowed to contain
      arbitrary binary data, including null characters and 8-bit
      characters. They are stored as a uint32 containing its length
      (number of bytes that follow) and zero (= empty string) or more
      bytes that are the value of the string. Terminating null
      characters are not used.

      Strings are also used to store text. In that case, US-ASCII is
      used for internal names, and ISO-10646 UTF-8 for text that might
      be displayed to the user. The terminating null character SHOULD
      NOT normally be stored in the string. For example: the US-ASCII
      string "testing" is represented as 00 00 00 07 t e s t i n g. The
      UTF-8 mapping does not alter the encoding of US-ASCII characters.
```

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,196 @@
...
+      cert_type:
+        1: user
+        2: host
+  packed_cstring:
+    seq:
+      - id: len
+        type: u4
+      - id: packed_strings
+        type: cstrings
+        size: len
```
generalmimon, line 137: Strictly speaking, `size` should be above `type` (https://doc.kaitai.io/ksy_style_guide.html#seq-attr):

```suggestion
      - id: packed_strings
        size: len
        type: cstrings
```

Also, consider renaming all `len` properties to `len_{FIELD}`, where `{FIELD}` is the identifier of a field that is delimited by the `len` size (e.g. `len` -> `len_packed_strings`, `len_value`, etc.). See https://doc.kaitai.io/ksy_style_guide.html#attr-id:

> * For fields that designate **size** of some particular data structure (in bytes or some other fixed units), use `len_` prefix and name of that data structure — i.e. `len_block` (length of a single block entry), `len_blocks` (total length of whole `blocks` array, made of `block` entries).

generalmimon (changes requested):

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,196 @@
...
+      - id: valid_before
+        doc: | 
+          Unix timestamp (seconds since 1970-01-01 00:00:00)
+          When "forever" is requested this is set to 0xFFFF_FFFF_FFFF_FFFF
+        type: u8
+      - id: critical_options
+        doc: | 
+          critical options is a set of zero or more key options encoded as
+          below. All such options are "critical" in the sense that an implementation
+          must refuse to authorise a key that has an unrecognised option.
```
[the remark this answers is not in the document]
N6UDP: Removed the reference to below

security/ssh_certificate.ksy
N6UDP: I'll admit to having done the entire initial go of this in the Kaitai web IDE and the copy/pastaing it to github ;) - I'll go clone and validate that the format and other style guidelines you've called out look to be in order.

security/ssh_certificate.ksy
N6UDP: was straight from the openbsd spec

N6UDP:
@generalmimon pretty sure I've addressed all the concerns, let me know if I missed anything.

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,159 @@
...
+      - id: name
+        type: cstring_utf8
+      - id: data
+        type: cstring_bytes
+  cstring_sshkey:
+    seq:
+      - id: len_cstring_sshkey
+        type: u4
+      - id: value
+        type: ssh_public_key
```
generalmimon, line 129: I'd go just for `len_value`:

```suggestion
      - id: len_value
        type: u4
      - id: value
        type: ssh_public_key
```

You don't need to specify that it relates to the `cstring_sshkey`, that's the name of the enclosing type.

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,159 @@
...
+  imports:
+    - /security/ssh_public_key
+  endian: be
+doc: |
+  OpenSSH Certificates are simple certs used by OpenSSH.
+doc-ref: 'https://cvsweb.openbsd.org/cgi-bin/cvsweb/~checkout~/src/usr.bin/ssh/PROTOCOL.certkeys?rev=1.17&content-type=text/plain'
+seq:
+  - id: cert_type
+    doc: type of ssh certificate
+    type: cstring_utf8
```
generalmimon, line 14: For all `seq`uence attributes and `instances`, the `doc` key should go **below** all other keys, see https://doc.kaitai.io/ksy_style_guide.html#seq-attr.

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,159 @@
...
+      string "testing" is represented as 00 00 00 07 t e s t i n g. The
+      UTF-8 mapping does not alter the encoding of US-ASCII characters.
+    seq:
+      - id: len_cstring_bytes
+        type: u4
+      - id: value
+        size: len_cstring_bytes
+  cstring_utf8:
+    doc: variant of cstring_bytes that decodes to UTF-8
+    seq:
```
generalmimon, line 153: This improves usability in the [Web IDE](https://ide.kaitai.io/), because you don't have to expand the type to see what's inside:

```suggestion
    -webide-representation: '{value}'
    seq:
```

N6UDP:
@generalmimon ksylint now passes with the exception of the case block which it just thinks is invalid according to the schema.

security/ssh_certificate.ksy
```diff
@@ -143,19 +142,20 @@ types:
...
       string "testing" is represented as 00 00 00 07 t e s t i n g. The
       UTF-8 mapping does not alter the encoding of US-ASCII characters.
-    -webide-representation: '{len_value:dec} bytes'
+    doc-ref: 'https://tools.ietf.org/html/rfc4251#section-5'
     seq:
       - id: len_value
         type: u4
       - id: value
         size: len_value
+    -webide-representation: '{len_value:dec} bytes'
```
generalmimon, line 151: I don't know what ksylint said, but I personally think that it makes sense to put the `-webide-representation` at the top of the type definition, i.e. below `doc` and `doc-ref`, but above `seq` and `instances`. On top of that, it usually makes sense to stick to the [order in the style guide](https://doc.kaitai.io/ksy_style_guide.html#type):

> ## 2. Order of sections in a type spec
> 
> Use the following order of sections:
> 
> - `meta`, if present, MUST go first
> 
> - `doc`
> 
> - `doc-ref`
> 
> - `seq`
> 
> - `instances`, `types`, `enums` — use one’s best judgement to order these 3 to maximize readability

generalmimon (changes requested):

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,161 @@
...
+      - id: strings
+        type: cstring_utf8
+        repeat: eos
+  cstring_tuples:
+    seq:
+      - id: strings
+        type: cstring_tuple
+        repeat: eos
+  cstring_tuple:
+    seq:
```
generalmimon, line 119:
```suggestion
  cstring_tuple:
    -webide-representation: '{name}: {data}'
    seq:
```

After this change, you'll be able to see inside the options without needing to expand them, which is a nice improvement I think:

image

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,162 @@
...
+        type: cstring_tuples
+        size: len_packed_strings
+  cstrings:
+    seq:
+      - id: strings
+        type: cstring_utf8
+        repeat: eos
+  cstring_tuples:
+    seq:
+      - id: strings
```
generalmimon, line 115: Shouldn't it be

```suggestion
      - id: tuples
```

...?

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,162 @@
...
+        1: user
+        2: host
+  packed_cstring:
+    seq:
+      - id: len_packed_strings
+        type: u4
+      - id: packed_strings
+        size: len_packed_strings
+        type: cstrings
+  packed_cstring_tuple:
```
generalmimon, line 101: Maybe

```suggestion
  packed_cstring_tuples:
```

...?

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,162 @@
...
+    seq:
+      - id: len_packed_strings
+        type: u4
+      - id: packed_strings
+        type: cstring_tuples
+        size: len_packed_strings
```
generalmimon, line 107: What about just

```suggestion
      - id: len_tuples
        type: u4
      - id: tuples
        type: cstring_tuples
        size: len_tuples
```

...? In other words, does the adjective **"packed"** give any information about the meaning of the value?

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,162 @@
...
+          cases:
+            '"ssh-rsa-cert-v01@openssh.com"': ssh_public_key::key_rsa
+            '"ssh-dss-cert-v01@openssh.com"': ssh_public_key::key_dsa
+            '"ecdsa-sha2-nistp256-cert-v01@openssh.com"': ssh_public_key::key_ecdsa
+            '"ecdsa-sha2-nistp384-cert-v01@openssh.com"': ssh_public_key::key_ecdsa
+            '"ecdsa-sha2-nistp521-cert-v01@openssh.com"': ssh_public_key::key_ecdsa
+            '"ssh-ed25519-cert-v01@openssh.com"': ssh_public_key::key_ed25519
+      - id: serial
+        type: u8
+        doc: optional serial number; if zero the CA doesn't use serial numbers
```
generalmimon, line 35:
```suggestion
        doc: optional serial number; if zero, the CA doesn't use serial numbers
```

generalmimon (changes requested):

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,162 @@
...
+      cert_type:
+        1: user
+        2: host
+  packed_cstring:
+    seq:
+      - id: len_packed_strings
+        type: u4
+      - id: packed_strings
+        size: len_packed_strings
+        type: cstrings
```
generalmimon, line 100: Same issue as https://github.com/kaitai-io/kaitai_struct_formats/pull/408#discussion_r591460340

security/ssh_certificate.ksy
N6UDP: Yeah I suppose, all the tuples and all the packed things are just nested "cstring"s

generalmimon (changes requested):

security/ssh_certificate.ksy
```diff
@@ -0,0 +1,162 @@
...
+          They are considered "critical" as implementations must refuse
+          to authorize a certificate that has unrecognized options. This
+          prevents an unknown restriction in the certificate from failing to
+          be applied.
+      - id: extensions
+        type: packed_cstring_tuples
+        doc: |
+          Contains zero or more optional extensions. These extensions
+          are not critical, and an implementation that encounters one
+          that it does not recognize may safely ignore it.
```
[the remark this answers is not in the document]
N6UDP: While I believe you are correct that all existing extensions / critical options are currently UTF8 strings the spec allows them to be "string" byte arrays as well.  It's possible that both future official extensions and existing non-openssh official extensions today use non-UTF8 string formats.  I'm inclined to leave parsing of them up to the calling implementation.. but if there's some easy way to get the raw bytes in addition to the string I am open to parsing them for convivence.  As for the signature field it varies as you noted by signature type and I didn't want to put the time into figuring out the format for all of them.  They likely should be parsed in their own ksy file as the signature format is used for those algorithms across various openssh protocols (ssh-agent, authentication, certificate signing, etc)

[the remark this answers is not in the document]
generalmimon: > While I believe you are correct that all existing extensions / critical options are currently UTF8 strings the spec allows them to be "string" byte arrays as well. It's possible that both future official extensions and existing non-openssh official extensions today use non-UTF8 string formats.

Ok, then let's go with `cstring_bytes` (instead `cstring_utf8`).

> As for the signature field it varies as you noted by signature type and I didn't want to put the time into figuring out the format for all of them. They likely should be parsed in their own ksy file as the signature format is used for those algorithms across various openssh protocols (ssh-agent, authentication, certificate signing, etc)

Yeah, that sounds reasonable. The signature formats are defined in different specifications depending the algorithm ([`PROTOCOL.certkeys:217-220`](https://github.com/openssh/openssh-portable/blob/2421a567a8862fe5102a4e7d60003ebffd1313dd/PROTOCOL.certkeys#L217-L220)):

> Signatures are computed and encoded according to the rules defined for the CA's public key algorithm (RFC4253 section 6.6 for ssh-rsa and ssh-dss, RFC5656 for the ECDSA types), and draft-josefsson-eddsa-ed25519-03 for Ed25519.

so they're quite out of scope of the `ssh_certificate.ksy` spec, and their parsing should be done in their own specs anyway, as you say.

N6UDP:
@generalmimon made some changes to make that parsing work; not sure I love the implementation or the naming but it does parse out better / display in the webui

generalmimon (changes requested):

PR #409, opened 2021-01-28
Added DIME message

Sijisu:
Adding the old-school DIME (Direct Internet Message Encapsulation) format. Based on the 2002 Microsoft draft [here](http://xml.coverpages.org/draft-nielsen-dime-02.txt).

I was unsure on the folder where to put it, but settled on `network/`.

Thanks for this awesome project!

### DIFF
--- a/network/dime_message.ksy
+++ b/network/dime_message.ksy
@@ -24,6 +24,7 @@ doc: |
 doc-ref: 
   - https://tools.ietf.org/html/draft-nielsen-dime-02
   - https://docs.microsoft.com/en-us/archive/msdn-magazine/2002/december/sending-files-attachments-and-soap-messages-via-dime
+  - http://imrannazar.com/Parsing-the-DIME-Message-Format
 seq:
   - id: records
     type: record
@@ -53,25 +54,35 @@ types:
     doc: each individual fragment of the message
     seq:
       - id: version
+        doc: DIME format version (always 1)
         type: b5
       - id: is_first_record
+        doc: Set if this is the first record in the message
         type: b1
       - id: is_last_record
+        doc: Set if this is the last record in the message
         type: b1
       - id: is_chunk_record
+        doc: Set if the file contained in this record is chunked into multiple records
         type: b1
       - id: type_format
+        doc: Indicates the structure and format of the value of the TYPE field
         enum: type_formats
         type: b4
       - id: reserved
+        doc: Reserved for future use
         type: b4
       - id: len_options
+        doc: Length of the Options field
         type: u2
       - id: len_id
+        doc: Length of the ID field
         type: u2
       - id: len_type
+        doc: Length of the Type field
         type: u2
       - id: len_data
+        doc: Length of the Data field
         type: u4
       - id: options
         size: len_options
@@ -79,16 +90,19 @@ types:
       - id: options_padding
         type: padding
       - id: id
+        doc: Unique identifier of the file (set in the first record of file)
         type: str
         size: len_id
       - id: id_padding
         type: padding
       - id: type
+        doc: Specified type in the format set with type_format
         type: str
         size: len_type
       - id: type_padding
         type: padding
       - id: data
+        doc: The file data
         size: len_data
       - id: data_padding
         type: padding
--- a/network/dime_message.ksy
+++ b/network/dime_message.ksy
@@ -23,48 +23,48 @@ types:
   padding:
     doc: padding to the next 4-byte boundary
     seq:
-    - id: boundary_padding
-      size: (4 - _io.pos) % 4
+      - id: boundary_padding
+        size: (4 - _io.pos) % 4
   record:
     doc: each individual fragment of the message
     seq:
-    - id: version
-      type: b5
-    - id: is_first_record
-      type: b1
-    - id: is_last_record
-      type: b1
-    - id: is_chunck_record
-      type: b1
-    - id: type_format
-      enum: type_formats
-      type: b4
-    - id: reserved
-      type: b4
-    - id: len_options
-      type: u2
-    - id: len_id
-      type: u2
-    - id: len_type
-      type: u2
-    - id: len_data
-      type: u4
-    - id: options
-      size: len_options
-    - id: options_padding
-      type: padding
-    - id: id
-      size: len_id
-    - id: id_padding
-      type: padding
-    - id: type
-      size: len_type
-    - id: type_padding
-      type: padding
-    - id: data
-      size: len_data
-    - id: data_padding
-      type: padding
+      - id: version
+        type: b5
+      - id: is_first_record
+        type: b1
+      - id: is_last_record
+        type: b1
+      - id: is_chunck_record
+        type: b1
+      - id: type_format
+        enum: type_formats
+        type: b4
+      - id: reserved
+        type: b4
+      - id: len_options
+        type: u2
+      - id: len_id
+        type: u2
+      - id: len_type
+        type: u2
+      - id: len_data
+        type: u4
+      - id: options
+        size: len_options
+      - id: options_padding
+        type: padding
+      - id: id
+        size: len_id
+      - id: id_padding
+        type: padding
+      - id: type
+        size: len_type
+      - id: type_padding
+        type: padding
+      - id: data
+        size: len_data
+      - id: data_padding
+        type: padding
 enums:
   type_formats:
     0: unchanged
--- a/network/dime_message.ksy
+++ b/network/dime_message.ksy
@@ -1,0 +1,61 @@
+meta:
+  id: dime_message
+  title: DIME (Direct Internet Message Encapsulation) Message
+  license: CC0-1.0
+  endian: be
+doc: |
+  Direct Internet Message Encapsulation (DIME)
+  is an old Microsoft specification for sending and receiving
+  SOAP messages along with additional attachments,
+  like binary files, XML fragments, and even other
+  SOAP messages, using standard transport protocols like HTTP.
+doc-ref: 
+  - http://xml.coverpages.org/draft-nielsen-dime-02.txt
+  - https://docs.microsoft.com/en-us/archive/msdn-magazine/2002/december/sending-files-attachments-and-soap-messages-via-dime
+seq:
+  - id: dime
+    type: message
+types:
+  message:
+    seq:
+    - id: message
+      type: record
+      repeat: eos
+  record:
+    seq:
+    - id: version
+      type: b5
+    - id: first_record
+      type: b1
+    - id: last_record
+      type: b1
+    - id: chunck_record
+      type: b1
+    - id: type_format
+      type: b4
+    - id: reserved
+      type: b4
+    - id: options_length
+      type: u2
+    - id: id_length
+      type: u2
+    - id: type_length
+      type: u2
+    - id: data_length
+      type: u4
+    - id: options
+      size: options_length
+    - id: options_padding
+      size: (4 - _io.pos) % 4
+    - id: id
+      size: id_length
+    - id: id_padding
+      size: (4 - _io.pos) % 4
+    - id: type
+      size: type_length
+    - id: type_padding
+      size: (4 - _io.pos) % 4
+    - id: data
+      size: data_length
+    - id: data_padding
+      size: (4 - _io.pos) % 4
--- a/network/dime_message.ksy
+++ b/network/dime_message.ksy
@@ -23,7 +23,7 @@ types:
   padding:
     doc: padding to the next 4-byte boundary
     seq:
-    - id: padding
+    - id: boundary_padding
       size: (4 - _io.pos) % 4
   record:
     doc: each individual fragment of the message
--- a/network/dime_message.ksy
+++ b/network/dime_message.ksy
@@ -1,7 +1,10 @@
 meta:
   id: dime_message
   title: DIME (Direct Internet Message Encapsulation) Message
+  xref:
+    wikidata: Q1227457
   license: CC0-1.0
+  bit-endian: be
   endian: be
 doc: |
   Direct Internet Message Encapsulation (DIME)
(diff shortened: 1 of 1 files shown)


Sijisu:
Forgot to include an example, I made this to parse the image response from my scanner, so here is an example response from my scanner:

https://d.sijisu.eu/scanner.dump

Works great in the Kaitai Web IDE

network/dime_message.ksy
```diff
@@ -0,0 +1,61 @@
...
+    type: message
+types:
+  message:
+    seq:
+    - id: message
+      type: record
+      repeat: eos
+  record:
+    seq:
+    - id: version
```
KOLANICH, line 26: May make sense to extract these fields, forming 2 bytes, into an own subtype.

generalmimon:
@Sijisu:

> so here is an example response from my scanner:
>
> https://d.sijisu.eu/scanner.dump

Thanks for the sample. I'm also uploading it GZIPped here on GitHub so that it stays available best forever: [**scanner.dump.gz**](https://github.com/kaitai-io/kaitai_struct_formats/files/5890499/scanner.dump.gz)

network/dime_message.ksy
```diff
@@ -0,0 +1,61 @@
...
+    - id: id_length
+      type: u2
+    - id: type_length
+      type: u2
+    - id: data_length
+      type: u4
+    - id: options
+      size: options_length
+    - id: options_padding
+      size: (4 - _io.pos) % 4
```
KOLANICH, line 49: Padding can be a subtype. Power of 2 padding is better to implement without `%`.  Something like `(_io.pos + 3) & ~3 - _io.pos`

network/dime_message.ksy
```diff
@@ -0,0 +1,61 @@
+meta:
+  id: dime_message
+  title: DIME (Direct Internet Message Encapsulation) Message
+  license: CC0-1.0
+  endian: be
```
generalmimon, line 5:
```suggestion
  bit-endian: be
  endian: be
```

You are using bit-sized integers in the spec and there are actually two layouts available (`bit-endian: be` is default, but that's only for backward compatibility), so you should make it explicit which one you want (see https://doc.kaitai.io/user_guide.html#bit-endian).

network/dime_message.ksy
```diff
@@ -0,0 +1,61 @@
...
+  - https://docs.microsoft.com/en-us/archive/msdn-magazine/2002/december/sending-files-attachments-and-soap-messages-via-dime
+seq:
+  - id: dime
+    type: message
+types:
+  message:
+    seq:
+    - id: message
+      type: record
+      repeat: eos
```
generalmimon, line 23: Why not just

```suggestion
seq:
  - id: messages
    type: record
    repeat: eos
```

?

BTW, it's best to avoid having any properties (i.e. `seq` or `instances`) named the same as the enclosing type to avoid name collision (see https://github.com/kaitai-io/kaitai_struct/issues/86#issuecomment-292676386).

network/dime_message.ksy
```diff
@@ -0,0 +1,61 @@
...
+  record:
+    seq:
+    - id: version
+      type: b5
+    - id: first_record
+      type: b1
+    - id: last_record
+      type: b1
+    - id: chunck_record
+      type: b1
```
generalmimon, line 33: There is a name convention of prefixing the boolean properties with `is_`:

```suggestion
    - id: is_first_record
      type: b1
    - id: is_last_record
      type: b1
    - id: is_chunck_record
      type: b1
```

network/dime_message.ksy
```diff
@@ -0,0 +1,61 @@
...
+    - id: type_format
+      type: b4
```
generalmimon, line 35: Is this an [enum](https://doc.kaitai.io/user_guide.html#enums)?

network/dime_message.ksy
```diff
@@ -0,0 +1,61 @@
...
+    - id: reserved
+      type: b4
+    - id: options_length
+      type: u2
+    - id: id_length
+      type: u2
+    - id: type_length
+      type: u2
+    - id: data_length
+      type: u4
```
generalmimon, line 45: According [to the KSY style guide](https://doc.kaitai.io/ksy_style_guide.html#attr-id), attributes that denote a byte size of some other attribute should start with `len_`:

> For fields that designate size of some particular data structure (in bytes or some other fixed units), use `len_` prefix and name of that data structure — i.e. `len_block` (length of a single `block` entry), `len_blocks` (total length of whole `blocks` array, made of `block` entries).

i.e. <del><code>options_length</code></del> <strong><code>len_options</code></strong> etc.

network/dime_message.ksy
```diff
@@ -0,0 +1,61 @@
+meta:
+  id: dime_message
+  title: DIME (Direct Internet Message Encapsulation) Message
```
generalmimon, line 3:
```suggestion
  title: DIME (Direct Internet Message Encapsulation) Message
  xref:
    wikidata: Q1227457
```

generalmimon (changes requested):

network/dime_message.ksy
generalmimon: > Power of 2 padding is better to implement without `%` (division is a costly operation).

IMHO these micro-optimizations are not worth it. To keep the code understandable and readable, we should primarily focus on expressing the intent.

For example, when the intent is to multiply `foo` by 63, you shouldn't do `(foo << 6) - foo`, but simply `foo * 63`.

If performance is a concern, choose a performant language like C++ and let the compiler do its job (it does a lot of transformations like from `foo * 63` to `(foo << 6) - foo` by itself). Additionally, if there is a great demand to have these for all languages, we can do them in the KSC as well, but I don't think it would really matter for most formats and usages.

network/dime_message.ksy
Sijisu: I have thought about it, but I think it makes more sense this way.

The draft doesn't split this header section at all as far as I looked, so i thing it would make it more confusing. (I don't even know how it would be named :smile:, since the lenght fields are part of the header as well it can't be header, so a preheader? part_of_the_header?)

So unless there is some practical reason to extract this part I don't know about, I would prefer not to.

network/dime_message.ksy
Sijisu: @generalmimon I missed your comment and I already made the changes, but I completely agree, readability is more important (especially with this kind of 2002 thing).

I would be for `(4 - _io.pos) % 4`. I will see what we will come out of the discussion and then either keep the newly commited `((_io.pos + 3) & ~3) - _io.pos` or go back to the original.

network/dime_message.ksy
Sijisu: It is kind of an enum. But with options `0x05-0x0F` reserved for future use, but you are right that nobody will probably want those ever indeed. So I made it a enum. Thanks!

network/dime_message.ksy
```diff
@@ -0,0 +1,76 @@
...
+seq:
+  - id: records
+    type: record
+    repeat: eos
+types:
+  padding:
+    doc: |
+      padding to the next 4-byte boundary
+    seq:
+    - id: padding
```
KOLANICH, line 27: Please, indent collections content relative to their key

Sijisu: I made the doc strings one-line.

generalmimon: @Sijisu:

> I made the doc strings one-line.

That's nice, but I think @KOLANICH meant this:

```diff
     seq:
-    - id: padding
-      size: (4 - _io.pos) % 4
```

```diff
     seq:
+      - id: padding
+        size: (4 - _io.pos) % 4
```

generalmimon: @Sijisu:

See also [KSY Style Guide > 1. General formatting](https://doc.kaitai.io/ksy_style_guide.html#general):


>
> <table>
> <thead>
> <tr>
> <th>Good</th>
> <th>Bad</th>
> </tr>
> </thead>
> <tbody>
> <tr>
> <td>
> 
> ```yaml
> imports:
>   - foo
>   - bar
>
> ```
>
> </td>
> <td>
> 
> ```yaml
> imports:
> - foo
> - bar
>
> ```
>
> </td>
> </tr>
> </tbody>
> </table>

Sijisu: Oh I see, I'm sorry :man_facepalming:

Sijisu: Fixed

network/dime_message.ksy
```diff
@@ -0,0 +1,74 @@
...
+seq:
+  - id: records
+    type: record
+    repeat: eos
+types:
+  padding:
+    doc: padding to the next 4-byte boundary
+    seq:
+      - id: boundary_padding
+        size: (4 - _io.pos) % 4
```
generalmimon, line 27: Wouldn't this make more sense? It is easier to understand for me because it does not apply modulo to a negative number.

```suggestion
        size: 4 - (_io.pos % 4)
```

This way, if `_io.pos` is `23`, you get `_io.pos % 4 == 3` (because <code>23 == 5 * 4 + <strong>3</strong></code>), but you want to align it to the next 4-byte boundary, which is `24 == 6 * 4`. To achieve that, you need the supplement of the remainder (which is `3`) to 4 and that can be calculated as `4 - 3 == 1`.

IMHO this is more understandable than `(4 - _io.pos) % 4`, because with `_io.pos == 23`, `(4 - _io.pos) == -19` and you're taking a modulo from a negative number here, which is a bit unnatural.

Sijisu: The simplification looks good.

I don't think much more can be done, from the logic of the matter it has to be in the format of `(someexpresion) % 4` (the modulo has to be applied last) and to my best knowledge there isn't a simpler or more expressive way than the simplified original one you proposed.

network/dime_message.ksy
```diff
@@ -0,0 +1,74 @@
...
+    wikidata: Q1227457
+  license: CC0-1.0
+  bit-endian: be
+  endian: be
+doc: |
+  Direct Internet Message Encapsulation (DIME)
+  is an old Microsoft specification for sending and receiving
+  SOAP messages along with additional attachments,
+  like binary files, XML fragments, and even other
+  SOAP messages, using standard transport protocols like HTTP.
```
generalmimon, line 14: IMHO it makes sense to include a command to get a sample file here:

```suggestion
  SOAP messages, using standard transport protocols like HTTP.
  
  Sample file: `curl -L
  https://github.com/kaitai-io/kaitai_struct_formats/files/5890499/scanner.dump.gz
  | gunzip -c > scanner.dump`
```

network/dime_message.ksy
```diff
@@ -0,0 +1,74 @@
...
+  record:
+    doc: each individual fragment of the message
+    seq:
+      - id: version
+        type: b5
+      - id: is_first_record
+        type: b1
+      - id: is_last_record
+        type: b1
+      - id: is_chunck_record
```
generalmimon, line 37: What does this mean? Is `chunck` an abbreviation of something or just a misspelled `chunk`?

network/dime_message.ksy
```diff
@@ -0,0 +1,74 @@
...
+      - id: len_options
+        type: u2
+      - id: len_id
+        type: u2
+      - id: len_type
+        type: u2
+      - id: len_data
+        type: u4
+      - id: options
+        size: len_options
```
generalmimon, line 53: Do `options` have some internal structure?

network/dime_message.ksy
Sijisu: Good idea! Thanks

network/dime_message.ksy
```diff
@@ -0,0 +1,74 @@
...
+      - id: options_padding
+        type: padding
+      - id: id
+        size: len_id
```
generalmimon, line 57: Isn't this always a string with some fixed encoding?

network/dime_message.ksy
```diff
@@ -0,0 +1,78 @@
...
+      - id: id_padding
+        type: padding
+      - id: type
+        size: len_type
```
generalmimon, line 65: The same as https://github.com/kaitai-io/kaitai_struct_formats/pull/409#discussion_r566843810

network/dime_message.ksy
Sijisu: It is indeed a misspelled `chunk`

network/dime_message.ksy
```diff
@@ -0,0 +1,78 @@
+meta:
+  id: dime_message
+  title: DIME (Direct Internet Message Encapsulation) Message
+  xref:
```
generalmimon, line 4: Does this make sense?

```suggestion
  xref:
    mime: application/dime
```

Sijisu: I don't think they are used very often, but you are right that if they're specified we should include them.

network/dime_message.ksy
Sijisu: Actually they do! Again, from what I read, it was never actually used for anything useful (to my knowledge there aren't any registered types), but is still in the spec and so it should be it struct as well...

I'm adding support for this in the next commit.
I don't know how to test it however, can't find a dime file using these options.

network/dime_message.ksy
Sijisu: It indeed is.

network/dime_message.ksy
Sijisu: I'm crafting a custom file to test this.

Sijisu: Alright, it works.

I noticed that even the original `scanner.dump` includes an option field in `records[1]`, but I edited the file to include multiple options (in `records[0]`) and it works as well.
[scanner_withoptions.dump.gz](https://github.com/kaitai-io/kaitai_struct_formats/files/5894723/scanner_withoptions.dump.gz)

network/dime_message.ksy
```diff
@@ -0,0 +1,115 @@
...
+      - id: len_element
+        type: u2
+      - id: element_data
+        size: len_element
+  record:
+    doc: each individual fragment of the message
+    seq:
+      - id: version
+        doc: DIME format version (always 1)
+        type: b5
```
generalmimon, line 58: Since 0.9, this can be checked using the `valid` key (kaitai-io/kaitai_struct#435):

```suggestion
      - id: version
        doc: DIME format version
        type: b5
        valid: 1
```

_Alternatively:_ 
```ksy
      - id: version
        doc: DIME format version
        type: b5
        valid:
          eq: 1
```

network/dime_message.ksy
```diff
@@ -0,0 +1,115 @@
...
+  encoding: ASCII
+doc: |
+  Direct Internet Message Encapsulation (DIME)
+  is an old Microsoft specification for sending and receiving
+  SOAP messages along with additional attachments,
+  like binary files, XML fragments, and even other
+  SOAP messages, using standard transport protocols like HTTP.
+  
+  Sample file: `curl -L
+  https://github.com/kaitai-io/kaitai_struct_formats/files/5890499/scanner.dump.gz
```
generalmimon, line 22: If the sample file you referenced in https://github.com/kaitai-io/kaitai_struct_formats/pull/409#discussion_r566909282 is better for testing, perhaps this one can be replaced?

```suggestion
  https://github.com/kaitai-io/kaitai_struct_formats/files/5894723/scanner_withoptions.dump.gz
```

network/dime_message.ksy
Sijisu: Didn't know this, cool, thanks.

KOLANICH: Probably it shouldn't. The format has reserved fields, and the point to have reserved fields is forward compatibility.

Sijisu: Yeah, that is true. But the format was also withdrawn before it even made RFC in 2002. There most certainly won't be other version (at least official). I don't think Microsoft will dig out this 20 years old spec and suddently make a 2.0. 

But you are right, there *could* be other versions. I don't know what the Kaitai's approach is on things like that.

network/dime_message.ksy
Sijisu: It can and it should.

network/dime_message.ksy
generalmimon: > But you are right, there _could_ be other versions. I don't know what the Kaitai's approach is on things like that.

I'm not sure where I stand. For best compatibility, it probably makes sense to follow the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) (_"Be conservative in what you do, be liberal in what you accept from others"_), which would mean to accept _any_ version specifier, but generate only `1` (after https://github.com/kaitai-io/kaitai_struct/issues/27 is implemented).

So the existing approach is totally OK I'd say.

generalmimon (approved):

Sijisu:
Thank you for the help and reviews, guys!

PR #410, opened 2021-01-31
Add battery_management_system_protocol

Jakeler:


hardware/battery/battery_management_system_protocol.ksy
```diff
@@ -0,0 +1,223 @@
...
+  - id: magic_start
+    contents: [0xdd]
+  - id: cmd
+    type: u1
+
+  - id: body
+    type:
+      switch-on: cmd
+      cases:
+        0xa5: read_req
```
KOLANICH, line 25: The values should probably belong to a enum.

hardware/battery/battery_management_system_protocol.ksy
```diff
@@ -0,0 +1,223 @@
+meta:
+  id: battery_management_system_protocol
+  title: Communication protocol of smart battery management systems from LLT power
+  license: CC0-1.0
+  ks-version: 0.8
+  endian: be
```
KOLANICH, line 6: `bit-endian: be`

hardware/battery/battery_management_system_protocol.ksy
```diff
@@ -0,0 +1,223 @@
...
+        doc: Cycle times
+      - id: prod_date
+        type: u2
+        doc: Production date
+      - id: balance_status
+        type: balance_list
+        doc: List of balance bits
+      - id: prot_status
+        type: prot_list
+        size: 2
```
KOLANICH, line 73: seems to be unneeded

hardware/battery/battery_management_system_protocol.ksy
```diff
@@ -0,0 +1,223 @@
...
+            type: b1
+          - id: ovp_cell
+            type: b1
+      fet_bits:
+        seq:
+          - id: reserved
+            type: b6
+          - id: discharge
+            type: b1
+            enum: fet_bit
```
KOLANICH, line 139: Isn't it already `b1`?

hardware/battery/battery_management_system_protocol.ksy
```diff
@@ -0,0 +1,223 @@
...
+            type: b1
+          - id: uvp_cell
+            type: b1
+          - id: ovp_cell
+            type: b1
+      fet_bits:
+        seq:
+          - id: reserved
+            type: b6
+          - id: discharge
```
KOLANICH, line 137: Boolean props name should start with `is_`

hardware/battery/battery_management_system_protocol.ksy
```diff
@@ -0,0 +1,223 @@
...
+            type: b1
+            enum: fet_bit
+          - id: charge
+            type: b1
+            enum: fet_bit
```
KOLANICH, line 142: `-affected-by: 522`
https://github.com/kaitai-io/kaitai_struct/issues/522

hardware/battery/battery_management_system_protocol.ksy
```diff
@@ -0,0 +1,223 @@
...
+      voltage:
```
KOLANICH, line 143: `-affected-by: 522`
https://github.com/kaitai-io/kaitai_struct/issues/522

hardware/battery/battery_management_system_protocol.ksy
```diff
@@ -0,0 +1,223 @@
...
+  After sending read requests they respond with various information's about the battery state in
+  a custom binary format.
+
+doc-ref: https://www.lithiumbatterypcb.com/Protocol%20English%20Version.rar
+
+seq:
+  - id: magic_start
+    contents: [0xdd]
+  - id: cmd
+    type: u1
```
KOLANICH, line 19: Should it be a enum?

hardware/battery/battery_management_system_protocol.ksy
```diff
@@ -0,0 +1,223 @@
...
+        instances:
+          amp:
+            value: raw * 0.01
+            doc: Actual current (A)
+      temp:
+        seq:
+          - id: raw
+            type: u2
+        instances:
+          celsius:
```
KOLANICH, line 175: maybe Kelvin then, not Celsius.

hardware/battery/battery_management_system_protocol.ksy
```diff
@@ -0,0 +1,223 @@
...
+  - id: body
+    type:
+      switch-on: cmd
+      cases:
+        0xa5: read_req
+        0x5a: write_req
+        _: response(cmd)
+  
+  - id: checksum
+    size: 2
```
generalmimon, line 30: Please write into `doc` key what format/algorithm the `checksum` field uses and from which bytes it is calculated.

Also, it's a good practice to add a [parse instance](https://doc.kaitai.io/user_guide.html#_instances_data_beyond_the_sequence) called `checksum_input` for example that can be used to obtain a byte array which serves as the basis for checksum calculation. If someone wants to check or write the checksum, they can easily access it.

It can look like this:

```ksy
instances:
  checksum_input:
    pos: 0
    size: 16 # or something like _sizeof - checksum._sizeof
```

Check https://github.com/kaitai-io/kaitai_struct/issues/84 for more information on the compile-time `sizeof` operator.

Jakeler: Doc added in 66778d83ab549a2fda38d8b2c65b79dc25902c36

A `checksum_input` would indeed be very useful. Thanks for the suggestion. 
But I can't get it to work, in theory this is just the raw bytes of `body`, I tried:
```ksy
  checksum_input:
    pos: 2
    size: body._sizeof
```
But this does not work with current implementation because `dynamic sized type`.

Should this not also work with substreams?
```ksy
  checksum_input:
    io: _root.body._io
    size-eos: true
```
Result: `/instances/checksum_input/io: don't know how to call method '_io' of object type 'KaitaiStructType'`
I also can't add a constant size there... sorry, I could not figure out how to do this.

hardware/battery/battery_management_system_protocol.ksy
Jakeler: Celsius is very useful in general, battery/cell specifications (operating range, etc) are usually in °C. I don't see why you would need K besides lab equipment. Normal battery users don't do thermodynamic calculations...
And no, it is just 273.1, because it is based on integers and this gives 0.1 resolution, they even mention this offset in the specification.

hardware/battery/battery_management_system_protocol.ksy
Jakeler: Could be, but can enums deal with 'other' values? I think this would throw errors here.

hardware/battery/battery_management_system_protocol.ksy
Jakeler: Don't know, what is a enum per default? In any case, does it hurt to specify the type?

hardware/battery/battery_management_system_protocol.ksy
Jakeler: b28c288
BTW: Is this `-affected-by` documented somewhere? I could not find it in the user guide

hardware/battery/battery_management_system_protocol.ksy
Jakeler: Why set minimum version to 0.9 if it is tested and working with 0.8?

generalmimon: @Jakeler:

> Why set minimum version to 0.9 if it is tested and working with 0.8?

See https://doc.kaitai.io/user_guide.html#bit-endian:

<table role="table">
<tbody>
<tr>
<th>IMPORTANT</th>
<td><p>Support for <code>meta/bit-endian</code> was added in <strong>v0.9</strong> (previously only BE direction was supported). To maintain backward compatibility, the big-endian order (<code>be</code>) is default.</p><div>However, if you don’t really need to support pre-0.9 KSC versions, it’s recommended to state <code>meta/bit-endian: be</code> explicitly, because it raises awareness about the existence of LE bit endianness.</div></td>
</td>
</tbody>
</table>

hardware/battery/battery_management_system_protocol.ksy
KOLANICH: A PR about it is there, but not yet merged.

hardware/battery/battery_management_system_protocol.ksy
KOLANICH: See https://github.com/kaitai-io/kaitai_struct/issues/778 for nuances about each of the supported language.

hardware/battery/battery_management_system_protocol.ksy
Jakeler: Okay, changed in 87ee4a0

hardware/battery/battery_management_system_protocol.ksy
Jakeler: I would not want to rely on that, If there is currently no consistent behavior across supported languages.

hardware/battery/battery_management_system_protocol.ksy
Jakeler: Is a `checksum_input` under these circumstances impossible in kaitai, or am I doing just something wrong here?

hardware/battery/battery_management_system_protocol.ksy
Jakeler: True, size is redundant, it is removed now 042fc91

hardware/battery/battery_management_system_protocol.ksy
generalmimon: > Don't know, what is a enum per default? In any case, does it hurt to specify the type?

I think that @KOLANICH meant that when you do just this

```diff
           - id: discharge
             type: b1
-            enum: fet_bit
```

, the attribute `discharge` will be a boolean (`true`/`false`), which is already enough to specify the value logic. Binding it with the enum `fet_bit`:

https://github.com/kaitai-io/kaitai_struct_formats/blob/042fc917646aa60d35edcd0ba100d4b05bbdf7c1/hardware/battery/battery_management_system_protocol.ksy#L97-L100

is redundant, because it just describes a non-standard duplicate of the `false`/`true` logic (`off` meaning `false` and `on` = `true`).

Jakeler: Ahh now I understand, that is indeed redundant, removed in a623ed0.

hardware/battery/battery_management_system_protocol.ksy
Jakeler: a623ed0

hardware/battery/battery_management_system_protocol.ksy
generalmimon: > I tried:
> 
> ```yaml
>   checksum_input:
>     pos: 2
>     size: body._sizeof
> ```
> 
> But this does not work with current implementation because `dynamic sized type`.

Yes, the [`sizeof` operator](https://github.com/kaitai-io/kaitai_struct/issues/84) currently works only for what can be derived at compile time.

> Should this not also work with substreams?
> 
> ```yaml
>   checksum_input:
>     io: _root.body._io
>     size-eos: true
> ```
> 
> Result: `/instances/checksum_input/io: don't know how to call method '_io' of object type 'KaitaiStructType'`

Well, it turns out that referencing `_io` of a generic base type `KaitaiStruct` (which arose by combining different user-defined `types`) is not implemented. That's not intentional, [`_io` is a common method](https://doc.kaitai.io/user_guide.html#_user_defined_types_2) of all user-defined types, in fact it's possible to do it manually from application code even in static-typed languages like [Java](https://github.com/kaitai-io/kaitai_struct_java_runtime/blob/ea3abf89ee7745ac56c814202c4ddb26fc668e11/src/main/java/io/kaitai/struct/KaitaiStruct.java#L41), only in KSY compiler nobody has implemented it yet.

However, even if it is implemented, it will not help in this case, because `body` doesn't have its own substream (and thus `_root.body._io` would just return the same as `_root._io`). Substream is created if and only if there is some byte `size` delimitation, which can be done using keys `size`, `size-eos: true` or `terminator`. 

> Is a `checksum_input` under these circumstances impossible in kaitai, or am I doing just something wrong here?

I think the best thing you can do is to store [`_io.pos`](https://doc.kaitai.io/user_guide.html#_streams) before and after parsing `body` and then just positioning the `checksum_input` on those exact offsets:

```ksy
seq:
  - id: magic_start
    contents: [0xdd]
  - id: cmd
    type: u1
  - size: 0
    if: ofs_body_start < 0 # storing current position
  - id: body
    type:
      switch-on: cmd
      # ...
  - size: 0
    if: ofs_body_end < 0 # storing current position
  - id: checksum
    type: u2
    # ...
  # ...
instances:
  ofs_body_start:
    value: _io.pos
  ofs_body_end:
    value: _io.pos
  checksum_input:
    pos: ofs_body_start
    size: ofs_body_end - ofs_body_start
```

In fact, it's a direct workaround for https://github.com/kaitai-io/kaitai_struct/issues/84 not supporting run-time calculations.

hardware/battery/battery_management_system_protocol.ksy
generalmimon: The values `0xa5` and `0x5a` used in this switch

https://github.com/kaitai-io/kaitai_struct_formats/blob/d96be77a2b6b6e6aa7f8d8fdbc0188bc0abd7be3/hardware/battery/battery_management_system_protocol.ksy#L22-L28

should really be part of an enum - this is exactly what [enums](https://doc.kaitai.io/user_guide.html#enums) are intended for.

> I would not want to rely on that, if there is currently no consistent behavior across supported languages.

That's fine, but in that case please attach <code>-affected-by: <a href="https://github.com/kaitai-io/kaitai_struct/issues/778">778</a></code> on the `cmd` field:

```diff
   - id: cmd
     type: u1
+    -affected-by: 778
```

and at least resolve the integer keys from the enum:

```diff
   - id: body
     type:
       switch-on: cmd
       cases:
-        0xa5: read_req
-        0x5a: write_req
+        commands::read.to_i: read_req
+        commands::write.to_i: write_req
         _: response(cmd)
```

```diff
+enums:
+  commands:
+    0xa5: read
+    0x5a: write
```

Having the integer keys centralized in the enum is good, because it is less error-prone and transparent to reference the enum values like `commands::read` than repeating integer keys like `0xa5` or `165` at many places.

hardware/battery/battery_management_system_protocol.ksy
Jakeler: Thanks again! Added in 26f169a.

hardware/battery/battery_management_system_protocol.ksy
Jakeler: Done: 064f90a

generalmimon:
@Jakeler Can you please share any sample files/captures for testing?

Jakeler:
@generalmimon 
Sure, I have just rewritten my test cases: https://github.com/Jakeler/bms-parser/blob/master/py/test.py
Also some more examples are here: https://github.com/Jakeler/bms-parser/blob/master/py/checksum.py#L19
And a full communication log: https://github.com/Jakeler/bms-parser/blob/master/dumps/pc.log

Jakeler:
Anything left that blocks merging?

hardware/battery/battery_management_system_protocol.ksy
```diff
@@ -0,0 +1,246 @@
...
+    pos: ofs_body_start
+    size: ofs_body_end - ofs_body_start
+
+enums:
+  commands:
+    0xa5: read
+    0x5a: write
+
+types:
+  read_req:
```
KOLANICH, line 61: I guess `read_req` is the same as `write_req`, just without a payload.

generalmimon:
> [`hardware/battery/battery_management_system_protocol.ksy`](https://github.com/kaitai-io/kaitai_struct_formats/blob/064f90a8c8dd7d2f50a624c7a1632493d79df162/hardware/battery/battery_management_system_protocol.ksy)

Please move the spec one level up to `hardware/battery_management_system_protocol.ksy`. `battery/` is not a great name for a subdirectory - it's too specific. It's quite likely that it would stay alone in this subfolder forever.

Besides, I don't like this third dimension (directory - directory - file) anarchy - in most cases it does not make any sense, like having a directory [`serialization/asn1/`](https://github.com/kaitai-io/kaitai_struct_formats/tree/ecd988d18703e256315e62e56b74bf44d4ad3146/serialization/asn1/) with a single file [`asn1_der.ksy`](https://github.com/kaitai-io/kaitai_struct_formats/tree/ecd988d18703e256315e62e56b74bf44d4ad3146/serialization/asn1/asn1_der.ksy).

hardware/battery/battery_management_system_protocol.ksy
```diff
@@ -0,0 +1,246 @@
...
+    seq:
+      - id: req_cmd
+        type: u1
```
KOLANICH, line 64: is it an enum?

generalmimon: Most likely it is, because there is the following docblock on this attribute:

https://github.com/kaitai-io/kaitai_struct_formats/blob/064f90a8c8dd7d2f50a624c7a1632493d79df162/hardware/battery/battery_management_system_protocol.ksy#L61-L65

which means that it supposedly should have the same value logic as shown below:

https://github.com/kaitai-io/kaitai_struct_formats/blob/064f90a8c8dd7d2f50a624c7a1632493d79df162/hardware/battery/battery_management_system_protocol.ksy#L239-L245

And I totally agree that it should be backed by an enum instead of relying on a comment to convey this important information.

KOLANICH (approved):
These were just some notes. Though I don't think that fixing them would accelerate the merging: if you looked at PR counter you will see that there is plenty of non-merged PRs here. Some are ignored for several years. Though having a PR into this repo is still beneficial since it increases visibility and discoverability of your spec (there were several cases that 2 specs for the same format were implemented by 2 people simultaneously).

KOLANICH:
>Please move the spec one level up to hardware/battery_management_system_protocol.ksy. battery/ is not a great name for a subdirectory - it's too specific. It's quite likely that it would stay alone in this subfolder forever.

It was done because I have asked that. There are other battery-related formats, like SmartBattery

generalmimon:
> There are other battery-related formats, like SmartBattery which I have plans to implement in some future (though IDK how far is that future, I'd surely want to play a bit with some laptops batteries, but don't have currently time for that).

I'm sure of that. But I think that for such still relatively low number of format specifications in KSF right now, the greater depth of the directory tree is rather detrimental than beneficial - the added level of subdirectories currently does not help in maintaining the directory structure balanced. It merely opens a room for a total anarchy - there are no established rules what qualifies for being a balanced subdirectory name (or rather a _topic_). If there is just a flat structure of top-level directories, most of which make some sense (though some less than others, that's true - for example `machine_code` is pretty useless, why not just `hardware`?), it's easier for the implementers to just pick the most appropriate directory and put their spec in there without worrying about whether they should introduce a new subdirectory because someday in the future (like 5-10 years from now) might arise 1-2 new specifications that fit into the same subfolder.

I would not try to predict future and solve problems that do not exist - only once we have too many files in the individual top-level categories and it will become apparent that they could be split into several balanced subgroups, it will perhaps make sense.

hardware/battery/battery_management_system_protocol.ksy
Jakeler: Yes, it is an enum in theory, but same as already discussed on `cmd` this is affected by https://github.com/kaitai-io/kaitai_struct/issues/778

generalmimon: > Yes, it is an enum in theory, but same as already discussed on `cmd` this is affected by [kaitai-io/kaitai_struct#778](https://github.com/kaitai-io/kaitai_struct/issues/778)

OK, but why you don't just add these values to the existing `commands` enum:

https://github.com/kaitai-io/kaitai_struct_formats/blob/064f90a8c8dd7d2f50a624c7a1632493d79df162/hardware/battery/battery_management_system_protocol.ksy#L55-L58

and resolve them here

https://github.com/kaitai-io/kaitai_struct_formats/blob/064f90a8c8dd7d2f50a624c7a1632493d79df162/hardware/battery/battery_management_system_protocol.ksy#L239-L245

the same way you do here?

https://github.com/kaitai-io/kaitai_struct_formats/blob/064f90a8c8dd7d2f50a624c7a1632493d79df162/hardware/battery/battery_management_system_protocol.ksy#L25-L31

generalmimon (changes requested):

hardware/battery/battery_management_system_protocol.ksy
```diff
@@ -0,0 +1,246 @@
...
+        commands::read.to_i: read_req
+        commands::write.to_i: write_req
+        _: response(cmd)
+  - size: 0
+    if: ofs_body_end < 0 # storing current position
+
+  - id: checksum
+    type: u2
+    doc: |
+      Should be equal to the result from: 0x10000 - sum(body)
```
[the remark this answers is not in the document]
Jakeler: I admit that I have simplified the formula a bit, but this is really just a theoretical problem. Even if we are assuming the complete data consists of `0xff...` you would need 256 bytes to reach the 16 bits max value. The data length field is just one byte, so the body can not be larger than 256 bytes. 
Yes, there is also a status byte (which is `0x00` on every valid packet) and the length itself, but real data is never always `0xff` of course, so this easily would work in practice even if the length is pushed to the limit.

In reality there are even earlier limits, for example the balancing flags are for max. 32 cells, that would produce only 64 + 2 = 66 bytes in the cell voltages response and the other structures are constant size.
But I can add a note to the doc...

hardware/battery/battery_management_system_protocol.ksy
Jakeler: Because I think that is misleading, you would have two commands in a request, for example: read (0xa5) basic_info (0x03).
I can make another `registers` enum though.

hardware/battery/battery_management_system_protocol.ksy
```diff
@@ -0,0 +1,246 @@
...
+      - id: remain_cap
+        type: capacity
+      - id: typ_cap
+        type: capacity
+      - id: cycles
+        type: u2
+        doc: Cycle times
+      - id: prod_date
+        type: u2
+        doc: Production date
```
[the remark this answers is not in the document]
Jakeler: Thanks for the structure, that works.

Jakeler:
>  the added level of subdirectories currently does not help in maintaining the directory structure balanced. It merely opens a room for a total anarchy - there are no established rules what qualifies for being a balanced subdirectory name (or rather a _topic_).

I agree with that, even more directories don't make it better, especially if there are only 1-2 specs in it. I will move this one up.

> If there is just a flat structure of top-level directories, most of which make some sense (though some less than others, that's true - for example machine_code is pretty useless, why not just hardware?), it's easier for the implementers to just pick the most appropriate directory and put their spec in there without worrying about whether they should introduce a new subdirectory because someday in the future (like 5-10 years from now) might arise 1-2 new specifications that fit into the same subfolder.

That being said, from my viewpoint (being pretty new in kaitai) the top level structure is in its current state is already incredibly confusing. Looking just at the folders I can't decide where some specs should be. For example what even means `hardware`, it is definitions for hardware (3d models, schematic formats) or stuff that runs close to hardware or everything somewhat related? First thing that comes to my mind for "hardware and digital formats" is firmware, but this is here in another top level folder... Looking at the current hardware contents it is more like communication protocols between hardware.
It is very inconsistent, most of them are general topics (where it is used) but then there are folder for specific products like `windows` and `macos`, suddenly grouping by a different metric (where it comes from) and creating conflicts. For example if I want the windows event log format I would look in `windows`, but it is in `log` instead... What is higher priority here?
At the same time there is no linux/unix folder and typical formats like systemd-journal or sudo are scattered over the general topics, again being inconsistent.Why is there no `os` (operating systems) for all of them?
Then there are also too generic folders like `security` - what should be in there? Many things are somewhat related to security, but it is again a different metric and causing conflicts with grouping by usage. For example I would expect efivar signatures in `firmware` not `security`... maybe call it cryptography for key formats etc.
IMHO the whole tree should be reorganized into a low number of general folders, grouped by usage of the formats. Some depth is not always bad (2-3 levels max). For example `hardware/firmware`, `media/image`, `media/font`, `media/game`, (`media/audio`?), `os/filesystem`, `os/executable`, `os/log` would be sensible subdirectories...

Also I am sure you deeply involved guys have an idea in your head about what these categories should contain, but especially for newcomers it would be helpful to have a short definition of every category/folder in the readme as well.

Jakeler:
> These were just some notes. Though I don't think that fixing them would accelerate the merging: if you looked at PR counter you will see that there is plenty of non-merged PRs here. Some are ignored for several years. Though having a PR into this repo is still beneficial since it increases visibility and discoverability of your spec (there were several cases that 2 specs for the same format were implemented by 2 people simultaneously because of lack of communication, essentially wasting the effort).

Yeah, I already saw that, but why? It would increase the visibility even more if stuff gets actually merged at some point...

KOLANICH:
>It is very inconsistent, most of them are general topics (where it is used) but then there are folder for specific products like windows and macos, suddenly grouping by a different metric (where it comes from) and creating conflicts. For example if I want the windows event log format I would look in windows, but it is in log instead... What is higher priority here?

I absolutely aggree that the current system is complete mess. I also would like to remove the dirs named after OSes, create `0d`, `1d`, `2d/{vector,raster}`, `3d/{mesh,voxel,structural}` in the root, move `image` into `2d` into the needed dir, so do with CAD for drawings. Then move 3d models from `media` into `3d`.

KOLANICH:
Here is the spec for dirs metadata. https://github.com/kaitai-io/kaitai_struct_doc/pull/48 . Now we should decide where and how (create a PR into this repo, create a separate temporary repo from which the file wil be copied into this repo discarding all the history of its editing, or maybe a wiki page with a yaml block) are we going to cooperate on it.

hardware/battery/battery_management_system_protocol.ksy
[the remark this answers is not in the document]
generalmimon: > I admit that I have simplified the formula a bit, **but this is really just a theoretical problem**.

Perhaps, but I don't perceive this as a solid reason to intentionally write a checksum formula there that does not work in general case, especially when the official one in the original specification **does**. That's not how reliable software is written. What is the point? To _simplify_ the calculation? Is it worth to study the entire format and evaluate all its edge cases to make sure that in absolutely no case cannot the sum of the payload bytes exceed 65536, only to be able to "simplify" the expression from `-sum(checksum_input) & 0xffff` to `0x10000 - sum(checksum_input)`? What if the format will be extended at some point in the future (some structure will be added) and the `sum(checksum_input)` will become able to go over that 65536?

hardware/battery_management_system_protocol.ksy
```diff
@@ -0,0 +1,269 @@
...
+    if: ofs_body_end < 0 # storing current position
+
+  - id: checksum
+    type: u2
+    doc: |
+      Should be equal to the result from: 0x10000 - sum(body)
+      Where sum() is calculated over the value of every byte individually. This works
+      for up to 256 bytes, for more it must be clamped to 16 bits: result & 0xffff
+      Body includes everything besides magic start/end byte, cmd and checksum,
+      so excluding 2 bytes at the beginning and 3 bytes at the end.
```
generalmimon, line 42: You don't need to write it here, it is sufficiently specified by the `checksum_input` attribute. Anyone can simply access `checksum_input` and they will directly obtain the byte array that's needed to calculate the checksum. It is also apparent from the spec where these bytes come from.

PR #415, opened 2021-02-09
Add android's DTB/DTBO partitions format

evdenis:
Documentation: https://source.android.com/devices/architecture/dto/partitions
Reference implementation: https://android.googlesource.com/platform/system/libufdt/+/refs/heads/master/utils/

Test files (gunzip before testing):
[dtb-exynos9820.img.gz](https://github.com/kaitai-io/kaitai_struct_formats/files/5954481/dtb-exynos9820.img.gz)
[dtbo-G973F.img.gz](https://github.com/kaitai-io/kaitai_struct_formats/files/5954482/dtbo-G973F.img.gz)

### DIFF
--- a/archive/android_dto.ksy
+++ b/archive/android_dto.ksy
@@ -1,0 +1,79 @@
+meta:
+  id: android_dto
+  title: Android DTB/DTBO Partition
+  license: CC0-1.0
+  file-extension: img
+  endian: be
+
+doc: |
+  Format for Android DTB/DTBO partitions. It's kind of archive with
+  dtb/dtbo files. Used only when there is a separate unique partition
+  (dtb, dtbo) on an android device to organize device tree files.
+  The format consists of a header with info about size and number
+  of device tree entries and the entries themselves. This format
+  description could be used to extract device tree entries from a
+  partition images and decompile them with dtc (device tree compiler).
+
+doc-ref:
+  - https://source.android.com/devices/architecture/dto/partitions
+  - https://android.googlesource.com/platform/system/libufdt/+/refs/tags/android-10.0.0_r47
+
+seq:
+  - id: header
+    type: dt_table_header
+  - id: entries
+    type: dt_table_entry
+    repeat: expr
+    repeat-expr: header.dt_entry_count
+
+types:
+  dt_table_header:
+    seq:
+      - id: magic
+        contents: [ 0xd7, 0xb7, 0xab, 0x1e ]
+      - id: total_size
+        type: u4
+        doc: includes dt_table_header + all dt_table_entry and all dtb/dtbo
+      - id: header_size
+        type: u4
+        doc: sizeof(dt_table_header)
+      - id: dt_entry_size
+        type: u4
+        doc: sizeof(dt_table_entry)
+      - id: dt_entry_count
+        type: u4
+        doc: number of dt_table_entry
+      - id: dt_entries_offset
+        type: u4
+        doc: offset to the first dt_table_entry from head of dt_table_header
+      - id: page_size
+        type: u4
+        doc: flash page size
+      - id: version
+        type: u4
+        doc: DTBO image version
+  dt_table_entry:
+    seq:
+      - id: dt_size
+        type: u4
+        doc: size of this entry
+      - id: dt_offset
+        type: u4
+        doc: offset from head of dt_table_header
+      - id: id
+        type: u4
+        doc: optional, must be zero if unused
+      - id: rev
+        type: u4
+        doc: optional, must be zero if unused
+      - id: custom
+        type: u4
+        repeat: expr
+        repeat-expr: 4
+        doc: optional, must be zero if unused
+    instances:
+      body:
+        io: _root._io
+        pos: dt_offset
+        size: dt_size
+        doc: DTB/DTBO file
--- a/archive/android_dto.ksy
+++ b/archive/android_dto.ksy
@@ -30,7 +30,7 @@ types:
   dt_table_header:
     seq:
       - id: magic
-        contents: [ 0xd7, 0xb7, 0xab, 0x1e ]
+        contents: [0xd7, 0xb7, 0xab, 0x1e]
       - id: total_size
         type: u4
         doc: includes dt_table_header + all dt_table_entry and all dtb/dtbo


archive/dto.ksy
```diff
@@ -0,0 +1,57 @@
+meta:
+  id: dto
+  title: DTB/DTBO
+  license: CC0-1.0
+  file-extension: img
+  endian: be
+
```
generalmimon, line 7: Would you add at least a few sentences about the format in [the `/doc` key](https://doc.kaitai.io/ksy_style_guide.html#doc), i.e. some general information about the format? What it is used for? What kind of data it stores? What is the main usage field?

archive/dto.ksy
```diff
@@ -0,0 +1,57 @@
...
+    - id: magic
+      contents: [ 0xd7, 0xb7, 0xab, 0x1e ]
+    - id: total_size
+      type: u4
+    - id: header_size
+      type: u4
+    - id: dt_entry_size
+      type: u4
+    - id: dt_entry_count
+      type: u4
```
generalmimon, line 24: `num_entries` is preferred, see the [KSY Style Guide](https://doc.kaitai.io/ksy_style_guide.html#attr-id). If you want to keep a reference to the original name, use [`-orig-id`](https://doc.kaitai.io/user_guide.html#orig-id).

archive/dto.ksy
```diff
@@ -0,0 +1,57 @@
...
+    - id: page_size
+      type: u4
+    - id: version
+      type: u4
+  entry:
+    seq:
+    - id: size
+      type: u4
+    - id: offset
+      type: u4
```
generalmimon, line 36: Names of these attributes are inadequate, because they lack any information about which field they relate to. It is recommended to call them `len_body` and `ofs_body`, see the [style guide](https://doc.kaitai.io/ksy_style_guide.html#attr-id).

archive/dto.ksy
```diff
@@ -0,0 +1,57 @@
+meta:
+  id: dto
```
generalmimon, line 2: Wouldn't `android_dto` be a more appropriate name, when we already have https://formats.kaitai.io/android_img/index.html?

archive/dto.ksy
```diff
@@ -0,0 +1,57 @@
...
+        pos: offset
+        size: size
+
+seq:
+  - id: header
+    type: header
+  - id: entries
+    type: entry
+    repeat: expr
+    repeat-expr: header.dt_entry_count
```
generalmimon, line 57: `seq` should go earlier in the file then `types`, see https://doc.kaitai.io/ksy_style_guide.html#type.

archive/dto.ksy
```diff
@@ -0,0 +1,57 @@
+meta:
+  id: dto
+  title: DTB/DTBO
+  license: CC0-1.0
+  file-extension: img
+  endian: be
+
+doc-ref:
+  - https://source.android.com/devices/architecture/dto/partitions
+  - https://android.googlesource.com/platform/system/libufdt/+/refs/heads/master/utils/
```
generalmimon, line 10: Can you craft a permanent link to a specific version? The tag `master` changes over time, so after 1-2 years it won't be clear what you were referring to (if one won't rely on internet archives to dig up what `master` is referring to at _this_ time). Moreover, they change the directory structure of the repository and the `utils/` folder will no longer exist, for example. So please, either link to a commit hash or some version tag so that we're future proof.

archive/dto.ksy
```diff
@@ -0,0 +1,57 @@
...
+      repeat-expr: 4
+    instances:
+      body:
+        io: _root._io
+        pos: offset
+        size: size
+
+seq:
+  - id: header
+    type: header
```
generalmimon, line 53: The same name of an attribute and an subtype will cause problems in C#, see https://github.com/kaitai-io/kaitai_struct/issues/86. Not that it would be a big deal, just saying :)

archive/dto.ksy
```diff
@@ -0,0 +1,57 @@
...
+  header:
+    seq:
+    - id: magic
+      contents: [ 0xd7, 0xb7, 0xab, 0x1e ]
+    - id: total_size
+      type: u4
+    - id: header_size
+      type: u4
+    - id: dt_entry_size
+      type: u4
```
generalmimon, line 22: Is it OK to ignore these fields? What do they mean?

generalmimon (changes requested):
In addition to the comments I wrote below, I would consider appropriate to add some short descriptions and notes to the individual fields whose meaning might be unclear - what is `page_size`, `version`, `id`, `rev` or `custom`?

archive/dto.ksy
KOLANICH: I feel like it may be not Android-specific. https://github.com/devicetree-org/devicetree-specification/tree/master/source

archive/dto.ksy
evdenis: Actually, I can't agree with you here. Names are known, they are fixed in documentation and they are "compatible"  with id naming restrictions. https://source.android.com/devices/architecture/dto/partitions I tried to replace them according to the style guide and adding -orig-id and it looks bad. Changing the fields names only introduces mess here. But I agree with you that I need to add fields-level docs.

archive/dto.ksy
```diff
@@ -0,0 +1,79 @@
+meta:
+  id: android_dto
```
generalmimon, line 2: Also please rename the file accordingly: ~`archive/dto.ksy`~ &xrarr; **`archive/android_dto.ksy`**.

archive/android_dto.ksy
```diff
@@ -0,0 +1,79 @@
...
+      type: u4
+      doc: flash page size
+    - id: version
+      type: u4
+      doc: DTBO image version
+  dt_table_entry:
+    seq:
+    - id: dt_size
+      type: u4
+      doc: size of this entry
```
generalmimon, line 59:
```suggestion
  dt_table_entry:
    seq:
      - id: dt_size
        type: u4
        doc: size of this entry
```

See [KSY Style Guide > 1. General formatting](https://doc.kaitai.io/ksy_style_guide.html#general):

>
> <table>
> <thead>
> <tr>
> <th align=left>Good</th>
> <th align=left>Bad</th>
> </tr>
> </thead>
> <tbody>
> <tr>
> <td>
> 
> ```yaml
> imports:
>   - foo
>   - bar
> ```
>
> </td>
> <td>
> 
> ```yaml
> imports:
> - foo
> - bar
> ```
>
> </td>
> </tr>
> </tbody>
> </table>

Please indent all `seq` fields like that.

evdenis:
Example usage: https://github.com/CruelKernel/scripts/blob/master/dumpbootimg#L250

generalmimon (approved):

Issue #209, opened 2019-07-25
glTF


KOLANICH:
```yaml
meta:
  id: gltf
  title: GL Transmission Format, binary container
  file-extension: glb
  xref:
    wikidata: Q28135989

doc: |
  glTF is a format for distribution of 3D models optimized for being used in software

doc-ref:
  - https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#glb-file-format-specification
```

generalmimon:
Oh, we've kind of forgot about this issue after merging https://github.com/kaitai-io/kaitai_struct_formats/pull/445. However, perhaps this issue should not be closed yet, because the current [`3d/gltf_binary.ksy`](https://github.com/kaitai-io/kaitai_struct_formats/blob/3e4be117e4c09aa4d4d206c9d4aa21bd8f3fcc5d/3d/gltf_binary.ksy) is to my taste unfinished and not very useful at this point - it merely parses one character string with JSON data and one byte array with binary data, without any attempt to provide a way of parsing the actual binary structures that are present in that byte block.

My proof of concept in https://github.com/kaitai-io/kaitai_struct_formats/pull/445#discussion_r598117321 can be used as a basis for creating a more in-depth KSY spec, which could then be considered more complete and usable out-of-the-box.

PR #420, opened 2021-02-13
Initial import of rendering / modelling formats

hakanai:
Imported from [my organisation's repository](https://github.com/ephemeral-laboratories/modelling_formats).

Since this is a new directory - is "rendering" a good name for this stuff? In IANA, these formats are classified as just "model", but the word has too many different meanings in software.

About these formats:

FBX - "Filmbox" format originally by Kaydara, now owned by Autodesk, standard interchange format between modelling software and game engines and the like.

GLB - Khronos group's standard binary format for model exchange. But unfortunately a terrible format for Kaitai. I implemented it as far in as was possible anyway.

The above two file types can be created using Blender. (Actually, Blender's own format is one that I want to take a look at as well.)

PMD, PMX - Model formats used by animation software MikuMikuDance (PMX newer, PMD older), many samples can be found on bowlrowl and similar sites.

VMD - Animation data format used by MikuMikuDance.

### DIFF
--- a/rendering/fbx.ksy
+++ b/rendering/fbx.ksy
@@ -1,0 +1,148 @@
+meta:
+  id: fbx
+  title: FBX 3D model interchange format
+  file-extension: fbx
+  encoding: UTF-8
+  endian: le
+  license: MIT
+
+seq:
+  - id: header
+    type: header
+  - id: records
+    type: node_record
+    repeat: until
+    repeat-until: _.end_offset == 0
+
+types:
+
+  header:
+    seq:
+      - id: magic
+        contents: "Kaydara FBX Binary  \0"
+        size: 20
+      - id: more_magic
+        contents: [0x1A, 0x00]
+      - id: version
+        type: u4
+
+  node_record:
+    seq:
+      - id: end_offset
+        type: u4
+      - id: property_count
+        type: u4
+      - id: property_list_length
+        type: u4
+      - id: name_length
+        type: u1
+      - id: name
+        type: str
+        size: name_length
+      - id: properties
+        type: property
+        repeat: expr
+        repeat-expr: property_count
+        size: property_list_length
+      - id: children
+        size: end_offset - _root._io.pos
+        type: node_record
+        repeat: until
+        repeat-until: _.end_offset == 0
+        if: not _io.eof
+
+  property:
+    seq:
+      - id: type_code
+        size: 1
+        type: str
+      - id: data
+        type:
+          switch-on: type_code
+          cases:
+            '"I"': s4
+            '"L"': s8
+            '"F"': f4
+            '"D"': f8
+            '"Y"': s2
+            '"C"': u1 # TODO boolean, should probably have a type for it
+            '"i"': array(1)
+            '"l"': array(2)
+            '"f"': array(3)
+            '"d"': array(4)
+            '"b"': array(5)
+            '"S"': string_data
+            '"R"': raw_binary_data
+
+  string_data:
+    seq:
+      - id: length
+        type: u4
+      - id: value
+        type: str
+        size: length
+  
+  raw_binary_data:
+    seq:
+      - id: length
+        type: u4
+      - id: value
+        size: length
+
+  array:
+    params:
+      # Works around lack of ability to parameterise the actual type by passing a fake one through.
+      # https://github.com/kaitai-io/kaitai_struct/issues/135
+      - id: pseudo_type
+        type: u1
+    seq:
+      - id: element_count
+        type: u4
+      - id: encoding
+        type: u1
+      - id: compressed_size
+        type: u4
+      - id: elements
+        # Can't switch for process, so switch for type as a workaround.
+        # https://github.com/kaitai-io/kaitai_struct/issues/374
+        type:
+          switch-on: encoding
+          cases:
+            0: array_data(pseudo_type, element_count)
+            1: array_data_compressed(pseudo_type, element_count)
+        size: compressed_size
+
+  array_data_compressed:
+    params:
+      - id: pseudo_type
+        type: u1
+      - id: element_count
+        type: u4
+    seq:
+      - id: data
+        type: array_data(pseudo_type, element_count)
+        process: zlib
+        size-eos: true
+    instances:
+      elements:
+        value: data.elements
+
+  array_data:
+    params:
+      - id: pseudo_type
+        type: u1
+      - id: element_count
+        type: u4
+    seq:
+      - id: elements
+        type:
+          switch-on: pseudo_type
+          cases:
+            1: s4
+            2: s8
+            3: f4
+            4: f8
+            5: u1
+        repeat: expr
+        repeat-expr: element_count
+
--- a/rendering/glb.ksy
+++ b/rendering/glb.ksy
@@ -4,8 +4,8 @@ meta:
   file-extension: glb
   endian: le
   license: MIT
-  doc-ref: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#binary-gltf-layout
   xref:
+    doc-ref: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#binary-gltf-layout
     mime: model/gltf-binary
 
 seq:
--- a/rendering/glb.ksy
+++ b/rendering/glb.ksy
@@ -1,0 +1,64 @@
+meta:
+  id: glb
+  title: glTF binary format
+  file-extension: glb
+  endian: le
+  license: MIT
+  doc-ref: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#binary-gltf-layout
+  xref:
+    mime: model/gltf-binary
+
+seq:
+  - id: header
+    type: header
+  - id: chunks
+    type: chunk
+    repeat: eos
+
+types:
+
+  header:
+    seq:
+      - id: magic
+        contents: 'glTF'
+      - id: version
+        type: u4
+      - id: length
+        type: u4
+
+  chunk:
+    seq:
+      - id: length
+        type: u4
+      - id: type
+        type: u4
+        enum: chunk_type
+      - id: data
+        size: length
+        type:
+          switch-on: type
+          cases:
+            'chunk_type::json': json
+            'chunk_type::bin': bin
+
+  json:
+    seq:
+      - id: data
+        type: str
+        encoding: UTF-8
+        size-eos: true
+        doc: |
+          This is where GLB deviates from being an elegant format.
+          To parse the rest of the file, you have to parse the JSON first.
+
+  bin:
+    seq:
+      - id: data
+        size-eos: true
+
+enums:
+  chunk_type:
+    # Literally "JSON"
+    0x4E4F534A: json
+    # Literally "BIN\0"
+    0x004E4942: bin
--- a/rendering/mmd_pmd.ksy
+++ b/rendering/mmd_pmd.ksy
@@ -1,0 +1,368 @@
+meta:
+  id: mmd_pmd
+  title: MMD model data (older format)
+  application: MikuMikuDance
+  file-extension: pmd
+  encoding: Shift_JIS
+  endian: le
+  license: MIT
+
+seq:
+  - id: header
+    type: header
+  - id: vertex_count
+    type: u4
+  - id: vertices
+    type: vertex
+    repeat: expr
+    repeat-expr: vertex_count
+  - id: face_vertex_count
+    type: u4
+  - id: faces
+    type: face
+    repeat: expr
+    repeat-expr: face_vertex_count / 3
+  - id: material_count
+    type: u4
+  - id: materials
+    type: material
+    repeat: expr
+    repeat-expr: material_count
+  - id: bone_count
+    type: u2
+  - id: bones
+    type: bone
+    repeat: expr
+    repeat-expr: bone_count
+  - id: ik_count
+    type: u2
+  - id: iks
+    type: ik
+    repeat: expr
+    repeat-expr: ik_count
+  - id: morph_count
+    type: u2
+  - id: morphs
+    type: morph
+    repeat: expr
+    repeat-expr: morph_count
+  - id: morph_frame_count
+    type: u1
+  - id: morph_frames
+    type: morph_frame
+    repeat: expr
+    repeat-expr: morph_frame_count
+  - id: bone_frame_name_count
+    type: u1
+  - id: bone_frame_names
+    type: bone_frame_name
+    repeat: expr
+    repeat-expr: bone_frame_name_count
+  - id: bone_frame_count
+    type: u4
+  - id: bone_frames
+    type: bone_frame
+    repeat: expr
+    repeat-expr: bone_frame_count
+  - id: english_header
+    type: english_header
+  - id: english_bone_names
+    type: english_bone_name
+    repeat: expr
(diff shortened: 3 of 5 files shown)


rendering/fbx.ksy
```diff
@@ -0,0 +1,148 @@
+meta:
+  id: fbx
```
KOLANICH, line 2: Maybe `kaydara_fbx`?

rendering/fbx.ksy
```diff
@@ -0,0 +1,148 @@
...
+  title: FBX 3D model interchange format
```
KOLANICH, line 3: The word `format` is usualy redundant. Maybe just `Kaydara FiLMBOX 3D Model`?

rendering/fbx.ksy
```diff
@@ -0,0 +1,148 @@
...
+  file-extension: fbx
```
KOLANICH, line 4: It'd be nice to add `application: Autodesk/Kaydara FiLMBOX`

rendering/fbx.ksy
```diff
@@ -0,0 +1,148 @@
...
+    type: node_record
+    repeat: until
+    repeat-until: _.end_offset == 0
+
+types:
+
+  header:
+    seq:
+      - id: magic
+        contents: "Kaydara FBX Binary  \0"
```
KOLANICH, line 22:
```yaml
type: strz
valid: "Kaydara FBX Binary "
```

rendering/fbx.ksy
```diff
@@ -0,0 +1,148 @@
...
+        size: 20
+      - id: more_magic
+        contents: [0x1A, 0x00]
```
KOLANICH, line 25: Looks like `u2`. You can use `valid` here too.

rendering/fbx.ksy
hakanai: I think I'll keep normal capitalisation of Filmbox but otherwise agreed.

rendering/fbx.ksy
```diff
@@ -0,0 +1,148 @@
...
+      - id: pseudo_type
+        type: u1
+      - id: element_count
+        type: u4
+    seq:
+      - id: elements
+        type:
+          switch-on: pseudo_type
+          cases:
+            1: s4
```
KOLANICH, line 141: The values should probably be members of a enum.

rendering/fbx.ksy
hakanai: Will add. I had wondered about this. The native application was Filmbox but now it's used by damn near everybody. Everybody using the Autodesk FBX SDK for a start, but then even implementations in open source like Blender's.

rendering/glb.ksy
```diff
@@ -0,0 +1,64 @@
+meta:
+  id: glb
+  title: glTF binary format
```
KOLANICH, line 3: Nice, long-waited format :+1:

rendering/fbx.ksy
```diff
@@ -0,0 +1,148 @@
+meta:
+  id: fbx
+  title: FBX 3D model interchange format
+  file-extension: fbx
+  encoding: UTF-8
+  endian: le
+  license: MIT
+
```
KOLANICH, line 8: Lacks `eoc` and `doc-ref`. Also may make sense to provide a link to a publicly-available sample file.

rendering/fbx.ksy
hakanai: Is valid just the same as contents but you get a different exception when it's wrong?

rendering/glb.ksy
```diff
@@ -0,0 +1,64 @@
...
+  file-extension: glb
+  endian: le
+  license: MIT
+  xref:
+    doc-ref: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#binary-gltf-layout
```
KOLANICH, line 8: Incorrect use of `doc-ref`. Feel free to merge your block with https://github.com/kaitai-io/kaitai_struct_formats/issues/209 Also add `Closes: #209` into PR text.

rendering/glb.ksy
```diff
@@ -0,0 +1,64 @@
...
+      - id: data
+        type: str
+        encoding: UTF-8
+        size-eos: true
+        doc: |
+          This is where GLB deviates from being an elegant format.
+          To parse the rest of the file, you have to parse the JSON first.
+
+  bin:
+    seq:
```
KOLANICH, line 55: Looks unfinished.

rendering/mmd_pmd.ksy
```diff
@@ -0,0 +1,368 @@
+meta:
+  id: mmd_pmd
+  title: MMD model data (older format)
```
KOLANICH, line 3: I guess here it shouldn't be abbreviated. Also please put each format into an own PR.

rendering/fbx.ksy
hakanai: I tried this, and I get an error that I don't fully understand.

```
/types/header/seq/0: parsing expression 'Kaydara FBX Binary  ' failed on 1:9, expected "or" | CharsWhile(Set( , n)) | "\\\n" | End
```

Wait, is `valid` expecting an _expression_?

rendering/mmd_pmx.ksy
```diff
@@ -0,0 +1,569 @@
...
+        type: len_string
+      - id: diffuse
+        type: f4_4
+      - id: specular
+        type: f4_3
+      - id: shininess
+        type: f4
+      - id: ambient
+        type: f4_3
+      - id: flag
```
KOLANICH, line 174: Very ambigous name. Is it a single boolean (in this case name should start from `is_` and describe the meaning). Or is it binary flags (in this case it should be not `u1`, but a type, in which bit-sized numbers reside, `b1` is a bit and has boolean type instead)?

rendering/mmd_pmx.ksy
```diff
@@ -0,0 +1,569 @@
...
+        type: u1
+      - id: edge_color
+        type: f4_4
+      - id: edge_size
+        type: f4
+      - id: texture_index
+        type:
+          switch-on: _root.header.texture_index_size
+          cases:
+            1: u1
```
KOLANICH, line 184: Again should be enums.

rendering/fbx.ksy
hakanai: `eoc`?

rendering/glb.ksy
hakanai: Incorrect use? I originally had it one level up but the compiler gave an error that didn't tell me how to fix it, and then didn't complain when I moved it inside the refs section.

What's the right place?

I have to admit, I'm not used to working with doc-ref because I usually work on formats with zero documentation.

rendering/fbx.ksy
KOLANICH: How about `'"Kaydara FBX Binary "'`?

rendering/glb.ksy
hakanai: In what sense? A GLB file is made up of data chunks. The chunk can be either a JSON chunk or a binary chunk. This is a binary chunk. There's no way to know what sort of data you're looking at in here though, because, for that, you'd have to first parse the JSON to get the offsets out of there.

rendering/fbx.ksy
KOLANICH: `valid` allows multiple ways of validation, `valid: something` is just a shortcut for `valid: eq: something`

rendering/fbx.ksy
KOLANICH: `doc`. A typo.

rendering/mmd_pmd.ksy
hakanai: I'll just put in both like `MMD (MikuMikuDance)` since MMD is what people are going to be using as their search term.

As for splitting the PR... ugh. Yeah, sure. Once I'm done with the first round of changes I'll split it and resubmit.

rendering/glb.ksy
KOLANICH: What's the right place?

Not in `meta`, just under `doc`, wuich is a sibling of `meta`.

rendering/mmd_pmx.ksy
hakanai: The meaning of the value is "the number of bytes in the index". Usually when you have a number of things you want to use an int, but here you're suggesting an enum.

Perhaps it might be better to keep as int but declare `valid` for the index size stuff as only valid if 1,2,4?

hakanai: So I will add new types for each index type like this:

```yaml
  bone_index:
    doc: Variable-length type storing a bone index
    seq:
      - id: value
        type: switch-on: _root.header.bone_index_size
        cases:
          1: u1
          2: u2
          4: u4
```

rendering/glb.ksy
KOLANICH: Please look at https://github.com/kaitai-io/kaitai_struct_formats/pull/183/files#diff-ce5a8a6f65e08325535c534e8222ecbd9f53a9749ba538b04cf857b5b07a0795R89 and https://github.com/kaitai-io/kaitai_struct_formats/pull/213/files#diff-35cf895592a530eee852a5354aaa16e650100fdfddb1e81752f7e6d0ef036786R115 

for the ideas how one can try to solve the issue of having data in some another format.

rendering/fbx.ksy
hakanai: Ah. Well, they don't document the format so there was nothing to put in I guess. Blender had an attempt at documenting the format - is it okay to put in a third-party document describing the format?

rendering/glb.ksy
hakanai: Aha. It would have been nice for the compiler to say that, instead of just saying that it was wrong.

rendering/fbx.ksy
hakanai: Works. Same workaround works for the array one as well.

rendering/fbx.ksy
hakanai: Ah nice, a declarative predicate language.

rendering/fbx.ksy
hakanai: Having issues getting this to compile with an enum... 

```
C:\Users\Trejkaz\Documents\GitHub\kaitai_struct_formats\rendering>kaitai-struct-compiler -t ruby fbx.ksy
/types/property/seq/1/type/cases/Str(i)/type: can't pass argument #0 of type Int1Type(true) into parameter `pseudo_type` of type EnumType(List(pseudo_type),Int1Type(false))
```

rendering/glb.ksy
hakanai: Neither of those appear to actually solve the issue, but rather push it out to a single-language solution, which doesn't really fit with the spirit of Kaitai anyway.

rendering/mmd_pmx.ksy
hakanai: Now I have removed repetition in the structures above, and have instead 6 repetitive types which only differ in three places:
- their name
- the doc comment
- the field they use to switch on

rendering/fbx.ksy
KOLANICH: They don't but you can put there some words from you, at least a brief summary of the format is needed: what its purpose is, what it is it used for, + anything you consider non-obvious.

rendering/glb.ksy
KOLANICH: > Is the compiler just not validating this section at all?

It is likely that in this section keys can be any strings.

rendering/glb.ksy
KOLANICH: >Neither of those appear to actually solve the issue, but rather push it out to a single-language solution which doesn't really fit with the spirit of Kaitai anyway.

I completely aggree.

>What would be nice is a native JSON type in Kaitai, so that we can then just traverse into the tree of JSON objects.

It would be nice to have KS extensible, because blessing JSON over other serialization formats, including custom ones, is incorrect. And the impl still will be language-dependent. The same way as runtimes are and KS itself is. All of this is just an abstraction layer around languages.

>So I guess one valid solution is, define a new format for JSON and import that format, so we can declare the type of this piece as json. And then say that implementing the JSON format is up to the reader. LOL

So a valid solution is finishing and landing interfaces proposal first, and then standardizing some interfaces and creating a standard library for each interface. See the discussion in #457.

rendering/fbx.ksy
KOLANICH: I have encountered this some times. To solve that I had to sometimes to cast to ints and back. But sometimes passing enums just works. Needs a bit of experimentation. But if a param is a enum, it should receve a enum, if it is just wn int, it should receive an int.

rendering/mmd_pmx.ksy
KOLANICH: You can pass the condition in `switch-on` as a param, BTW.

hakanai: I think I tried that in this situation once in the past and it didn't work, but it will be tried.

hakanai: OK, I got it down to:

```yaml
      - id: indices
        type: sized_index(_root.header.vertex_index_size)
```

I think I'll call this _good enough_ but it would be nice if I could write

```yaml
      - id: indices
        type: sized_index(index_id::vertex)
```

But (a) I'd be declaring an enum with fictional values just to use it as a param and (b) I'm 99% sure that the last time I tried using enums in params they didn't work anyway.

generalmimon: > But (a) I'd be declaring an enum with fictional values just to use it as a param and (b) I'm 99% sure that the last time I tried using enums in params they didn't work anyway.

They were added in the 0.9 release, see the [release notes](https://kaitai.io/news/2020/10/16/kaitai-struct-v0.9-released.html):

> - Major bugfixes:
>   - `params/type` - add support for:
>     - specific user types
>     - `enum` types (#413)
>     - byte arrays (`bytes`)
>     - arrays (`u2[]`, `struct[]`, etc.)

rendering/glb.ksy
hakanai: I wasn't saying that json would be the only format to implement, nor was I saying that json would be implemented by kaitai itself. I was just saying that we could implement a placeholder 'type' for any given unknown type, and then how that's implemented is up to the future. If the future happens to be some interface where the type is converted to something else using external code, that's fine. If someone figures out how to complete a kts file for json, that's fine too.

rendering/glb.ksy
hakanai: Sounds like a recipe for disaster...

In any case, it's fixed now.

rendering/fbx.ksy
hakanai: Got it.

rendering/fbx.ksy
hakanai: Should I be changing the filename to match? I notice the generated code ends up using this ID as the classname instead of the filename of the ksy. (Sucks that we have essentially the same info in two places...)

rendering/mmd_pmx.ksy
hakanai: Similar problem with the ones below as well.

I was able to figure out that this one contains _multiple_ flags, but have no docs on what the flags are. At least for now I'll rename it to `flags`. The ones below are more interesting though.

hakanai:
Closing as I'm splitting into multiple PRs as previously discussed.

PR #421, opened 2021-02-19
Update swf.ksy

Whiteseeker:
As I'm digging more and more into swf, I've added some of the tag that are missing (a lot more are missing, if I get time I might make further modification to it).
Cool stuff with kaitai 👍

### DIFF
--- a/executable/swf.ksy
+++ b/executable/swf.ksy
@@ -4,17 +4,9 @@ meta:
   file-extension: swf
   xref:
     justsolve: SWF
-    pronom: fmt/507
-      # - fmt/{505..507} # Adobe Flash {8..10}
-      # - fmt/{757..776} # Adobe Flash {11..30}
-    loc:
-      - fdd000130 # SWF 7
-      - fdd000248 # SWF 8
+    loc: fmt/507
     mime: application/x-shockwave-flash
     wikidata: Q594447
-  tags:
-    - executable
-    - media
   license: CC0-1.0
   endian: le
 #  imports:
@@ -90,6 +82,20 @@ types:
         type: u1
       - id: b
         type: u1
+  rgba:
+    seq:
+      - id: r
+        type: u1
+      - id: g
+        type: u1
+      - id: b
+        type: u1
+      - id: a
+        type: u1
+  matrix:
+    seq:
+      - id: has_scale
+        type: b1
   tag:
     seq:
       - id: record_header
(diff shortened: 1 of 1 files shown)


executable/swf.ksy
```diff
@@ -104,7 +110,94 @@ types:
             'tag_type::script_limits': script_limits_body
             'tag_type::symbol_class': symbol_class_body
             'tag_type::set_background_color': rgb
+            'tag_type::define_sprite': define_sprite_body
             'tag_type::export_assets': symbol_class_body
+            'tag_type::define_shape4': define_shape4_body
+  define_shape4_body:
```
KOLANICH, line 116: :+1:

executable/swf.ksy
```diff
@@ -104,7 +110,94 @@ types:
...
+      - id: fill_styles
+        type: fill_style
+        repeat: expr
+        repeat-expr: fill_style_count
+  fill_style:
+    seq:
+      - id: fill_style_type
+        type: u1
+        enum: fill_style_type_enum
+      - id: color_a
```
KOLANICH, line 155: Plese add some words (and maybe change `id`s) about how these 2 colors are different and why we need both.

executable/swf.ksy
```diff
@@ -104,7 +110,94 @@ types:
...
+        type: shape_with_style
+  shape_with_style:
+    seq:
+      - id: fill_styles
+        type: fill_style_array
+  fill_style_array:
+    seq:
+      - id: fill_style_count
+        type: u1
+      - id: fill_style_count_extended
```
KOLANICH, line 143: Is this a variable-length integer?

executable/swf.ksy
```diff
@@ -104,7 +110,94 @@ types:
...
+      - id: gradient_matrix
+        type: matrix
+        if: fill_style_type == fill_style_type_enum::linear_gradient or fill_style_type == fill_style_type_enum::radial_gradient or fill_style_type == fill_style_type_enum::focal_gradient
+      - id: gradient
+        type: gradient_type
+        if: fill_style_type == fill_style_type_enum::linear_gradient or fill_style_type == fill_style_type_enum::radial_gradient
+  gradient_type:
+    seq:
+      - id: spread_mode
+        type: b2
```
KOLANICH, line 170: are these enums?

executable/swf.ksy
```diff
@@ -104,7 +110,94 @@ types:
...
+        type: grad_record
+        repeat: expr
+        repeat-expr: num_gradient
+  grad_record:
+    seq:
+      - id: ratio
+        type: u1
+      - id: color_a
+        type: rgba
+        if:  _parent._parent._parent._parent._parent._parent.record_header.tag_type == tag_type::define_shape4
```
KOLANICH, line 185: Can it be shortened anyhow?

KOLANICH (changes requested):
LGTM

executable/swf.ksy
```diff
@@ -4,17 +4,9 @@ meta:
   file-extension: swf
   xref:
     justsolve: SWF
-    pronom: fmt/507
```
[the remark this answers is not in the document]
Whiteseeker: I edited the version of swf.ksy in the web editor and that come from there:
image

I was kind of expecting that the webeditor would fetch the last version of the grammars from github but I think I was wrong and they aren't in sync with github's grammar. 
I'll just cancel the PR and redo it later.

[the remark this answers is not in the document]
Whiteseeker: Ah true I can't close the PR. Feel free to close it and I will redo one later when I implemented a bit more of the 215 pages of specifications of the swf format :)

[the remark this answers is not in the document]
KOLANICH: I see no reason to close, just force-push.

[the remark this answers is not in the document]
Whiteseeker: Well I've done quite more work on the SWF format. I'll update the PR next week.

executable/swf.ksy
Whiteseeker: No. The specs of SWF says just: 
image

Whiteseeker: So it has both in case it's extended, fill_style_count is set at 0xFF and the real value would be in fill_style_count_extended.
They do that to save size as most of the have value below 0xFF

executable/swf.ksy
Whiteseeker: yes I'll add them

executable/swf.ksy
Whiteseeker: Yeah I found a way, will fix

Whiteseeker (commented):
I made a new commit to fix these comments.

executable/swf.ksy
KOLANICH: It is still unclear what it means.

executable/swf.ksy
```diff
@@ -104,7 +152,184 @@ types:
...
+    seq:
+      - id: fill_style_count
+        type: u1
+      - id: fill_style_count_extended
+        type: u2
+        if: fill_style_count == 0xFF
+      - id: fill_styles
+        type: fill_style(shape_version)
+        repeat: expr
+        repeat-expr: fill_style_count
```
KOLANICH, line 270: Is it really a `fill_style_count` but not `(fill_style_count!=0xff?fill_style_count:fill_style_count_extended)`?

executable/swf.ksy
KOLANICH: OK. Also, do you take into account

>Supported only for Shape2 and Shape3

?

executable/swf.ksy
```diff
@@ -90,6 +96,48 @@ types:
         type: u1
       - id: b
         type: u1
+  # Swf specs has 3 types for color, rgb, rgba and arg.
```
KOLANICH, line 99: This comment doesn't bring much clarity.

Whiteseeker: True that. Happy to remove it. I think It's not obvious how to explain a 215 page specs in a few comments 😓

executable/swf.ksy
Whiteseeker: You are right, I had no idea kaitai supporterd if / ternary operator in the repeat-expr field. Will fix (still new to kaitai here 😅 )

Whiteseeker: Fixed in 1efe4b15c689ea88561c6191116d47a976736882

executable/swf.ksy
```diff
@@ -192,14 +417,44 @@ enums:
     0x46: none # F
     0x43: zlib # C
     0x5a: lzma # Z
+  join_style:
+    0: round_join
```
KOLANICH, line 421: the postfixes are unneeded

executable/swf.ksy
```diff
@@ -192,14 +417,44 @@ enums:
...
+    0x43: non_smoothed_clipped_bitmap
+  spread_mode_enum:
+    0: pad_mode
+    1: reflect_mode
+    2: repeat_mode
+    3: reserved
+  interpolation_mode_enum:
+    0: normal_rgb
+    1: linear_rgb
+    2: reserved
```
KOLANICH, line 445: won't the same ids cause issues?

executable/swf.ksy
```diff
@@ -104,7 +152,184 @@ types:
...
+  fill_style:
+    params:
+      - id: shape_version
+        type: u1
+    seq:
+      - id: fill_style_type
+        type: u1
+        enum: fill_style_type_enum
+      - id: color_a
+        type: rgba
```
KOLANICH, line 280:
```yaml
- id: color
  type:
    switch-on: shape_version >= 4
    cases:
      true: rgba
      false: rgb
  if: fill_style_type == fill_style_type_enum::solid
```

executable/swf.ksy
generalmimon: > won't the same ids cause issues?

Yes, they certainly do. I don't know a target language where this works as intended - usually a parse error is thrown (C++, C#, Go, Java, Nim, Python, PHP) or the last value shadows all previous ones (silently in JavaScript, Lua, Ruby; in Perl it shows a warning).

executable/swf.ksy
```diff
@@ -104,7 +152,184 @@ types:
...
+  grad_record:
+    params:
+      - id: shape_version
+        type: u1
+    seq:
+      - id: ratio
+        type: u1
+      - id: color_a
+        type: rgba
+        if:  shape_version == 4
```
generalmimon, line 317:
```suggestion
        if: shape_version == 4
```

executable/swf.ksy
```diff
@@ -104,7 +152,184 @@ types:
...
+        enum: cap_type
+      - id: meter_limit_factor
+        type: u2
+        if: join_style == join_style::miter_join
+      - id: color
+        type: rgba
+        if: has_fill_flag == false
+      - id: fill_type
+        type: fill_style(shape_version)
+        if: has_fill_flag == true
```
generalmimon, line 256: It's not wrong, but I'd simplify this:

```suggestion
      - id: color
        type: rgba
        if: not has_fill_flag
      - id: fill_type
        type: fill_style(shape_version)
        if: has_fill_flag
```

PR #423, opened 2021-02-20
vp8_ivf format improvement

frankpolte:
Now every frame is parsed. Works well with all tested ivf files. e.g.: https://github.com/webmproject/vp8-test-vectors/blob/master/vp80-00-comprehensive-015.ivf


media/vp8_ivf.ksy
```diff
@@ -49,23 +50,62 @@ seq:
...
+        type: b24
+      - id: width
+        type: b14
+      - id: horizontal_scale
+        type: b2
+      - id: height
+        type: b14
+      - id: vertical_scale
+        type: b2
+
```
generalmimon, line 111: There should not be any whitespace "after the newline at EOF" (actually not even at the end of _any_ line - that's why we have the [`trim_trailing_whitespace`](https://github.com/kaitai-io/kaitai_struct_formats/blob/9aa3c997fe6555c3c4e79ec34cd9c108960d1d26/.editorconfig#L8) option in [`.editorconfig`](https://editorconfig.org/) for .ksy files, but not many people have their editors configured to honor it):

```suggestion

```

media/vp8_ivf.ksy
```diff
@@ -7,6 +7,7 @@ meta:
   license: CC0-1.0
   ks-version: 0.7
```
generalmimon, line 8: Given that you're adding the key `bit-endian: le` two lines below, which was added in the version 0.9 (see https://doc.kaitai.io/user_guide.html#bit-endian), please bump the declared minimal Kaitai Struct version needed to compile the spec:

```suggestion
  ks-version: 0.9
```

generalmimon (changes requested):
What specification you used to look up the information? It's apparently not the one referenced in the `/doc-ref` key:

https://github.com/kaitai-io/kaitai_struct_formats/blob/4c370acf8306edab11a0f44be1b90340aa65b888/media/vp8_ivf.ksy#L17

I ask you to add a link to a spec which best describes the structure of format pieces that you're adding. `doc-ref` can be a list, so 

```diff
https://tools.ietf.org/html/rfc6386#section-9
```

media/vp8_ivf.ksy
```diff
@@ -49,23 +50,62 @@ seq:
...
+        type: b19
+      - id: keyframe_hdr
+        type: keyframe_tag
+        if: not is_interframe
+      - id: compressed_keyframe
+        size: len_frame - 10
+        if: not is_interframe
+      - id: compressed_interframe
+        size: len_frame - 3
+        if: is_interframe
```
generalmimon, line 98: What are these constants `10` and `3` that you're subtracting? What fields they are sizes of?

You know, calculating the sizes manually like that with subtraction is almost always a bad idea, and it can be usually done better with [substreams](https://doc.kaitai.io/user_guide.html#stream).

media/vp8_ivf.ksy
frankpolte: * I used the rfc6386 as well. It is the "official" specification I think. (regarding source)
* I see. I will configure my editor accordingly. (regarding trailing whitespace)

media/vp8_ivf.ksy
frankpolte: I see. Is there a way to check those issues?

media/vp8_ivf.ksy
frankpolte: @generalmimon Every frame has a 3-byte frame tag: https://tools.ietf.org/html/rfc6386#section-9.1 ,if it is a keyframe (defined by the first bit of that tag), then another 7 bytes are used for further information. On the same page: "...For key frames, the frame tag is followed by a further 7 bytes of uncompressed data, ..."
As they are fixed in the specification I see no case where this could be troublesome. May be a comment?

media/vp8_ivf.ksy
generalmimon: > @generalmimon : I see. Is there a way to check those issues?

Not really. Sometimes we try to write the version when some feature was introduced to the [User Guide](https://doc.kaitai.io/user_guide.html) (search for `v0.` on the page to see all such occurences), but we still don't even have all features documented (i.e. particularly the ones added in the [0.9 release](https://kaitai.io/news/2020/10/16/kaitai-struct-v0.9-released.html)), so some are a bit esoteric.

But the `ks-version` key is not something to care about too much, because with every new release, we treat all older versions as EOL and we discourage to use them.

PR #425, opened 2021-02-24
Removed the quotes where they are not needed.

KOLANICH:
Replaced http -> https somewhere.
Fixed indentation somewhere.


serialization/bson.ksy
```diff
@@ -140,23 +140,23 @@ types:
...
+            bson_type::document: bson
+            bson_type::array: bson
+            bson_type::bin_data: bin_data
+            bson_type::object_id: object_id
+            bson_type::boolean: u1
+            bson_type::utc_datetime: s8   #The int64 is UTC milliseconds since the Unix epoch.
+            bson_type::reg_ex: reg_ex
+            bson_type::db_pointer: db_pointer
+            bson_type::javascript: string
+            bson_type::symbol: string   # a programming language (e.g., Python) symbol
```
generalmimon, line 154: Why do you introduce triple spaces here?

Let's follow this rule for KSY comments - exactly one space before the number sign (`#`), one space after it and then the actual message.

```suggestion
            bson_type::utc_datetime: s8 # The int64 is UTC milliseconds since the Unix epoch.
            bson_type::reg_ex: reg_ex
            bson_type::db_pointer: db_pointer
            bson_type::javascript: string
            bson_type::symbol: string # a programming language (e.g., Python) symbol
```

database/sqlite3.ksy
```diff
@@ -33,7 +33,7 @@ doc: |
 doc-ref: https://www.sqlite.org/fileformat.html
 seq:
   - id: magic
-    contents: ["SQLite format 3", 0]
+    contents: [SQLite format 3, 0]
```
generalmimon, line 36: IMHO this change impairs intelligibility - seeing <code>[SQLite format <strong>3, 0</strong>]</code> can lead to false impression that the digit `3` must have the same logical meaning as `0`, because there is nothing that would differentiate them. Sure, if the viewer you use for looking at the code uses advanced YAML highlighting (like VS Code), you'll see a change in color:

image

But most viewers don't go in so much detail, and some crappy ones (like the Ace editor used in the [Web IDE](https://ide.kaitai.io/devel/)) even do it wrong:

image

So I'd suggest either reverting to the original state:

```suggestion
    contents: ["SQLite format 3", 0]
```

or at least writing `0` in hexadecimal so that it's clear that it's a byte value:

```suggestion
    contents: [SQLite format 3, 0x00]
```

hardware/mifare/mifare_classic.ksy
```diff
@@ -72,19 +72,19 @@ types:
                 repeat-expr: 4
             instances:
               value_valid:
-                value: 'valuez[0]==~valuez[1] and valuez[0]==valuez[2]'
+                value: valuez[0]==~valuez[1] and valuez[0]==valuez[2]
```
generalmimon, line 75:
```suggestion
                value: valuez[0] == ~valuez[1] and valuez[0] == valuez[2]
```

hardware/mifare/mifare_classic.ksy
```diff
@@ -72,19 +72,19 @@ types:
...
               addr_valid:
-                value: 'addrz[0]==~addrz[1] and addrz[0]==addrz[2] and addrz[1]==addrz[3]'
+                value: addrz[0]==~addrz[1] and addrz[0]==addrz[2] and addrz[1]==addrz[3]
```
generalmimon, line 77:
```suggestion
                value: addrz[0] == ~addrz[1] and addrz[0] == addrz[2] and addrz[1] == addrz[3]
```

hardware/mifare/mifare_classic.ksy
```diff
@@ -171,16 +171,16 @@ types:
               shift_value:
                 value: (bit_no==1?-1:1)
               chunk_no:
-                value: '((inv_chunk_no+shift_value+_parent._parent.ac_count_of_chunks)%_parent._parent.ac_count_of_chunks)'
+                value: ((inv_chunk_no+shift_value+_parent._parent.ac_count_of_chunks)%_parent._parent.ac_count_of_chunks)
```
generalmimon, line 174:
```suggestion
                value: ((inv_chunk_no + shift_value + _parent._parent.ac_count_of_chunks) % _parent._parent.ac_count_of_chunks)
```

hardware/mifare/mifare_classic.ksy
```diff
@@ -171,16 +171,16 @@ types:
...
               inv_chunk_no:
-                value: 'bit_no+shift_value'
+                value: bit_no+shift_value
```
generalmimon, line 176:
```suggestion
                value: bit_no + shift_value
```

hardware/mifare/mifare_classic.ksy
```diff
@@ -223,7 +223,7 @@ types:
                 value: ac.inv_shift_val <= 0b010
                 doc: key A is required
               can_write_keys:
-                value: "(ac.inv_shift_val+1)%3 != 0 and (ac.inv_shift_val<6)"
+                value: (ac.inv_shift_val+1)%3 != 0 and (ac.inv_shift_val<6)
```
generalmimon, line 226:
```suggestion
                value: (ac.inv_shift_val + 1) % 3 != 0 and (ac.inv_shift_val < 6)
```

image/bmp.ksy
```diff
@@ -419,10 +419,7 @@ types:
           cases:
             true: strz
         if: has_profile
-        doc-ref: https://docs.microsoft.com/en-us/previous-versions/windows/desktop/wcs/using-structures-in-wcs-1-0 "If the profile is embedded,
-          profile data is the actual profile, and if it is linked, the profile data is the
-          null-terminated file name of the profile. This cannot be a Unicode string. It must be composed exclusively
-          of characters from the Windows character set (code page 1252)."
+        doc-ref: https://docs.microsoft.com/en-us/previous-versions/windows/desktop/wcs/using-structures-in-wcs-1-0 "If the profile is embedded, profile data is the actual profile, and if it is linked, the profile data is the null-terminated file name of the profile. This cannot be a Unicode string. It must be composed exclusively of characters from the Windows character set (code page 1252)."
```
generalmimon, line 422:
```suggestion
        doc-ref: >-
          https://docs.microsoft.com/en-us/previous-versions/windows/desktop/wcs/using-structures-in-wcs-1-0
          "If the profile is embedded, profile data is the actual profile, and if it is linked,
          the profile data is the null-terminated file name of the profile.
          This cannot be a Unicode string. It must be composed exclusively of characters
          from the Windows character set (code page 1252)."
```

macos/ds_store.ksy
```diff
@@ -35,7 +35,7 @@ types:
   buddy_allocator_header:
     seq:
       - id: magic
-        contents: ["Bud1"]
+        contents: [Bud1]
```
generalmimon, line 38:
```suggestion
        contents: Bud1
```

media/android_opengl_shaders_cache.ksy
```diff
@@ -46,7 +46,7 @@ types:
     seq:
       - id: magic
         -orig-id: mMagicNumber, blobCacheMagic
-        contents: ["$bB_"]
+        contents: [$bB_]
```
generalmimon, line 49:
```suggestion
        contents: $bB_
```

scientific/nt_mdt/nt_mdt.ksy
```diff
@@ -53,7 +53,7 @@ types:
       - id: frames
         type: frame
         repeat: expr
-        repeat-expr: '_root.last_frame+1'
+        repeat-expr: _root.last_frame+1
```
generalmimon, line 56:
```suggestion
        repeat-expr: _root.last_frame + 1
```

hardware/mifare/mifare_classic.ksy
```diff
@@ -171,16 +171,16 @@ types:
               shift_value:
                 value: (bit_no==1?-1:1)
```
generalmimon, line 172:
```suggestion
                value: bit_no == 1 ?-1 : 1
```

image/icc_4.ksy
```diff
@@ -96,7 +96,7 @@ types:
         0x6172676C: argyll_cms_cmm #argl
         0x44676f53: logosync_cmm #LgoS
         0x48444d20: heidelberg_cmm #HDM
-        0x6C636d73: little_cms_cmm #lcms
+        0x6C636D73: little_cms_cmm #lcms
```
generalmimon, line 99: While you do this change, please also normalize the other enum values:

https://github.com/kaitai-io/kaitai_struct_formats/blob/e14dc4a2ed47609e2f33798670819d0d3591fa0c/image/icc_4.ksy#L95-L109

```diff
-        0x48434d4D: harlequin_rip_cmm #HCMM
+        0x48434D4D: harlequin_rip_cmm #HCMM
         0x6172676C: argyll_cms_cmm #argl
-        0x44676f53: logosync_cmm #LgoS
+        0x44676F53: logosync_cmm #LgoS
-        0x48444d20: heidelberg_cmm #HDM
+        0x48444D20: heidelberg_cmm #HDM
         0x6C636D73: little_cms_cmm #lcms
-        0x4b434d53: kodak_cmm #kcms
+        0x4B434D53: kodak_cmm #kcms
-        0x4d434d44: konica_minolta_cmm #MCML
+        0x4D434D44: konica_minolta_cmm #MCML
         0x57435320: windows_color_system_cmm #WCS
         0x5349474E: mutoh_cmm #SIGN
-        0x52474d53: device_link_cmm #RGMS
+        0x52474D53: device_link_cmm #RGMS
         0x53494343: sample_icc_cmm #SICC
-        0x54434d4d: toshiba_cmm #TCMM
+        0x54434D4D: toshiba_cmm #TCMM
         0x33324254: the_imaging_factory_cmm #32BT
         0x57544720: ware_to_go_cmm #WTG
-        0x7a633030: zoran_cmm #zc00
+        0x7A633030: zoran_cmm #zc00
```

https://github.com/kaitai-io/kaitai_struct_formats/blob/e14dc4a2ed47609e2f33798670819d0d3591fa0c/image/icc_4.ksy#L284

https://github.com/kaitai-io/kaitai_struct_formats/blob/e14dc4a2ed47609e2f33798670819d0d3591fa0c/image/icc_4.ksy#L1658

dgelessus (commented):
> Removed the quotes where they are not needed.
> Replaced http -> https somewhere.
> Fixed indentation somewhere.

It would be very helpful to have these as separate commits, or perhaps even as separate PRs, so they can be reviewed and discussed more easily. Right now I can't easily find where the HTTPS and indentation changes are without reading through the many quote-related changes.

Also, I'm personally not a fan of huge reformatting actions with no prior discussion or agreement. As @generalmimon already pointed out in [one comment](https://github.com/kaitai-io/kaitai_struct_formats/pull/425#discussion_r583034821), removing all quotes where possible doesn't always make the code more readable.

If we are going to do this, we should first decide when exactly things should/shouldn't be quoted, and write that in the style guide, so that it's easy to look up and follow the agreed formatting convention in the future.

image/bmp.ksy
KOLANICH: I feel like it may be better to put the text into `doc` and leave `doc-ref` for the link only

KOLANICH:
>It would be very helpful to have these as separate commits, or perhaps even as separate PRs

Maybe. I have fixed them while reviewing the changes the script has done with the files. Decided that if we still modify these lines, it may make sense to replace http -> https in them too.

>Fixed indentation somewhere.

was mainly done by YAML parsing and serializing library (`rumal.yaml`), I just preferred to keep that.

>Also, I'm personally not a fan of huge reformatting actions with no prior discussion or agreement.

TBH I just processed the whole code base witha script for #426, and found a lot of changes that were out of scope of that PR. Though lot of them looked beneficial... So I have done a pass doing only a round trip, then reviewed the changes manually and reverted some of them and generated a patch undoing the reverted ones. Then done a yet another pass, but with the code refactoring `repeat-*`s, and then applied the generated patch and solved conflicts, then rebased.

So the style is the one `ruamel.yaml` lib outputs, and IMHO it is not bad (the only issue I see is that I'd prefer double quotes, but the lib uses single ones, need to divve a bit deeper) since the lib tries to optimize the generated document.

So, if we gonna apply automatic changes to the code base, I feel like this PR is a prereq.

image/icc_4.ksy
KOLANICH: Thanks for mentioning that. I have missed this.

image/bmp.ksy
generalmimon: > I feel like it may be better to put the text into `doc` and leave `doc-ref` for the link only

Yes, you're right. I intended to emphasize the fact that it is a quote from that specific URL, but it really should be rather in `doc`.

Also, I've just found that the link address no longer works (Microsoft does a bad job of keeping URLs permanent) and it has moved to https://docs.microsoft.com/en-us/windows/win32/wcs/using-structures-in-wcs-1-0.

KOLANICH:
@dgelessus, @generalmimon, I have separated the changes into different commits.

PR #426, opened 2021-02-24
Modernized the repeats

KOLANICH:
`repeat-expr` -> `repeat: expr`
`repeat-until` -> `repeat: until`


dgelessus (changes requested):
This syntax was discussed in kaitai-io/kaitai_struct#776, but as far as I can tell it's not implemented in KS 0.9 or any development version:

```ksy
meta:
  id: test
seq:
  - id: foo
    type: s1
    repeat:
      expr: 5
```

```sh
$ ksc --target python test.ksy
/seq/0/repeat: expected string, got Map(expr -> 5) (class scala.collection.immutable.MapLike$$anon$2)
```

generalmimon:
@dgelessus:

> This syntax was discussed in [kaitai-io/kaitai_struct#776](https://github.com/kaitai-io/kaitai_struct/issues/776), but as far as I can tell it's not implemented in KS 0.9 or any development version

Yes, for that reason, let's mark the PR as draft.

PR #427, opened 2021-03-03
add .ksy for the GIMP brush file format

armijnhemel:
This is a .ksy for version 2 of the GIMP brush file format

### DIFF
--- a/image/gimp_brush.ksy
+++ b/image/gimp_brush.ksy
@@ -1,6 +1,7 @@
 meta:
   id: gimp_brush
-  title: GIMP (GNU Image Manipulation Program) brush version 2 file
+  title: GIMP brush file version 2
+  application: GIMP (GNU Image Manipulation Program)
   file-extension: gbr
   xref:
     mime: image/x-gimp-gbr
--- a/image/gimp_brush.ksy
+++ b/image/gimp_brush.ksy
@@ -6,11 +6,11 @@ meta:
   endian: be
 doc-ref: https://gitlab.gnome.org/GNOME/gimp/-/raw/441631322be109da6489b2aad670bdba916315c0/devel-docs/gbr.txt
 seq:
-  - id: header_size
+  - id: len_header
     type: u4
   - id: header
     type: header
-    size: header_size
+    size: len_header
 types:
   header:
     seq:
--- a/image/gimp_brush.ksy
+++ b/image/gimp_brush.ksy
@@ -9,6 +9,17 @@ meta:
     wikidata: Q28206177
   license: CC0-1.0
   endian: be
+doc: |
+  GIMP brush format is native to the GIMP image editor for storing a brush or a texture.
+  It can be used in all [Paint Tools](https://docs.gimp.org/2.10/en/gimp-tools-paint.html),
+  for example Pencil and Paintbrush. It works by repeating the brush bitmap as you move
+  the tool. The Spacing parameter sets the distance between the brush marks as a percentage
+  of brush width. Its default value can be set in the brush file.
+
+  You can also use GIMP to create new brushes in this format. Custom brushes can be loaded
+  into GIMP for use in the paint tools by copying them into one of the Brush Folders -
+  select **Edit** > **Preferences** in the menu bar, expand the **Folders** section
+  and choose **Brushes** to see the recognized Brush Folders or to add new ones.
 doc-ref: https://gitlab.gnome.org/GNOME/gimp/-/raw/4416313/devel-docs/gbr.txt
 seq:
   - id: len_header
--- a/image/gimp_brush.ksy
+++ b/image/gimp_brush.ksy
@@ -4,7 +4,7 @@ meta:
   file-extension: gbr
   license: CC0-1.0
   endian: be
-doc-ref: https://gitlab.gnome.org/GNOME/gimp/-/blob/master/devel-docs/gbr.txt
+doc-ref: https://gitlab.gnome.org/GNOME/gimp/-/raw/master/devel-docs/gbr.txt
 seq:
   - id: header_size
     type: u4be
--- a/image/gimp_brush.ksy
+++ b/image/gimp_brush.ksy
@@ -4,7 +4,7 @@ meta:
   file-extension: gbr
   license: CC0-1.0
   endian: be
-doc-ref: https://gitlab.gnome.org/GNOME/gimp/-/raw/441631322be109da6489b2aad670bdba916315c0/devel-docs/gbr.txt
+doc-ref: https://gitlab.gnome.org/GNOME/gimp/-/raw/4416313/devel-docs/gbr.txt
 seq:
   - id: len_header
     type: u4
--- a/image/gimp_brush.ksy
+++ b/image/gimp_brush.ksy
@@ -1,0 +1,32 @@
+meta:
+  id: gimp_brush
+  title: GIMP (GNU Image Manipulation Program) brush version 2 file
+  file-extension: gbr
+  license: CC0-1.0
+  endian: be
+doc-ref: https://gitlab.gnome.org/GNOME/gimp/-/blob/master/devel-docs/gbr.txt
+seq:
+  - id: header_size
+    type: u4be
+  - id: version
+    type: u4be
+  - id: width
+    type: u4be
+  - id: height
+    type: u4be
+  - id: color_depth
+    type: u4be
+  - id: magic
+    contents: GIMP
+  - id: spacing
+    type: u4be
+  - id: brush_name
+    type: strz
+    size: header_size-1 - 28
+    encoding: UTF-8
+instances:
+  body_size:
+    value: width * height * color_depth
+  body:
+    pos: header_size
+    size: body_size
--- a/image/gimp_brush.ksy
+++ b/image/gimp_brush.ksy
@@ -45,6 +45,7 @@ types:
         contents: GIMP
       - id: spacing
         type: u4
+        doc: Default spacing to be used for brush. Percentage of brush width.
       - id: brush_name
         type: strz
         size-eos: true
--- a/image/gimp_brush.ksy
+++ b/image/gimp_brush.ksy
@@ -8,25 +8,31 @@ doc-ref: https://gitlab.gnome.org/GNOME/gimp/-/raw/master/devel-docs/gbr.txt
 seq:
   - id: header_size
     type: u4be
-  - id: version
-    type: u4be
-  - id: width
-    type: u4be
-  - id: height
-    type: u4be
-  - id: color_depth
-    type: u4be
-  - id: magic
-    contents: GIMP
-  - id: spacing
-    type: u4be
-  - id: brush_name
-    type: strz
-    size: header_size-1 - 28
-    encoding: UTF-8
+  - id: header
+    type: header
+    size: header_size
+types:
+  header:
+    seq:
+      - id: version
+        type: u4be
+      - id: width
+        type: u4be
+      - id: height
+        type: u4be
+      - id: color_depth
+        type: u4be
+      - id: magic
+        contents: GIMP
+      - id: spacing
+        type: u4be
+      - id: brush_name
+        type: strz
+        size-eos: true
+        encoding: UTF-8
 instances:
   body_size:
-    value: width * height * color_depth
+    value: header.width * header.height * header.color_depth
   body:
     pos: header_size
     size: body_size
--- a/image/gimp_brush.ksy
+++ b/image/gimp_brush.ksy
@@ -2,6 +2,8 @@ meta:
   id: gimp_brush
   title: GIMP (GNU Image Manipulation Program) brush version 2 file
   file-extension: gbr
+  xref:
+    mime: image/x-gimp-gbr
   license: CC0-1.0
   endian: be
 doc-ref: https://gitlab.gnome.org/GNOME/gimp/-/raw/4416313/devel-docs/gbr.txt
@@ -16,10 +18,23 @@ types:
     seq:
       - id: version
         type: u4
+        valid: 2
       - id: width
         type: u4
+        valid:
+          min: 1
+          max: 10000
+        doc-ref:
+          - https://github.com/GNOME/gimp/blob/9b6d59f/app/core/gimpbrush-load.c#L170 # valid/min
+          - https://github.com/GNOME/gimp/blob/9b6d59f/app/core/gimpbrush-header.h#L24 # valid/max
       - id: height
         type: u4
+        valid:
+          min: 1
+          max: 10000
+        doc-ref:
+          - https://github.com/GNOME/gimp/blob/9b6d59f/app/core/gimpbrush-load.c#L177 # valid/min
+          - https://github.com/GNOME/gimp/blob/9b6d59f/app/core/gimpbrush-header.h#L24 # valid/max
       - id: bytes_per_pixel
         type: u4
         enum: color_depth
--- a/image/gimp_brush.ksy
+++ b/image/gimp_brush.ksy
@@ -10,7 +10,7 @@ seq:
     type: u4
   - id: header
     type: header
-    size: len_header
+    size: len_header - len_header._sizeof
 types:
   header:
     seq:
--- a/image/gimp_brush.ksy
+++ b/image/gimp_brush.ksy
@@ -4,7 +4,9 @@ meta:
   application: GIMP (GNU Image Manipulation Program)
   file-extension: gbr
   xref:
+    justsolve: GIMP_Brush
     mime: image/x-gimp-gbr
+    wikidata: Q28206177
   license: CC0-1.0
   endian: be
 doc-ref: https://gitlab.gnome.org/GNOME/gimp/-/raw/4416313/devel-docs/gbr.txt
--- a/image/gimp_brush.ksy
+++ b/image/gimp_brush.ksy
@@ -20,7 +20,7 @@ types:
         type: u4
       - id: height
         type: u4
-      - id: color_depth
+      - id: bytes_per_pixel
         type: u4
       - id: magic
         contents: GIMP
--- a/image/gimp_brush.ksy
+++ b/image/gimp_brush.ksy
@@ -62,11 +62,11 @@ types:
         size-eos: true
         encoding: UTF-8
 instances:
-  body_size:
+  len_body:
     value: header.width * header.height * header.bytes_per_pixel.to_i
   body:
     pos: len_header
-    size: body_size
+    size: len_body
 enums:
   color_depth:
     1: grayscale
--- a/image/gimp_brush.ksy
+++ b/image/gimp_brush.ksy
@@ -34,5 +34,5 @@ instances:
   body_size:
     value: header.width * header.height * header.bytes_per_pixel
   body:
-    pos: header_size
+    pos: len_header
     size: body_size
--- a/image/gimp_brush.ksy
+++ b/image/gimp_brush.ksy
@@ -61,12 +61,63 @@ types:
         type: strz
         size-eos: true
         encoding: UTF-8
+  bitmap:
+    seq:
+      - id: rows
+        type: row
+        repeat: expr
+        repeat-expr: _root.header.height
+  row:
+    seq:
+      - id: pixels
+        type:
+          switch-on: _root.header.bytes_per_pixel
+          cases:
+            color_depth::grayscale: pixel_gray
+            color_depth::rgba: pixel_rgba
+        repeat: expr
+        repeat-expr: _root.header.width
+    types:
+      pixel_gray:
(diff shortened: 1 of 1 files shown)


image/gimp_brush.ksy
```diff
@@ -0,0 +1,32 @@
+meta:
+  id: gimp_brush
+  title: GIMP (GNU Image Manipulation Program) brush version 2 file
+  file-extension: gbr
+  license: CC0-1.0
```
KOLANICH, line 5: This may (or may not, if you have clean room or black box reverse engineered it) violate the terms of GPL, under which that repo is licensed.

image/gimp_brush.ksy
```diff
@@ -0,0 +1,32 @@
...
+  endian: be
+doc-ref: https://gitlab.gnome.org/GNOME/gimp/-/blob/master/devel-docs/gbr.txt
```
KOLANICH, line 7: https://gitlab.gnome.org/GNOME/gimp/-/raw/master/devel-docs/gbr.txt

GitLab forcibly promotes JS in favour of GitLab inc. and their partners' commercial interests. Fortunately they have not yet restricted direct access to raw files and fortunately MarkDown or designed to be readable and writeable as plain text. If you want HTML markup, we can use GIMP GitHub mirror. https://github.com/GNOME/gimp/blob/master/devel-docs/gbr.txt Though GH is proprietary and belongs to M$,  it hehaves much more nicer.

image/gimp_brush.ksy
```diff
@@ -0,0 +1,32 @@
...
+    type: u4be
+  - id: color_depth
+    type: u4be
+  - id: magic
+    contents: GIMP
+  - id: spacing
+    type: u4be
+  - id: brush_name
+    type: strz
+    size: header_size-1 - 28
```
KOLANICH, line 25: What is `28`? Is it a size of some type? For sizes of types we have `sizeof<>`

image/gimp_brush.ksy
```diff
@@ -0,0 +1,32 @@
...
+    encoding: UTF-8
+instances:
+  body_size:
+    value: width * height * color_depth
```
KOLANICH, line 29: Please consider structuring it, since now it is not clear, what is the order of the pixels and components within pixels.

image/gimp_brush.ksy
armijnhemel: This was derived from a documentation file released by the GIMP developers themselves. I am pretty confident that this would not constitute a license violation. I will verify with the GIMP team.

image/gimp_brush.ksy
KOLANICH, line 25: You are doing this wrong.

```yaml
seq:
  - id: header_size
    type: u4be
  - id: header
    type: header
    size: header_size

types:
  header
    seq:
      - id: version
        type: u4be
      - id: width
        type: u4be
     - id: height
       type: u4be
     - id: color_depth
       type: u4be
    - id: magic
      contents: GIMP
    - id: spacing
      type: u4be
    - id: brush_name
      type: strz
      size-eos: true
```

armijnhemel (commented):
Thank you for the comments. I will review them carefully.

image/gimp_brush.ksy
armijnhemel: I am not sure I entirely understand this one (this is my first .ksy, so please bear with me). It is basically a row based format. Depending on the color depth there might be a single byte (grayscale) or four bytes (RGBA). How would you structure this?

KOLANICH:
```yaml
seq:
...
  - id: body
    type: row
    repeat: expr
    repeat-expr: header.height

types:
  row:
    seq:
      - id: pixels
        type:
          switch-on: _root.header.color_depth
          cases:
            color_depth::grayscale: pixel_gray
            color_depth::rgba: pixel_rgba
        repeat: expr
        repeat-expr: _root.header.width
    types:
      pixel_gray:
        seq:
          - id: gray
            type: 1
        instances:
          red:
            value: gray
          green:
            value: gray
          blue:
            value: gray
          alpha:
            value: 1

      pixel_rgba:
        seq:
          - id: red
            type: u1
          - id: green
            type: u1
          - id: blue
            type: u1
          - id: alpha
            type: u1
```

generalmimon (changes requested):

armijnhemel:
oops, sorry, I shouldn't have pushed that last particular commit. I will undo tomorrow and resubmit.

generalmimon:
@armijnhemel:

> oops, sorry, I shouldn't have pushed that last particular commit. I will undo tomorrow and resubmit.

Yeah, I totally understand it. That's why you should never open PRs from `master` of your fork - otherwise the pull request blocks your `master` branch and you can't use it normally (for pulling upstream `master` changes, for example) until the PR is merged.

For your own convenience, remember a rule of thumb when making pull requests from your fork repository: **never commit directly on `master` of your fork**. Create a topic branch for every PR you create. See https://contribute.jquery.org/commits-and-pull-requests/#never-commit-on-master:

image/gimp_brush.ksy
```diff
@@ -0,0 +1,32 @@
...
+  - id: width
+    type: u4be
+  - id: height
+    type: u4be
+  - id: color_depth
+    type: u4be
+  - id: magic
+    contents: GIMP
+  - id: spacing
+    type: u4be
```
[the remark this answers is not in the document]
KOLANICH: If there was any possibility that the spec was reverse-engineered in a copyrigh-clean way, quoting the original doc will surely make more reasons to call it a derivative work and require infection with GPL of it and any software that would depend on it.

image/gimp_brush.ksy
```diff
@@ -0,0 +1,38 @@
...
+  header:
+    seq:
+      - id: version
+        type: u4be
+      - id: width
+        type: u4be
+      - id: height
+        type: u4be
+      - id: color_depth
+        type: u4be
```
generalmimon, line 24: I'd rename this to `bytes_per_pixel`, because it's more explicit and less abstract and vague than `color_depth`. All the more so that the term "color depth" usually refers to the number of **bits per pixel** (not ~bytes~), so abusing the same term here might be actually deceptive and ambiguous. See https://en.wikipedia.org/wiki/Color_depth:

> **Color depth** or **colour depth** (see [spelling differences](https://en.wikipedia.org/wiki/American_and_British_English_spelling_differences#-our,_-or)), also known as [**bit depth**](https://en.wikipedia.org/wiki/Bit_depth_(computer_graphics)), is either the number of bits used to indicate the color of a single pixel

armijnhemel:
> @armijnhemel:
> 
> > oops, sorry, I shouldn't have pushed that last particular commit. I will undo tomorrow and resubmit.
> 
> Yeah, I totally understand it. That's why you should never open PRs from `master` of your fork - in that case, the pull request blocks your `master` branch and you can't use it normally (for pulling upstream `master` changes, for example) until the PR is merged.
> 
> For your own convenience, remember a rule of thumb when making pull requests from your fork repository: **never commit directly on `master` of your fork**. Create a topic branch for every PR you create. See https://contribute.jquery.org/commits-and-pull-requests/#never-commit-on-master:
> 
> > When you're working on a fork, you should always think of your `master` branch as a "landing place" for upstream changes. You should only ever make your commits to topic branches, and your own commits should only ever end up on `master` after they've been merged in upstream by a maintainer.

Talking to lawyers all day melted my brain and I thought I hadn't pushed from master. I only realised later. Won't happen again! :-)

image/gimp_brush.ksy
```diff
@@ -0,0 +1,38 @@
+meta:
+  id: gimp_brush
+  title: GIMP (GNU Image Manipulation Program) brush version 2 file
+  file-extension: gbr
+  license: CC0-1.0
+  endian: be
+doc-ref: https://gitlab.gnome.org/GNOME/gimp/-/raw/master/devel-docs/gbr.txt
```
generalmimon, line 7: Please don't link to the `master` branch that changes all the time - all URLs that you provide should be permanent (if possible). The target document should not change under our hands, which is something you have to expect when you refer to the `master` branch. There's also no guarantee that the document will remain available on the same path - it may be renamed or moved (to another path in the same repository or another repository).

I think that the reference docs should reflect the version that was used at the time of writing the spec. If a new version of the format is released, which has a different format structure than the former version, one would just open the URL from `doc-ref` and they would be presented with essentially a totally different format that doesn't reflect the one implemented by the KSY spec.

If you go to https://gitlab.gnome.org/GNOME/gimp/-/blob/master/devel-docs/gbr.txt, you can press <kbd>Y</kbd> on the keyboard, which will change the URL in the address bar to the permanent URL, which you can copy. Alternatively, there is also the following button in the top right corner (which does the same thing as <kbd>Y</kbd>):

<table>
<tbody>
<tr>
<td>Permalink</td>
</tr>
</tbody>
</table>

Then if you want to link to the raw file as @KOLANICH suggested in https://github.com/kaitai-io/kaitai_struct_formats/pull/427#discussion_r586725441, just change `/blob` to `/raw` in the URL:

```diff
- https://gitlab.gnome.org/GNOME/gimp/-/blob/441631322be109da6489b2aad670bdba916315c0/devel-docs/gbr.txt
+ https://gitlab.gnome.org/GNOME/gimp/-/raw/441631322be109da6489b2aad670bdba916315c0/devel-docs/gbr.txt
```

image/gimp_brush.ksy
```diff
@@ -0,0 +1,38 @@
...
+seq:
+  - id: header_size
```
generalmimon, line 9:
```
../formats/gimp_brush.ksy: /seq/0/id: warning: use `len_header` instead of `header_size`, given that it's only used as a byte size of `header` (see https://doc.kaitai.io/ksy_style_guide.html#attr-id)
```

image/gimp_brush.ksy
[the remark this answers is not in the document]
armijnhemel: This has been changed to 'eos' instead.

image/gimp_brush.ksy
armijnhemel: ok, fixed

image/gimp_brush.ksy
generalmimon: @armijnhemel This has not yet been resolved.

armijnhemel: It is now :-) Sorry for being slow, lots of other tasks to juggle.

image/gimp_brush.ksy
```diff
@@ -0,0 +1,38 @@
...
+        contents: GIMP
+      - id: spacing
+        type: u4
+      - id: brush_name
+        type: strz
+        size-eos: true
+        encoding: UTF-8
+instances:
+  body_size:
+    value: header.width * header.height * header.color_depth
```
generalmimon, line 35: You need to rename `color_depth` here as well (https://github.com/kaitai-io/kaitai_struct_formats/pull/427/commits/56f7c1c4d877f49547f2b7dcb0b3305d0860196e):

```suggestion
    value: header.width * header.height * header.bytes_per_pixel
```

image/gimp_brush.ksy
```diff
@@ -0,0 +1,38 @@
...
+  header:
+    seq:
+      - id: version
+        type: u4
+      - id: width
+        type: u4
+      - id: height
+        type: u4
+      - id: bytes_per_pixel
+        type: u4
```
generalmimon, line 24: As @KOLANICH implicitly suggests in https://github.com/kaitai-io/kaitai_struct_formats/pull/427#discussion_r586804286, it really makes sense to bind this field with an enum, because only values `1` and `4` can be used (see [`devel-docs/gbr.txt:41`](https://github.com/GNOME/gimp/blob/0ef5d672b7bd81b50e38f3cfb4bdce2df80f5458/devel-docs/gbr.txt#L41)):

```ksy
          cases:
            color_depth::grayscale: pixel_gray
            color_depth::rgba: pixel_rgba
```

So you should create the following enum:

```ksy
enums:
  color_depth:
    1: grayscale
    4: rgba
```

and attach it to `bytes_per_pixel`:

```diff
       - id: bytes_per_pixel
         type: u4
+        enum: color_depth
```

If you need `bytes_per_pixel` as an integer again (e.g. for the calculation of `body_size`), you can use the enum method [`to_i`](https://doc.kaitai.io/user_guide.html#_enums):

```diff
 instances:
   body_size:
-    value: header.width * header.height * header.bytes_per_pixel
+    value: header.width * header.height * header.bytes_per_pixel.to_i
```

generalmimon (changes requested):
Also see the unresolved comments above: https://github.com/kaitai-io/kaitai_struct_formats/pull/427#discussion_r586804286, https://github.com/kaitai-io/kaitai_struct_formats/pull/427#discussion_r586723543, https://github.com/kaitai-io/kaitai_struct_formats/pull/427#discussion_r586727531

image/gimp_brush.ksy
generalmimon: Thank you @KOLANICH for the pixel structures! They work fine except the `pixel_gray` type, which should be like this instead:

```ksy
      pixel_gray:
        seq:
          - id: gray
            type: u1
        instances:
          red:
            value: 0
          green:
            value: 0
          blue:
            value: 0
          alpha:
            value: gray
```

I found this experimentally: after you edit the pixel value (actually the last byte) of the brush [`Basic/pixel.gbr`](https://gitlab.gnome.org/GNOME/gimp/-/blob/0ef5d67/data/brushes/Basic/pixel.gbr) in a hex editor (e.g. to `0x55` or `85` in decimal), load it into GIMP and try to draw something with it (but I recommend setting _Size_ to some large value like 50, because with the default size 1 the edges of the brush stroke will be smoothed and you might not be able to pick up the right color).

Then if you press the <kbd>O</kbd> (as in **O**scar, not zero: <kbd>0</kbd>) key to select the Color Picker and pick the color while holding <kbd>Shift</kbd>, the info window that pops up should show this:

<table>
<thead>
<tr>
<th colspan=3 align=center>Color Picker</th>
</tr>
<tr>
<th colspan=2>Pixel</th>
<th>RGB (%)</th>
</tr>
</thead>
<tbody>
<tr>
<td align=left>R:</td>
<td align=right>0</td>
<td rowspan=4 align=center>...</td>
</tr>
<tr>
<td align=left>G:</td>
<td align=right>0</td>
</tr>
<tr>
<td align=left>B:</td>
<td align=right>0</td>
</tr>
<tr>
<td align=left>A:</td>
<td align=right>85</td>
</tr>
</tbody>
</table>

---

However, although it kind of makes sense to _describe_ all structures in the KS language that can be described in it (at least for documentation/educational purposes), the question is if they are usable in the real world scenarios (whether the performance penalty and memory usage will be acceptable even for larger files).

I haven't measured anything, but I suspect that creating a class instance for each pixel will slightly (I estimate 3-4 times) increase the memory consumption and the overhead with creating them will probably slow down the parsing a bit (in comparison with creating a fixed-size byte array in advance and just gradually filling it with RGBA values as you parse the bitmap, just as you would've done when parsing the file with a hand-written code). Naturally, it would be best to have https://github.com/kaitai-io/kaitai_struct/issues/188, which would be ultimately exactly just filling the fixed-size buffer, but I wonder if the parsing of bitmap pixels is feasible using purely (or _mostly_) KS in the current state.

KOLANICH: >However, although it kind of makes sense to describe all structures in the KS language that can be described in it (at least for documentation/educational purposes), the question is if they are usable in the real world scenarios (whether the performance penalty and memory usage will be acceptable even for larger files).

Absolutely valid concern. The solution may be to mark some types and fields as "hardware-implemented", so the compiler won't usually generate code for them, but to keep them as a blob.

And an another mitigation is to redesign runtimes to have structs with less overhead for simple cases (i.e. when `_io` is not needed) and make the compiler to use them and generate the code allocating them properly.

About kaitai-io/kaitai_struct#188 : I don't really beleive in it for images, because there are quite some standard formats for pixels (with different encodings, not necessarily RGB, HSV or YCrCb), and there may be a lot of non-standard, especially in old games. It is impossible and makes no sense to support them all in KS. (Though I guess it may be possible to create a machine-readable spec DSL for pixel formats and let people encode them themeself, but I guess it is out of scope of KS project)

image/gimp_brush.ksy
armijnhemel: this has been fixed (properly I hope)

image/gimp_brush.ksy
```diff
@@ -0,0 +1,32 @@
...
+  title: GIMP (GNU Image Manipulation Program) brush version 2 file
+  file-extension: gbr
+  license: CC0-1.0
+  endian: be
+doc-ref: https://gitlab.gnome.org/GNOME/gimp/-/blob/master/devel-docs/gbr.txt
+seq:
+  - id: header_size
+    type: u4be
+  - id: version
+    type: u4be
```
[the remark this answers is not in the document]
generalmimon: > what should they set this value to in order to be recognized in GIMP or other applications? How does GIMP treat this value - is it currently just ignored, or does it switch some features on/off or changes the expected byte layout?

As I anticipated, the source code of GIMP basically answers all these questions on these lines: [GNOME / **gimp** `/app/core/gimpbrush-load.c:201-237`](https://github.com/GNOME/gimp/blob/9b6d59f38aaef8c560e9fd2a19011da8e9dc6264/app/core/gimpbrush-load.c#L201-L237)

Apparently, only three values are possible - 1, 2 and 3. However, with the current implementation limited to the version 2 (and you also state it in `/meta/title`:

https://github.com/kaitai-io/kaitai_struct_formats/blob/db1486016fe04c546b589f1b07708099a69eb172/image/gimp_brush.ksy#L3

), it would make sense to make this even more explicit with `valid: 2`. But if you'd like to extend the specification a bit to support also versions 1 and 3 like GIMP does, it would be appreciated :)

image/gimp_brush.ksy
```diff
@@ -0,0 +1,43 @@
+meta:
+  id: gimp_brush
+  title: GIMP (GNU Image Manipulation Program) brush version 2 file
+  file-extension: gbr
```
generalmimon, line 4:
```suggestion
  file-extension: gbr
  xref:
    mime: image/x-gimp-gbr
```

according to GIMP source code (https://github.com/GNOME/gimp/blob/9b6d59f/app/core/gimpbrush-load.c#L283):

```c
"mime-type", "image/x-gimp-gbr",
```

image/gimp_brush.ksy
```diff
@@ -0,0 +1,43 @@
...
+    size: len_header - len_header._sizeof
+types:
+  header:
+    seq:
+      - id: version
+        type: u4
+      - id: width
+        type: u4
+      - id: height
+        type: u4
```
generalmimon, line 22: There are assertions for these values in the GIMP code ([`gimpbrush-load.c:191-192`](https://github.com/GNOME/gimp/blob/9b6d59f/app/core/gimpbrush-load.c#L191-L192)):

```c
  if (header.width  > GIMP_BRUSH_MAX_SIZE ||
      header.height > GIMP_BRUSH_MAX_SIZE ||
      /* ... */)
    {
      g_set_error (error, GIMP_DATA_ERROR, GIMP_DATA_ERROR_READ,
                   _("Fatal parse error in brush file: %dx%d over max size."),
                   header.width, header.height);
      return NULL;
    }
```

[`gimpbrush-header.h:24`](https://github.com/GNOME/gimp/blob/9b6d59f/app/core/gimpbrush-header.h#L24):

```c
#define GIMP_BRUSH_MAX_SIZE 10000 /* Max size in either dimension in px */
```

[`gimpbrush-load.c:170-182`](https://github.com/GNOME/gimp/blob/9b6d59f/app/core/gimpbrush-load.c#L170-L182):

```c
  if (header.width == 0)
    {
      g_set_error (error, GIMP_DATA_ERROR, GIMP_DATA_ERROR_READ,
                   _("Fatal parse error in brush file: Width = 0."));
      return NULL;
    }


  if (header.height == 0)
    {
      g_set_error (error, GIMP_DATA_ERROR, GIMP_DATA_ERROR_READ,
                   _("Fatal parse error in brush file: Height = 0."));
      return NULL;
    }
```

If we want to ensure that the GIMP brush is compatible with GIMP, I'd add the assertions too:

```suggestion
      - id: width
        type: u4
        valid:
          min: 1
          max: 10000
        doc-ref:
          - https://github.com/GNOME/gimp/blob/9b6d59f/app/core/gimpbrush-load.c#L170 # valid/min
          - https://github.com/GNOME/gimp/blob/9b6d59f/app/core/gimpbrush-header.h#L24 # valid/max
      - id: height
        type: u4
        valid:
          min: 1
          max: 10000
        doc-ref:
          - https://github.com/GNOME/gimp/blob/9b6d59f/app/core/gimpbrush-load.c#L177 # valid/min
          - https://github.com/GNOME/gimp/blob/9b6d59f/app/core/gimpbrush-header.h#L24 # valid/max
```

generalmimon (approved):
Thanks for contributing!

Issue #429, opened 2021-03-05
suggestion: rename asus_trx.ksy to broadcom_trx.ksy


armijnhemel:
The TRX format is not specific to ASUS devices, but instead it is the standard format of firmware files on Broadcom based devices:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/devicetree/bindings/mtd/partitions/brcm,trx.txt

A more complete specification of the two TRX formats can be found here:

http://web.archive.org/web/20190127154419/https://openwrt.org/docs/techref/header

KOLANICH:
Thanks for the info!

PR #431, opened 2021-03-05
Renamed asus_trx to broadcom_trx

KOLANICH:
Fixes: #429

### DIFF
--- a/firmware/broadcom_trx.ksy
+++ b/firmware/broadcom_trx.ksy
@@ -1,6 +1,6 @@
 meta:
-  id: asus_trx
-  title: ASUS routers .trx official firmware packaging
+  id: broadcom_trx
+  title: Broadcom devices .trx firmware packaging
   file-extension: trx
   license: GPL-2.0
   ks-version: 0.9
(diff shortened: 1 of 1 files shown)


firmware/broadcom_trx.ksy
```diff
@@ -1,6 +1,6 @@
 meta:
-  id: asus_trx
-  title: ASUS routers .trx official firmware packaging
+  id: broadcom_trx
+  title: Broadcom devices .trx firmware packaging
   file-extension: trx
   license: GPL-2.0
```
generalmimon, line 5: When you are already editing the file, please also fix this deprecated SPDX identifier [`GPL-2.0`](https://spdx.org/licenses/GPL-2.0.html): is it supposed to be [`GPL-2.0-only`](https://spdx.org/licenses/GPL-2.0-only.html) or [`GPL-2.0-or-later`](https://spdx.org/licenses/GPL-2.0-or-later.html)?

generalmimon (approved):
Thanks!

PR #432, opened 2021-03-05
PNG: include definitions for animated PNG (APNG)

armijnhemel:
Animated PNG (APNG) is a not very widely used PNG format, but sometimes encountered in the wild. This PR adds support for the three chunks that are defined in the APNG specification:

https://wiki.mozilla.org/APNG_Specification

I have tested this with the Kaitai Web IDE and test files from:

http://littlesvr.ca/apng/

### DIFF
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -372,9 +372,6 @@ types:
         doc: |
           Sequence number of the animation chunk. The fcTL and fdAT chunks
           have a 4 byte sequence number. Both chunk types share the sequence.
-          The purpose of this number is to detect (and optionally correct)
-          sequence errors in an Animated PNG, since the PNG specification
-          does not impose ordering restrictions on ancillary chunks.
           The first fcTL chunk must contain sequence number 0, and the sequence
           numbers in the remaining fcTL and fdAT chunks must be in order, with
           no gaps or duplicates.
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -360,9 +360,9 @@ enums:
   compression_methods:
     0: zlib
   dispose_op_values:
-   0: none
-   1: background
-   2: previous
+    0: none
+    1: background
+    2: previous
   blend_op_values:
-   0: source
-   1: over
+    0: source
+    1: over
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -310,7 +310,7 @@ types:
     seq:
       - id: num_frames
         type: u4
-        doc: Number of frames
+        doc: Number of frames, must be equal to the number of `frame_control_chunk`s
       - id: num_plays
         type: u4
         doc: Number of times to loop, 0 indicates infinite looping.
@@ -323,11 +323,13 @@ types:
       - id: width
         type: u4
         valid:
+          min: 1
           max: _root.ihdr.width
         doc: Width of the following frame
       - id: height
         type: u4
         valid:
+          min: 1
           max: _root.ihdr.height
         doc: Height of the following frame
       - id: x_offset
@@ -357,7 +359,7 @@ types:
     instances:
       delay:
         value: 'delay_num / (delay_den == 0 ? 100.0 : delay_den)'
-        doc: delay in seconds
+        doc: Time to display this frame, in seconds
   frame_data_chunk:
     doc-ref: https://wiki.mozilla.org/APNG_Specification#.60fdAT.60:_The_Frame_Data_Chunk
     seq:
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -71,6 +71,11 @@ types:
             '"iTXt"': international_text_chunk
             '"tEXt"': text_chunk
             '"zTXt"': compressed_text_chunk
+
+            # animated PNG chunks
+            '"acTL"': animation_control_chunk
+            '"fcTL"': frame_control_chunk
+            '"fdAT"': frame_data_chunk
       - id: crc
         size: 4
   ihdr_chunk:
@@ -292,6 +297,56 @@ types:
       - id: text_datastream
         process: zlib
         size-eos: true
+  animation_control_chunk:
+    doc-ref: https://wiki.mozilla.org/APNG_Specification#.60acTL.60:_The_Animation_Control_Chunk
+    seq:
+      - id: num_frames
+        type: u4
+        doc: Number of frames
+      - id: num_plays
+        type: u4
+        doc: Number of times to loop, 0 indicates infinite looping.
+  frame_control_chunk:
+    doc-ref: https://wiki.mozilla.org/APNG_Specification#.60fcTL.60:_The_Frame_Control_Chunk
+    seq:
+      - id: sequence_number
+        type: u4
+        doc: Sequence number of the animation chunk
+      - id: width
+        type: u4
+        doc: Width of the following frame
+      - id: height
+        type: u4
+        doc: Height of the following frame
+      - id: x_offset
+        type: u4
+        doc: X position at which to render the following frame
+      - id: y_offset
+        type: u4
+        doc: Y position at which to render the following frame
+      - id: delay_num
+        type: u2
+        doc: Frame delay fraction numerator
+      - id: delay_den
+        type: u2
+        doc: Frame delay fraction denominator
+      - id: dispose_op
+        type: u1
+        enum: dispose_op_values
+        doc: Type of frame area disposal to be done after rendering this frame
+      - id: blend_op
+        type: u1
+        enum: blend_op_values
+        doc: Type of frame area rendering for this frame
+  frame_data_chunk:
+    doc-ref: https://wiki.mozilla.org/APNG_Specification#.60fdAT.60:_The_Frame_Data_Chunk
+    seq:
+      - id: sequence_number
+        type: u4
+        doc: Sequence number of the animation chunk
+      - id: frame_data
+        size-eos: true
+        doc: Frame data for the frame
 enums:
   color_type:
     0: greyscale
@@ -304,3 +359,10 @@ enums:
     1: meter
   compression_methods:
     0: zlib
+  dispose_op_values:
+   0: none
+   1: background
+   2: previous
+  blend_op_values:
+   0: source
+   1: over
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -322,9 +322,13 @@ types:
         doc: Height of the following frame
       - id: x_offset
         type: u4
+        valid:
+          max: _root.ihdr.width
         doc: X position at which to render the following frame
       - id: y_offset
         type: u4
+        valid:
+          max: _root.ihdr.height
         doc: Y position at which to render the following frame
       - id: delay_num
         type: u2
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -348,6 +348,10 @@ types:
         type: u1
         enum: blend_op_values
         doc: Type of frame area rendering for this frame
+    instances:
+      delay:
+        value: delay_num / (delay_den == 0 ? 100.0 : delay_den)
+        doc: delay in seconds
   frame_data_chunk:
     doc-ref: https://wiki.mozilla.org/APNG_Specification#.60fdAT.60:_The_Frame_Data_Chunk
     seq:
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -316,19 +316,23 @@ types:
         doc: Sequence number of the animation chunk
       - id: width
         type: u4
+        valid:
+          max: _root.ihdr.width
         doc: Width of the following frame
       - id: height
         type: u4
+        valid:
+          max: _root.ihdr.height
         doc: Height of the following frame
       - id: x_offset
         type: u4
         valid:
-          max: _root.ihdr.width
+          max: _root.ihdr.width - width
         doc: X position at which to render the following frame
       - id: y_offset
         type: u4
         valid:
-          max: _root.ihdr.height
+          max: _root.ihdr.height - height
         doc: Y position at which to render the following frame
       - id: delay_num
         type: u2
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -350,7 +350,7 @@ types:
         doc: Type of frame area rendering for this frame
     instances:
       delay:
-        value: delay_num / (delay_den == 0 ? 100.0 : delay_den)
+        value: 'delay_num / (delay_den == 0 ? 100.0 : delay_den)'
         doc: delay in seconds
   frame_data_chunk:
     doc-ref: https://wiki.mozilla.org/APNG_Specification#.60fdAT.60:_The_Frame_Data_Chunk
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -7,7 +7,9 @@ meta:
     iso: 15948:2004
     justsolve: PNG
     loc: fdd000153
-    mime: image/png
+    mime:
+      - image/png
+      - image/apng
     pronom:
       - fmt/11 # PNG 1.0
       - fmt/12 # PNG 1.1
@@ -15,7 +17,7 @@ meta:
     rfc: 2083
     wikidata: Q178051
   license: CC0-1.0
-  ks-version: 0.8
+  ks-version: 0.9
   endian: be
 seq:
   # https://www.w3.org/TR/PNG/#5PNG-file-signature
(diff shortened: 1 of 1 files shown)


armijnhemel:
As a side note: there are quite a few files out there with different PNG chunks, including some very exotic ones (ZDoom used PNG chunks for save games). Should I add these as well?

generalmimon:
> As a side note: there are quite a few files out there with different PNG chunks, including some very exotic ones (ZDoom used PNG chunks for save games). Should I add these as well?

It's up to you, of course, but I think that anything that makes the format specification more complete and thorough is beneficial.

armijnhemel:
> > As a side note: there are quite a few files out there with different PNG chunks, including some very exotic ones (ZDoom used PNG chunks for save games). Should I add these as well?
> 
> It's up to you, of course, but I think that anything that makes the format specification more complete and thorough is beneficial.

Definitely. I will dig up specifications of various formats I have identified so far and work on that this weekend.

image/png.ksy
```diff
@@ -304,3 +359,10 @@ enums:
     1: meter
   compression_methods:
     0: zlib
+  dispose_op_values:
+   0: none
+   1: background
+   2: previous
+  blend_op_values:
+   0: source
+   1: over
```
GreyCat, line 368: Nit: please use 2-space indent to these values.

image/png.ksy
```diff
@@ -292,6 +297,56 @@ types:
...
+        doc: X position at which to render the following frame
+      - id: y_offset
+        type: u4
+        doc: Y position at which to render the following frame
+      - id: delay_num
+        type: u2
+        doc: Frame delay fraction numerator
+      - id: delay_den
+        type: u2
+        doc: Frame delay fraction denominator
```
GreyCat, line 332: It would be great to clarify which time units does frame delay use in this case.

GreyCat (approved):

image/png.ksy
armijnhemel: oh, oops. I will fix that.

image/png.ksy
armijnhemel: I will try to clarify this. The specification says:

"he `delay_num` and `delay_den` parameters together specify a fraction indicating the time to display the current frame, in seconds."

generalmimon, line 332: These two values actually form one value, so it would be reasonable to create a [value instance](https://doc.kaitai.io/user_guide.html#_value_instances) that will calculate the final value of `delay` from the values of `delay_num` and `delay_den`.

If this pattern of a fractional value would repeat, I'd recommend to create a self-contained subtype for it. You can do it here too, but it's not necessary - if you prefer to stick to the specification as closely, it's OK and just add the value instance for convenient access.

image/png.ksy
```diff
@@ -304,3 +359,10 @@ enums:
     1: meter
   compression_methods:
     0: zlib
+  dispose_op_values:
+    0: none
+    1: background
+    2: previous
+  blend_op_values:
+    0: source
+    1: over
```
generalmimon, line 368: You could also add descriptions to these values with ["verbose enum"](https://doc.kaitai.io/user_guide.html#verbose-enums) declaration.

image/png.ksy
armijnhemel: I have not seen it in any other PNG chunk that I have encountered so far. I will look into the value instance

image/png.ksy
```diff
@@ -292,6 +297,56 @@ types:
...
+    doc-ref: https://wiki.mozilla.org/APNG_Specification#.60acTL.60:_The_Animation_Control_Chunk
+    seq:
+      - id: num_frames
+        type: u4
+        doc: Number of frames
+      - id: num_plays
+        type: u4
+        doc: Number of times to loop, 0 indicates infinite looping.
+  frame_control_chunk:
+    doc-ref: https://wiki.mozilla.org/APNG_Specification#.60fcTL.60:_The_Frame_Control_Chunk
```
generalmimon, line 310: The spec lists these constraints that the values have to match:

> Constraints on frame regions:
> 
> ```
>    `x_offset` >= 0
>    `y_offset` >= 0
>    `width`    > 0
>    `height`   > 0
>    `x_offset` + `width`  <= `IHDR` width
>    `y_offset` + `height` <= `IHDR` height
> ```

The first 4 are obviously pointless, because that's implicitly determined by the unsigned-ness of these integers. But it would make sense to check the bottom 2 constraints with the `valid/max` key (kaitai-io/kaitai_struct#435). It is available since the 0.9 release, so be sure to first increment the declared minimum version in [`/meta/ks-version`](https://github.com/kaitai-io/kaitai_struct_formats/blob/4a97145510b2bf3a7dcecf64da46ee363b657e89/image/png.ksy#L18):

```diff
-  ks-version: 0.8
+  ks-version: 0.9
```

I suggest writing the checks like this:

```ksy
      - id: x_offset
        type: u4
        valid:
          max: _root.ihdr.width
        doc: X position at which to render the following frame
```

```ksy
      - id: width
        type: u4
        valid:
          max: _root.ihdr.width - x_offset
        doc: Width of the following frame
```

Of course, it would be actually enough to do the check only once in `width` as the spec suggests (`` `x_offset` + `width`  <= `IHDR` width `` obviously cannot be true if `` `x_offset` > `IHDR` width `` unless `width` is negative, but that can't happen because we're working with unsigned integers). But I think that it's reasonable to report errors on the field that is actually wrong - if there is only a check on `width`, an eventual `ValidationGreaterThanError` will pinpoint the YAML path of `width`, although already `x_offset` might have been invalid.

generalmimon:
From the [spec](https://wiki.mozilla.org/APNG_Specification#MIME_type):

> ## MIME type
> APNG can be identified using the `image/apng` MIME type.

Given that the `image/png.ksy` spec also becomes a spec for the APNG format, what about adding that MIME type into [`/meta/xref/mime`](https://github.com/kaitai-io/kaitai_struct_formats/blob/4a97145510b2bf3a7dcecf64da46ee363b657e89/image/png.ksy#L10)?

```diff
-    mime: image/png
+    mime:
+      - image/png
+      - image/apng
```

image/png.ksy
KOLANICH: I guess it may make sense to introduce the type `fraction`

armijnhemel:
The generated Python code generates this:

```
    class DisposeOpValues(Enum):
        none = 0
        background = 1
        previous = 2
```
"none" is very close to the reserved keyword None. Is it advisable to rename this?

generalmimon:
> "none" is very close to the reserved keyword None. Is it advisable to rename this?

No. Python is case sensitive, so it will not cause any problem there, and using the identifier `none` is already quite common in the existing .ksy specifications in the format gallery, see https://github.com/kaitai-io/kaitai_struct_formats/search?q=%22none%22.

image/png.ksy
armijnhemel: I added a few checks. Unfortunately the web IDE barfs as soon as I add:

```
        valid:
          max: _root.ihdr.width - x_offset
```

generalmimon: > Unfortunately the web IDE barfs as soon as I add

Oh, sorry, my bad, I haven't realized that `width` comes before `x_offset`, not after (I was confused by the order of the constraints as presented in the spec). In this case, just do it the other way around:

```ksy
      - id: width
        type: u4
        valid:
          max: _root.ihdr.width
        doc: Width of the following frame
```

```ksy
      - id: x_offset
        type: u4
        valid:
          max: _root.ihdr.width - width
        doc: X position at which to render the following frame
```

It makes actually no real difference - either you consider `x_offset` static and assert that `width` must fit into the remaining space:

```
|            _root.ihdr.width            |
| x_offset | _root.ihdr.width - x_offset |
           | ------> |                   |
              width
```

or you consider `width` static and assert that `x_offset` must adapt:

```
|            _root.ihdr.width            |
| _root.ihdr.width - width |    width    |
| ----------> |            |
   x_offset
```

image/png.ksy
armijnhemel: I am at a bit at a loss of how to do this properly, as there is a special case in case delay_den is 0. I don't know how to capture that in kaitai (still learning).

generalmimon: > as there is a special case in case delay_den is 0

Are you looking for the [ternary (conditional) operator](https://doc.kaitai.io/user_guide.html#_ternary_if_then_else_operator)?

If you have any struggles with putting the expression together, just describe what you need and we'll help you.

armijnhemel: It is this part of the specification:

```
The `delay_num` and `delay_den` parameters together specify a fraction indicating the time to display the current frame, in seconds. If the denominator is 0, it is to be treated as if it were 100 (that is, `delay_num` then specifies 1/100ths of a second). If the the value of the numerator is 0 the decoder should render the next frame as quickly as possible, though viewers may impose a reasonable lower bound. 
```

So in pseudo code it would be something like this:

if delay_den == 0, then delay_den = 100
value = delay_num/delay_den

generalmimon: > if delay_den == 0, then delay_den = 100
> value = delay_num/delay_den

That's an easy one:

```ksy
instances:
  delay:
    value: delay_num / (delay_den == 0 ? 100.0 : delay_den)
```

The only thing that you need to be aware of is that given that both `delay_num` and `delay_den` are **integers**, you need to convert one of the division operands to a float (`double`) in order to get _float division_ (not integer division, which happens if you do `A / B` with both `A` and `B` integers). See https://doc.kaitai.io/user_guide.html#_arithmetic_operators:

> <table>
> <tbody>
> <tr>
> <th>NOTE</th>
> <td>If both operands are integers, the result of an arithmetic operation is an integer, otherwise it is a floating point number. For example, that means that <code>7 / 2</code> is <code>3</code>, and <code>7 / 2.0</code> is <code>3.5</code>.</td>
> </tr>
> </tbody>
> </table>

Kaitai Struct has a strong type system under the hood and it tracks type of every symbol in all expressions. Here, I used the fact that you can make the overall derived type of a ternary operation (`condition ? if_true : if_false`) a float by making either `if_true` or `if_false` a float - I deliberately use `100.0` (instead of just `100`) so that the division works as intended. If there was just `100`, the result would be truncated to an integer (you could just get `0` seconds or `1` second, but not anything between that).

generalmimon: > I think this is probably best documented in the value instance as discussed below.

The value instance helps, but @GreyCat probably meant that you could also write `doc: in seconds` to the value instance so that it is clear in what units the final decimal value is measured.

armijnhemel: OK, I added it, including the doc as requested. I hope that everything is fine now. I tested it with the web IDE and it seems to work.

armijnhemel:
Odd. The kaitai compiler gives me the following error:

    png.ksy:353:52: mapping values are not allowed here

I am using the 0.9 version:

```
$ kaitai-struct-compiler --version
kaitai-struct-compiler 0.9
```

armijnhemel:
> Odd. The kaitai compiler gives me the following error:
> 
> ```
> png.ksy:353:52: mapping values are not allowed here
> ```
> 
> I am using the 0.9 version:
> 
> ```
> $ kaitai-struct-compiler --version
> kaitai-struct-compiler 0.9
> ```

Never mind, fixed.

armijnhemel:
If I would want to add more PNG data, not related to APNG, should I do that in this PR, or wait for this one to be merged?

generalmimon:
> If I would want to add more PNG data, not related to APNG, should I do that in this PR, or wait for this one to be merged?

If it's not related to APNG, it should be done in another PR.

armijnhemel:
> > If I would want to add more PNG data, not related to APNG, should I do that in this PR, or wait for this one to be merged?
> 
> If it's not related to APNG, it should be done in another PR.

I will do this as soon as this particular one has been merged. Thanks for all the help so far, I really appreciate it.

generalmimon:
Let's add more metadata for the APNG format ([`image/png.ksy:1-21`](https://github.com/kaitai-io/kaitai_struct_formats/blob/a80aac83f10c1dc8913e8620996fb5d4e6f7172e/image/png.ksy#L1-L21)):

```diff
 meta:
   id: png
   title: PNG (Portable Network Graphics) file
-  file-extension: png
+  file-extension:
+    - png
+    - apng
   xref:
     forensicswiki: Portable_Network_Graphics_(PNG)
     iso: 15948:2004
-    justsolve: PNG
+    justsolve:
+      - PNG
+      - APNG
     loc: fdd000153
     mime:
       - image/png
       - image/apng
+      - image/vnd.mozilla.apng
     pronom:
       - fmt/11 # PNG 1.0
       - fmt/12 # PNG 1.1
       - fmt/13 # PNG 1.2
+      - fmt/935 # APNG
     rfc: 2083
     wikidata: Q178051
   license: CC0-1.0
   ks-version: 0.9
   endian: be
```

The file extension `.apng` does not appear to be very common (APNG files usually use the standard `.png` extension too), but it probably makes sense to include it anyway.

image/png.ksy
```diff
@@ -292,6 +305,68 @@ types:
...
+        doc: delay in seconds
+  frame_data_chunk:
+    doc-ref: https://wiki.mozilla.org/APNG_Specification#.60fdAT.60:_The_Frame_Data_Chunk
+    seq:
+      - id: sequence_number
+        type: u4
+        doc: Sequence number of the animation chunk
+      - id: frame_data
+        size-eos: true
+        doc: Frame data for the frame
```
generalmimon, line 369: https://wiki.mozilla.org/APNG_Specification#.60fdAT.60:_The_Frame_Data_Chunk:

> At least one `fdAT` chunk is required for each frame. The compressed datastream is then the **concatenation** of the contents of the data fields **of all the `fdAT` chunks within a frame**.

This is currently not clear from the KSY spec, so please include it into this `doc`. It also applies to the `IDAT` chunk, but this is currently handled just implicitly as a raw byte array, so there is no space for `doc`:

https://github.com/kaitai-io/kaitai_struct_formats/blob/a4f0d393c97163708d53b557b4c46c23f7de9300/image/png.ksy#L64

Sample files using this data splitting across multiple `IDAT`/`fdAT` chunks are here: [https://philip.html5.org/tests/apng/tests.html#:~:text=IDAT, fdAT splitting](https://philip.html5.org/tests/apng/tests.html#:~:text=IDAT,%20fdAT%20splitting)

image/png.ksy
```diff
@@ -292,6 +305,68 @@ types:
...
+        enum: dispose_op_values
+        doc: Type of frame area disposal to be done after rendering this frame
+      - id: blend_op
+        type: u1
+        enum: blend_op_values
+        doc: Type of frame area rendering for this frame
+    instances:
+      delay:
+        value: 'delay_num / (delay_den == 0 ? 100.0 : delay_den)'
+        doc: delay in seconds
```
generalmimon, line 360: Probably it makes sense to write this a little bit more descriptively. The [spec](https://wiki.mozilla.org/APNG_Specification#.60fcTL.60:_The_Frame_Control_Chunk) says

> The `delay_num` and `delay_den` parameters together specify a fraction **indicating the time to display the current frame, in seconds**.

and I suggest

```suggestion
        doc: Time to display this frame, in seconds
```

generalmimon (changes requested):
It would be good to add https://philip.html5.org/tests/apng/tests.html into the top-level `/doc` (which does not exist yet, so just create it) as a source of APNG test files.

image/png.ksy
```diff
@@ -292,6 +305,68 @@ types:
...
+        doc: Number of frames
+      - id: num_plays
+        type: u4
+        doc: Number of times to loop, 0 indicates infinite looping.
+  frame_control_chunk:
+    doc-ref: https://wiki.mozilla.org/APNG_Specification#.60fcTL.60:_The_Frame_Control_Chunk
+    seq:
+      - id: sequence_number
+        type: u4
+        doc: Sequence number of the animation chunk
```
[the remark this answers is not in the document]
armijnhemel: The specification says it nicely, so I have just included it. I couldn't word it better myself.

image/png.ksy
```diff
@@ -292,6 +309,81 @@ types:
...
+    doc-ref: https://wiki.mozilla.org/APNG_Specification#.60fdAT.60:_The_Frame_Data_Chunk
+    seq:
+      - id: sequence_number
+        type: u4
+        doc: |
+          Sequence number of the animation chunk. The fcTL and fdAT chunks
+          have a 4 byte sequence number. Both chunk types share the sequence.
+          The purpose of this number is to detect (and optionally correct)
+          sequence errors in an Animated PNG, since the PNG specification
+          does not impose ordering restrictions on ancillary chunks.
```
generalmimon, line 377: This sentence is probably not necessary.

generalmimon (approved):
Looks good to me, thanks for your contribution!

Issue #361, opened 2021-01-01
`file` command's magic pattern file


generalmimon:
```ksy
meta:
  id: magic_file
  title: "`file` command's magic pattern file"
  application: "`file` command"
  file-extension: mgc
  tags:
    - linux
doc: |
  A database of "magic patterns" that the `file` command uses to identify
  the type of a file. It is usually located in `/usr/share/misc/magic.mgc`.
doc-ref: https://manpages.ubuntu.com/manpages/bionic/man5/magic.5.html
```

dgelessus:
Made some edits:

* Clarified that this is about the compiled binary database format, not the text-based source format for magic patterns.
* Mentioned libmagic, the library behind the `file` command. Most of the time you only use the command-line interface, but libmagic also provides a C and Python API.
* Replaced `linux` tag with `unix`, because it's not Linux-specific in any way. If Wikipedia is to be believed, the first `file` command appeared on an early Unix version in 1973, and even the open-source implementation used by most Linux distros (and macOS) predates Linux by five years.
* Added another common path for the default magic.mgc file (found on macOS 10.14 and Ubuntu 20.04).
* Added link to the `file`/libmagic source code that reads/writes the binary database format.

generalmimon:
> * Replaced `linux` tag with `unix`, because it's not Linux-specific in any way.

Perhaps, I used `linux` only because it is an established tag that KSF site recognizes and includes the format into the https://formats.kaitai.io/#:~:text=GNU/Linux-specific category. I didn't think of factual correctness too much. I am a Windows user and the `file` utility definitely seems to me to be more connected with the Linux world 😁

The downside of using `unix` tag right now is that it won't be recognized by the KSF site. Maybe we can rename the `linux` tag to `unix`, now when the set of `/meta/tags` values is still pretty recent (it was introduced by @GreyCat a few months ago IIRC)?

dgelessus:
> Maybe we can rename the `linux` tag to `unix`

Hm, I'm wondering if it's worth keeping the separate `linux` tag in addition to `unix`. Of the specs currently tagged as `linux`, many are indeed Linux-specific, or at least associated much more with Linux than other Unixes (e. g. `btrfs_stream`, `cramfs`, `ext2`, `luks`, `lvm2`, `systemd_journal`), so a separate `linux` tag would still have some use.

[...]

PR #433, opened 2021-03-09
Add Time Zone Information Format (TZif) - common/tzif.ksy

ian-abbott:
TZif files are commonly used by the C standard libraries on most UNIX
systems to calculate local time from universal time for a time zone, or
(with some amount of ambiguity for negative transitions) vice versa.

### DIFF
--- a/common/tzif.ksy
+++ b/common/tzif.ksy
@@ -149,20 +149,16 @@ types:
             false: s4
             true: s8
         repeat: expr
-        repeat-expr: num_transitions + 0
-        # Appending '+ 0' to suppress a warning "use `num_transition_times`
-        # instead of `num_transitions` [...]".  We want the same parameter
-        # value to be the repeat count for two different arrays.
+        repeat-expr: num_transitions
+        # Note: `num_transitions` also as repeat count for `transition_types`.
         doc: |
           Array of POSIX time values sorted in ascending order. Each is used as
           a transition time at which the rules for computing local time change.
       - id: transition_types
         type: u1
         repeat: expr
-        repeat-expr: num_transitions + 0
-        # Appending '+ 0' to suppress a warning (see `transition_times`
-        # attribute above). This array has the same number of elements as
-        # `transition_times` above.
+        repeat-expr: num_transitions
+        # Note: `num_transitions` also as repeat count for `transition_times`.
         doc: |
           Array of transition type numbers. All but the last one map transition
           times in the `transition_times` array to local time type records in


common/tzif.ksy
```diff
@@ -0,0 +1,256 @@
...
+  ks-version: 0.8
+  endian: be
+doc: |
+  TZif is the binary file format used to represent data for a single time zone
+  compiled by the _zic_ compiler from information in the IANA time zone
+  database. It is used by most UNIX systems to calculate local time for a time
+  zone, usually identified by a time zone identifier string referring to a
+  relatively populous location within the geographic region covered by the time
+  zone, for example: "America/New_York" or "Europe/Paris". This identifier is
+  external to the TZif file itself.
```
armijnhemel, line 18: It might be good to include a link to the specifications. The file format is described in section 5 of the manpage of tzfile:

```
man 5 tzfile
```

with an online reference here:

http://man7.org/linux/man-pages/man5/tzfile.5.html

ian-abbott: The ultimate specification is RFC 8536. I tagged that in meta/xref/rfc. I could add a doc-ref linking to the html version of the RFC. I'm reluctant to link to the Linux manpage because it is not definitive. The source for the definitive man page is in the tzcode or tzdb package, but I don't know of a nice, formatted version online.

ian-abbott: I've added a doc-ref to the top level pointing to the html version of the RFC, and added some other doc-ref lines pointing to sections of the RFC where appropriate.

common/tzif.ksy
```diff
@@ -0,0 +1,262 @@
+meta:
+  id: tzif
+  title: The Time Zone Information Format (TZif)
+  xref:
+    mime: application/tzif
```
KOLANICH, line 5:
```yaml
mime:
  - application/tzif
  - application/tzif-leap
```

common/tzif.ksy
KOLANICH: I guess, if it is in xrefs, it shouldn't be in doc-refs. @generalmimon, what do you think?

common/tzif.ksy
```diff
@@ -0,0 +1,262 @@
...
+  tz_header:
+    doc-ref: https://tools.ietf.org/html/rfc8536#section-3.1
+    seq:
+      - id: magic
+        contents: TZif
+      - id: ver_byte
+        -orig-id: ver
+        type: u1
+        doc: |
+          File format version: 0x00 Version 1; 0x32 Version 2; 0x33 Version 3.
```
KOLANICH, line 55: should be either or a enum, or the undocumented internal structure decomposed anyhow.

common/tzif.ksy
```diff
@@ -0,0 +1,262 @@
...
+          block MUST have the same version as as the first header block.
+      - id: reserved
+        size: 15
+        doc: Reserved bytes MUST be 0.
+      - id: num_is_ut_flags
+        -orig-id: isutcnt
+        type: u4
+        doc: |
+          Number of UT/local indicators contained in the data block. MUST
+          either be 0 or equal to `num_local_time_types`.
```
KOLANICH, line 66: `valid`

common/tzif.ksy
```diff
@@ -0,0 +1,262 @@
...
+        type: u4
+        doc: |
+          Number of local time type records contained in the data block. MUST
+          NOT be 0.
+      - id: len_tz_designations
+        -orig-id: charcnt
+        type: u4
+        doc: |
+          Total number of octets used by the set of time zone designations
+          contained in the data block. MUST NOT be 0. Includes the trailing NUL
```
KOLANICH, line 92: `valid`

common/tzif.ksy
```diff
@@ -0,0 +1,262 @@
...
+          switch-on: is_ts64
+          cases:
+            false: s4
+            true: s8
+        repeat: expr
+        repeat-expr: _parent.header.num_transition_times
+        doc: |
+          A series of UNIX leap-time values sorted in strictly ascending order.
+          Each value is used as a transition time at which the rules for
+          computing local time may change. Each time value SHOULD be at least
```
KOLANICH, line 118: ???

common/tzif.ksy
```diff
@@ -0,0 +1,262 @@
...
+          local time types were specified as standard time or wall-clock time.
+          The value MUST be 0 or 1. A value of one (1) indicates standard time.
+          The value MUST be set to one (1) if the corresponding UT/local
+          indicator is set to one (1). A value of zero (0) indicates wall time.
+          If `_parent.header.num_is_std_flags` is zero (0), all transition times
+          associated with local time types are assumed to be specified as wall
+          time.
+      - id: is_ut_flags
+        # UT/local indicators
+        type: b8
```
KOLANICH, line 165: `b8` is `u1`,  but less efficient.

This probably should be an own type full of bit flags.

common/tzif.ksy
```diff
@@ -0,0 +1,262 @@
+meta:
+  id: tzif
+  title: The Time Zone Information Format (TZif)
+  xref:
+    mime: application/tzif
+    # mime: "application/tzif-leap" if leap-second records are included
+    rfc: 8536
+  license: CC0-1.0
```
KOLANICH, line 8: Are the comments copued from somewhere? IETF RFCs are copyrighted to IETF.

armijnhemel: Yes they are BUT IETF grants a license if you attribute:

https://trustee.ietf.org/documents/trust-legal-provisions/tlp-5/

It is in section 3.

common/tzif.ksy
generalmimon: @KOLANICH:

> I guess, if it is in xrefs, it shouldn't be in doc-refs. @generalmimon, what do you think?

I agree with you, there is no need for that. At least the top-level `/doc-ref` is useless given that it's already in `/meta/xref`. `doc-ref`s in subtypes pointing to a specific section with a hash in the URL (like `https://tools.ietf.org/html/rfc8536#section-3.1`) kind of make sense because they provide additional information, and you don't have to find the relevant section in the entire RFC spec.

common/tzif.ksy
ian-abbott: I don't want it to break if/when the version gets incremented in the future.

common/tzif.ksy
ian-abbott: Actually, that's a bug. It should be `u1` like `is_std_flags`.

common/tzif.ksy
ian-abbott: Taken from the RFC.

common/tzif.ksy
ian-abbott: I probably need to reword all the comments to avoid copyright issues.

common/tzif.ksy
KOLANICH: https://github.com/kaitai-io/kaitai_struct/issues/435

common/tzif.ksy
ian-abbott: Done.

common/tzif.ksy
ian-abbott: Removed the top-level `/doc-ref`.

common/tzif.ksy
```diff
@@ -0,0 +1,262 @@
...
+      - id: ver_byte
+        -orig-id: ver
+        type: u1
+        doc: |
+          File format version: 0x00 Version 1; 0x32 Version 2; 0x33 Version 3.
+          The additional header block for version 2+ after the version 1 data
+          block MUST have the same version as as the first header block.
+      - id: reserved
+        size: 15
+        doc: Reserved bytes MUST be 0.
```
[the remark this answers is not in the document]
ian-abbott: The reserved bytes _might_ be used by a future version, but should be 0 in the current version (3) and earlier versions. It is probably safer for readers to ignore the reserved field, but writers should set it to all zeros (at least for current versions).

common/tzif.ksy
ian-abbott: What should `ks-version` be set to if validation is used?

common/tzif.ksy
```diff
@@ -0,0 +1,262 @@
...
+          -2**59.
+      - id: transition_types
+        type: u1
+        repeat: expr
+        repeat-expr: _parent.header.num_transition_times
+        doc: |
+          Transition types specifying the type of local time of the
+          corresponding transition time. These values serve as zero-based
+          indices into the array of local time type records. Each type index
+          MUST be in the range [0, `_parent.header.num_local_time_types` - 1].
```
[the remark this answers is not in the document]
ian-abbott: Is it possible to validate all the elements of an array are in range? Does it need some sort of iteration loop?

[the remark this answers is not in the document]
ian-abbott: The only way I can think of to validate the elements of an array of simple type is to change the simple type to a custom `seq` type containing a single attribute plus validation check, but that would make the overall type structure more complicated than it should be.

common/tzif.ksy
ian-abbott: To avoid forward references, I have validated `num_local_time_types` instead of `num_is_ut_flags`.

common/tzif.ksy
```diff
@@ -0,0 +1,262 @@
...
+        type: u4
+        doc: |
+          Number of UT/local indicators contained in the data block. MUST
+          either be 0 or equal to `num_local_time_types`.
+      - id: num_is_std_flags
+        -orig-id: isstdcnt
+        type: u4
+        doc: |
+          Number of standard/wall indicators contained in the data block. MUST
+          either be 0 or equal to `num_local_time_types`.
```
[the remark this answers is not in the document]
ian-abbott: To avoid forward references, I have validated `num_local_time_types` instead of `num_is_std_flags`.

common/tzif.ksy
ian-abbott: Validated as non-zero.

common/tzif.ksy
ian-abbott: These would be hard to validate!

common/tzif.ksy
```diff
@@ -0,0 +1,279 @@
...
+          `local_time_types` array).
+      - id: local_time_types
+        type: tz_local_time_type
+        repeat: expr
+        repeat-expr: _parent.header.num_local_time_types
+        doc: Local time type records.
+      - id: tz_designations
+        size: _parent.header.len_tz_designations
+        valid:
+          expr: '_[_parent.header.len_tz_designations - 1] == 0'
```
generalmimon, line 152: You can simplify this to

```suggestion
          expr: '_.last == 0'
```

..., see https://github.com/kaitai-io/kaitai_struct_tests/blob/97d68e9847aea0783eee21c27933ebf01fd4e441/formats/expr_bytes_ops.ksy#L16-L17.

One of the points in the 0.9 version was that we achieved a feature parity (in all languages except Perl) between "normal" arrays and byte arrays (i.e. implemented the [array methods](https://doc.kaitai.io/user_guide.html#_array_types) also for byte arrays - https://github.com/kaitai-io/kaitai_struct/issues/633), so you can take advantage of that.

common/tzif.ksy
```diff
@@ -0,0 +1,279 @@
...
+      - id: is_dst
+        -orig-id: dst
+        type: u1
+        valid:
+          max: 1
+        doc: |
+          Indicates whether local time should be considered Daylight Saving Time
+          (DST). The value MUST be 0 or 1. A value of one (1) indicates that
+          this time type is DST. A value of zero (0) indicates that this time
+          type is standard time.
```
generalmimon, line 219: A singular attribute beginning with `is_` should be a boolean, not an integer. You should provide a boolean value instance as a supplement to this attribute:

```diff
-      - id: is_dst
+      - id: dst
-        -orig-id: dst
         type: u1
```

```ksy
    instances:
      is_dst:
        value: dst != 0
```

Also I don't really like `valid/max: 1`, it does not express the intent well. Be explicit rather than implicit (i.e. use [`valid/any-of`](https://github.com/kaitai-io/kaitai_struct/issues/435#issue-323332353)):

```ksy
  - id: dst
    type: u1
    valid:
      any-of:
        - 0
        - 1
```

alternatively (if you prefer):

```ksy
  - id: dst
    type: u1
    valid:
      any-of: [0, 1]
```

common/tzif.ksy
```diff
@@ -0,0 +1,279 @@
+meta:
+  id: tzif
+  title: The Time Zone Information Format (TZif)
+  xref:
+    mime:
+      - application/tzif
+      - application/tzif-leap
+    rfc: 8536
```
generalmimon, line 8: You said in `/doc` that

> It is used by most **UNIX** systems to calculate local time for a time zone

..., so you might tag the format as `unix`-related:

```suggestion
    rfc: 8536
  tags:
    - unix
```

Note: in https://github.com/kaitai-io/kaitai_struct_formats/issues/361#issuecomment-753707340 shares @dgelessus his opinion whether to keep both `linux` and `unix` tags (currently existing specs in the format gallery [only use `linux`](https://github.com/kaitai-io/kaitai_struct_formats/search?q=tags+linux)).

KOLANICH: How about `nix` tag?

common/tzif.ksy
ian-abbott: Thanks, I've changed it.

common/tzif.ksy
ian-abbott: It's definitely not Linux-specific, and not really Unix-specific either. Other systems such as Android and iOS use TZif files. There are also TZif readers for other software systems, for example Python's `zoneinfo` module and the timezone support in PHP.

common/tzif.ksy
ian-abbott: The TZif data also has arrays of u8 that contain [0, 1] values. I'm not sure how to deal with those unless value instances can be used to convert arrays of u8 to arrays of bool.

ian-abbott: For now, I have renamed it to `dst_flag` and changed the validity to `any-of: [0, 1]`, but have not mapped it to a boolean value instance. I have also renamed the other `is_` attributes of type `u1`.

common/tzif.ksy
generalmimon: > It's definitely not Linux-specific, and not really Unix-specific either. Other systems such as Android and iOS use TZif files. There are also TZif readers for other software systems, for example Python's `zoneinfo` module and the timezone support in PHP.

OK, then it probably doesn't make sense to add it.

common/tzif.ksy
ian-abbott: There is an unreleased TZif version 4 that has no changes to the format, just some relaxations in the contents of the leap-second records (if they are present). I plan to update the doc-strings for v4 when it gets released officially.

KOLANICH:
I guess it was just a warning, not an error.

generalmimon:
@ian-abbott:

> It wanted me to rename it to both `num_transition_times` and `num_transition_types`. I worked around it by using `num_transitions + 0`.

There is no point in suppressing these harmless warnings at any cost. These warnings are only meant to help a bit with sticking some of the https://doc.kaitai.io/ksy_style_guide.html#attr-id recommendations, but they are implemented quite naively (and can sometimes contradict each other, as in your case), so they cannot be followed without thinking.

So just use `num_transitions` and ignore the warning. It'll disappear once the compiler gets a bit more robust logic on whether to issue the warning or not.

ian-abbott:
@generalmimon OK, I have removed the workaround that suppressed those harmless warnings. The only thing that is bugging me now is that I get this error when compiling for **nim**, **perl**, or **rust**:

```
tzif: /:
	error: an implementation is missing
```

moreati:
I wrote tzif.ksy https://github.com/kaitai-io/kaitai_struct_formats/compare/master...moreati:kaitai_struct_formats:tzif. before I thought to check for open PRs. Sorry for the duplication.

Ian, do you still want to get yours merged? Would you like me to compare them and add review comments? Alternatively  would you be interested in a PR against your branch - to then feed into this PR? Otherwise I'm happy to leave you in peace. You have priority, I don't want to usurp your effort/credit.

PR #435, opened 2021-03-09
Added bencode

KOLANICH:


KOLANICH:
@generalmimon, @dgelessius

KOLANICH:
@generalmimon

arvidn:
it seems I was pinged on this, but I can't see it anywhere. I don't really read this file format, but one thing I notice is the "encoding: utf-8" in the header.

It's important to know that not all strings are UTF-8 encoded. bencoding doesn't distinguish between blobs of raw bytes and human readable strings, they are both encoded as strings.

The human readable strings are UTF-8 encoded, but there's nothing in the bencoding format itself that hints on which strings are meant to be human readable. Not even dictionary keys are guaranteed to be human readable strings (and are sometimes hash digests, in binary).

KOLANICH:
Thanks. Would fix that. In fact this utf-8 is not really used for binary strings, but only for metainformation (sizes, `:` chars). It's better to use ascii there, since the alphabet is limited to it. Do you know any adopted enough extesions of bencode it may make sense to include here?

arvidn:
The only extension I'm aware of is `rencode` which Deluge uses in its RPC protocol. It attempts to have a more compact representation of integers, and maybe some other types. I doubt it's prominent enough to include here. afaik it's only used in that one place.

KOLANICH:
Thanks.

PR #71, opened 2017-10-25
Added Odin3 protocol

KOLANICH:

PR #443, opened 2021-03-11
Added Odin3 protocol

KOLANICH:
Replaces #71

### DIFF
--- a/firmware/odin3.ksy
+++ b/firmware/odin3.ksy
@@ -1,0 +1,204 @@
+meta:
+  id: odin3
+  title: ODIN3
+  license: MIT
+  endian: le
+  application:
+    - Samsung ODIN3
+    - Samsung LOKE
+    - Heimdall
+  xref:
+    wikidata: Q19599499
+doc: |
+  [Samsung Odin](https://en.wikipedia.org/wiki/Odin_(firmware_flashing_software)) is a proprietary piece of software developed by Samsung and used to flash firmware into Samsung devices. Odin utility has a Loke counterpart in the device bootloader. This ksy documents the message format of the protocol they talk to each other.
+  The protocol was reverse engineered by Benjamin Dobell who have created a MIT-licensed utility Heimdall, on which source code this ksy is based.
+  If you wanna test and augment this spec keep in mind that a lot of websites spreading leaked versions of Odin utility were created with the sole purpose of spreading malware glued with the tool. The most trustworthy website I know is the one belonging to chainfire, a well-known dev on Android scene.
+doc-ref: https://gitlab.com/BenjaminDobell/Heimdall/tree/master/heimdall/source
+enums:
+  chip_type:
+    0: ram
+    1: nand
+  tr_request:
+    0: flash
+    1: dump
+    2: part
+    3: end
+    0x2000: unknown2000
+  destination:
+    0: phone
+    1: modem
+  packet_type:
+    0x00000000: send_file_part
+    0x00000064: session
+    0x00000065: pit_file
+    0x00000066: file_transfer
+    0x00000067: end_session
+instances:
+  odin_handshake_message:
+    contents:   ["ODIN", 0x00]
+  loke_handshake_message:
+    contents:   ["LOKE"]
+
+seq:
+  - id: type
+    type: u4
+    enum: packet_type
+  - id: content
+    size-eos: true
+    type:
+      switch-on: type
+      cases:
+        "packet_type::session": session
+        "packet_type::pit_file": pit_file
+        "packet_type::file_transfer": file_transfer
+        "packet_type::end_session": end_session
+
+types:
+  end_session:
+    seq:
+      - id: request
+        type: u4
+        enum: request
+    enums:
+      request:
+        0: end_session
+        1: reboot_device
+
+  file_transfer:
+    seq:
+      - id: request
+        type: u4
+        enum: tr_request
+      - id: content
+        type:
+          switch-on: request
+          cases:
+            "tr_request::flash": flash
+            "tr_request::dump": dump
+            "tr_request::part": part
+            "tr_request::end": end
+    types:
+      flash:
+        seq:
+          - id: sequence_byte_count
+            type: u4
+      part:
+        seq:
+          - id: sequence_byte_count
+            type: u4
+           # or (recheck this!)
+          - id: part_index
+            type: u4
+      dump:
+        seq:
+          - id: chip_type
+            type: u4
+            enum: chip_type
+          - id: chip_id
+            type: u4
+      end:
+        seq:
+          - id: destination
+            type: u4
+            enum: destination
+          - id: sequence_byte_count
+            type: u4
+          - id: unknown1 # efs?
+            type: u4
+          - id: device_type
+            type: u4
+          - id: content
+            type:
+              switch-on: destination
+              cases:
+                "destination::phone": phone
+                #"destination::modem": modem
+          - id: end_of_file
+            type: u4
+        types:
+          phone:
+            seq:
+              - id: file_identifier
+                type: u4
+                enum: file
+            enums:
+              file:
+                0x00: primary_bootloader
+                0x01: pit # don't flash the pit this way!
+                0x03: secondary_bootloader
+                0x04: secondary_bootloader_backup
+                0x06: kernel
+                0x07: recovery
+                0x08: tablet_modem
+
+                0x12: unknown12
+                0x14: efs
+                0x15: param_lfs
+                0x16: factory_file_system
+                0x17: database_data
+                0x18: cache
+
+                0x0b: modem # kies flashes the modem this way
+
+
+
+  pit_file:
+    seq:
+      - id: request
+        type: u4
+        enum: tr_request
+      - id: content
+        type:
+          switch-on: request
+          cases:
+            "tr_request::flash": flash
+            "tr_request::part": part
+            "tr_request::end": end
+    types:
+      flash:
+        seq:
+          - id: part_size
+            type: u4
+      part:
+        seq:
+          - id: part_index
+            type: u4
+      end:
+        seq:
+          - id: file_size
+            type: u4
+
+  session:
+    enums:
+      request:
+        0: begin_session
+        1: device_type
+        2: total_bytes
+        3: enable_some_sort_of_flag
+        5: file_part_size
+        8: enable_tflash
+    seq:
+      - id: request
+        type: u4
+        enum: request
+      #- id: unknown_parameter
+      #  type: u4
+      - id: content
+        type:
+          switch-on: request
+          cases:
+            #"request::begin_session": begin_session
+            #"request::device_type": device_type
+            "request::total_bytes": total_bytes
+            #"request::enable_some_sort_of_flag": enable_some_sort_of_flag
+            "request::file_part_size": file_part_size
+            #"request::enable_tflash": enable_tflash
+    types:
+      file_part_size:
+        seq:
+          - id: file_part_size
+            type: u4
+      total_bytes:
+        seq:
+          - id: total_bytes
+            type: u4


KOLANICH:
@TheAirBlow @nickelc

TheAirBlow:
> @TheAirBlow @nickelc

Hello! I already documented most of the stuff on https://github.com/samsung-loki/samsung-docs, no need in this.

KOLANICH:
Thanks for the info, but using the info from the leaks can be illegal, so I cannot use it, because of the same reasons developers of Wine and ReactOS stay away of any Micro$oft leaks.

TheAirBlow:
> Thanks for the info, but using the info from the leaks can be illegal, so I cannot use it, because of the same reasons developers of Wine and ReactOS stay away of any Micro$oft leaks.

I have only used the leak to make the PIT more accurate, all of the other stuff is reverse-engineered

TheAirBlow:
We could've probably get to it by ourselves, as V1/V2 was such a logical thing man

KOLANICH:
@TheAirBlow, I have added the info about variables names in the impls (Heimdall, Thor, wuotan) to the spec. It seems that not all the impls have all the features (or maybe the code is not very clean and contains hardcoded numbers like in `DeviceInfo.cs`). You may want to check your Thor for the features missing, also maybe we can upstream something (missing enums + DeviceInfo) into Heimdall.

Also it may make sense to document all the enums in the repo in the form of ksy files, it will allow to generate source code automatically (the enums used in ODIN binary protocol are already documented in this ksy, but I guess it can be useful to do the same with the enums used in the text one (it makes no sense to describe the structures used in the text protocol in the form of ksys, but enums are different beasts)).

TheAirBlow:
@KOLANICH, and why you're even using software as a reference point? https://github.com/samsung-loki/samsung-docs has a lot more information, mostly discovered by using IDA Pro on the `aboot.mbn`.

KOLANICH:
@TheAirBlow,
1. software is far less ambigious, for example in the docs (not backed by any open-source software I have managed to find) it is written that the command changing region code uses 3 byte string, but it doesn't tell if the region code is 2 bytes + null terminator (and if it can be more in that case?) or 3 bytes of the code, it doesn't say anything about the case being used, about if this code matches some ISO standard. Software could contain answers to these questions.
2. working software is the primary source of truth. docs can lie, working software is always true to some extent
3. I find this free-form docs is inconvenient to use. These flat tables with inconsistent field names have caused me to do parser's job manually to map the structures and fields in my spec to the structures and fields in the tables.
4. For legal reasons.

I propose
1. relicense the docs to MIT license, as Dobel's Heimdall is. Any other permissive license is OK too, but just to play safe I have licensed the spec under MIT, since I have derived it from Heimdall.
2. migrate all the stuff from there that is not yet in my spec but should be there into it.
3. replace the free-form docs with my spec.
4. maybe even replace the parsers within the impls to the generated parsers, so the spec becomes the primary source of truth and benefiting the spec automatically benefits all the software.

TheAirBlow:
> @TheAirBlow,
> 1. software is far less ambigious, for example in the docs (not backed by any open-source software I have managed to find) it is written that the command changing region code uses 3 byte string, but it doesn't tell if the region code is 2 bytes + null terminator (and if it can be more in that case?) or 3 bytes of the code, it doesn't say anything about the case being used, about if this code matches some ISO standard. Software could contain answers to these questions.
> 2. working software is the primary source of truth. docs can lie, working software is always true to some extent
> 3. I find this free-form docs is inconvenient to use. These flat tables with inconsistent field names have caused me to do parser's job manually to map the structures and fields in my spec to the structures and fields in the tables.
> 4. For legal reasons.
> 
> I propose
> 1. relicense the docs to MIT license, as Dobel's Heimdall is. Any other permissive license is OK too, but just to play safe I have licensed the spec under MIT, since I have derived it from Heimdall.
> 2. migrate all the stuff from there that is not yet in my spec but should be there into it.
> 3. replace the free-form docs with my spec.
> 4. maybe even replace the parsers within the impls to the generated parsers, so the spec becomes the primary source of truth and benefiting the spec automatically benefits all the software.
> 

1) It doesn't say "null-terminated" specifically, so:
* Everyone knows that a region code is 3 chars long (SER for example)
* Docs tell "String, 3 bytes", so it is **not null-terminated**.
* I would make a glossary to answer such questions, people are different and I will take this in account.
2) Docs is the best way to gather information, no need to dive into a lot of code, and most of rhe comments would probably just repeat stuff from the docs, or in C# there is `<inheritdocs/>`
3) Inconsistency? The docs are updated to match reality, so it is to be expected. Or what did you mean?

I propose section
1) As I remember, all of my stuff is licensed under a Mozilla licence, I do not remember it's full name. It is FOSS and permits non-commercial use, isn't it fine.
3) How does it actually generate code? All of these parsers? There is no reason to have 1 sextillion of parsers for each response, which are pretty much the same, with a single being changed most of the time (packet type), so it is better to just create generic parsers/packers. If response/request has arguments, no problem, another generic parser/packer.

Prove me wrong, send me the generated code. In theory it would just make the life harder.

TheAirBlow:
> >1. about region code
> 
> Thanks for the info.
> 
> >3. Inconsistency?
> 
> `| 0x01 | 32-bit integer | Modem/CP |`
> `| 0x00 | 32-bit integer | Phone/AP |`
> 
> But how it really should look like, is how it is written in the spec, which one can argue it be translated in the following table row
> 
> ```
> | (un)signed 32-bit integer | `destination` | destination |
> ```
> 
> and then
> 
> ```
> Destinations
> Phone
> <the rest of phone table ....>
> modem
> <the rest of modem table ....>
> ```
> 
> IETF RFCs are written kinda like this.
> 
> >As I remember, all of my stuff is licensed under a Mozilla licence, I do not remember it's full name. It is FOSS and permits non-commercial use, isn't it fine.
> 
> It is MPL, which is a weak copyleft license, but not permissive. It means I'm not allowed to take quotes out of your docs in any significant amount.
> 
> BTW, inmorder to relicense legally you need consent of all the contributors.
> 
> BTW, I prefer Unlicense license.
> 
> >How does it actually generate code?
> 
> The whole `kaitai-io` org on GH is about the tooling automatically generating parsers for a set of languages. Each spec (ksy file) can be automatically converted into a parser usable from one of the supported languages. You can download the nightly version of the tool here: https://cloudsmith.io/~kaitai/repos/debian-unstable for Debian/Ubuntu and https://ci.appveyor.com/project/kaitai-io/kaitai-struct/build/artifacts for Windows.
> 
> >Prove me wrong, send me the generated code.
> 
> https://github.com/KOLANICH-tools/USBPcapOdinDumper/blob/master/USBPcapOdinDumper/kaitai/odin3.py
> 
> (I have retested the tool against tue pcap files provided by people in Heimdall repo, it errors (I guess the spec for parsing usbmon/usbpcap uas some issues, but it is not now to be solved) , but I remember it worked against my pcap file (I don't have it anymore, but here is the old dump of raw packets into files using that tool on the pcap file https://github.com/Benjamin-Dobell/Heimdall/files/1414758/ODIN_flash_capture.zip , you can try to test your impl by parsing these files, each file corresponds to 1 usb transaction))
> 

As I thought, a parser for each packet type. It is not the best thing to do :/

TheAirBlow:
> > 3. Inconsistency?
> 
> `| 0x01 | 32-bit integer | Modem/CP |` `| 0x00 | 32-bit integer | Phone/AP |`
> 
> But how it really should look like, is how it is written in the spec, which one can argue it be translated in the following table row
> 
> ```
> | (un)signed 32-bit integer | `destination` | destination |
> ```
> 
> and then
> 
> ```
> Destinations
> Phone
> <the rest of phone table ....>
> modem
> <the rest of modem table ....>
> ```

This is a massive misunderstanding. It is not the intended way to think about this.
(btw, integer is signed by default, so it is normal to say `32-bit integer` instead of `signed 32-bit integer`
Why does it have to be this foolproof?)
We have something like this in docs:
```
| dynamic | 32-bit integer | Binary Type (more info below) |

Below:
| 0x00 | 32-bit integer | Phone/AP |
| 0x01 | 32-bit integer | Modem/CP |

Doc's is structured in this way:
| Value (may be dynamic) | Type | Description |
```
So, the intended path is go check Binary Type section below,
and then use the Value and Description as possible values and it's description.

KOLANICH:
>I did document these in the docs properly, why do you hate them so much? The're skipped cuz it is not needed with C# being managed, thw same thing is probably also possible to do in C++

Thanks. Currently I'm fixing my USBPcapOdinDumper (and also writing a usbmon dissector for scapy), so I have not yet uploaded the new spec and have not tested it yet. Do you have a pcap file with these packets and could you document the base of the offsets?

>... uh, what? Handshake is a standard thing.

When designing the spec I have noticed the pattern that the most of messages from LOKE to ODIN are of the same format as the messages from ODIN to LOKE (packet), it is a very logical thing as it simplifies design and extending of the parser and debug tooling (and I guess they had them) because one can use the same parser. But some messages, like the one starting a session, don't follow this pattern. They are exceptions from the rules I have to process separately. It is a bit disappointing.

>The're skipped cuz it is not needed with C# being managed, thw same thing is probably also possible to do in C++

I don't understand what you mean, offsets can be used from any lang, managed or not.

>btw, integer is signed by default

For me integers are unsigned by default because signed integers make sense only in very specific use cases. For example a size cannot be negative. If a size is negative, then negative "sizes" are imbued with a special meaning that they are not sizes but something else. If only one negative value is meaningful (usually -1, in other words the greatest possible value of a signed integer), it is not a reason to make the whole field signed rather then just compare to that value (it is a more future-proof approach because in a very hypothetical situation of exhaustion of the positive range of a signed int, the very natural thing is just to say that the int is unsigned and get almost the same amount of values).

But endiannes ... again, little endian is the only natural way to deal with endianness when numbers are stored in memory buffers. That's why all the modern hardware is LE, so the formats are designed to be LE to avoid conversions.

TheAirBlow:
@KOLANICH 

> When designing the spec I have noticed the pattern that the most of messages from LOKE to ODIN are of the same format as the messages from ODIN to LOKE (packet), it is a very logical thing as it simplifies design and extending of the parser and debug tooling (and I guess they had them) because one can use the same parser. But some messages, like the one starting a session, don't follow this pattern. They are exceptions from the rules I have to process separately. It is a bit disappointing.

It wouldn't be reliable to use strings as commands in the protocol itself.
The handshake is designed this way to differentiate different protocols easily.

> For me integers are unsigned by default because signed integers make sense only in very specific use cases. For example a size cannot be negative. If a size is negative, then negative "sizes" are imbued with a special meaning that they are not sizes but something else. If only one negative value is meaningful (usually -1, in other words the greatest possible value of a signed integer), it is not a reason to make the whole field signed rather then just compare to that value (it is a more future-proof approach because in a very hypothetical situation of exhaustion of the positive range of a signed int, the very natural thing is just to say that the int is unsigned and get almost the same amount of values).

Again, you're a very, very different person. Most languages use `int` 
for `Int32` (signed), and `uint` for `UInt32` (unsigned), so signed integer is today's default.

> I don't understand what you mean, offsets can be used from any lang, managed or not.

In C++ you need more pointer fuckery to achieve the same thing.

> Thanks. Currently I'm fixing my USBPcapOdinDumper (and also writing a usbmon dissector for scapy), so I have not yet uploaded the new spec and have not tested it yet. Do you have a pcap file with these packets and could you document the base of the offsets?

Forgot about the base. I'll look into it soon.

> But endiannes ... again, little endian is the only natural way to deal with endianness when numbers are stored in memory buffers. That's why all the modern hardware is LE, so the formats are designed to be LE to avoid conversions.

I already threw that out of the way.

TheAirBlow:
> I'm not speaking about handshake strings. I'm speaking about the responses of LOKE to the ODIN messages with `type == packet_type::session and content.requests session::request::begin_session`. **I may be wrong** (I know this understanding contradicts your docs, I need to check it again against Heimdall sources), but my current mental model of that, that I got while originally reading sources of Heimdall (it was in 2017), is the following. Usual `session` payloads have `request` field acting as a tag, in both requests and responses, and in responses the same tag is used as in the requests they answer to. It is very convenient: one doesn't have to track state of the convo and can use the same parser for reqs and responses. But in the answers to `session::request::begin_session` that field seems to be ommitted. It immediately makes it very inconvenient since the way of parsing of the responses becomes depending on the type of the requests preceeding them, the protocol stops being stateless and we may need a state machine to properly parse it (I've made a mistake in the spec that I treated all `type == packet_type::session and is_loki_to_odin` the same, disrespect to the history).

Responses of LOKE to ODIN... I still don't get what you mean.
The protocol works like that:
1) First, we send a handshake, `ODIN` is one of them
2) Then we need to expect a response, in this case `LOKE`
3) After that we need to start the session
4) Do all of the stuff we need
5) End the session, and possibly reboot

My docs don't have "guides", and it is also something I would do soon.

Here is my point about Kaitai Structs - it is pointless, because every single response and request is a different class, and it is not a very good thing. It makes it for developers harder to use and it would still need to be modified, because requests/responses should inherit classes `Request`/`Response`, etc.

Also, after reading the generated code **parses requests**, instead of **packing** them. 
Idk if it is true, I don't know python well. I asked for C# and C++ last time, why did you send me Python?

TheAirBlow:
@KOLANICH, `Do you have a pcap file`
No, I don't. I don't want to test stuff on my main phone, which has a broken bootloader.
It already spooked the crap out of me. Never touching Odin on my phone again.

I am looking for testers, because I just finished implementing flashing.

TheAirBlow:
@KOLANICH, может лучше на русском? Нам обоим проще будет.

PR #445, opened 2021-03-13
Rendering format: GLB

hakanai:
glTF's binary container


rendering/glb.ksy
```diff
@@ -0,0 +1,69 @@
...
+  - id: chunks
+    type: chunk
+    repeat: eos
+
+types:
+
+  header:
+    seq:
+      - id: magic
+        contents: 'glTF'
```
KOLANICH, line 28: the quotes are not needed.

rendering/glb.ksy
```diff
@@ -0,0 +1,69 @@
+meta:
+  id: glb
```
generalmimon, line 2: Isn't it better to call the format `gltf` or `gltf_binary`? It seems to be much more common name than `glb`. From what I've seen, `.glb` is only used as a file extension.

The `/meta/id` should reflect the format name rather than file extension.

rendering/glb.ksy
```diff
@@ -0,0 +1,69 @@
...
+  title: GL Transmission Format, binary container
+  file-extension: glb
+  endian: le
+  license: MIT
+  xref:
+    mime: model/gltf-binary
+    wikidata: Q28135989
```
generalmimon, line 9: Let's reorder the keys per the recommendation in the [style guide](https://doc.kaitai.io/ksy_style_guide.html#meta):

```suggestion
  title: GL Transmission Format, binary container
  file-extension: glb
  xref:
    mime: model/gltf-binary
    wikidata: Q28135989
  license: MIT
  endian: le
```

rendering/glb.ksy
```diff
@@ -0,0 +1,69 @@
...
+
+doc: |
+  glTF is a format for distribution of 3D models optimized for being used in software
+
+doc-ref: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#binary-gltf-layout
```
generalmimon, line 14: Link to a specific commit rather than `master` which changes over time:

```suggestion
doc-ref: https://github.com/KhronosGroup/glTF/tree/2354846/specification/2.0#binary-gltf-layout
```

rendering/glb.ksy
```diff
@@ -0,0 +1,69 @@
...
+        encoding: UTF-8
+        size-eos: true
+        doc: |
+          This is where GLB deviates from being an elegant format.
+          To parse the rest of the file, you have to parse the JSON first.
+
+  bin:
+    seq:
+      - id: data
+        size-eos: true
```
generalmimon, line 62: What is the structure of `data`? Can it be described by the KS language?

rendering/glb.ksy
```diff
@@ -0,0 +1,69 @@
...
+
+  chunk:
+    seq:
+      - id: length
+        type: u4
+      - id: type
+        type: u4
+        enum: chunk_type
+      - id: data
+        size: length
```
generalmimon, line 42: `len_data` per [style guide](https://doc.kaitai.io/ksy_style_guide.html#attr-id):

```suggestion
      - id: len_data
        type: u4
      - id: type
        type: u4
        enum: chunk_type
      - id: data
        size: len_data
```

rendering/glb.ksy
```diff
@@ -0,0 +1,69 @@
...
+    seq:
+      - id: data
+        size-eos: true
+
+enums:
+  chunk_type:
+    # Literally "JSON"
+    0x4E4F534A: json
+    # Literally "BIN\0"
+    0x004E4942: bin
```
generalmimon, line 69: Stating the textual representation of a four-character code in a comment is a common practice, and repeating _"Literally"_ everywhere is not scalable. Just omit it and write the text representation on the same line, it's still clear:

```suggestion
    0x4E4F534A: json # "JSON"
    0x004E4942: bin # "BIN\0"
```

rendering/glb.ksy
```diff
@@ -0,0 +1,69 @@
...
+    repeat: eos
+
+types:
+
+  header:
+    seq:
+      - id: magic
+        contents: glTF
+      - id: version
+        type: u4
```
generalmimon, line 30: If I wanted to use this KSY spec for serialization, what I should set this field to in order to have the files recognized by other applications? Is there only one version that everybody uses, or several ones? How does the version affect the format structure/logic?

Please write this info into the `doc` key under this field.

rendering/glb.ksy
```diff
@@ -0,0 +1,69 @@
...
+      - id: length
+        type: u4
```
generalmimon, line 32: This is unused? Is it supposed to mean the byte length of the `header`, or what?

generalmimon (changes requested):
> <code><strong>rendering/</strong>glb.ksy</code>

I don't really like that you create a top-level folder in KSF without prior discussion. _Rendering_ is a terrible name (it's way too broad - nothing says that all specs from the [`image/`](https://github.com/kaitai-io/kaitai_struct_formats/tree/master/image) folder wouldn't fit into `rendering/`; also, I can't imagine anything under a "rendering format" or "format for rendering"), and it simply can't stay in the official format gallery.

rendering/glb.ksy
```diff
@@ -0,0 +1,69 @@
...
+          cases:
+            'chunk_type::json': json
+            'chunk_type::bin': bin
+
+  json:
+    seq:
+      - id: data
+        type: str
+        encoding: UTF-8
+        size-eos: true
```
generalmimon, line 54: The order à la [style guide](https://doc.kaitai.io/ksy_style_guide.html#seq-attr):

```suggestion
      - id: data
        size-eos: true
        type: str
        encoding: UTF-8
```

There is a logic behind it: when this field is parsed, first an intermediate byte array till the `e`nd `o`f `s`tream is read, which is then interpreted as a `str`ing in the specified `encoding`.

rendering/glb.ksy
hakanai: gltf_binary seems better, will rename to that

hakanai:
> I don't really like that you create a top-level folder in KSF without prior discussion.
>
>  _Rendering_ is a terrible name (it's way too broad - nothing says that all specs from the [`image/`](https://github.com/kaitai-io/kaitai_struct_formats/tree/master/image) folder wouldn't fit into `rendering/`; also, I can't imagine anything under a "rendering format" or "format for rendering"), and it simply can't stay in the official format gallery.

I don't like the name either.

Although... I will also point out that it isn't my fault that there wasn't a folder for these sorts of files yet. 3D model formats aren't exactly new.

generalmimon:
> Although... I will also point out that it isn't my fault that there wasn't a folder for these sorts of files yet. 3D model formats aren't exactly new.

Yes, you're exactly right with that. Right now they're in [`media/`](https://github.com/kaitai-io/kaitai_struct_formats/tree/c7658a6/media) ([`blender_blend.ksy`](https://github.com/kaitai-io/kaitai_struct_formats/blob/c7658a6/media/blender_blend.ksy), [`stl.ksy`](https://github.com/kaitai-io/kaitai_struct_formats/blob/c7658a6/media/stl.ksy)), but I agree that it's not ideal.

rendering/glb.ksy
hakanai: > `length` is the total length of the Binary glTF, including Header and all Chunks, in bytes.

Could add as docs I guess?

rendering/glb.ksy
hakanai: Possibly, but the problem is that getting the offsets requires parsing the JSON.

hakanai:
I have to admit that I hadn't thought to look in media for similar formats, because they're not media formats. 🤔

rendering/glb.ksy
KOLANICH: Here is the discussion from the previous issue: https://github.com/kaitai-io/kaitai_struct_formats/pull/420#discussion_r576577074

generalmimon: > Possibly, but the problem is that getting the offsets requires parsing the JSON.

I see. Nevertheless, you can provide parametric types to help the programmer with parsing once the JSON is parsed. Here's an outline of what I mean (I used https://github.com/KhronosGroup/glTF-Sample-Models/blob/0bc2ad7/2.0/BoxVertexColors/glTF-Binary/BoxVertexColors.glb for testing and I hardcoded some values from the JSON into the KSY for convenient testing in the [Web IDE](https://ide.kaitai.io/) - these will need to be removed and provided by the application code instead to work for a generic file):

```ksy
seq:
  # ...
  - id: buffer_views
    type: buffer_view([0, 72, 360, 648, 1032][_index], [72, 288, 288, 384, 192][_index])
    repeat: expr
    repeat-expr: 5
  - id: accessors
    type: >-
      accessor(buffer_views[ [0, 1, 2, 3, 4][_index] ], 0, _index == 0 ? component_type::unsigned_short : component_type::float, _index == 0 ? 36 : 24, ["SCALAR", "VEC3", "VEC3", "VEC4", "VEC2"][_index])
    repeat: expr
    repeat-expr: 5

types:
  # ...
  
  slot: {}

  buffer_view:
    params:
      - id: byte_offset
        type: s4
      - id: byte_length
        type: s4
    instances:
      bytes:
        io: _root.chunks[1].data.as<bin>._io
        pos: byte_offset
        size: byte_length
        type: slot
        if: _root.chunks.size >= 2 and _root.chunks[1].type == chunk_type::bin
  
  accessor:
    params:
      - id: buffer_view
        type: buffer_view
      - id: byte_offset
        type: s4
      - id: component_type # bug in the Web IDE: enum params don't show the corresponding enum value name
        type: s4
        enum: component_type
      - id: num_items
        type: s4
      - id: type
        type: str
    instances:
      num_components_per_item:
        value: |
          type == "SCALAR" ? 1
          : type == "VEC2" ? 2
          : type == "VEC3" ? 3
          : type == "VEC4" ? 4
          : type == "MAT2" ? 4
          : type == "MAT3" ? 9
          : type == "MAT4" ? 16
          : 0
      items:
        io: buffer_view.bytes._io
        pos: byte_offset
        type:
          switch-on: component_type
          cases:
            component_type::byte: s1
            component_type::unsigned_byte: u1
            component_type::short: s2
            component_type::unsigned_short: u2
            component_type::unsigned_int: u4
            component_type::float: f4
        repeat: expr
        repeat-expr: num_items * num_components_per_item
      t:
        value: component_type

enums:
  chunk_type:
    0x4E4F534A: json # "JSON"
    0x004E4942: bin  # "BIN\0"
  component_type:
    5120: byte
    5121: unsigned_byte
    5122: short
    5123: unsigned_short
    5125: unsigned_int
    5126: float
```

Don't consider it done, it's only a proof of concept. If there is any other important structure or logic that I missed in the [specification](https://github.com/KhronosGroup/glTF/tree/2354846/specification/2.0#binary-data-storage), please implement it. For example, how does the [data alignment](https://github.com/KhronosGroup/glTF/tree/2354846/specification/2.0#data-alignment) come into play for `accessor`s? I also don't use the `byteStride` property of `bufferView`, because I frankly have no idea what it means ;-)

hakanai: All of that sounds like large change, future improvement stuff. Getting the base structure down first still makes sense.

generalmimon:
> As a side-note I had wondered whether `3d/` would be an OK name for it.

Yep, that will do. We probably can't come up with a better name: it's clear what it's trying to represent (this group of 3d modelling formats is also almost exclusively referred to as "3d file formats", see https://www.google.com/search?q=3d+formats) and it's not too broad or too narrow.

rendering/glb.ksy
generalmimon: > All of that sounds like large change, future improvement stuff. Getting the base structure down first still makes sense.

OK, as you wish. Let's leave it to the future.

generalmimon (approved):

generalmimon:
@hakanai Well, actually, how are CAD formats related to 3D formats? Don't 3D formats fit into the broader term CAD? I've just discovered that we actually have the folder [`cad/`](https://github.com/kaitai-io/kaitai_struct_formats/tree/3e4be117e4c09aa4d4d206c9d4aa21bd8f3fcc5d/cad) in the format gallery, completely forgot about it before...

To me, it sounds very similar: the [STL format](https://en.wikipedia.org/wiki/STL_(file_format)) (which I perceive as an exemplary 3D format) is described on Wikipedia as follows:

> **STL** is a [file format](https://en.wikipedia.org/wiki/File_format) native to the [stereolithography](https://en.wikipedia.org/wiki/Stereolithography) <ins>[CAD](https://en.wikipedia.org/wiki/Computer-aided_design)</ins> software created by [3D Systems](https://en.wikipedia.org/wiki/3D_Systems).

So IMHO odds are that the glTF format can also be classified as CAD-related format.

KOLANICH:
IMHO it is 3D CAD formats (solid modelling or meshes, doesn't matter) should be classified as `3d`

PR #446, opened 2021-03-13
Rendering formats from MMD (MikuMikuDance)

hakanai:


rendering/mmd_pmd.ksy
```diff
@@ -0,0 +1,371 @@
...
+      - id: name
+        type: strz
+        size: 20
+      - id: parent_index
+        type: u2
+      - id: tail_index
+        type: u2
+      - id: type
+        type: u1
+      - id: ik_index
```
KOLANICH, line 170: probably can be connected to the ik via an instance.

hakanai: Nice idea

hakanai: It could just be the web UI, but I get "Call stack: undefined RangeError: Maximum call stack size exceeded" trying to add even just:

```yaml
      parent:
        value: '_parent.bones[parent_index]'
      tail:
        value: '_parent.bones[tail_index]'
```

Whereas just this is fine:
```yaml
      parent:
        value: '_parent.bones[parent_index]'
```

3d/mmd_pmd.ksy
```diff
@@ -0,0 +1,371 @@
...
+        type: vec3
+      - id: rotation_limitation2
+        type: vec3
+      - id: spring_position
+        type: vec3
+      - id: spring_rotation
+        type: vec3
+
+  u2_2:
+    seq:
```
KOLANICH, line 331: Does it make much sense to name the fields individually rather than `repeat-expr: 2`?

3d/mmd_pmd.ksy
```diff
@@ -0,0 +1,371 @@
...
+        type: f4
+      - id: type
+        type: u1
+
+  constraint:
+    seq:
+      - id: name
+        type: strz
+        size: 20
+      - id: rigid_body_index1
```
hakanai, line 309: Might change these to repeats as well because it seems like it might be easier to use that way.

3d/mmd_pmd.ksy
```diff
@@ -0,0 +1,371 @@
...
+        type: u4
+      - id: rigid_body_index2
+        type: u4
+      - id: position
+        type: vec3
+      - id: rotation
+        type: vec3
+      - id: translation_limitation1
```
hakanai, line 317: These on the other hand I would leave as-is because I strongly suspect that it's a lower/upper pair and that naming would make more sense. To rename it, I would have to figure out which is which, though.

3d/mmd_pmd.ksy
KOLANICH: >I guess I could have both, using a repeat and then adding instances for the individual properties?

Yes, I usually do this, but it would cause overhead because the compiler is not smart enough to not to cache instance evaluation resu.ts in this case. But I guess it would be fixed somewhen.

3d/mmd_pmx.ksy
```diff
@@ -0,0 +1,653 @@
...
+        type: len_string
+      - id: diffuse
+        type: vec4
+      - id: specular
+        type: vec3
+      - id: shininess
+        type: f4
+      - id: ambient
+        type: vec3
+      - id: flags
```
KOLANICH, line 231: Should it be a type of `b1`s?

3d/mmd_pmx.ksy
```diff
@@ -0,0 +1,653 @@
...
+        type: len_string
+      - id: english_name
+        type: len_string
+      - id: position
+        type: vec3
+      - id: parent_index
+        type: sized_index(_root.header.bone_index_size)
+      - id: transformation_class
+        type: u4
+      - id: flags
```
KOLANICH, line 279: Again, can be a type of `b1`s (you will need `bit-endian`).

3d/mmd_pmd.ksy
hakanai: Incidentally is there a good way to refactor these out to a common location?

hakanai: Figured out a way... PR has been updated. I went back on using `instances` though because it did bloat the class size and memory usage to crazy levels. I separated colour structures so that those still have intuitive field names.

3d/vector_types.ksy
```diff
@@ -0,0 +1,54 @@
+meta:
```
KOLANICH, line 1: 1. It may make sense to move it into `/common/`
2. I guess each type name should contain the types of elements.
3. I wonder if we should embed a 2d vector into 3d one.
4.
```yaml
-affected-by:
  - 135  # would allow to drop the types with hardcoded types of fields
  - 65  # implementation of constant-size types via raw structs would make embedding one vec into another one more efficient (effectively zero-cost for some languages) 
`

https://github.com/kaitai-io/kaitai_struct/issues/135 https://github.com/kaitai-io/kaitai_struct/issues/65

hakanai: Embedding a vector so that you can access `.xy` more efficiently is probably no good either because it assumes that you prefer the X and Y coords over Z and that is not always the case because it really depends on which way the object is oriented. In graphics programming I have seen things like `.zx` or even things like `.xyxx`.

So really all I’d want here is straight call-through methods which avoid caching so that I can do those swizzles.

hakanai: As for moving to common, yeah, I worried about moving it to common _exactly because_ then I’d have to deal with people commenting about the types of elements, or even the endianness, or other things along those lines.

hakanai: What would _really_ be convenient here is parametrised imports so that I can import vector types here and pass in the element type and endianness.

PR #447, opened 2021-03-13
Rendering format: FBX

hakanai:
Filmbox, now owned by Autocad, but produced and consumed by lots of others.

### DIFF
--- a/rendering/fbx.ksy
+++ b/rendering/fbx.ksy
@@ -1,0 +1,167 @@
+meta:
+  id: kaydara_fbx
+  title: Kaydara FBX (Filmbox) 3D model
+  file-extension: fbx
+  application: Autodesk/Kaydara Filmbox
+  encoding: UTF-8
+  endian: le
+  license: MIT
+
+doc: |
+  A proprietary file format developed by Kaydara and owned by Autodesk
+  since 2006. It is used to provide interoperability between digital content
+  creation applications. FBX is also part of Autodesk Gameware, a series of
+  video game middleware.
+
+seq:
+  - id: header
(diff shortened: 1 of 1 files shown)


rendering/fbx.ksy
```diff
@@ -0,0 +1,167 @@
...
+    seq:
+      - id: length
+        type: u4
+      - id: value
+        size: length
+
+  array:
+    params:
+      # Works around lack of ability to parameterise the actual type by passing a fake one through.
+      # https://github.com/kaitai-io/kaitai_struct/issues/135
```
KOLANICH, line 104: `--affected-by`

hakanai: How does `-affected-by` work, i.e. where exactly can it go? I can't seem find it in the docs.

KOLANICH: Props starting from `-` are not processed by KSK and are for documentation purposes and for software other than KSC. We would like to have a bot notifying spec authors about the issues affecting their specs being fixed and counting stats on issues to prioritize their fixing. But this bot is not yet implemented.

the proposed annotation format is `-affected-by: <the number of the issue in KS main repo>`. It can also be an array of such numbers.

hakanai: Can it be in any position in the tree then?

KOLANICH: Likely no. I guess only in dicts that are not expected to have user-supplied identifiers as keys.

hakanai:
Latest push fixes parsing for the relatively large and complex FBX file for my character model.

The odds of this working for all models should now be significantly better. :)

PR #448, opened 2021-03-14
openpgp message: rewrite duplicate "reserved" entries

armijnhemel:
The word "reserved" was used multiple times. Although the compiler compiles it just fine I ran into trouble when using it with Python. This commit should resolve this issue.

### DIFF
--- a/security/openpgp_message.ksy
+++ b/security/openpgp_message.ksy
@@ -448,26 +448,26 @@ enums:
     110: private_experimental_algorithm_10
 
   subpacket_types:
-    0: reserved
-    1: reserved
+    0: reserved_0
+    1: reserved_1
     2: signature_creation_time
     3: signature_expiration_time
     4: exportable_certification
     5: trust_signature
     6: regular_expression
     7: revocable
-    8: reserved
+    8: reserved_8
     9: key_expiration_time
     10: placeholder_for_backward_compatibility
     11: preferred_symmetric_algorithms
     12: revocation_key
-    13: reserved
-    14: reserved
-    15: reserved
+    13: reserved_13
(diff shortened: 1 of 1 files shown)


generalmimon:
@armijnhemel Yes, thanks for that, duplicate enum keys are incorrect (they don't work properly in any language), and the fact that compiler silently allows them is a bug: https://github.com/kaitai-io/kaitai_struct_formats/pull/421#discussion_r588939814

security/openpgp_message.ksy
```diff
@@ -448,26 +448,26 @@ enums:
     110: private_experimental_algorithm_10
 
   subpacket_types:
-    0: reserved
-    1: reserved
+    0: reserved_0
+    1: reserved_1
```
generalmimon, line 452: Should be rather `reserved0`, `reserved1` per the [style guide](https://doc.kaitai.io/ksy_style_guide.html#attr-id):

> For `reserved` fields which are known to be unused, use reserved name (or `reserved1`, `reserved2`, etc, if there are many of them)

Although this is not a field but an enum key, there is no reason to use a different convention.

generalmimon (changes requested):

security/openpgp_message.ksy
armijnhemel: Alrighty then. I also fixed a few other instances of that.

generalmimon (approved):
Thanks!

security/openpgp_message.ksy
KOLANICH: For enums it is proposed to use a different convention - to use enum number value instead of sequence numbers. But here they match.

PR #456, opened 2021-03-21
ELF: parse note sections + other ELF edits

### LINKED ISSUE #253: elf has no sh_type::dynstr definition but used in it
int elf.ksy sh_type defines without dynstr:
```ksy
 sh_type:
    0: null_type
    1: progbits
    2: symtab
    3: strtab
    4: rela
    5: hash
    6: dynamic
    7: note
    8: nobits
    9: rel
    10: shlib
    11: dynsym
    14: init_array
    15: fini_array
    16: preinit_array
    17: group
    18: symtab_shndx
```
but then use it in:
```ksy
         body:
            io: _root._io
            pos: ofs_body
            size: len_body
            type:
              switch-on: type
              cases:
                'sh_type::dynamic': dynamic_section
                'sh_type::strtab': strings_struct
                'sh_type::dynsym': dynsym_section
                'sh_type::dynstr': strings_struct
          name:
```
the generated python code:
```python
 class ShType(Enum):
        null_type = 0
        progbits = 1
        symtab = 2
        strtab = 3
        rela = 4
        hash = 5
        dynamic = 6
        note = 7
        nobits = 8
        rel = 9
        shlib = 10
```
and generated python use the dynstr:
```python
  @property
            def body(self):
                if hasattr(self, '_m_body'):
                    return self._m_body if hasattr(self, '_m_body') else None

                io = self._root._io
                _pos = io.pos()
                io.seek(self.ofs_body)
                if self._is_le:
                    _on = self.type
                    if _on == self._root.ShType.dynamic:
                        self._raw__m_body = io.read_bytes(self.len_body)
                        io = KaitaiStream(BytesIO(self._raw__m_body))
                        self._m_body = self._root.EndianElf.DynamicSection(io, self, self._root, self._is_le)
                    elif _on == self._root.ShType.strtab:
                        self._raw__m_body = io.read_bytes(self.len_body)
                        io = KaitaiStream(BytesIO(self._raw__m_body))
                        self._m_body = self._root.EndianElf.StringsStruct(io, self, self._root, self._is_le)
                    elif _on == self._root.ShType.dynstr: # !!!!!!here
                        self._raw__m_body = io.read_bytes(self.len_body)
                        io = KaitaiStream(BytesIO(self._raw__m_body))
                        self._m_body = self._root.EndianElf.StringsStruct(io, self, self._root, self._is_le)
                    elif _on == self._root.ShType.dynsym:
                        self._raw__m_body = io.read_bytes(self.len_body)
                        io = KaitaiStream(BytesIO(self._raw__m_body))
                        self._m_body = self._root.EndianElf.DynsymSection(io, self, self._root, self._is_le)
                    else:
                        self._m_body = io.read_bytes(self.len_body)
                else:
```
the generated python code would get an error:

> Traceback (most recent call last):
>   File "/root/remote/mdeobfs/kk.py", line 171, in <module>
>     k25(486546801)
>   File "/root/remote/mdeobfs/kk.py", line 33, in k25
>     eu.write_mem(image_base + __text_sec.addr, __text_sec.body)
>   File "/root/remote/mdeobfs/elf.py", line 1069, in body
>     elif _on == self._root.ShType.dynstr:
>   File "/root/.pyenv/versions/3.7.4/lib/python3.7/enum.py", line 349, in __getattr__
>     raise AttributeError(name) from None
> AttributeError: dynstr

@GreyCat I updated the details

armijnhemel:
Parse note sections in ELF files. The fields note_name and note_description are 4 byte aligned (hence the ugly calculation). Although the ELF specification seems to hint that the note_name field is ASCII I have found that this isn't the case on Linux binaries that have the .gnu.build.attributes section:

https://fedoraproject.org/wiki/Toolchain/Watermark#Proposed_Specification_for_non-loaded_notes

Fixes #253

### DIFF
--- a/executable/elf.ksy
+++ b/executable/elf.ksy
@@ -504,6 +504,7 @@ types:
             type: u4
           - id: name
             size: len_name
+            terminator: 0
             doc: |
               Although the ELF specification seems to hint that the `note_name` field
               is ASCII this isn't the case for Linux binaries that have a
--- a/executable/elf.ksy
+++ b/executable/elf.ksy
@@ -506,7 +506,7 @@ types:
             size: name_size + (-name_size % 4)
             doc: |
               Although the ELF specification seems to hint that the note_name field
-              is ASCII this isn't the case on Linux binaries that have the
+              is ASCII this isn't the case for Linux binaries that have a
               .gnu.build.attributes section.
             doc-ref: https://fedoraproject.org/wiki/Toolchain/Watermark#Proposed_Specification_for_non-loaded_notes
           - id: note_description
--- a/executable/elf.ksy
+++ b/executable/elf.ksy
@@ -20,7 +20,7 @@ meta:
     - executable
     - linux
   license: CC0-1.0
-  ks-version: 0.8
+  ks-version: 0.9
 doc-ref: https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/elf.h;hb=HEAD
 seq:
   - id: magic
@@ -495,6 +495,14 @@ types:
             type: note_section_entry
             repeat: eos
       note_section_entry:
+        doc-ref:
+          - https://docs.oracle.com/cd/E23824_01/html/819-0690/chapter6-18048.html
+          # The following source claims that note's `name` and `descriptor` should be padded
+          # to 8 bytes in 64-bit ELFs, not always to 4 - although this seems to be an idea of
+          # the original spec, it did not catch on in the real world and most implementations
+          # always use 4-byte alignment - see
+          # <https://fa.linux.kernel.narkive.com/2Za5xb58/patch-01-02-elf-always-define-elf-addr-t-in-linux-elf-h#post13>
+          - https://refspecs.linuxfoundation.org/elf/gabi4+/ch5.pheader.html#note_section
         seq:
           - id: len_name
             type: u4
@@ -517,6 +525,9 @@ types:
           - id: descriptor_padding
             size: -len_descriptor % 4
       relocation_section:
+        doc-ref:
+          - https://docs.oracle.com/cd/E23824_01/html/819-0690/chapter6-54839.html
+          - https://refspecs.linuxfoundation.org/elf/gabi4+/ch4.reloc.html
         params:
           - id: has_addend
             type: bool
--- a/executable/elf.ksy
+++ b/executable/elf.ksy
@@ -405,6 +405,7 @@ types:
                 'sh_type::strtab': strings_struct
                 'sh_type::dynsym': dynsym_section
                 'sh_type::dynstr': strings_struct
+                'sh_type::note': note_section
           name:
             io: _root.header.strings._io
             pos: ofs_name
@@ -486,6 +487,23 @@ types:
             type: u8
           - id: size
             type: u8
+      note_section:
+        seq:
+          - id: entries
+            type: note_section_entry
+            repeat: eos
+      note_section_entry:
+        seq:
+          - id: name_size
+            type: u4
+          - id: description_size
+            type: u4
+          - id: note_type
+            type: u4
+          - id: note_name
+            size: "name_size % 4 == 0 ? name_size: name_size + (4 - name_size % 4)"
+          - id: note_description
+            size: "description_size % 4 == 0 ? description_size: description_size + (4 - description_size % 4)"
     instances:
       program_headers:
         pos: program_header_offset
--- a/executable/elf.ksy
+++ b/executable/elf.ksy
@@ -498,7 +498,7 @@ types:
         seq:
           - id: len_name
             type: u4
-          - id: len_description
+          - id: len_descriptor
             type: u4
           - id: type
             type: u4
@@ -511,10 +511,10 @@ types:
             doc-ref: https://fedoraproject.org/wiki/Toolchain/Watermark#Proposed_Specification_for_non-loaded_notes
           - id: name_padding
             size: -len_name % 4
-          - id: description
-            size: len_description
-          - id: description_padding
-            size: -len_description % 4
+          - id: descriptor
+            size: len_descriptor
+          - id: descriptor_padding
+            size: -len_descriptor % 4
       relocation_section:
         params:
           - id: has_addend
--- a/executable/elf.ksy
+++ b/executable/elf.ksy
@@ -500,16 +500,16 @@ types:
             type: u4
           - id: len_description
             type: u4
-          - id: note_type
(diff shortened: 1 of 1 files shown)


executable/elf.ksy
```diff
@@ -486,6 +487,23 @@ types:
...
+          - id: name_size
+            type: u4
+          - id: description_size
+            type: u4
+          - id: note_type
+            type: u4
+          - id: note_name
+            size: "name_size % 4 == 0 ? name_size: name_size + (4 - name_size % 4)"
+          - id: note_description
+            size: "description_size % 4 == 0 ? description_size: description_size + (4 - description_size % 4)"
```
generalmimon, line 506: 1. Put spaces around both `?` and `:` operators of the ternary operation.
2. If it's not justified to use double quotes `"` (e.g. to be able to use character escapes), use single `'` quotes instead to enclose KS expressions. A potential problem with double quotes `"` is that any [escape sequences](https://yaml.org/spec/1.2/spec.html#id2776092) used in the expression will be interpreted by the YAML parser, not the Kaitai Struct compiler. YAML recognizes some sequences that KSC doesn't, so that can create inconsistency between using `value: '"foo"'` and `value: "'foo'"` syntax.
   For example `value: "'\U00000041'"` will work (it is interpreted by the YAML parser and the compiler only gets the decoded string), but `value: '"\U00000041"'` will not (interpreted by KSC).
3. BTW, you can simplify it to:
   
   ```suggestion
             - id: note_name
               size: name_size + (-name_size % 4)
             - id: note_description
               size: description_size + (-description_size % 4)
   ```
   
   We make extra effort to ensure that modulo of negative integers works just the way the [**modulo** operation](https://en.wikipedia.org/wiki/Modulo_operation) is intended to (see the test [`expr_mod.ksy`](https://github.com/kaitai-io/kaitai_struct_tests/blob/97d68e9847aea0783eee21c27933ebf01fd4e441/formats/expr_mod.ksy#L1), which passes in all languages we're testing except Nim - see https://ci.kaitai.io/ and search test `ExprMod`), so you can safely use it.

   See discussion in https://github.com/kaitai-io/kaitai_struct_formats/pull/409#discussion_r566748681 on this alignment calculation approach.

generalmimon (changes requested):
> Although the ELF specification seems to hint that the note_name field is ASCII I have found that this isn't the case on Linux binaries that have the .gnu.build.attributes section:
> 
> https://fedoraproject.org/wiki/Toolchain/Watermark#Proposed_Specification_for_non-loaded_notes

You should write this information into the KSY to a `doc` key of the `note_section_entry` or the individual attributes which are concerned. Or if it's something that affects all strings across the entire format, write it into the top-level `/doc`.

armijnhemel:
For SHT_REL and SHT_RELA I followed the naming conventions from the specifications, but I was thinking if I shouldn't just rename it from rel_offs and rela_offs to just "offset".

generalmimon:
@armijnhemel Do you know some good sources about the ELF format structure? In the top-level `/doc-ref`, there is just

https://github.com/kaitai-io/kaitai_struct_formats/blob/5b6327312a3e20a2646e6954f6febbbd95729a6d/executable/elf.ksy#L24

..., but there must be some better documentation somewhere.

armijnhemel:
https://docs.oracle.com/cd/E19683-01/816-1386/6m7qcoblh/index.html
https://refspecs.linuxfoundation.org/

Those are usually the documents that I keep returning to.

armijnhemel:
Specifically for this PR, note sections are documented in:

https://refspecs.linuxfoundation.org/elf/gabi4+/ch5.pheader.html#note_section
https://docs.oracle.com/cd/E19683-01/816-1386/chapter6-18048/index.html

Relocation (SHT_REL and SHT_RELA) are documented in:

https://refspecs.linuxfoundation.org/elf/gabi4+/ch4.reloc.html
https://docs.oracle.com/cd/E19683-01/816-1386/chapter6-54839/index.html

Regarding test files for note sections, any recent Linux distribution should have plenty of files. For example, in /bin/ls on my system I see the following when running `readelf`:

```
  [ 2] .note.gnu.property NOTE            0000000000000338 000338 000020 00   A  0   0  8
  [ 3] .note.gnu.build-id NOTE            0000000000000358 000358 000024 00   A  0   0  4
  [ 4] .note.ABI-tag     NOTE            000000000000037c 00037c 000020 00   A  0   0  4

...

Displaying notes found in: .note.gnu.property
  Owner                Data size        Description
  GNU                  0x00000010       NT_GNU_PROPERTY_TYPE_0        Properties: x86 feature: IBT, SHSTK

Displaying notes found in: .note.gnu.build-id
  Owner                Data size        Description
  GNU                  0x00000014       NT_GNU_BUILD_ID (unique build ID bitstring)         Build ID: 16b217835012f85440656d8e4d5c3fa9b7554cb8

Displaying notes found in: .note.ABI-tag
  Owner                Data size        Description
  GNU                  0x00000010       NT_GNU_ABI_TAG (ABI version tag)            OS: Linux, ABI: 3.2.0

```

executable/elf.ksy
```diff
@@ -486,6 +489,73 @@ types:
...
+            type: note_section_entry
+            repeat: eos
+      note_section_entry:
+        doc-ref:
+          - https://docs.oracle.com/cd/E23824_01/html/819-0690/chapter6-18048.html
+          # The following source claims that note's `name` and `descriptor` should be padded
+          # to 8 bytes in 64-bit ELFs, not always to 4 - although this seems to be an idea of
+          # the original spec, it did not catch on in the real world and most implementations
+          # always use 4-byte alignment - see
+          # <https://fa.linux.kernel.narkive.com/2Za5xb58/patch-01-02-elf-always-define-elf-addr-t-in-linux-elf-h#post13>
```
generalmimon, line 504: Hmm, this disagreement whether to use 4- or 8-byte alignment in 64-bit ELFs might be less settled than I thought it was - I've just stumbled upon https://reviews.llvm.org/D61296:

> While the generic ABI requires notes to be 8-byte aligned in ELF64, many
> vendor-specific notes (from Linux, NetBSD, Solaris, etc) use 4-byte
> alignment.

Having a dictionary of known note types/names that use 4-byte alignment (otherwise use 8-byte) doesn't sound reasonable. I guess that the 4-byte alignment is fine to do in KSY, but maybe this comment that I came up with could be updated to reflect this finding. However, I still don't understand the whole situation and don't know how other parsers handle it, if this what I found is a universally adopted strategy or an exception.

generalmimon (approved):
Thanks!

PR #460, opened 2021-03-29
add Android sparse format

armijnhemel:
This commit adds a description of the Android sparse file format.

### DIFF
--- a/archive/android_sparse.ksy
+++ b/archive/android_sparse.ksy
@@ -25,7 +25,7 @@ seq:
   - id: chunks
     type: chunk
     repeat: expr
-    repeat-expr: header.total_chunks
+    repeat-expr: header.num_chunks
 types:
   file_header:
      seq:
@@ -45,11 +45,11 @@ types:
          -orig-id: blk_sz
          type: u4
          doc: block size in bytes, multiple of 4
-       - id: total_blocks
+       - id: num_blocks
          -orig-id: total_blks
          type: u4
          doc: blocks in the original data
-       - id: total_chunks
+       - id: num_chunks
          -orig-id: total_chunks
          type: u4
        - id: checksum
--- a/archive/android_sparse.ksy
+++ b/archive/android_sparse.ksy
@@ -19,14 +19,14 @@ doc: |
     Note: this is not the same as the Android sparse data image format.
 doc-ref: https://android.googlesource.com/platform/system/core/+/7b444f08c1/libsparse/sparse_format.h
 seq:
-  - id: img_header
-    type: header
+  - id: header
+    type: file_header
   - id: chunks
     type: chunk
     repeat: expr
-    repeat-expr: img_header.total_chunks
+    repeat-expr: header.total_chunks
 types:
-  header:
+  file_header:
      seq:
        - id: magic
          contents: [0x3a, 0xff, 0x26, 0xed]
@@ -58,11 +58,11 @@ types:
   chunk:
     seq:
       - id: header
-        type: fixed_header
+        type: chunk_header
       - id: body
         size: header.total_size - header._sizeof
     types:
-      fixed_header:
+      chunk_header:
         seq:
           - id: chunk_type
             type: u2
--- a/archive/android_sparse.ksy
+++ b/archive/android_sparse.ksy
@@ -30,10 +30,8 @@ types:
      seq:
        - id: magic
          contents: [0x3a, 0xff, 0x26, 0xed]
-       - id: major_version
-         type: u2
-       - id: minor_version
-         type: u2
+       - id: version
+         type: version
        - id: file_header_size
          type: u2
          doc: size of file header, should be 28
@@ -67,6 +65,12 @@ types:
         doc: in bytes of chunk input file including chunk header and data
       - id: body
         size: total_size - 12
+  version:
+    seq:
+      - id: major
+        type: u2
+      - id: minor
+        type: u2
 enums:
   chunk_types:
     0xcac1: raw
--- a/archive/android_sparse.ksy
+++ b/archive/android_sparse.ksy
@@ -1,6 +1,6 @@
 meta:
   id: android_sparse
-  title: Android Sparse image
+  title: Android sparse image
   file-extension: img
   tags:
     - archive
--- a/archive/android_sparse.ksy
+++ b/archive/android_sparse.ksy
@@ -18,7 +18,9 @@ doc: |
     <https://android.googlesource.com/platform/system/core/+/7b444f08c1/libsparse> - `img2simg.c`
 
     Note: this is not the same as the Android sparse data image format.
-doc-ref: https://android.googlesource.com/platform/system/core/+/7b444f08c1/libsparse/sparse_format.h
+doc-ref:
+  - https://android.googlesource.com/platform/system/core/+/7b444f08c1/libsparse/sparse_format.h
+  - https://source.android.com/devices/bootloader/images#sparse-image-format
 seq:
   - id: header
     type: file_header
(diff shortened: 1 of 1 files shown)


archive/android_sparse.ksy
```diff
@@ -0,0 +1,75 @@
...
+  - id: img_header_entries
+    type: image_header_entry
+    repeat: expr
+    repeat-expr: img_header.total_chunks
+types:
+  header:
+     seq:
+       - id: magic
+         contents: [0x3a, 0xff, 0x26, 0xed]
+       - id: major_version
```
KOLANICH, line 33:
```yaml
version:
  seq:
    - id: major
      type: u2
  - id: minor
    type: u2
```

archive/android_sparse.ksy
```diff
@@ -0,0 +1,75 @@
...
+      - id: reserved
+        type: u2
+      - id: chunk_size
+        type: u4
+        doc: in blocks in output image
+      - id: total_size
+        type: u4
+        doc: in bytes of chunk input file including chunk header and data
+      - id: body
+        size: total_size - 12
```
KOLANICH, line 69: `sizeof<>`?

armijnhemel: Yes, this is something that I probably need to fix. I looked at the manual, but totally overlooked sizeof.

armijnhemel: Not sure how I should do this :-/

generalmimon: > I looked at the manual, but totally overlooked sizeof.

It's not yet there, sorry for that. The only documentation right now is the issue kaitai-io/kaitai_struct#84 (but note that the "functions" `sizeof()` and `lea()` are not implemented yet). We have also written it into the 0.9 release notes, but that's it:

>   * Implement compile-time `sizeof` and `bitsizeof` operators ([#84](https://github.com/kaitai-io/kaitai_struct/issues/84))
>     * Type-based: `sizeof<u4>`, `bitsizeof<b13>`, `sizeof<user_type>`
>     * Value-based: `file_header._sizeof`, `flags._bitsizeof` (`file_header`, `flags` are fields defined in the current type)

---

> Not sure how I should do this :-/

Well, when you have all the fields that needs to be subtracted on the same level, it's gonna be clumsy - you need to subtract every one of them:

```suggestion
        size: total_size - (chunk_type._sizeof + reserved._sizeof + chunk_size._sizeof + total_size._sizeof)
```

Probably a better way would be to put all fields with fixed size into a common subtype, whose total size can then be accessed by a single `sizeof` operator.

```ksy
  image_header_entry:
    seq:
      - id: header
        type: fixed_header
      - id: body
        size: header.total_size - header._sizeof
        # size: header.total_size - sizeof<fixed_header> # if you prefer
    types:
      fixed_header:
        seq:
          - id: chunk_type
            type: u2
            enum: chunk_types
            doc: chunk type
          - id: reserved
            type: u2
          - id: chunk_size
            type: u4
            doc: in blocks in output image
          - id: total_size
            type: u4
            doc: in bytes of chunk input file including chunk header and data
```

armijnhemel: OK, fixed.

archive/android_sparse.ksy
```diff
@@ -51,20 +51,25 @@ types:
...
-        type: u4
-        doc: in blocks in output image
-      - id: total_size
-        type: u4
-        doc: in bytes of chunk input file including chunk header and data
+      - id: header
+        type: fixed_header
       - id: body
-        size: total_size - 12
+        size: header.total_size - header._sizeof
```
KOLANICH, line 57: @generalmimon, is there any benefit to use `sizeof<type>` in such cases? The fact that in this case we need `sizeof<type>` can probably be inferred automatically, when the type of the value is constant-size and can be resolved. Should we encourage usage of `_sizeof` in these cases in KSYs (but the compiler must generate the code as if `sizeof<type>` is used in these cases, maybe using `decltype`, where available, not lookups in runtime) because they give a bit more semantically-valuable information?

archive/android_sparse.ksy
```diff
@@ -0,0 +1,93 @@
...
+       - id: checksum
+         -orig-id: image_checksum
+         type: u4
+         doc: CRC32 checksum of the original data
+  image_header_entry:
+    seq:
+      - id: header
+        type: fixed_header
+      - id: body
+        size: header.total_size - header._sizeof
```
generalmimon, line 63: I've just stumbled upon a file called [`oem_ru_base.img`](https://herelinkfw.cubepilot.org/tools/oem_ru_base.img) from https://docs.cubepilot.org/user-guides/herelink/herelink-user-guides/oem-image-setup:

> ### [Steps](https://docs.cubepilot.org/user-guides/herelink/herelink-user-guides/oem-image-setup#steps)
> 
> Download and unzip following files:
> 
> * (...)
> * Remote Unit: [oem_ru_base.img](https://herelinkfw.cubepilot.org/tools/oem_ru_base.img)

Parsing this sparse file based on the `.ksy` spec fails on the very first `imageHeader` entry with an error:

```
RangeError: Invalid typed array length: -11
```

After doing a temporary modification:

```diff
--- 1/android_sparse@old.ksy
+++ 2/android_sparse@new.ksy
@@ -24,7 +24,8 @@ seq:
   - id: img_header_entries
     type: image_header_entry
     repeat: expr
-    repeat-expr: img_header.total_chunks
+    repeat-expr: 1
+    # repeat-expr: img_header.total_chunks
 types:
   header:
      seq:
@@ -59,8 +60,8 @@ types:
     seq:
       - id: header
         type: fixed_header
-      - id: body
-        size: header.total_size - header._sizeof
+      # - id: body
+      #   size: header.total_size - header._sizeof
     types:
       fixed_header:
         seq:
```

the following structure appears:

```js
{
  "imgHeader": {
    "magic": [58, 255, 38, 237],
    "version": {
      "major": 1,
      "minor": 0
    },
    "fileHeaderSize": 32,
    "chunkHeaderSize": 16,
    "blockSize": 4096,
    "totalBlocks": 76800,
    "totalChunks": 26,
    "checksum": 0
  },
  "imgHeaderEntries": [
    {
      "header": {
        "chunkType": 0,
        "reserved": 0,
        "chunkSize": 0xcac1,
        "totalSize": 1
      }
    }
  ]
}
```

So the parsing errorred on `totalSize` being 1, which makes `header.total_size - header._sizeof` negative (`header.total_size - header._sizeof = 1 - 12 = -11`). And `chunkType` and `chunkSize` have clearly swapped places 🤷

The `simg2img` has no problem with this file though, so it must be readable somehow. I will investigate further.

Anyway, I think this is a great example why we need sample files, and not only from one source or generator, but ideally diverse from many different sources. Implementing a format in a `.ksy` file according to some existing parser or specification is one thing, but whether it actually is capable of parsing real-life files is a totally different thing, and there is no other way to find out than to collect a variety of samples and test them. Not always are specifications compatible with how the format is actually used in the wild. Moreover, omissions and errors are common among both format descriptions and parsers/generators.

armijnhemel: I used simg2img and then img2simg and getting a significantly smaller result, which then is parsed correctly. I looked at the two, and also ran simg_dump.py ( https://github.com/anestisb/android-simg2img/blob/master/simg_dump.py ) which complains about the file:

```
$ python2 simg_dump.py oem_ru_base.img
simg_dump.py: oem_ru_base.img: The file header size was expected to be 28, but is 32.
```

and indeed, the header size (`file_header_size`) is 32, whereas in "normal" sparse image files it is supposed to be 28.

armijnhemel: Added four padding bytes to fill the header to 32 and the file is parsed correctly. So the problem is `file_header_size`. I can see a few solutions:

1. adding padding bytes depending on the size of the header
2. splitting the header in two parts
3. making `image_header_entries` an instance starting at pos `file_header_size` (probably after some sanity checks)

My personal preference would be 1.

generalmimon (approved):
Thanks!

PR #461, opened 2021-03-29
add GRUB2 font description

armijnhemel:
This description adds support for the GRUB2 font file format. This does not yet process the character definitions itself.

### DIFF
--- a/font/grub2font.ksy
+++ b/font/grub2font.ksy
@@ -5,6 +5,7 @@ meta:
   tags:
     - font
   license: CC0-1.0
+  encoding: ASCII
   endian: be
 doc: |
   Bitmap font format for the GRUB 2 bootloader.
@@ -28,11 +29,10 @@ types:
        - id: section_name
          size: 4
          type: str
-         encoding: ASCII
-       - id: section_length
+       - id: len_section
          type: u4
        - id: body
-         size: section_length
+         size: len_section
          type:
            switch-on: section_name
            cases:
@@ -51,22 +51,18 @@ types:
     seq:
       - id: name
         type: strz
-        encoding: ASCII
   font_family_name:
     seq:
       - id: name
         type: strz
-        encoding: ASCII
   font_weight:
     seq:
       - id: name
         type: strz
-        encoding: ASCII
   font_slant:
     seq:
       - id: name
         type: strz
-        encoding: ASCII
   font_point_size:
     seq:
       - id: point_size
@@ -92,7 +88,7 @@ types:
       - id: entries
         type: character
         repeat: expr
-        repeat-expr: _parent.section_length/sizeof<character>
+        repeat-expr: _parent.len_section/sizeof<character>
     types:
       character:
         seq:
@@ -101,12 +97,12 @@ types:
             doc: Unicode code point
           - id: flags
             type: u1
-          - id: offset
+          - id: ofs_character
             type: u4
         instances:
           bitmap:
             io: _root._io
-            pos: offset
+            pos: ofs_character
             type: character_definition
       character_definition:
         seq:
--- a/font/grub2font.ksy
+++ b/font/grub2font.ksy
@@ -17,6 +17,11 @@ seq:
     type: font_section
     repeat: until
     repeat-until: _.section_name == "DATA"
+    doc: |
+      The "DATA" section acts as a terminator. The documentation says:
+      "A marker that indicates the remainder of the file is data accessed
+      via the character index (CHIX) section. When reading this font file,
+      the rest of the file can be ignored when scanning the sections."
 types:
   font_section:
      seq:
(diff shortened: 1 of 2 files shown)


font/grub2font.ksy
```diff
@@ -0,0 +1,100 @@
...
+  Bitmap font format for the GRUB 2 bootloader.
+doc-ref: http://grub.gibibit.com/New_font_format
+seq:
+  - id: font_header
+    contents: ['FILE', 0, 0, 0, 4, 'PFF2']
+    size: 12
+  - id: font_sections
+    type: font_section
+    repeat: until
+    repeat-until: _.section_name == "DATA"
```
KOLANICH, line 19: Why and how is `data` so special ?

font/grub2font.ksy
```diff
@@ -0,0 +1,100 @@
...
+      - id: descent
+        type: u2
+  character_index:
+    seq:
+      - id: character_entries
+        type: character_entry
+        repeat: expr
+        repeat-expr: _parent.section_length/9
+    types:
+      character_entry:
```
KOLANICH, line 92: too long, maybe `char`?

font/grub2font.ksy
```diff
@@ -0,0 +1,100 @@
...
+        seq:
+          - id: code_point
+            type: u4
+            doc: Unicode code point
+          - id: flags
+            type: u1
```
KOLANICH, line 98: should it be a bit-fields type?

font/grub2font.ksy
```diff
@@ -0,0 +1,100 @@
...
+             '"MAXW"': maximum_character_width
+             '"MAXH"': maximum_character_height
+             '"ASCE"': ascent_in_pixels
+             '"DESC"': descent_in_pixels
+             '"CHIX"': character_index
+         if: section_name != "DATA"
+  font_name:
+    seq:
+      - id: name
+        type: strz
```
KOLANICH, line 48: Allows garbage after the end of a string. Should we parse it too?

font/grub2font.ksy
armijnhemel: It acts as a terminator. From the documentation:

"A marker that indicates the remainder of the file is data accessed via the character index (CHIX) section. When reading this font file, the rest of the file can be ignored when scanning the sections. "

KOLANICH: Can some words be added into the `doc` of that enum value?

font/grub2font.ksy
armijnhemel: So far I have only focused on the structure (I have no need for the rest of the information).

font/grub2font.ksy
armijnhemel: I would prefer "character" as that would be more in line with the documentation.

font/grub2font.ksy
generalmimon: @KOLANICH You know my attitude to this: https://github.com/kaitai-io/kaitai_struct_formats/pull/381#discussion_r554371635

font/grub2font.ksy
armijnhemel: I added it near the repeat as I think it makes more sense there, but happy to move it.

font/grub2font.ksy
```diff
@@ -103,3 +103,22 @@ types:
...
+          - id: height
+            type: u2
+          - id: x_offset
+            type: u2
+          - id: y_offset
+            type: u2
+          - id: device_width
+            type: u2
+          - id: bitmap_data
+            size: (width * height + 7) / 8
```
generalmimon, line 124:
```suggestion
            size: (width * height + 7) / 8 # ceiled integer division
```

I'd add this comment that might save time and reduce confusion when stumbling across this magical calulcation and trying to figure out what you meant by this.

Also, this means that every pixel always takes up 1 bit, right? What is the color palette, then? Is it `0 = black` and `1 = white`, or `0 = primary color` (it's a font, so it's probably meant to be dyed an arbitrary color I suppose) and `1 = transparent` or vice versa? Write in the `doc` key, please.

In addition to the comment `# ceiled integer division`, it would make sense to mention in `doc` what this ceiling does in practice: as I understand it, `bitmap_data` is an actually a [bit array](https://en.wikipedia.org/wiki/Bit_array) of `width * height` bits whose end position is aligned to the closest next byte boundary. I think it's beneficial to acquaint the spec readers/users with this concept, because it's an integral trait of the format.

armijnhemel: The documentation doesn't say anything about colour. The source code says each pixel is 1 bit.

font/grub2font.ksy
armijnhemel: I have now worked through the rest of the format. The "flags" are irrelevant: compressed fonts have never been implemented.

font/grub2font.ksy
armijnhemel: Regarding 2 & 3: I wrote something, not sure if it makes sense to you.

generalmimon (approved):

PR #462, opened 2021-03-30
add AU (Sun/NeXT audio)

armijnhemel:
a simple parser for the AU audio format. The contents of the streams themselves are not parsed.

### DIFF
--- a/media/au.ksy
+++ b/media/au.ksy
@@ -5,8 +5,7 @@ meta:
     - au
     - snd
   xref:
-    justsolve:
-      - AU
+    justsolve: AU
     mime:
       - audio/basic
     pronom:
@@ -26,6 +25,7 @@ seq:
     contents: ".snd"
   - id: header_size
     type: u4
+    min: 24
   - id: data_size
     type: u4
   - id: encoding
--- a/media/au.ksy
+++ b/media/au.ksy
@@ -45,11 +45,18 @@ seq:
   - id: header
     type: header
     size: len_header - magic._sizeof - len_header._sizeof
+instances:
+  len_data:
+    value: 'header.data_size == 0xffff_ffff ? _io.size - len_header : header.data_size'
 types:
   header:
     seq:
       - id: data_size
         type: u4
+        doc: |
+          don't read this field, access `_root.len_data` instead
+
+          value `0xffff_ffff` means unspecified size
       - id: encoding
         type: u4
         enum: encodings
--- a/media/au.ksy
+++ b/media/au.ksy
@@ -25,7 +25,8 @@ seq:
     contents: ".snd"
   - id: header_size
     type: u4
-    min: 24
+    valid:
+      min: 24
   - id: data_size
     type: u4
   - id: encoding
--- a/media/au.ksy
+++ b/media/au.ksy
@@ -13,13 +13,14 @@ meta:
   ks-version: 0.9
   endian: be
 doc: |
-  Test files for AU can be found in the Python Git repository:
+  The NeXT/Sun audio file format.
 
-  https://github.com/python/cpython/tree/b8a7daf077dab18e9e3701c5380b542ae0aa9a94/Lib/test/sndhdrdata
+  Sample files:
 
-  as well as:
+  * <https://github.com/python/cpython/tree/b8a7daf077da/Lib/test/sndhdrdata>
+  * <ftp://ftp-ccrma.stanford.edu/pub/Lisp/sf.tar.gz>
+  * <http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/AU/Samples.html>
 
-  ftp://ftp-ccrma.stanford.edu/pub/Lisp/sf.tar.gz
 doc-ref:
   # - https://en.wikipedia.org/wiki/Au_file_format # incorrect encoding enum, don't use
   - http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/AU/AU.html
--- a/media/au.ksy
+++ b/media/au.ksy
@@ -43,7 +43,7 @@ types:
         type: u4
       - id: num_channels
         type: u4
-      - id: description
+      - id: comment
         size-eos: true
         type: strz
         encoding: ASCII
--- a/media/au.ksy
+++ b/media/au.ksy
@@ -16,27 +16,41 @@ meta:
   ks-version: 0.9
   endian: be
 doc: |
-  Test files for AU can be found in any Python release in the
-  directory Lib/test/sndhdrdata/
+  Test files for AU can be found in the Python Git repository:
+
+  https://github.com/python/cpython/tree/b8a7daf077dab18e9e3701c5380b542ae0aa9a94/Lib/test/sndhdrdata
+
+  as well as:
+
+  ftp://ftp-ccrma.stanford.edu/pub/Lisp/sf.tar.gz
 doc-ref:
   - https://en.wikipedia.org/wiki/Au_file_format
   - https://web.archive.org/web/20121028010008/http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/AU/AU.html
 seq:
-  - id: magic
-    contents: ".snd"
-  - id: header_size
-    type: u4
-    valid:
-      min: 24
-  - id: data_size
-    type: u4
-  - id: encoding
-    type: u4
-    enum: encodings
-  - id: sample_rate
-    type: u4
-  - id: channels
-    type: u4
+  - id: header
+    type: header
+  - id: description
+    type: strz
+    encoding: ASCII
+    size: header.header_size - sizeof<header>
+types:
+  header:
+    seq:
+      - id: magic
+        contents: ".snd"
+      - id: header_size
+        type: u4
+        valid:
+          min: 24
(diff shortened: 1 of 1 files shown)


media/au.ksy
```diff
@@ -0,0 +1,61 @@
+meta:
+  id: au
+  title: AU audio format
+  file-extension:
+    - au
+    - snd
+  xref:
+    justsolve:
+      - AU
```
KOLANICH, line 9: For a single item the list is unneeded

media/au.ksy
```diff
@@ -0,0 +1,61 @@
...
+    pronom:
+      - x-fmt/139
+    wikidata:
+      - Q672985
+  license: CC0-1.0
+  ks-version: 0.9
+  endian: be
+doc: |
+  Test files for AU can be found in any Python release in the
+  directory Lib/test/sndhdrdata/
```
KOLANICH, line 21: 1. a direct link should be put
2. Gonna be removed: https://www.python.org/dev/peps/pep-0594/#sunau . Tests should likely be moved to https://github.com/tiran/legacylib/tree/master/sunau tool but it is not done yet

media/au.ksy
```diff
@@ -0,0 +1,61 @@
...
+doc-ref: https://en.wikipedia.org/wiki/Au_file_format
+doc-ref: https://web.archive.org/web/20121028010008/http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/AU/AU.html
+seq:
+  - id: magic
+    contents: ".snd"
+  - id: header_size
```
KOLANICH, line 27: I don't see it used in the spec.

armijnhemel: It is named differently in the different specs. In this one it is "hdr_size:

https://pubs.opengroup.org/external/auformat.html

In this one it is "offset to data":

https://web.archive.org/web/20121028010008/http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/AU/AU.html

because the audio data immediately follows the header.

KOLANICH: Immediately follows the header of the specific size. The header must be of that size then. It is done intentionally for forward compatibility.

armijnhemel: Very likely. In practice the header always seems to have been 24 though and the format is no longer in widespread use. One check that could be added is that the value of that field is at least 24.

KOLANICH: >One check that could be added is that the value of that field is at least 24.

It violates the goal of forward comatibility.

KOLANICH: You are doing it wrong.

```yaml
types:
  preheader:
    seq:
       - id: magic
         contents: ".snd"
       - id: header_size
         type: u4
  header:
    seq:
      - id: preheadet
        type: preheader
     - id: rest
       type: rest_of_header
       size: preheader.header_size - sizeof<preheader>
```

armijnhemel: If there were lots of different variations of the format I would have agreed with you. But this format hasn't changed in over 30 years and has effectively been abandoned by the industry. There has been NO file that has anything else in that header except the value 24. So why try to make it work for variants that don't exist and that won't exist?

But you are right, I am doing it wrong. I should not say that the minimum value is 24, but that the value equals 24 :-)

armijnhemel: > If there were lots of different variations of the format I would have agreed with you. But this format hasn't changed in over 30 years and has effectively been abandoned by the industry. There has been NO file that has anything else in that header except the value 24. So why try to make it work for variants that don't exist and that won't exist?
> 
> But you are right, I am doing it wrong. I should not say that the minimum value is 24, but that the value equals 24 :-)

I am taking this back. After some more searching I found a header with size 40. I will change it as suggested. Thank you!

armijnhemel: I went through the specification again. Everything that is part of the header but after byte 24 is part of a description or textual information (ASCII encoded). I have reworked the header to reflect this. I have kept the "min: 24" statement as it is a hard restriction that directly follows from the specification. It absolutely makes no sense to make it more generic.

media/au.ksy
```diff
@@ -0,0 +1,77 @@
...
+
+  ftp://ftp-ccrma.stanford.edu/pub/Lisp/sf.tar.gz
+doc-ref:
+  - https://en.wikipedia.org/wiki/Au_file_format
+  - https://web.archive.org/web/20121028010008/http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/AU/AU.html
+seq:
+  - id: header
+    type: header
+  - id: description
+    type: strz
```
KOLANICH, line 33: It is incorrect, anything can be here.

KOLANICH: Also, `header_size` means it must be a part of a header, not a field after it.

media/au.ksy
armijnhemel: As said, I found other files with a different header, so this has been fixed.

media/au.ksy
armijnhemel: The name "header_size" is directly copied from the specifications (there it is "hdr_size") and it actually is part of the header, so I have no idea what you are talking about.

KOLANICH: I have already given you the snippet about how IMHO it should look like:

* `header_size` is within the header, but still determines `header` size.
* `sizeof<>` is used, no hardcoded handcoded numbers

armijnhemel: > It is incorrect, anything can be here.

Although the documentation says this:

```
Further, there are no absolute guarantees that the 'info' is ASCII text, (non-ASCII info may eventually be far more useful anyway).
```

but the SunOS manpage that is being referred to ( audio_filehdr.3 ) specifically mentions a "char *infop" argument to the functions and says:

```
The infop argument, if not NULL, specifies the address of  a
buffer, whose size is given by ilen, into which the informa-
tion field of the audio  file  is  copied.   The  buffer  is
always  returned  NULL-terminated,  even if it was not large
enough to hold the entire file information field.
```

So even though they might have envisioned something else than ASCII this apparently wasn't implemented.

armijnhemel: > I have already given you the snippet about how IMHO it should look like:
> 
>     * `header_size` is within the header, but still determines `header` size.
> 
>     * `sizeof<>` is used, no hardcoded handcoded numbers
> 
>     * header can be extended by anyone. If you don't know how to parse the extension, it is safely ignored, otherwise you can add code into the end of header parsing the stuff using `if`, but it would require some info passed, either off-band (file extension) or maybe by modifying signature.

I don't think you understand what I did. The only hardcoded value is that the minimum value of header_size should be 24 because that is what directly follows from the specification. Any file that has a value lower than 24 in "header_size" should be rejected. Please read the AU specification.

KOLANICH: I understand what you did. But`header_size` says that it is a size of a header, not a size of a header with something else after it. So a field of `header` type should occupy that size. You can add something after it, but to have there a `strz` with ASCII you should know ahead of time it is there, something else can be there, the format is forward compatible, it means third-party devs could have implemented own extensions before the spec mentioned ASCII, the format using ASCII text there is only one case of a valid au file.

>So even though they might have envisioned something else than ASCII this apparently wasn't implemented.

>The variable-length annotation field is currently ignored by most audio applications. It is present for interchange compatibility.  In the future, applications may use this field to store structured data annotation, such as markers and edit points.

armijnhemel: So if I rename it from header_size to "data_offset" it should be fine then :-P

The specifications that I found are actually conflicting. One says it is the header size, the other says that it is the size of the data and that there can be data between the header (24 bytes) and the start of the audio data. The original header file calls it "hdr_size". I'd be more than happy to rename it. Right now we are really splitting hairs.

media/au.ksy
```diff
@@ -0,0 +1,63 @@
...
+    - au
+    - snd
+  xref:
+    justsolve: AU
+    mime:
+      - audio/basic
+    pronom:
+      - x-fmt/139
+    wikidata:
+      - Q672985
```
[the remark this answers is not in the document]
armijnhemel: whoops, overlooked that. Fixed.

media/au.ksy
generalmimon: I agree with @KOLANICH. In principle, I'm not a fan of the `valid/min` check that merely emulates the standard EOF checks done automatically by the Kaitai Struct runtime libraries:

https://github.com/kaitai-io/kaitai_struct_formats/blob/98f38afdde472b97aad9f79bf8b3b16e4fda2aa9/media/au.ksy#L38-L41

I think it's better to put the header into a substream with the length of `header_size` bytes. If `header_size < 24`, the parsing will just fail with an `EOFError` pointing to the first element that couldn't fit into the `header_size`-byte substream.

https://github.com/kaitai-io/kaitai_struct_formats/blob/98f38afdde472b97aad9f79bf8b3b16e4fda2aa9/media/au.ksy#L34-L50

generalmimon:
@armijnhemel I've done some modifications, can you please check them out? Otherwise I think it's OK to merge.

armijnhemel:
> @armijnhemel I've done some modifications, can you please check them out? Otherwise I think it's OK to merge.

It looks good to me!

generalmimon (approved):
Thanks!

PR #463, opened 2021-03-31
wav: fix spelling mistake, process more chunks

armijnhemel:
This PR fixes a simple spelling mistakes and adds chunk names that are not in the standard spec, namely ID3 and PEAK. The data of these chunks is not yet processed any further.

Test files for ID3: https://github.com/alexa/alexa-auto-sdk/tree/3.1/samples/cpp/assets/inputs
Test files for PEAK: https://github.com/scipy/scipy/tree/master/scipy/io/tests/data

### DIFF
--- a/media/wav.ksy
+++ b/media/wav.ksy
@@ -137,6 +137,7 @@ types:
         type: strz
 
   bext_chunk_type:
+    doc-ref: https://en.wikipedia.org/wiki/Broadcast_Wave_Format
     seq:
       - id: description
         size: 256
@@ -171,22 +172,21 @@ types:
         type: u2
       - id: max_short_term_loudness
         type: u2
-    doc-ref: https://en.wikipedia.org/wiki/Broadcast_Wave_Format
 
   axml_chunk_type:
+    doc-ref: https://tech.ebu.ch/docs/tech/tech3285s5.pdf
     seq:
       - id: body
         type: str
         size-eos: true
         encoding: UTF-8
-    doc-ref: https://tech.ebu.ch/docs/tech/tech3285s5.pdf
 
   ixml_chunk_type:
+    doc-ref: https://en.wikipedia.org/wiki/IXML
     seq:
       - id: body
         type: strz
         encoding: UTF-8
-    doc-ref: https://en.wikipedia.org/wiki/IXML
 
   cue_chunk_type:
     seq:
@@ -346,13 +346,13 @@ types:
         doc: chunk containing XMP data
         doc-ref: https://wwwimages2.adobe.com/content/dam/acom/en/devnet/xmp/pdfs/XMP%20SDK%20Release%20cc-2016-08/XMPSpecificationPart3.pdf
   afsp_chunk_type:
+    doc-ref: http://www-mmsp.ece.mcgill.ca/Documents/Downloads/AFsp/
     seq:
       - id: afspid
         contents: "AFsp"
       - id: text
         type: strz
         encoding: UTF-8
-    doc-ref: http://www-mmsp.ece.mcgill.ca/Documents/Downloads/AFsp/
 
 enums:
   w_format_tag_type:
--- a/media/wav.ksy
+++ b/media/wav.ksy
@@ -219,9 +219,9 @@ types:
 
   fact_chunk_type:
     seq:
-      - id: sample_length
+      - id: num_samples_per_channel
+        -orig-id: dwSampleLength
         type: u4
-        doc: Number of samples per channel
         doc-ref: http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/WAVE.html
 
   format_chunk_type:
--- a/media/wav.ksy
+++ b/media/wav.ksy
@@ -2,8 +2,8 @@ meta:
   id: wav
   title: Microsoft WAVE audio file
   file-extension:
-    - bwf
     - wav
+    - bwf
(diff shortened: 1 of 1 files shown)


armijnhemel:
test file for _pmx and fact chunks: https://github.com/varunchitre15/MT6589_kernel_source/tree/master/mediatek/custom/arima89_we_s_jb2/factory/res/sound

armijnhemel:
> test file for _pmx and fact chunks: https://github.com/varunchitre15/MT6589_kernel_source/tree/master/mediatek/custom/arima89_we_s_jb2/factory/res/sound

Specification for _PMX: https://wwwimages2.adobe.com/content/dam/acom/en/devnet/xmp/pdfs/XMP%20SDK%20Release%20cc-2016-08/XMPSpecificationPart3.pdf (page 24)

armijnhemel:
Should chunks that are generic for RIFF go into riff.ksy? Will they automatically be picked up by wav.ksy and available? I am confused as there is some duplication between riff.ksy and wav.ksy (info and list chunks).

KOLANICH:
They should, IMHO. IMHO we currently cannot have custom ksys for each riff-based format when deduplicating riff to an own file. We need one big `riff` and [the way to disable unneeded types compilation](https://github.com/kaitai-io/kaitai_struct/issues/339) and [templates](https://github.com/kaitai-io/kaitai_struct/issues/135) to be able to record such parameterization into dedicated files.=

armijnhemel:
There are lots of other chunks out there from different specifications. I have not been able to obtain test files for those yet. There are a few chunks that could possibly benefit from more parsing (such as the "chna" chunk and "id3" chunk) but currently I have no time to dig deep into the specifications. I have left references to specifications and/or test files if anyone feels like adding these.

media/wav.ksy
```diff
@@ -26,11 +45,19 @@ doc: |
...
+
+  https://www.bbc.co.uk/rd/publications/saqas
 
   This Kaitai implementation was written by John Byrd of Gigantic Software
   (jbyrd@giganticsoftware.com), and it is likely to contain bugs.
-doc-ref: http://soundfile.sapp.org/doc/WaveFormat/
+doc-ref:
+  - http://soundfile.sapp.org/doc/WaveFormat/
+  - https://web.archive.org/web/20101031101749/http://www.ebu.ch/fr/technical/publications/userguides/bwf_user_guide.php
+  - https://www.itu.int/rec/R-REC-BS.2076-2-201910-I/en
```
generalmimon, line 60: @armijnhemel This only describes the `chna` chunk which is not even described yet, right? I would move it to the particular type (or enum key if there is no type yet) that it describes, because currently it's confusing - if it's in the top-level `/doc-ref` without any further description, I'd expect that it documents the overall WAV/BWF format and/or many of its chunks, not just a single one.

media/wav.ksy
```diff
@@ -556,10 +628,21 @@ enums:
...
-    0x74786562: bext
     0x20657563: cue
     0x61746164: data
     0x64696d75: umid
     0x666e696d: minf
     0x6e676572: regn
+    0x20336469: id3
+    0x4b414550: peak
+    0x584d505f: pmx
+    0x4c4d5869: ixml
```
generalmimon, line 641: This also looks like a "BWF chunk" to me - from http://www.gallery.co.uk/ixml/:

> iXML is primarily designed to be used as a [RIFF](http://www.gallery.co.uk/ixml/glossary.html) (embedded tagged data) chunk inside a **Broadcast Wave file** (although it can be optionally included in other file types), to **supercede** the metadata currently written in the **standard Broadcast Wave 'bext' chunk** description field in a non-standardised way by several manufacturers.

armijnhemel: Yes, BWF (and related) are based on .wav. So the question is: should this be in the .wav format, or in a separate format that will be largely identical? If there would be some sort of inheritance mechanism in KS it would be a lot simpler.

media/wav.ksy
armijnhemel: It is described in chapter 7 of https://www.itu.int/dms_pubrec/itu-r/rec/bs/R-REC-BS.2076-2-201910-I!!PDF-E.pdf

Moving it to the enum makes perfect sense.

media/wav.ksy
generalmimon: > So the question is: should this be in the .wav format, or in a separate format that will be largely identical? If there would be some sort of inheritance mechanism in KS it would be a lot simpler.

I think it should stay in the .wav format, that's completely OK. This is almost the same thing as PNG/APNG - https://github.com/kaitai-io/kaitai_struct_formats/blob/5f3ce8dd442557996d6a7370526ea1fe00848c90/image/png.ksy#L5-L6. APNG still has the exact same structure as PNG, as well as BWF has the same structure as WAV, so keeping them in the same spec makes sense to me.

generalmimon: Instead, I had in mind that if `ixml` is rather a BWF chunk, it should be below this comment

https://github.com/kaitai-io/kaitai_struct_formats/blob/48e78529e6034c2c2e497aa027d92b747613197e/media/wav.ksy#L642-L644

which attempts to separate the BWF chunks from the "pure WAV" ones, as far as I understand.

armijnhemel: yes, makes sense

generalmimon (approved):
Thanks for contributing!

PR #465, opened 2021-04-02
add Chrome PAK specification

armijnhemel:
A fairly simple specification for the Chrome PAK versions 4 and 5, as frequently found on Android systems from Google. This parses the structure of the header and all of the offsets, as well as aliases (version 5 only, as they don't exist in version 4).

At the end of the resource table there is one element of which the id is always 0, and this acts as a terminator. I have given it its own data type, but it could also be moved in the resource table itself, if this is deemed to be cleaner.

### DIFF
--- a/serialization/chrome_pak.ksy
+++ b/serialization/chrome_pak.ksy
@@ -1,0 +1,90 @@
+meta:
+  id: chrome_pak
+  title: Chrome PAK serialization format
+  file-extension: pak
+  license: CC0-1.0
+  endian: le
+doc: |
+  Format mostly used by Google Chrome and various Android apps to store
+  resources such as translated strings, help messages and images.
+doc-ref:
+  - http://dev.chromium.org/developers/design-documents/linuxresourcesandlocalizedstrings # version 4
+  - https://chromium.googlesource.com/chromium/src/tools/grit/+/22f7a68bb5ad68fe4192d0f34466049038735b9c/grit/format/data_pack.py # version 4
+  - https://chromium.googlesource.com/chromium/src/tools/grit/+/c1a76d6143016e9b2c292236bab623d1497cd31f/grit/format/data_pack.py # version 5
+seq:
+  - id: version
+    type: u4
+    doc: Support versions are 4 and 5
+  - id: header
+    type:
+      switch-on: version
+      cases:
+        4: header4
+        5: header5
+types:
+  header4:
+    seq:
+      - id: number_of_resources
+        type: u4
+      - id: encoding
+        type: u1
+        enum: encoding
+      - id: resources
+        type: resource
+        repeat: expr
+        repeat-expr: number_of_resources
+      - id: end_of_resource
+        type: end_of_resource
+        doc: |
+          The length is calculated by looking at the offset of
+          the next item, so an extra entry is stored with id 0
+          and offset pointing to the end of the resources.
+  header5:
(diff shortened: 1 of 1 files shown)


serialization/chrome_pak.ksy
```diff
@@ -0,0 +1,90 @@
...
+        type: u1
+        enum: encoding
+      - id: ignore
+        size: 3
+        doc: These three bytes have no meaning in the file
+      - id: number_of_resources
+        type: u2
+      - id: number_of_aliases
+        type: u2
+      - id: resources
```
KOLANICH, line 54: Duplication

serialization/chrome_pak.ksy
```diff
@@ -0,0 +1,90 @@
...
+        type: u2
+      - id: index
+        type: u2
+  resource:
+    seq:
+      - id: id
+        type: u2
+      - id: offset
+        type: u4
+  end_of_resource:
```
KOLANICH, line 80:
```yaml
seq:
  - id: end_resource
    type: resource
    valid:
      expr: _.id == 0
```

serialization/chrome_pak.ksy
armijnhemel: What is duplicated?

serialization/chrome_pak.ksy
armijnhemel: That is probably better. Thanks!

KOLANICH: No, it is not. Just the array should have `number_of_resources + 1` items. And the actual array of resources should have `number_of_resources` items, then you can safely use `[_index + 1]` to calculate length.

serialization/chrome_pak.ksy
KOLANICH: `number_of_resources` `resources` `end_of_resource` in both types. It is not how this is supposed to be specified. Instead there should be a single type, with `if`s conditioned on version.

serialization/chrome_pak.ksy
armijnhemel: Luckily I haven't changed it yet. As said, making the array bigger makes more sense, but I want to make sure that the index of the last is 0.

serialization/chrome_pak.ksy
armijnhemel: I don't really see the issue. I want to remove 'end_of_resource' anyway (as discussed) and then it is really just two very simple headers that have the various elements in a different order. Using "ifs" everywhere doesn't really help with readability either. Abstraction is good, but there is a thing as too much abstraction.

serialization/chrome_pak.ksy
KOLANICH: Infortunately we don't have `valid` for `value` `instance`s. But the check still can be done in `valid` in some other place than the field itself. So you need to validate that the last element id is 0.

serialization/chrome_pak.ksy
KOLANICH: I see the issue. Having 2 different formats for headers would cause unnecessary issues for strongly typed languages and even for non-strongly-typed ones and would be not enough clearly carrying the information how the format has been changed when the version has changed from 3 to 5.

armijnhemel: I was actually talking about it being more readable for humans for maintainability.

PR #467, opened 2021-04-03
add flattened device tree/devicetree blobs specification

armijnhemel:
This is a specification for DTB (devicetree blobs/flattened device tree), the compiled result of .dts files as found in the Linux kernel. Test files can be found here:

https://github.com/qemu/qemu/tree/master/pc-bios

Note: this is NOT the same as the dtb/dtbo image format that was pulled here: https://github.com/kaitai-io/kaitai_struct_formats/pull/415


armijnhemel:
Note: this specification checks the structure of the dtb, but does not process the individual properties (as there are lots of them).

armijnhemel:
Also, I have added this to the directory "database", but that doesn't seem to be the right place.

database/dtb.ksy
```diff
@@ -0,0 +1,94 @@
...
+  title: Flattened Devicetree Format
+  file-extension: dtb
+  license: CC0-1.0
+  endian: be
+doc: |
+  Also referred to as Devicetree Blob (DTB). It is a flat
+  binary encoding of data (primarily devicetree data, although
+  other data is possible as well).
+doc-ref:
+  - https://github.com/devicetree-org/devicetree-specification/releases/download/v0.3/devicetree-specification-v0.3.pdf
```
KOLANICH, line 12: I guess https://github.com/devicetree-org/devicetree-specification/blob/master/source/chapter5-flattened-format.rst makes more sense.

database/dtb.ksy
```diff
@@ -0,0 +1,94 @@
...
+  - https://elinux.org/images/f/f4/Elc2013_Fernandes.pdf
+seq:
+  - id: magic
```
KOLANICH, line 15: Please add `-orig-id`s everywhere.

database/dtb.ksy
```diff
@@ -0,0 +1,94 @@
...
+  strings_block:
+    pos: strings_block_offset
+    size: size_dt_strings
+    type: strings
+types:
+  strings:
+    seq:
+      - id: strings
+        type: strz
+        encoding: ASCII
```
KOLANICH, line 52: citation needed. There is no sense assume anything except utf-8 in 2021.

database/dtb.ksy
```diff
@@ -0,0 +1,94 @@
...
+    type: u4
+  - id: memory_reservation_block_offset
+    type: u4
+  - id: version
+    type: u4
+  - id: last_compatible_version
+    type: u4
+  - id: boot_cpuid_phys
+    type: u4
+  - id: size_dt_strings
```
KOLANICH, line 31: the `id`s should be renamed: they go to the generated parser. Information about original ids should be preserved using `-orig-id`s.

KOLANICH:
>but that doesn't seem to be the right place.

We have a directory `firmware`, but we should really to invent a better structure for the repo (with more formal criteria what goes where which can be used to predict the location of a spec with onpy basic knowledge agout the format it implements). I guess `serialization` will be better for this format.

database/dtb.ksy
armijnhemel: https://github.com/devicetree-org/devicetree-specification/blob/master/source/chapter2-devicetree-basics.rst

It is actually stricter than ASCII:
Valid characters for node names 

Character | Description
-- | --
0-9 | digit
a-z | lowercase letter
A-Z | uppercase letter
, | comma
. | period
_ | underscore
+ | plus sign
- | dash

database/dtb.ksy
armijnhemel: I will add this. The PDF has all data in one document, which can be convenient for some.

database/dtb.ksy
KOLANICH: Thanks. Can this be added into the spec?

database/dtb.ksy
armijnhemel: I renamed some to give them more useful names, some I have kept as I don't have a better name for them.

database/dtb.ksy
```diff
@@ -2,39 +2,55 @@ meta:
...
   license: CC0-1.0
   endian: be
 doc: |
   Also referred to as Devicetree Blob (DTB). It is a flat
   binary encoding of data (primarily devicetree data, although
   other data is possible as well).
+
+  The encoding of strings used in the specification is actually a subset of ASCII:
+
+  https://github.com/devicetree-org/devicetree-specification/blob/master/source/chapter2-devicetree-basics.rst
```
KOLANICH, line 16: And maybe put a link to examples here too and also the info that it may be available as `/sys/firmware/fdt` on OpenWRT systems.

KOLANICH: Here are some from Keenetic routers: [keenetic.zip](https://github.com/kaitai-io/kaitai_struct_formats/files/6252867/keenetic.zip)

armijnhemel:
> Note: this specification checks the structure of the dtb, but does not process the individual properties (as there are lots of them).

This specification also treats the tokens in the structure block as a list of tokens. It would be more accurate to have it as a tree.

database/dtb.ksy
```diff
@@ -0,0 +1,124 @@
+meta:
+  id: dtb
+  title: Flattened Devicetree Format
+  file-extension: dtb
```
KOLANICH, line 4:
```yaml
application:
  - Das U-Boot
  - Linux
```

armijnhemel: added, there are probably more that use it. We can add those later.

database/dtb.ksy
armijnhemel: > The encoding of the strings should likely be the docs to strings table, not to the whole doc.

Yes, you are correct. I have changed this.

database/dtb.ksy
```diff
@@ -0,0 +1,128 @@
...
+  application:
+    - Linux
+    - Das U-Boot
+  xref:
+    - wikidata: Q16960371
```
generalmimon, line 9: That's not right - the `meta/xref` must be an object (map), not a list. The compiler should at least warn about this...

```suggestion
  xref:
    wikidata: Q16960371
```

See the [KSY diagram](https://doc.kaitai.io/ksy_diagram.html) for reference.

database/dtb.ksy
```diff
@@ -0,0 +1,128 @@
...
+  endian: be
+doc: |
+  Also referred to as Devicetree Blob (DTB). It is a flat
+  binary encoding of data (primarily devicetree data, although
+  other data is possible as well).
+
+  On Linux systems that support this the blobs can be accessed in
+  /sys/firmware/fdt
+
+  https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-firmware-ofw
```
generalmimon, line 20: Is this stray URL related to the above or below paragraph? Can you make it clear?

database/dtb.ksy
```diff
@@ -0,0 +1,128 @@
...
+  - https://github.com/devicetree-org/devicetree-specification/releases/tag/v0.3
+  - https://github.com/devicetree-org/devicetree-specification/blob/ba2aa679679fc4fedf67130f18a6f0ecc4cf0382/source/flattened-format.rst
+  - https://elinux.org/images/f/f4/Elc2013_Fernandes.pdf
+seq:
+  - id: magic
+    contents: [0xd0, 0x0d, 0xfe, 0xed]
+    -orig-id: magic
+  - id: total_size
+    type: u4
+    -orig-id: totalsize
```
generalmimon, line 40: The key `-orig-id` should go right below `id`, see https://doc.kaitai.io/ksy_style_guide.html#seq-attr:

> ## 5. Sequence attributes
>
> When specifying an attribute, one MUST use the following order of keys:
>
> 1. Identifier(s)
>    * `id`
>    * `-orig-id` — use to specify original ID spelling if transcribing a structure from existing software and/or official spec
> 1. `size`
> 1. `size-eos`
> 1. (...)

database/dtb.ksy
```diff
@@ -0,0 +1,128 @@
...
+  strings_block:
+    pos: strings_block_offset
+    size: strings_block_size
+    type: strings
+types:
+  strings:
+    seq:
+      - id: strings
+        type: strz
+        encoding: ASCII
```
generalmimon, line 84: Given that all strings in the spec use `ASCII` encoding, what about making it default in `/meta/encoding` instead of repeating it all around the file?

database/dtb.ksy
armijnhemel: The compiler didn't warn about it.

database/dtb.ksy
armijnhemel: I made it a bit clearer (I hope). Please verify.

database/dtb.ksy
generalmimon: > The compiler didn't warn about it.

I know. When I wrote

> The compiler should at least warn about this...

I meant that it would make sense to implement such a check in the future. We should issue _at least_ a warning (it's not a real obstacle to proper compilation currently, because the compiler does not make any use of `meta/xref` in the generated code anyway) or maybe a fatal error (but that smells like putting artificial obstacles in user's way, which is not great - warning is probably a much better option).

database/dtb.ksy
generalmimon: Yes, looks better.

database/dtb.ksy
```diff
@@ -0,0 +1,129 @@
...
+  actually a subset of ASCII:
+
+  https://github.com/devicetree-org/devicetree-specification/blob/master/source/chapter2-devicetree-basics.rst
+
+  Example files:
+
+  - https://github.com/qemu/qemu/tree/master/pc-bios
+doc-ref:
+  - https://github.com/devicetree-org/devicetree-specification/releases/tag/v0.3
+  - https://github.com/devicetree-org/devicetree-specification/blob/ba2aa679679fc4fedf67130f18a6f0ecc4cf0382/source/flattened-format.rst
```
generalmimon, line 33: Please shorten the commit hash to 7 characters, like

```suggestion
  - https://github.com/devicetree-org/devicetree-specification/blob/ba2aa67/source/flattened-format.rst
```

database/dtb.ksy
```diff
@@ -0,0 +1,129 @@
...
+    seq:
+      - id: fdt_nodes
+        type: fdt_node
+        repeat: until
+        repeat-until: _.token_type == fdt::end
+  fdt_begin_node:
+    seq:
+      - id: name
+        type: strz
+        encoding: ASCII
```
generalmimon, line 108: You have put the`encoding: ASCII` into top-level `/meta`, so you can remove it everywhere in the file:

```suggestion
```

There is also another occurrence above this one:

https://github.com/kaitai-io/kaitai_struct_formats/blob/0ea5632153a181f19769958407ff89b9e2cc3789/database/dtb.ksy#L85

armijnhemel: Yes, sorry, I forgot, currently juggling too many things at the same time.

database/dtb.ksy
```diff
@@ -0,0 +1,127 @@
...
+    type: u4
+  - id: structure_block_offset
+    -orig-id: off_dt_struct
+    type: u4
+  - id: strings_block_offset
+    -orig-id: off_dt_strings
+    type: u4
+  - id: memory_reservation_block_offset
+    -orig-id: off_mem_rsvmap
+    type: u4
```
generalmimon, line 50: According to the KSY Style Guide (https://doc.kaitai.io/ksy_style_guide.html#attr-id), these three fields and also this one:

https://github.com/kaitai-io/kaitai_struct_formats/blob/7fc9effd35ff7106f3e52ed1746e7a075068e200/database/dtb.ksy#L114-L116

should use the prefix `ofs_`:

> For fields that designate **offset** to some particular data structure, use `ofs_` prefix and name of that data structure (as it would appear in the file) — i.e. `ofs_block`, `ofs_queries`, `ofs_path`

so:

```shell
sed -i 's/structure_block_offset/ofs_structure_block/g' dtb.ksy
sed -i 's/strings_block_offset/ofs_strings_block/g' dtb.ksy
sed -i 's/memory_reservation_block_offset/ofs_memory_reservation_block/g' dtb.ksy
sed -i 's/name_offset/ofs_name/g' dtb.ksy
```

database/dtb.ksy
```diff
@@ -0,0 +1,127 @@
...
+  fdt_prop:
+    seq:
+      - id: length
+        -orig-id: len
+        type: u4
+      - id: ofs_name
+        -orig-id: nameoff
+        type: u4
+      - id: property
+        size: length
```
generalmimon, line 118: https://doc.kaitai.io/ksy_style_guide.html#attr-id:

> * For fields that designate **size** of some particular data structure (in bytes or some other fixed units), use `len_` prefix and name of that data structure — i.e. `len_block` (length of a single `block` entry), `len_blocks` (total length of whole `blocks` array, made of `block` entries).

```suggestion
      - id: len_property
        -orig-id: len
        type: u4
      - id: ofs_name
        -orig-id: nameoff
        type: u4
      - id: property
        size: len_property
```

armijnhemel: OK, I will change. I will wait for you to finish the complete review :-)

database/dtb.ksy
```diff
@@ -0,0 +1,127 @@
...
+      - id: boundary_padding
+        size: (- _io.pos) % 4
+  fdt_prop:
+    seq:
+      - id: length
+        -orig-id: len
+        type: u4
+      - id: ofs_name
+        -orig-id: nameoff
+        type: u4
```
generalmimon, line 116: You should use this property in an `instance` to provide access to the `name` string:

```ksy
instances:
  name:
    io: _root.strings_block._io
    pos: ofs_name
    type: strz
```

(please verify if I understood it correctly).

armijnhemel: Yes, this is correct.

database/dtb.ksy
```diff
@@ -0,0 +1,127 @@
...
+  - id: strings_block_size
+    -orig-id: size_dt_strings
+    type: u4
+  - id: structure_block_size
+    -orig-id: size_dt_struct
+    type: u4
+instances:
+  memory_reservation_block:
+    pos: ofs_memory_reservation_block
+    size: ofs_structure_block - ofs_memory_reservation_block
```
generalmimon, line 71: According to https://github.com/devicetree-org/devicetree-specification/releases/download/v0.3/devicetree-specification-v0.3.pdf (section **5.3 Memory Reservation Block > 5.3.2 Format**, page 47/56), this lacks inner structure:

> ### 5.3.2 Format
> 
> The memory reservation block consists of a list of pairs of 64-bit big-endian integers, each pair being represented by the following C structure.
> 
> ```c
> struct fdt_reserve_entry {
>     uint64_t address;
>     uint64_t size;
> };
> ```
> Each pair gives the physical address and size in bytes of a reserved memory region. These given regions shall not overlap each other. The list of reserved blocks shall be terminated with an entry where both address and size are equal to 0. Note that the address and size values are always 64-bit. On 32-bit CPUs the upper 32-bits of the value are ignored.
> 
> Each uint64_t in the memory reservation block, and thus the memory reservation block as a whole, shall be located at an 8-byte aligned offset from the beginning of the devicetree blob (see section 5.6).

database/dtb.ksy
```diff
@@ -0,0 +1,127 @@
...
+  structure_block:
+    pos: ofs_structure_block
+    size: ofs_strings_block - ofs_structure_block
```
generalmimon, line 74: Is there a reason why you don't use `len_structure_block` instead (counting on that the above suggestion `'s/structure_block_size/len_structure_block/g'` is already applied)?

database/dtb.ksy
```diff
@@ -0,0 +1,127 @@
...
+    seq:
+      - id: strings
+        type: strz
+        repeat: eos
+  fdt_node:
+    seq:
+      - id: token_type
+        type: u4
+        enum: fdt
+      - id: fdt_node_body
```
generalmimon, line 91: IMHO just `body` is enough - prepending the superior type name is unnecessary.

database/dtb.ksy
```diff
@@ -0,0 +1,127 @@
...
+    -orig-id: off_dt_struct
+    type: u4
+  - id: ofs_strings_block
+    -orig-id: off_dt_strings
+    type: u4
+  - id: ofs_memory_reservation_block
+    -orig-id: off_mem_rsvmap
+    type: u4
+  - id: version
+    -orig-id: version
```
generalmimon, line 52: No need for same `-orig-id` as `id` - I believe it can be omitted.

database/dtb.ksy
```diff
@@ -0,0 +1,127 @@
+meta:
+  id: dtb
+  title: Flattened Devicetree Format
+  file-extension: dtb
+  application:
+    - Linux
+    - Das U-Boot
+  xref:
+    wikidata: Q16960371
+  license: CC0-1.0
```
generalmimon, line 10: It is probably a good idea to add `ks-version: 0.9`, because the `doc-ref` key allows a list as of 0.9:

https://github.com/kaitai-io/kaitai_struct_formats/blob/c4551e5a3d5cafdbee23c885c78a795d42380397/database/dtb.ksy#L31-L34

The `/meta/ks-version` is supposed to check if the compiler version is sufficient, so that the user gets a clear error message

```
/meta: this ksy requires compiler version at least 0.9, but you have 0.8
```

and not cryptic

```
/doc-ref: expected string, got List(https://github.com/devicetree-org/devicetree-specification/releases/
tag/v0.3, https://github.com/devicetree-org/devicetree-specification/blob/ba2aa67/source/flattened
-format.rst, https://elinux.org/images/f/f4/Elc2013_Fernandes.pdf) (class scala.collection.immutable
.$colon$colon)
```

which looks rather like a bug.

database/dtb.ksy
```diff
@@ -0,0 +1,127 @@
...
+        type: u4
+      - id: ofs_name
+        -orig-id: nameoff
+        type: u4
+      - id: property
+        size: length
+      - id: boundary_padding
+        size: (- _io.pos) % 4
+enums:
+  fdt:
```
generalmimon, line 122: Isn't there a more meaningful enum name? I guess that `fdt` is an abbreviation of _Flattened Devicetree Format_, not something like **token type**.

Please don't forget about 

https://github.com/kaitai-io/kaitai_struct_formats/blob/c4551e5a3d5cafdbee23c885c78a795d42380397/database/dtb.ksy#L88-L90

if you rename it.

database/dtb.ksy
```diff
@@ -0,0 +1,127 @@
...
+
+  On Linux systems that support this the blobs can be accessed in
+  /sys/firmware/fdt :
+
+  - https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-firmware-ofw
+
+  The encoding of strings used in the strings block and struct block is
+  actually a subset of ASCII:
+
+  https://github.com/devicetree-org/devicetree-specification/blob/master/source/chapter2-devicetree-basics.rst
```
generalmimon, line 26: 1. It should point to a specific revision,
2. it can link to a concrete section by including the hashtag:

```suggestion
  <https://github.com/devicetree-org/devicetree-specification/blob/deb8f23/source/chapter2-devicetree-basics.rst#devicetree-structure-and-conventions>
```

database/dtb.ksy
```diff
@@ -0,0 +1,127 @@
...
+    pos: ofs_strings_block
+    size: strings_block_size
+    type: strings
+types:
+  strings:
+    seq:
+      - id: strings
+        type: strz
+        repeat: eos
+  fdt_node:
```
generalmimon, line 86: Adding a key `-webide-representation` on repeated types giving an insight into the inner structure saves time (and money) in the [Web IDE](https://ide.kaitai.io/) (you don't need to expand all elements one by one just to see what's inside them):

```suggestion
  fdt_node:
    -webide-representation: '{token_type} {body}'
```

> **Note:** using `body`, because I'm counting on the previous suggestion `s/fdt_node_body/body/g` to be applied.

```diff
   fdt_prop:
+    -webide-representation: '{name}'
     seq:
       - id: length
@@ ... @@
     instances:
       name:
         io: _root.strings_block._io
         pos: ofs_name
         type: strz
+        -webide-parse-mode: eager
```

generalmimon (changes requested):
Hopefully it will be easier for you to address everything at once...

database/dtb.ksy
```diff
@@ -0,0 +1,127 @@
+meta:
+  id: dtb
+  title: Flattened Devicetree Format
+  file-extension: dtb
+  application:
+    - Linux
+    - Das U-Boot
+  xref:
+    wikidata: Q16960371
```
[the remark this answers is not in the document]
armijnhemel: No, it is not just for Linux. It is also being ported to other operating systems (FreeBSD has some support for it as well), bootloaders, and so on, and becoming more of a universal way to describe hardware in a more generic way. Linux (and especially Linux on ARM) is the most prominent user if this particular format though, so I will add the tag.

Some other users: U-Boot (the bootloader) has support for it and many of the Android bootloaders as well. they pass this information on to whatever kernel they are booting.

database/dtb.ksy
armijnhemel: That is correct. I skipped this part, as for my particular use case it is not interesting. Also, none of the test files I looked at actually had anything useful in there, except the terminator.

database/dtb.ksy
armijnhemel: I think this is a brainfart.

database/dtb.ksy
KOLANICH: I don't like the word `token` here, maybe just `type`?

armijnhemel: The reason I called it token is because they are called that in the documentation:

https://github.com/devicetree-org/devicetree-specification/blob/v0.3/source/flattened-format.rst#lexical-structure

```
The five token types are as follows:
```

Perhaps because they are in an element `fdt_note` it can indeed be just `token`. @generalmimon any thoughts?

database/dtb.ksy
armijnhemel: I have added this now as well.

generalmimon:
@armijnhemel 

> Also, I have added this to the directory "database", but that doesn't seem to be the right place.

@KOLANICH 

> We have a directory `firmware`, but [we should really to invent a better structure for the repo](https://github.com/kaitai-io/kaitai_struct_doc/pull/48) (with more formal criteria what goes where which can be used to predict the location of a spec with onpy basic knowledge agout the format it implements). I guess `serialization` (but not `firmware`) will be better for this format.

I also think `serialization` is quite reasonable - it's definitely better than `database`. However, as for https://github.com/kaitai-io/kaitai_struct_formats/pull/404#issuecomment-766454630, it doesn't really matter what the format looks from inside but rather what it's mainly used for and where people would expect it. From this perspective, I'm thinking about `hardware` - a _devicetree_ is used to describe system hardware. However, the `hardware` folder is classified as _Hardware Protocols_ at https://formats.kaitai.io, which is not really what we mean. I don't see any connection with `firmware`.

database/dtb.ksy
```diff
@@ -0,0 +1,152 @@
...
+    type: u4
+  - id: ofs_strings_block
+    -orig-id: off_dt_strings
+    type: u4
+  - id: ofs_memory_reservation_block
+    -orig-id: off_mem_rsvmap
+    type: u4
+  - id: version
+    type: u4
+  - id: last_compatible_version
```
generalmimon, line 56: This name is misleading, at least to me - if I see **last** compatible version I imagine the latest one (i.e. the one that came _last_), not the lowest or earliest. However, it's clear that the [spec](https://github.com/devicetree-org/devicetree-specification/releases/download/v0.3/devicetree-specification-v0.3.pdf) intended the exact opposite:

> <dl>
> <dt>last_comp_version</dt>
> <dd>This field shall contain the lowest version of the devicetree data structure with which the version
> used is backwards compatible. So, for the structure as defined in this document (version 17), this field shall contain
> 16 because version 17 is backwards compatible with version 16, but not earlier versions. As per section 5.1, a
> DTSpec boot program should provide a devicetree in a format which is backwards compatible with version 16, and
> thus this field shall always contain 16.</dd>

I'm not sure why did the spec authors choose `last_` adjective in the name. Probably when you imagine the latest version as the "first", then you go further into the version history and declare the oldest compatible version as the "last" that would the current version be compatible with (if we went beyond the "last", any older versions would no longer work). However, this concept seems to be upside down - I guess it's natural that "last" should mean latest (the last one that has been released so far) and the "first" would mean the initial version, i.e. the oldest version. In this justification for the spec's naming that I came up with, I needed to reverse the timeline and the role of "first" and "last" pointers (we go _back in time_).

Does `first_compatible_version`, `oldest_compatible_version`, `lowest_compatible_version` or `min_compatible_version` make more sense? Which one would you choose?

database/dtb.ksy
```diff
@@ -0,0 +1,151 @@
...
+  actually a subset of ASCII:
+
+  <https://github.com/devicetree-org/devicetree-specification/blob/v0.3/source/devicetree-basics.rst>
+
+  Example files:
+
+  - <https://github.com/qemu/qemu/tree/master/pc-bios>
+doc-ref:
+  - https://github.com/devicetree-org/devicetree-specification/releases/tag/v0.3
+  - https://github.com/devicetree-org/devicetree-specification/blob/ba2aa67/source/flattened-format.rst
```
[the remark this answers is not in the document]
generalmimon: Resolved in https://github.com/kaitai-io/kaitai_struct_formats/pull/467/commits/41155f67cd9c452eb71ae56f7c701c852a58ef6d

generalmimon:
As I wrote before in https://github.com/kaitai-io/kaitai_struct_formats/pull/467#issuecomment-882021127, I moved the spec to `hardware/` directory (https://github.com/kaitai-io/kaitai_struct_formats/pull/467/commits/a8331cab26e5849cd12850857c9e7152d8fbf287) because I think it's the first place where people would look for it.

armijnhemel:
> As I wrote before in [#467 (comment)](https://github.com/kaitai-io/kaitai_struct_formats/pull/467#issuecomment-882021127), I moved the spec to the `hardware/` directory ([a8331ca](https://github.com/kaitai-io/kaitai_struct_formats/commit/a8331cab26e5849cd12850857c9e7152d8fbf287)) because I think it's the first place where people would look for it.

For pure dtb: yes, but there are, as always, exceptions, namely FIT images where entire Linux kernel images and file systems are included and dtb purely acts as a container.

generalmimon (approved):
Thanks for contributing! Merging in.

PR #469, opened 2021-04-05
add definition for RPM files

armijnhemel:
This is a fairly complete parser for version 3 of the RPM file format which is the currently maintained version of RPM. Other versions are either very old or are unmaintained and not covered by this specification.

There is some duplication in the format that cannot easily be removed )see https://github.com/kaitai-io/kaitai_struct/issues/872 ).

This doesn't parse the entire format. Specifically the following are not parsed or reported:

* the size of the payload. This is dependent on some values in the signature section of the RPM. I left notes about how I *think* it should be done and I would appreciate some comments there.
*  the individual tags are not split out further or given any meaning, as this about a few hundred tags that would potentially need to be processed. I think it is more useful to have a program using the parser process the output.
* the payload itself is not processed, because that is usually some cpio compressed archive (and could be compressed in a number of ways, depending on values in the header) or (likely compressed) drpm data. I am leaving this for someone else.

Currently all tags are reported as an array of values, even though tags that have the tag 'string' will only ever have one string. I have decided to return an array nonetheless to have a more consistent result.

I know that this is a big one to go through and digest. Comments would be much appreciated.

### DIFF
--- a/archive/rpm.ksy
+++ b/archive/rpm.ksy
@@ -16,7 +16,7 @@ doc: |
 doc-ref:
   - https://github.com/rpm-software-management/rpm/blob/master/doc/manual/format.md
   - https://github.com/rpm-software-management/rpm/blob/master/doc/manual/tags.md
-  - https://refspecs.linuxbase.org/LSB_4.1.0/LSB-Core-generic/LSB-Core-generic/pkgformat.html
+  - https://refspecs.linuxbase.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/pkgformat.html
   - http://ftp.rpm.org/max-rpm/
 seq:
   - id: lead
@@ -28,8 +28,11 @@ seq:
   - id: header
     type: header
   #- id: payload
-    #type: payload
+    # size: ??
+    # doc: if signature has a SIZE value, then it is:
+    # signature[SIZE][0] - sizeof<header>
 types:
+  dummy: {}
   lead:
     seq:
       - id: magic
@@ -76,7 +79,7 @@ types:
         repeat: expr
         repeat-expr: header_record.index_record_count
       - id: storage_section
-        #type: storage_section
+        type: dummy
         size: header_record.index_storage_size
   signature_index_record:
     seq:
@@ -90,6 +93,73 @@ types:
         type: u4
       - id: count
         type: u4
+    instances:
+       body:
+          io: _parent.storage_section._io
+          pos: record_offset
+          type:
+            switch-on: record_type
+            cases:
+              header_types::int8: record_type_int8(count)
+              header_types::int16: record_type_int16(count)
+              header_types::int32: record_type_int32(count)
+              header_types::string: record_type_string
+              header_types::bin: record_type_bin(count)
+              header_types::string_array: record_type_string_array(count)
+              header_types::i18nstring: record_type_string_array(count)
+  record_type_int8:
+    params:
+      - id: count
+        type: u4
+    seq:
+      - id: values
+        type: u2
+        repeat: expr
+        repeat-expr: count
+  record_type_int16:
+    params:
+      - id: count
+        type: u4
+    seq:
+      - id: values
+        type: u2
+        repeat: expr
+        repeat-expr: count
+  record_type_int32:
+    params:
+      - id: count
+        type: u4
+    seq:
+      - id: values
+        type: u4
+        repeat: expr
+        repeat-expr: count
+  record_type_string:
+    seq:
+      - id: values
+        type: strz
+        encoding: UTF-8
+        repeat: expr
+        repeat-expr: 1
+  record_type_bin:
+    params:
+      - id: count
+        type: u4
+    seq:
+      - id: values
+        size: count
+        repeat: expr
+        repeat-expr: 1
+  record_type_string_array:
+    params:
+      - id: count
+        type: u4
+    seq:
+      - id: values
+        type: strz
+        encoding: UTF-8
+        repeat: expr
+        repeat-expr: count
   # header, which is almost identical to signature
   # except that some of the tags have a different
   # meaning in signature and header.
@@ -102,7 +172,7 @@ types:
         repeat: expr
         repeat-expr: header_record.index_record_count
       - id: storage_section
-        #type: storage_section
+        type: dummy
         size: header_record.index_storage_size
   header_index_record:
     seq:
@@ -116,6 +186,20 @@ types:
         type: u4
       - id: count
         type: u4
+    instances:
+       body:
+          io: _parent.storage_section._io
+          pos: record_offset
+          type:
+            switch-on: record_type
+            cases:
+              header_types::int8: record_type_int8(count)
+              header_types::int16: record_type_int16(count)
+              header_types::int32: record_type_int32(count)
+              header_types::string: record_type_string
+              header_types::bin: record_type_bin(count)
+              header_types::string_array: record_type_string_array(count)
+              header_types::i18nstring: record_type_string_array(count)
   header_record:
     seq:
       - id: magic
--- a/archive/rpm.ksy
+++ b/archive/rpm.ksy
@@ -19,8 +19,8 @@ doc: |
   RPM file format, as well as a currently abandoned fork (rpm5). These formats
   are not covered by this specification.
 doc-ref:
-  - https://github.com/rpm-software-management/rpm/blob/master/doc/manual/format.md
-  - https://github.com/rpm-software-management/rpm/blob/master/doc/manual/tags.md
+  - https://github.com/rpm-software-management/rpm/blob/911448/doc/manual/format.md
+  - https://github.com/rpm-software-management/rpm/blob/911448/doc/manual/tags.md
   - https://refspecs.linuxbase.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/pkgformat.html
   - http://ftp.rpm.org/max-rpm/
 seq:
--- a/archive/rpm.ksy
+++ b/archive/rpm.ksy
@@ -78,23 +78,6 @@ types:
       - id: storage_section
         #type: storage_section
         size: header_record.index_storage_size
-  header_record:
-    seq:
-      - id: magic
-        contents: [0x8e, 0xad, 0xe8, 0x01]
-      - id: reserved
-        contents: [0, 0, 0, 0]
-      - id: index_record_count
-        -orig-id: nindex
-        type: u4
-        valid:
-          min: 1
-      - id: index_storage_size
-        -orig-id: hsize
-        type: u4
-        doc: | 
-          Size of the storage area for the data
-          pointed to by the Index Records.
   signature_index_record:
     seq:
       - id: tag
@@ -133,6 +116,23 @@ types:
         type: u4
       - id: count
         type: u4
+  header_record:
+    seq:
+      - id: magic
+        contents: [0x8e, 0xad, 0xe8, 0x01]
+      - id: reserved
+        contents: [0, 0, 0, 0]
+      - id: index_record_count
+        -orig-id: nindex
+        type: u4
+        valid:
+          min: 1
+      - id: index_storage_size
+        -orig-id: hsize
+        type: u4
+        doc: |
+          Size of the storage area for the data
+          pointed to by the Index Records.
 enums:
   rpm_types:
     0: binary
@@ -229,6 +229,10 @@ enums:
     1053: conflictflags
     1054: conflictname
     1055: conflictversion
+    1059: excludearch # from lib/rpmtag.h
+    1060: excludeos # from lib/rpmtag.h
+    1061: exclusivearch # from lib/rpmtag.h
+    1062: exclusiveos # from lib/rpmtag.h
     1064: rpmversion
     1065: triggerscripts # from lib/rpmtag.h
     1066: triggername # from lib/rpmtag.h
@@ -242,12 +246,14 @@ enums:
     1086: postinstall_interpreter # /bin/sh
     1087: preuninstall_interpreter # /bin/sh
     1088: postuninstall_interpreter # /bin/sh
+    1089: buildarchs # from lib/rpmtag.h
     1090: obsoletename
     1092: triggerscriptprog # from lib/rpmtag.h
     1094: cookie
     1095: filedevices
     1096: fileinodes
     1097: filelangs
+    1106: sourcepackage # from lib/rpmtag.h
     1112: provideflags
     1113: provideversion
     1114: obsoleteflags
@@ -260,6 +266,9 @@ enums:
     1124: payload_format
     1125: payload_compressor
     1126: payload_flags
+    1127: installcolor # from lib/rpmtag.h
+    1128: installtid # from lib/rpmtag.h
+    1129: removetid # from lib/rpmtag.h
     1131: rhnplatform
     1132: platform
     # below are all from lib/rpmtag.h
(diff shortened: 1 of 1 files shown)


archive/rpm.ksy
```diff
@@ -0,0 +1,472 @@
...
+    5054: supplementflags
+    5055: enhancename
+    5056: enhanceversion
+    5057: enhanceflags
+    5058: recommendnevrs
+    5059: suggestnevrs
+    5060: supplementnevrs
+    5061: enhancenevrs
+    5062: encoding
+    5066: filetriggerscripts
```
KOLANICH, line 435: please format the ids properly. In order to use `-orig-id` you would have to convert it to extended enum value, i.e.:

```yaml
5066:
  id: file_trigger_scripts
  -orig-id: file_trigger_scripts
```

KOLANICH (approved):
Except to the noticed flaws, LGTM.

archive/rpm.ksy
```diff
@@ -0,0 +1,472 @@
...
+    - src.rpm
+    - drpm
+  xref:
+    pronom: fmt/795 # v3
+    wikidata: Q492650
+  license: CC0-1.0
+  ks-version: 0.9
+  endian: be
+doc: |
+  This parser isr for RPM version 3 which is the current version of RPM. There
```
KOLANICH, line 15: `is`

KOLANICH:
@pmatilai, @Conan-Kudo

archive/rpm.ksy
```diff
@@ -0,0 +1,472 @@
+meta:
+  id: rpm
```
generalmimon, line 2: Missing `title`.

archive/rpm.ksy
```diff
@@ -0,0 +1,472 @@
...
+      - id: package_name
+        type: strz
+        encoding: UTF-8
+        size: 66
+      - id: os
+        -orig-id: osnum
+        type: u2
+        enum: operating_systems
+      - id: signature_type
+        -orig-id: signature_type
```
generalmimon, line 60: No need for `-orig-id` if it's the same as `id`.

```suggestion
```

archive/rpm.ksy
Conan-Kudo, line 15: RPM is currently on _version 4_, not _version 3_.

Conan-Kudo (changes requested):

Conan-Kudo:
cc: @pmatilai, @ffesti, @dmach

archive/rpm.ksy
armijnhemel: https://github.com/rpm-software-management/rpm/blob/master/doc/manual/format.md

```
This document describes the RPM file format version 3.0, which is used by RPM versions 2.1 and greater.
```

The major version in the header says 3 as well. Of course the version of the rpm tool is 4.x, but the file *format* is still version 3.

armijnhemel: I have clarified this in the description.

archive/rpm.ksy
armijnhemel: added -orig-ids everywhere, should be clearer now.

archive/rpm.ksy
```diff
@@ -0,0 +1,1033 @@
...
+  endian: be
+doc: |
+  This parser is for the RPM version 3 file format which is the current version
+  of the file format used by RPM 2.1 and later (including RPM version 4.x, which
+  is the current version of the RPM tool). There are historical versions of the
+  RPM file format, as well as a currently abandoned fork (rpm5). These formats
+  are not covered by this specification.
+doc-ref:
+  - https://github.com/rpm-software-management/rpm/blob/master/doc/manual/format.md
+  - https://github.com/rpm-software-management/rpm/blob/master/doc/manual/tags.md
```
Conan-Kudo, line 23: You should probably have versioned reference here.

Conan-Kudo (changes requested):

archive/rpm.ksy
armijnhemel: Urgh, yes, you are correct. Fixed.

armijnhemel:
Is there any way I can (cleanly) calculate the size of the payload? I have been trying all kinds of things, but to no avail.

archive/rpm.ksy
```diff
@@ -0,0 +1,1068 @@
...
+      - id: signature_type
+        type: u2
+        valid: 5
+      - id: reserved
+        size: 16
+  rpm_version:
+    seq:
+      - id: major
+        type: u1
+        valid: 0x3
```
generalmimon, line 72: > The major version in the header says 3 as well.

I don't know, but the `rpm` software actually allows 4 too ([`lib/rpmlead.c:102-105`](https://github.com/rpm-software-management/rpm/blob/4afe2d14d33db82ccb41c0a8d5eb1a4db90762fc/lib/rpmlead.c#L102-L105)):

```c
    if (lead->major < 3 || lead->major > 4) {
	*msg = xstrdup(_("unsupported RPM package version"));
	return RPMRC_FAIL;
    }
```

But writes 3 ([`lib/rpmlead.c:56`](https://github.com/rpm-software-management/rpm/blob/4afe2d14d33db82ccb41c0a8d5eb1a4db90762fc/lib/rpmlead.c#L56)):

```c
	l->major = 3;
```

armijnhemel: It has been version 3 for something like 20 years, so I think it is a safe assumption to make. I have never seen a valid file with major version 4.

KOLANICH: IMHO we should also accept `4`. It may be a plan if RPM developers to make `4` somehow compatible to `3`. Since `rpm` tool accepting `4` is already rolled out, there is no way for rpm devs to reconsider. It would be easier for them just use `5` if they wanted to make incompatible changes.

armijnhemel: > IMHO we should also accept `4`. It may be a plan if RPM developers to make `4` somehow compatible to `3`. Since `rpm` tool accepting `4` is already rolled out, there is no way for rpm devs to reconsider. It would be easier for them just use `5` if they wanted to make incompatible changes.

I have not seen any version with format version 4. I just went through 21 years of RPM source history (it is all in the RPM Git repository) and it has defaulted to format version 3 ever since and it has never defaulted to 4 at any given time. Note: this is about the RPM version in the `lead` part of the file format.

The reason I put that check in there is for sanity checks during carving.

I would not be opposed to allowing version 4 to be marked as valid, but I just think it is a bit unnecessary, as I am confident the RPM developers will not change it (instead, they will work inside the current format, as they have done in the last 20 years).

archive/rpm.ksy
```diff
@@ -0,0 +1,1076 @@
...
+      id: payload_digest_alt
+      -orig-id: RPMTAG_PAYLOADDIGESTALT
+  header_types:
+    # from LSB
+    0: not_implemented
+    1: char
+    2: int8
+    3: int16
+    4: int32
+    5: int64 # reserved
```
generalmimon, line 1072: I think that _Reserved_ in https://refspecs.linuxbase.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/pkgformat.html#AEN63192 rather means that it wasn't used by any RPM Tag yet at the time when the LSB RPM spec was up to date. This is no longer the case, there _are_ a few `int64` tags now (search for "`int64`" in https://github.com/rpm-software-management/rpm/blob/911448f2/doc/manual/tags.md) - Longarchivesize (271), Longsize (5009), Longfilesizes (5008) and Longsigsize (270).

Also `int64` is fully implemented in `rpm` (see https://github.com/rpm-software-management/rpm/search?q=RPM_INT64_TYPE), so I don't see reason not to implement it in the `.ksy`.

armijnhemel: LSB seems to be horribly outdated when it comes to RPM (also only talks about gzip compression, although there has been support for other formats for years) so it should be taken with a grain of salt. I agree with you that it can be changed. I missed it because the files I tested with didn't contain RPMs that are *that* big.

archive/rpm.ksy
generalmimon: > I have not seen any version with format version 4. I just went through 21 years of RPM source history (it is all in the RPM Git repository) and it has defaulted to format version 3 ever since and it has never defaulted to 4 at any given time.

Hmm, interestingly enough, I also dug into its Git history and discovered this commit https://github.com/rpm-software-management/rpm/commit/0bac1d421ccdcc719e2e60bcfb4859d0f602510c from 17 Dec 2009 with the following change (apparently reverting back to 3 hardcoded as opposed to switching between 3 and 4 based on some feature added in the past):

```diff
     rpmGetArchInfo(NULL, &archnum);
     rpmGetOsInfo(NULL, &osnum);
 
-    l->major = (_noDirTokens ? 4: 3);
+    l->major = 3;
     l->minor = 0;
     l->archnum = archnum;
     l->osnum = osnum;
```

Change of major version setting to `(_noDirTokens ? 4: 3)` was introduced [in `92698df7`](https://github.com/rpm-software-management/rpm/commit/92698df7236d0fc6c1d9c683210ddc12984764b3#diff-e7052a5203b7bc010c0d314bd59e04f5c7cc17fd4f933465813e3189690ec762R47) (from 1 Dec 2007). So it was in the `rpm` codebase for 2 years (from Dec 2007 to Dec 2009).

pmatilai: Folks, the real answer to all this is quite different:
The lead as a structure is 25 years obsolete, the data there is meaningless. Seriously.
Except to check for the magic to detect that it's an rpm file in the first place, just ignore everything in it.

archive/rpm.ksy
pmatilai: Yes, rpm >= 4.6 uses int64 for large sizes and you will need to implement it to deal with real-world packages in todays world. LSB seems stuck in the last millenium when it comes to rpm. See https://rpm.org/devel_doc/large_files.html for documentation.

generalmimon:
> Is there any way I can (cleanly) calculate the size of the payload? I have been trying all kinds of things, but to no avail. The problem is that I want to be able to access a specific value, namely `signature[SIZE][0]` (in pseudocode).

@armijnhemel The following modification does exactly that (but I don't know if it counts as _cleanly_ 🙂):

```diff
diff --git i/archive/rpm.ksy w/archive/rpm.ksy
index 5e4875a..a4c6879 100644
--- i/archive/rpm.ksy
+++ w/archive/rpm.ksy
@@ -30,14 +30,46 @@ seq:
     type: header(true)
   - id: boundary_padding
     size: (- _io.pos) % 8
+  - size: 0
+    if: ofs_header < 0
   - id: header
     type: header(false)
-  # - id: payload
-  #   size: ??
-  #   doc: |
-  #     if signature has a SIZE value, then it is:
-  #     signature[SIZE][0] - sizeof<header>
+  - size: 0
+    if: len_header < 0
+  - id: signature_tags_steps
+    type: 'signature_tags_step(_index, _index < 1 ? -1 : signature_tags_steps[_index - 1].size_tag_idx)'
+    repeat: expr
+    repeat-expr: signature.header_record.num_index_records
+  - id: payload
+    size: len_payload
+    if: has_signature_size_rec
+instances:
+  len_payload:
+    value: 'signature_size_rec.body.as<record_type_int32>.values[0] - len_header'
+    if: has_signature_size_rec
+  ofs_header:
+    value: _io.pos
+  len_header:
+    value: _io.pos - ofs_header
+  has_signature_size_rec:
+    value: signature_tags_steps.last.size_tag_idx != -1
+  signature_size_rec:
+    value: signature.index_records[signature_tags_steps.last.size_tag_idx]
+    if: has_signature_size_rec
 types:
+  signature_tags_step:
+    params:
+      - id: idx
+        type: s4
+      - id: prev_size_rec_idx
+        type: s4
+    instances:
+      size_tag_idx:
+        value: |
+          prev_size_rec_idx != -1 ? prev_size_rec_idx :
+            (_parent.signature.index_records[idx].signature_tag == signature_tags::size
+            and _parent.signature.index_records[idx].record_type == record_types::int32
+            and _parent.signature.index_records[idx].count >= 1 ? idx : -1)
   dummy: {}
   lead:
     doc: |
```

To my knowledge, this is the best thing we can do right now and I'm quite happy with it. If we implement features that would make this code cleaner at some time, we can always refactor it.

generalmimon:
Perhaps only parse `payload` with an instance instead of a `seq` attribute, because not every `.ksy` spec user will want to use it - some won't need to parse it at all (e.g. metadata in `header` are enough for their use) and some may prefer parsing `payload` in their application code instead for some reason.

armijnhemel:
> > @armijnhemel Thanks for contributing this spec. Can you tell me if you see any shortcomings in the current implementation? Otherwise I think it's good to merge.
> > If anyone also wants to comment on the latest version, I'd appreciate that too :)
> 
> > > Is there any way I can (cleanly) calculate the size of the payload? I have been trying all kinds of things, but to no avail. The problem is that I want to be able to access a specific value, namely `signature[SIZE][0]` (in pseudocode).
> > 
> > 
> > @armijnhemel The following modification does exactly that (but I don't know if it counts as _cleanly_ slightly_smiling_face):
> > ```diff
> > diff --git i/archive/rpm.ksy w/archive/rpm.ksy
> > index 5e4875a..a4c6879 100644
> > --- i/archive/rpm.ksy
> > +++ w/archive/rpm.ksy
> > @@ -30,14 +30,46 @@ seq:
> >      type: header(true)
> >    - id: boundary_padding
> >      size: (- _io.pos) % 8
> > +  - size: 0
> > +    if: ofs_header < 0
> >    - id: header
> >      type: header(false)
> > -  # - id: payload
> > -  #   size: ??
> > -  #   doc: |
> > -  #     if signature has a SIZE value, then it is:
> > -  #     signature[SIZE][0] - sizeof<header>
> > +  - size: 0
> > +    if: len_header < 0
> > +  - id: signature_tags_steps
> > +    type: 'signature_tags_step(_index, _index < 1 ? -1 : signature_tags_steps[_index - 1].size_tag_idx)'
> > +    repeat: expr
> > +    repeat-expr: signature.header_record.num_index_records
> > +  - id: payload
> > +    size: len_payload
> > +    if: has_signature_size_rec
> > +instances:
> > +  len_payload:
> > +    value: 'signature_size_rec.body.as<record_type_int32>.values[0] - len_header'
> > +    if: has_signature_size_rec
> > +  ofs_header:
> > +    value: _io.pos
> > +  len_header:
> > +    value: _io.pos - ofs_header
> > +  has_signature_size_rec:
> > +    value: signature_tags_steps.last.size_tag_idx != -1
> > +  signature_size_rec:
> > +    value: signature.index_records[signature_tags_steps.last.size_tag_idx]
> > +    if: has_signature_size_rec
> >  types:
> > +  signature_tags_step:
> > +    params:
> > +      - id: idx
> > +        type: s4
> > +      - id: prev_size_rec_idx
> > +        type: s4
> > +    instances:
> > +      size_tag_idx:
> > +        value: |
> > +          prev_size_rec_idx != -1 ? prev_size_rec_idx :
> > +            (_parent.signature.index_records[idx].signature_tag == signature_tags::size
> > +            and _parent.signature.index_records[idx].record_type == record_types::int32
> > +            and _parent.signature.index_records[idx].count >= 1 ? idx : -1)
> >    dummy: {}
> >    lead:
> >      doc: |
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > To my knowledge, this is the best thing we can do right now and I'm quite happy with it. If we implement features in the future that would make this code cleaner, we can always refactor it.
> 
> I need to test with this and wrap my head around it. It is not immediately obvious what it does, so perhaps we should add some comments.

The head wrapping is still something I need to do, but I tested with several files. So far it looks good and I can extract the payload correctly.

I would like to ask if the obsolete tags that were removed could be reintroduced. I sometimes (but seldomly) still encounter RPM files from 20 years ago in my work (don't ask).

There are also old files that have some internal tags, like this one that uses tag 1131 (internal, obsolete):

https://archives.fedoraproject.org/pub/archive/fedora/linux/core/1/i386/os/Fedora/RPMS/4Suite-1.0-0.0.a3.i386.rpm

It is definitely not a blocker and I can add them to my own local copy if needed, so if you want to merge, please go ahead!

archive/rpm.ksy
```diff
@@ -0,0 +1,1177 @@
...
+    # 1003 - obsolete
+    1004:
+      id: md5
+      -orig-id: RPMSIGTAG_MD5
+      doc: MD5 signature
+    1005:
+      id: gpg
+      -orig-id: RPMSIGTAG_GPG
+      doc: GnuPG signature
+    # 1006 - obsolete
```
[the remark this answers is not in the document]
generalmimon: Resolved in https://github.com/kaitai-io/kaitai_struct_formats/pull/469/commits/cbbf40b7df674e1f324260b9904566b9b03cd6f1

generalmimon:
@armijnhemel 

> The head wrapping is still something I need to do

My comment above (https://github.com/kaitai-io/kaitai_struct_formats/pull/469#issuecomment-950219229) didn't help?

> I would like to ask if the obsolete tags that were removed could be reintroduced.

Can you please look at the latest changes?

armijnhemel:
> @armijnhemel
> 
> > The head wrapping is still something I need to do
> 
> My comment above ([#469 (comment)](https://github.com/kaitai-io/kaitai_struct_formats/pull/469#issuecomment-950219229)) didn't help?

Yes, but I still need to wrap my head around it. It works, I just need to make sure I understand what is going on :-)

> 
> > I would like to ask if the obsolete tags that were removed could be reintroduced.
> 
> Can you please look at the latest changes?

Looks good to me!

generalmimon:
@armijnhemel:

> Looks good to me!

Thanks!

archive/rpm.ksy
```diff
@@ -0,0 +1,1458 @@
...
+      id: old_enhances_flags_obsolete
+      -orig-id: RPMTAG_OLDENHANCESFLAGS
+    1162:
+      id: priority_unimplemented
+      -orig-id: RPMTAG_PRIORITY
+    1163:
+      id: cvsid_unimplemented
+      -orig-id:
+        - RPMTAG_CVSID
+        - RPMTAG_SVNID
```
Mingun, line 1053: Interestingly, this is the first (merged) spec where `-orig-id` represented as a list of two elements, which looks strange.

generalmimon: @Mingun

> Interestingly, this is the first (merged) spec where `-orig-id` represented as a list of two elements

No, it isn't. Search for regex "`-orig-id:\n`".

```
3 results - 2 files

archive\rpm.ksy:
  1050        id: cvsid_unimplemented
  1051:       -orig-id:
  1052:         - RPMTAG_CVSID
  1053          - RPMTAG_SVNID

image\bmp.ksy:
  206    bitmap_header:
  207:     -orig-id:
  208:       - BITMAPCOREHEADER
  209        - OS21XBITMAPHEADER

  460    rgb_record:
  461:     -orig-id:
  462:       - RGB_TRIPLE
  463        - RGB_QUAD
```

> which looks strange

Is it a problem? How would _you_ represent such a thing?

KOLANICH: I use multielement `-orig-id`s extensively in my specs.

Mingun: Ok, the first used in the enum value. The other `-orig-id` fields my parser just completely ignore.

> Is it a problem? 

No, I've just noticed that my parser failed to parse that. It is strange in that sense why the same thing is referenced by different names in the original spec?

KOLANICH: >It is strange in that sense why the same thing is referenced by different names in the original spec?

There may be more than 1 "official" stuff. I.e. a PDF document and an official header file. And a lot of non-official, sometimes no official docs at all, but few reverse-engineered open source impls.

archive/rpm.ksy
```diff
@@ -0,0 +1,1177 @@
...
+        type: u2
+        enum: architectures
+      - id: package_name
+        size: 66
+        type: strz
+      - id: os
+        -orig-id: osnum
+        type: u2
+        enum: operating_systems
+      - id: signature_type
```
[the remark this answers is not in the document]
pmatilai: Yup, the stuff here is so old that rpm itself hasn't supported the format in two decades, and I've never seen an actual specimen of these tags. My testing-archives go back to 1999 material. Don't bother.

PR #470, opened 2021-04-06
Resource fork for macOs

tonal:
The resource fork is a fork or section of a file on Apple's classic Mac OS operating system, which was also carried over to the modern macOS for compatibility, used to store structured data along with the unstructured data stored within the data fork.
See https://en.wikipedia.org/wiki/Resource_fork
Port from macfork - http://freshmeat.sourceforge.net/projects/macfork

### DIFF
--- a/filesystem/mac_fork.ksy
+++ b/filesystem/mac_fork.ksy
@@ -1,0 +1,95 @@
+meta:
+  id: mak_res
+  file-extension: rsr
+  endian: be
+  title: Resource fork
+  # http://freshmeat.sourceforge.net/projects/macfork
+  # https://en.wikipedia.org/wiki/Resource_fork
+seq:
+  - id: res_header
+    type: res_header
+instances:
+  map: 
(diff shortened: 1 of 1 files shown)


generalmimon (changes requested):
I think that the Mac OS resource fork format has been already described by an existing spec [`macos/resource_fork.ksy`](https://github.com/kaitai-io/kaitai_struct_formats/blob/3e4be117e4c09aa4d4d206c9d4aa21bd8f3fcc5d/macos/resource_fork.ksy).

Cc @dgelessus 

Instead of creating a new spec, perhaps you could check out the existing version, take the best from **your** implementation and try to improve the existing one?

tonal:
Sorry. The existing spec ([macos/resource_fork.ksy](https://github.com/kaitai-io/kaitai_struct_formats/blob/3e4be117e4c09aa4d4d206c9d4aa21bd8f3fcc5d/macos/resource_fork.ksy)) does describe this format much more fully.

PR #471, opened 2021-04-09
add Qualcomm Snapdragon bootloader for Android.

armijnhemel:
This file processes the header of the Qualcomm Snapdragon boot loader as used on some Android devices.


KOLANICH:
Could you specify the whole format, not only its header?

armijnhemel:
> Could you specify the whole format, not only its header?

No, I cannot, as the contents could be anything: ELF files, proprietary Qualcomm formats, etc.

KOLANICH:
At least splitting the file into chunks each one has a well-defined size and position is neeeded.

archive/android_bootldr.ksy
```diff
@@ -0,0 +1,36 @@
+meta:
+  id: android_bootldr
```
generalmimon, line 2: The file name and `/meta/id` should include some identifier of environment for which this format is specific, because I guess this is not anything like the "main" Android `bootloader.img` format or superior to all other ones. After all, you yourself described another variation of the `bootloader.img` format for another class of Android devices in https://github.com/kaitai-io/kaitai_struct_formats/pull/472.

archive/android_bootldr.ksy
```diff
@@ -0,0 +1,36 @@
...
+  title: Qualcomm Snapdragon (MSM) bootloader.img format
```
generalmimon, line 3: BTW, the Wikipedia article [Devices using Qualcomm Snapdragon processors](https://en.wikipedia.org/wiki/Devices_using_Qualcomm_Snapdragon_processors) lists devices grouped by Qualcomm model number, so it may be helpful for finding out which devices will ultimately support and use this format (by checking if the model number starts with `MSM`, e.g. `MSM8996`). You might want to include this link into `/doc` (please don't forget about these angle brackets `<>`, see https://github.com/kaitai-io/kaitai_struct_formats/pull/467#discussion_r609723301).

archive/android_bootldr.ksy
```diff
@@ -0,0 +1,36 @@
...
+  file-extension: img
+  tags:
+    - archive
+    - android
+  license: CC0-1.0
+  endian: le
+doc-ref: https://android.googlesource.com/device/lge/hammerhead/+/7618a7/releasetools.py
```
generalmimon, line 10: I suggest adding a `doc` with some resources where one can get sample files:

- https://developers.google.com/android/images
- https://tmp.androidfilehost.com/?w=files&flid=300713#:~:text=bootloader.img

I like the second link particularly because you don't have to download 1+ GB .zip archive just to obtain a single `bootloader.img` file a with few MBs from it (which is the case when you use the Google website; you also need to pick the right device there because only several of them use this "archive" format for `bootloader.img`). However, it is a bit inconvenient to work with it in the Web IDE because of its relatively large size (~36 MB). It also uses different partitions than the _Hammerhead's `bootloader.img`_ that you link in `/doc-ref`, but the overall "archive" structure is the same.

---

Anyway, you should also add at least a few sentences about the format in [the `/doc` key](https://doc.kaitai.io/ksy_style_guide.html#doc), i.e. some general information about the format? What it is used for? What kind of data it stores? What is the main usage field?

archive/android_bootldr.ksy
```diff
@@ -0,0 +1,36 @@
...
+seq:
+  - id: magic
+    contents: BOOTLDR!
+  - id: images
+    -orig-id: num_images
+    type: u4
```
generalmimon, line 16: Actually, `num_images` would be a far better name - you can't tell that `images` is only a number of images just from the name (I'd expect an attribute called `images` to have type `image[]`, i.e. an array of objects of type `image`). This also corresponds to recommended conventions in the [KSY Style Guide](https://doc.kaitai.io/ksy_style_guide.html#attr-id):

> One SHOULD use the following rules to maintain consistency across various KSY files. Doing that would maintain the "principle of least surprise" and make life easier to end-users, reducing amount of guesswork.
> 
> * For simple non-repeated fields, use a simple singular form — e.g. `width`, `header`, `transaction_id`, file.
> * For an array of objects (i.e. with `repeat: something`), use plural form — e.g. `files`, `transactions`.
> * (...)

```suggestion
  - id: num_images
    type: u4
```

Also don't forget about

https://github.com/kaitai-io/kaitai_struct_formats/blob/d2500791ffe1b3d8b6af58552d3ddb275e2beb1b/archive/android_bootldr.ksy#L23-L26

while renaming (or just do `sed -i 's/\<images\>/num_images/g' <ksy-file>`).

archive/android_bootldr.ksy
```diff
@@ -0,0 +1,36 @@
...
+  - id: ofs_start
+    -orig-id: start_offset
+    type: u4
```
generalmimon, line 19: Please rename it to `ofs_img_bodies` (see the following review comment):

```shell
sed -i 's/\<ofs_start\>/ofs_img_bodies/g' <ksy-file>
```

archive/android_bootldr.ksy
```diff
@@ -0,0 +1,36 @@
...
+  - id: bootloader_size
+    -orig-id: bootldr_size
+    type: u4
```
generalmimon, line 22: Please rename it to `len_img_bodies` (see the following review comment):

```shell
sed -i 's/\<bootloader_size\>/len_img_bodies/g' <ksy-file>
```

archive/android_bootldr.ksy
```diff
@@ -0,0 +1,36 @@
...
+  - id: img_info
+    type: img_info
+    repeat: expr
+    repeat-expr: images
```
generalmimon, line 26: I know that the noun `info`(rmation) doesn't have a plural form in English, but the name of each repeated field should ideally be distinguishable from the singular form.

https://doc.kaitai.io/ksy_style_guide.html#attr-id:

> * For an array of objects (i.e. with repeat: something), use plural form — e.g. `files`, `transactions`.

The original identifier of this attribute doesn't matter, you can always preserve it in `-orig-id`. I wouldn't even mind if you use `img_infos`, regardless of what Standard English says about it. At least, there would be a clear distinction.

Or you can replace `info` with some noun that can be pluralized, e.g. `header`.

Besides that, making the subtype and attribute names different will make the KSC-generated code from this format spec usable in C# - when there is an attribute `img_info` with the same name as the subtype `img_info`, the C# codegen would be affected by https://github.com/kaitai-io/kaitai_struct/issues/86. In complex formats, this is somewhat tiresome to eliminate (so we usually don't even mention it as there are dozens of more serious issues), but dealing with this occurrence in this simple format makes sense I think.

archive/android_bootldr.ksy
```diff
@@ -0,0 +1,36 @@
...
+types:
+  img_info:
+    -webide-representation: '{name}'
+    seq:
+      - id: name
+        type: strz
+        encoding: UTF-8
+        size: 64
```
generalmimon, line 34:
```suggestion
      - id: name
        size: 64
        type: strz
        encoding: UTF-8
```

See https://doc.kaitai.io/ksy_style_guide.html#seq-attr:

> ## 5. Sequence attributes
> 
> When specifying an attribute, one MUST use the following order of keys:
> 
> 1. Identifier(s)
>    * <ins>**`id`**</ins>
>    * `-orig-id` — use to specify original ID spelling if transcribing a structure from existing software and/or official `spec`
> 1. <ins>**`size`**</ins>
> 1. `size-eos`
> 1. <ins>**`type`**</ins>
> 1. Type-related keys:
>    a. (...)
>    h. <ins>**`encoding`**</ins>

archive/android_bootldr.ksy
```diff
@@ -0,0 +1,36 @@
...
+      - id: size
+        type: u4
```
generalmimon, line 36: 1. I'd rename it to `len_body`, since just calling it `size` does not necessarily say what size it is.

   ```suggestion
         - id: len_body
           type: u4
   ```

2. You could add an `instance` providing access to all byte arrays corresponding to each particular image:

   ```diff
      - id: img_infos
        type: img_info
        repeat: expr
        repeat-expr: num_images
   +instances:
   +  img_bodies_wrapper:
   +    pos: ofs_img_bodies
   +    size: len_img_bodies
   +    type: img_bodies_type
   +  img_bodies:
   +    value: img_bodies_wrapper.img_bodies
   +  is_valid:
   +    value: img_bodies_wrapper.is_eof
    types:
      img_info:
   @@ ... @@
   +  img_bodies_type:
   +    seq:
   +      - id: img_bodies
   +        size: _root.img_infos[_index].size
   +        repeat: expr
   +        repeat-expr: _root.num_images
   +      - size: 0
   +        if: is_eof < 0
   +    instances:
   +      is_eof:
   +        value: _io.eof
   ```
   
   _(uses previous suggestions on renaming the attribute `img_info` &rarr; <code>img_info**s**</code>, `'s/\<images\>/num_images/g'`, `'s/\<ofs_start\>/ofs_img_bodies/g'` and `'s/\<bootloader_size\>/len_img_bodies/g'`)._

   I consider the `is_valid` instance desirable because the Python parser linked in [`/doc-ref`](https://github.com/kaitai-io/kaitai_struct_formats/blob/d2500791ffe1b3d8b6af58552d3ddb275e2beb1b/archive/android_bootldr.ksy#L10) is really strict about `len_img_bodies`, originally `bootloader_size` ([`releasetools.py:162`](https://android.googlesource.com/device/lge/hammerhead/+/7618a7/releasetools.py#162)):

   ```python
     p = start_offset
     img_dict = {}
     for name, size in imgs:
       img_dict[trunc_to_null(name)] = p, size
       p += size
     assert p - start_offset == bootloader_size, "bootloader.img corrupted"
   ```

   To ensure that this KSY implementation is compatible with the reference parser, it should check the same thing - that's what the `is_valid` attribute is for. The check for `_io.eof` ensures that there are no additional bytes in the `len_img_bodies`-byte substream after the last image body.

generalmimon (changes requested):

archive/android_bootldr.ksy
armijnhemel: I will search for information. Not every device with a Snapdragon used it, just a few did. Welcome to the mess that is Android! ;-)

generalmimon: > Not every device with a Snapdragon used it, just a few did.

Yes, but I downloaded and inspected a few factory images from https://developers.google.com/android/images and all models claimed to have the `MSM*` Snapdragon processor that I checked actually did use this format, so it seems to be correlated:

* ["hammerhead" for Nexus 5](https://developers.google.com/android/images#hammerhead): ✔️ (processor `MSM8974AA`, see [wiki](https://en.wikipedia.org/wiki/Devices_using_Qualcomm_Snapdragon_processors#Snapdragon_800,_801_and_805_(2013/14)))
* ["bullhead" for Nexus 5X](https://developers.google.com/android/images#bullhead): ✔️ (processor `MSM8992`, see [wiki](https://en.wikipedia.org/wiki/Devices_using_Qualcomm_Snapdragon_processors#Snapdragon_808_and_810_(2015)))
* ["sailfish" for Pixel](https://developers.google.com/android/images#sailfish): ✔️ (processor `MSM8996 Pro-AB`, see [wiki](https://en.wikipedia.org/wiki/Devices_using_Qualcomm_Snapdragon_processors#Snapdragon_820_and_821_(2016)))
* ["sargo" for Pixel 3a](https://developers.google.com/android/images#sargo): ❌ (processor `SDM670`, see [wiki](https://en.wikipedia.org/wiki/Devices_using_Qualcomm_Snapdragon_processors#Snapdragon_632_and_670_(2018)))
* ["redfin" for Pixel 5](https://developers.google.com/android/images#redfin): ❌ (processor `SM7250-AB`, see [wiki](https://en.wikipedia.org/wiki/Devices_using_Qualcomm_Snapdragon_processors#Snapdragon_750G_and_765/765G/768G_5G_(2020)))

> Welcome to the mess that is Android! ;-)

You're certainly right about that :)

archive/android_bootldr.ksy
armijnhemel: I am hesitant to use `_io.eof` because it complicates carving if there actually are additional bytes.

archive/android_bootldr.ksy
armijnhemel: I have renamed this. Please see if the new name is acceptable.

archive/android_bootldr.ksy
armijnhemel: I have added a short description, although not an in depth one.

archive/android_bootldr.ksy
armijnhemel: I have added some stuff in the description. If you want more, let me know.

generalmimon:
Links related to this format:

* https://github.com/NVISOsecurity/nexus_5_bootloader_unpacker
* https://github.com/ReFirmLabs/binwalk/issues/299
* https://github.com/Gallopsled/pwntools/blob/bd12d18/pwnlib/adb/bootloader.py

archive/android_bootldr_qcom.ksy
```diff
@@ -0,0 +1,179 @@
...
+    repeat: expr
+    repeat-expr: num_images
+types:
+  img_header:
+    -webide-representation: '{name}'
+    seq:
+      - id: name
+        size: 64
+        type: strz
+        encoding: UTF-8
```
generalmimon, line 166: @armijnhemel Doesn't

https://github.com/kaitai-io/kaitai_struct_formats/blob/82d9493383d9790eff5c3053f778d079b1509f00/archive/android_asus_bootldr.ksy#L14-L15

in fact apply here as well?

generalmimon (approved):

PR #472, opened 2021-04-09
add a specification for the bootloader used on several ASUS Android devices

armijnhemel:
This is a specification for the bootloader used on several ASUS Android devices. The magic is the same as some other Android bootloaders. I have only implemented support for version 2 and later, as I couldn't quickly find test files that use version 1 (and I am not even sure if any of those files were even publicly released).

### DIFF
--- a/archive/android_asus_bootldr.ksy
+++ b/archive/android_asus_bootldr.ksy
@@ -40,13 +40,13 @@ types:
     -webide-representation: '{file_name}'
     seq:
       - id: magic
+        size: 8
         type: str
         valid:
           any-of:
             - '"IFWI!!!!"'
             - '"DROIDBT!"'
             - '"SPLASHS!"'
-        size: 8
       - id: len_body
         type: u4
       - id: flags
--- a/archive/android_asus_bootldr.ksy
+++ b/archive/android_asus_bootldr.ksy
@@ -39,7 +39,7 @@ types:
   image:
     -webide-representation: '{file_name}'
     seq:
-      - id: magic
+      - id: chunk_id
         size: 8
         type: str
         valid:
@@ -64,7 +64,7 @@ types:
     instances:
       file_name:
         value: |
-          magic == "IFWI!!!!" ? "ifwi.bin" :
-          magic == "DROIDBT!" ? "droidboot.img" :
-          magic == "SPLASHS!" ? "splashscreen.img" :
+          chunk_id == "IFWI!!!!" ? "ifwi.bin" :
+          chunk_id == "DROIDBT!" ? "droidboot.img" :
+          chunk_id == "SPLASHS!" ? "splashscreen.img" :
           ""
--- a/archive/android_asus_bootldr.ksy
+++ b/archive/android_asus_bootldr.ksy
@@ -12,6 +12,9 @@ doc: |
   A bootloader image which only seems to have been used on a few ASUS
   devices. The encoding is ASCII, because the `releasetools.py` script
   is written using Python 2, where the default encoding is ASCII.
+
+  A test file can be found in the firmware files for the "fugu" device,
+  which can be downloaded from <https://developers.google.com/android/images>
 doc-ref: https://android.googlesource.com/device/asus/fugu/+/android-8.1.0_r5/releasetools.py
 seq:
   - id: magic
--- a/archive/android_asus_bootldr.ksy
+++ b/archive/android_asus_bootldr.ksy
@@ -1,0 +1,51 @@
+meta:
+  id: android_asus_bootldr
+  title: ASUS Fugu bootloader.img format (version 2 and later)
+  file-extension: img
+  tags:
+    - archive
+    - android
+  license: CC0-1.0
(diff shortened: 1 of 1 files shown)


archive/android_asus_bootldr.ksy
```diff
@@ -0,0 +1,51 @@
...
+types:
+  image:
+    seq:
+      - id: magic
+        type: str
+        size: 8
+      - id: len_image
+        type: u4
+      - id: flags
+        type: u1
```
KOLANICH, line 43: likely should be a type of bit fields.

archive/android_asus_bootldr.ksy
```diff
@@ -0,0 +1,51 @@
...
+    type: image
+    repeat: expr
+    repeat-expr: 3
+    doc: |
+      Only three images are included: `ifwi.bin`, `droidboot.img`
+      and `splashscreen.img`
+types:
+  image:
+    seq:
+      - id: magic
```
KOLANICH, line 37: Is it validated or is it dispatched on?

armijnhemel: I am not quite sure what you mean. There are only three values that are allowed, so I guess I could add them. The contents themselves are not specified, at least not in the sources where I found them. They might be formats without public specifications (although I am not 100% sure).

KOLANICH: I mean this field looks more like an identifier of a type of data in `body` rather than a signature. Sould it be renamed and a `switch-on`-`cases` added into `body` `type` (maybe commented-out for now)?

armijnhemel: If you look at the data, then it can be argued that it actually is a signature.

A test file can be found in the firmware files for the "fugu" device which can be downloaded from https://developers.google.com/android/images

For example, I downloaded fugu-lrx21m-factory-e012394c.zip

```
$ unzip -l fugu-lrx21m-factory-e012394c.zip 
Archive:  fugu-lrx21m-factory-e012394c.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  10-31-2014 00:32   fugu-lrx21m/
      853  10-31-2014 00:32   fugu-lrx21m/flash-all.bat
      700  10-31-2014 00:32   fugu-lrx21m/flash-base.sh
 12459008  10-31-2014 00:32   fugu-lrx21m/bootloader-fugu-fugu-02.05.img
437449865  10-31-2014 00:32   fugu-lrx21m/image-fugu-lrx21m.zip
      741  10-31-2014 00:32   fugu-lrx21m/flash-all.sh
---------                     -------
449911167                     6 files
```

The file `bootloader-fugu-fugu-02.05.img` is in this format.

The first part of the header looks like this:

```
$ hexdump -C bootloader-fugu-fugu-02.05.img | head
00000000  42 4f 4f 54 4c 44 52 21  03 00 a2 00 00 00 00 00  |BOOTLDR!........|
00000010  49 46 57 49 21 21 21 21  a4 00 40 00 01 00 00 00  |IFWI!!!!..@.....|
```

The script that I refer to also calls these "magic", so it kind of make sense to call it "magic" here as well, but personally I don't really care :-)

archive/android_asus_bootldr.ksy
armijnhemel: This might be overkill. The only sanity check that is being done in the script I refer to is `flags & 1 == 1`. The rest of the bits is not defined.

archive/android_asus_bootldr.ksy
generalmimon: @armijnhemel 

> If you look at the data, then it can be argued that it actually is a signature.

@KOLANICH 

> I mean this field looks more like an identifier of a type of data in body rather than a signature.

I renamed `magic` to `chunk_id`, because I agree with @KOLANICH - it is an _"identifier of a type of data in `body` rather than a signature"_ (https://android.googlesource.com/device/asus/fugu/+/android-8.1.0_r5/releasetools.py#66): 

```python
    if c_magic == "IFWI!!!!":
      common.ZipWriteStr(info.output_zip, "ifwi.bin", buf)
      WriteIfwi(info);
      continue
    if c_magic == "DROIDBT!":
      common.ZipWriteStr(info.output_zip, "droidboot.img", buf)
      WriteDroidboot(info);
      continue
    if c_magic == "SPLASHS!":
      common.ZipWriteStr(info.output_zip, "splashscreen.img", buf)
      WriteSplashscreen(info);
      continue
```

It is used to identify the type of the chunk (storing a particular file), so it is not really a "magic number" identifying the type of the format.

archive/android_asus_bootldr.ksy
generalmimon: > The only sanity check that is being done in the script I refer to is `flags & 1 == 1`.

I added this sanity check to `valid/expr`.

generalmimon (approved):
Let's merge this - the Android ecosystem is dominated with closed-source software, so this is probably the best we can do.

@armijnhemel Thanks!

PR #473, opened 2021-04-10
add XAR header format

armijnhemel:
The XAR format is an archive that consists of a header, a zlib compressed XML document and then the payload. The structure of the payload (the files, etc.) is described in the XML file. Because Kaitai Struct does not have a built in XML parser this parser is limited to just the header.

### DIFF
--- a/archive/xar.ksy
+++ b/archive/xar.ksy
@@ -27,18 +27,31 @@ doc: |
   supports additional checksums, but doesn't have checksum_name.
 doc-ref: https://github.com/mackyle/xar/wiki/xarformat
 seq:
-  - id: magic
-    contents: 'xar!'
-  - id: len_header
-    type: u2
+  - id: header_prefix
+    type: file_header_prefix
+    doc: internal; access `_root.header` instead
   - id: header
-    type:
-      switch-on: len_header
-      cases:
-        28: apple_header
-        _: regular_header
+    size: header_prefix.len_header - header_prefix._sizeof
+    type: file_header
+  - id: toc
+    size: header.len_toc_compressed
+    type: toc_type
+    process: zlib
+    doc: zlib compressed XML further describing the content of the archive
+instances:
+  checksum_algorithm_other:
+    -orig-id: XAR_CKSUM_OTHER
+    value: 3
+    doc-ref: https://github.com/mackyle/xar/blob/66d451d/xar/include/xar.h.in#L85
 types:
-  regular_header:
+  file_header_prefix:
+    seq:
+      - id: magic
+        contents: 'xar!'
+      - id: len_header
+        type: u2
+        doc: internal; access `_root.header.len_header` instead
+  file_header:
     seq:
       - id: version
         type: u2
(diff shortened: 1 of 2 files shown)


armijnhemel:
The version from Apple (which is in current use) has the same version number in the header. I am not sure how to wiggle these into this particular specification. Basically, it should ignore one field, but only if that field is not a valid strz. Ideally I would use some sort of "lookahead" operation, or some other conditional.

KOLANICH:
>Because Kaitai Struct does not have a built in XML parser this parser is limited to just the header.

You can try to implement the approach I use in `numpy_npy` format.

archive/xar_header.ksy
```diff
@@ -0,0 +1,58 @@
...
+
+  It should be noted that there is a different version from Apple uses the
+  same version number as the (unmaintained) version described here, but that
+  supports additional checksums, but doesn't have checksum_name.
+versions
+doc-ref: https://github.com/mackyle/xar/wiki/xarformat
+seq:
+  - id: magic
+    contents: 'xar!'
+  - id: header_size
```
KOLANICH, line 32: I don't see any `header` specified that occupues exactly the count of bytes that is stored in this field.

archive/xar_header.ksy
```diff
@@ -0,0 +1,58 @@
...
+  - id: toc_length_uncompressed
+    type: u8
+  - id: checksum_algorithm
+    type: u4
+    enum: checksum_algorithms
+  - id: checksum_name
+    type: strz
+    size: 36
+    if: checksum_algorithm == checksum_algorithms::other
+  - id: toc
```
KOLANICH, line 49: is it a `str`?

archive/xar_header.ksy
armijnhemel: It is in the header itself.

archive/xar_header.ksy
armijnhemel: The XML you mean? It is zlib compressed data. Uncompressed it should be a str or strz

archive/xar_header.ksy
KOLANICH: It can be moved into an own subtype, which size is determined by this field within it.

generalmimon:
@armijnhemel:

> The structure of the payload (the files, etc.) is described in the XML file. 

@KOLANICH:

> > Because Kaitai Struct does not have a built in XML parser this parser is limited to just the header.
> 
> You can try to implement the approach I use in `numpy_npy` format.

This is a very similar concept to the glTF binary format (#445), where one needs to parse JSON to be able to parse the binary structure. However, this does not mean it can't be described with the KS language - you can either outsource the XML parsing to opaque types (which I don't really recommend though, because you'll have lose any type checks / derivations that the compiler does and you'll have to resort to unsafe `.as<>` type casts, which are easy to screw up) or create a parametric type that can be instantiated from the application code. The application code will serve as a "bonding agent": first, it will invoke the initial parsing, access the XML string, parse it using a native or external XML parsing library of choice and then puts the needed data pulled from the XML "back" to the KSY-generated parser via the parametric types.

Which means that although the KSY won't read the whole file down to bottom (so it won't be so enjoyable to use in the Web IDE, for example), it's quite easy to use and you don't lose any type checks and you won't have to use type castings. For the glTF binary format, I have created a proof of concept for this approach, so you can see what I mean: https://github.com/kaitai-io/kaitai_struct_formats/pull/445#discussion_r598117321

archive/xar_header.ksy
armijnhemel: This would probably solve my issue of the variants as well: if the header size is larger than 28, then I can switch to the format described here, otherwise to the format that Apple uses. The header of the Apple variant will (according to the source code), always set it to 28.

archive/xar_header.ksy
KOLANICH: Then it is a strz.

dgelessus: > Then it is a strz.

Is the field really supposed to be zero-terminated? If not, I think it would be better to declare it as just `str`. That way if the data does contain a zero byte for some reason, you'll get an error when you parse the XML data (zero bytes are disallowed in XML). If you declare it as `strz` even though it isn't, the data will instead get silently truncated if a zero byte appears.

Also, are you sure that the XML data always has a fixed encoding? XML can also be encoded as UTF-16, not just UTF-8. So unless the XAR format requires one specific encoding for the XML data, I would parse it as raw bytes and let the XML parser handle the encoding.

armijnhemel: I looked at the source code of two versions of the program (the one on GitHub and the one from Apple) and it is explicitly set to UTF-8.

armijnhemel:
I fixed a few things. One thing I am not happy about is that currently the toc is defined as part of the header, but it shouldn't be.

I tried this:

```
seq:
  - id: magic
    contents: 'xar!'
  - id: len_header
    type: u2
  - id: header
    type:
      switch-on: len_header
      cases:
        28: apple_header
        _: regular_header
  - id: toc
    size: header.len_toc_compressed
    process: zlib
    doc: zlib compressed XML further describing the content of the archive

```

but that doesn't seem to work:

```
Call stack: undefined io.kaitai.struct.precompile.ErrorInInput: xar_header: /seq/3/size: don't know how to call method 'len_toc_compressed' of object type 'KaitaiStructType'
```

I don't know where I am going wrong there.

KOLANICH:
It would require interfaces proposal to implement it the way you wanted. There may be another way to implement it though

archive/xar_header.ksy
```diff
@@ -0,0 +1,89 @@
+meta:
+  id: xar_header
```
generalmimon, line 2: I understand that you tried to clarify that it can only parse the header, but this applies to a number of formats already in KSF and none of them use the main identifier to denote that. In reality, the vast majority of existing `.ksy` specifications are somehow incomplete or incorrect, but let's use other means to indicate the implementation status rather than the name.

Actually I hope that the repo https://github.com/kaitai-io/kaitai_struct_samples will eventually help meet this goal - if we collect enough quality sample files and their manually curated expected interpretations (in practice probably a YAML tree of expected values), we could use these to test the `.ksy` specifications and display the score as the "implementation status". Thanks to that, it would be transparent how each specification is doing, which one requires care and what missing KS features block many of the `.ksy` implementations from being better.

armijnhemel: I am not attached to the name, so feel free to rename.

generalmimon:
@armijnhemel Sample files, both with the Apple header and the regular one?

armijnhemel:
> @armijnhemel Sample files, both with the Apple header and the regular one?

Test files with gzip, bzip2 and no compression:

https://github.com/armijnhemel/binaryanalysis-ng/blob/master/src/test/testdata/unpackers/xar/test-bzip2.xar
https://github.com/armijnhemel/binaryanalysis-ng/blob/master/src/test/testdata/unpackers/xar/test-gzip.xar
https://github.com/armijnhemel/binaryanalysis-ng/blob/master/src/test/testdata/unpackers/xar/test-none.xar

I would need to search for test files for the Apple version.

archive/xar_header.ksy
```diff
@@ -0,0 +1,89 @@
...
+  From Wikipedia:
+
+  "XAR (short for eXtensible ARchive format) is an open source file archiver
+  and the archiver’s file format. It was created within the OpenDarwin project
+  and is used in macOS X 10.5 and up for software installation routines, as
+  well as browser extensions in Safari 5.0 and up."
+
+  It should be noted that there is a different version from Apple uses the
+  same version number as the (unmaintained) version described here, but that
+  supports additional checksums, but doesn't have checksum_name.
```
generalmimon, line 26: Pretty complicated sentence :-)

archive/xar.ksy
```diff
@@ -0,0 +1,124 @@
...
+  From Wikipedia:
+
+  "XAR (short for eXtensible ARchive format) is an open source file archiver
+  and the archiver’s file format. It was created within the OpenDarwin project
+  and is used in macOS X 10.5 and up for software installation routines, as
+  well as browser extensions in Safari 5.0 and up."
+
+  It should be noted that there is a different version from Apple which uses
+  the same version number as the (unmaintained) version described here and
+  supports additional checksums, but doesn't have `checksum_alg_name`.
```
generalmimon, line 27: @armijnhemel Is there a source of that information, a sample file, description or a source code snippet supporting this?

armijnhemel: Of course. The original:

https://github.com/mackyle/xar/blob/master/xar/include/xar.h.in#L54

and the Apple version:

https://github.com/apple-opensource-mirror/xar/blob/master/xar/include/xar.h.in#L53

The important difference is `struct xar_header_ex`. The header file says:

```
Only used when cksum_alg == XAR_CKSUM_OTHER
```
and includes an extra field `toc_cksum_name[36]`.

It is fairly easy to generate a new test file that shows this, for example:

```
$ xar -c --file-cksum=sha224 -f ls.xar ls
```

generalmimon: OK, I took the sentence verbatim so I thought that there is a yet another format that _"supports additional checksums **but** doesn't have `checksum_alg_name`"_. 

I would remove this sentence entirely, I guess this is specified enough by the .ksy logic.

archive/xar.ksy
```diff
@@ -0,0 +1,120 @@
+meta:
+  id: xar
+  title: XAR (eXtensible ARchive)
+  file-extension:
+    - pkg
+    - xar
+    - xip
```
generalmimon, line 7: By convention, `file-extension` entries should be listed in the order from the most typical/popular extension to the least common one (https://doc.kaitai.io/ksy_style_guide.html#meta):

> `file-extension` — if there’s only one extension, MUST be a string; if there are several, MUST be an sequence in block form and SHOULD order extensions from most popular extension to least popular.

I wonder if it makes sense to put `pkg` above `xar` here?

armijnhemel: because `pkg` is used for many other formats as well I am not so sure about that. I have no idea which extension is more popular these days. I would say that `xar` is at least the most descriptive.

PR #475, opened 2021-04-15
GIF: process application authentication (GIF89a specification section 26)

armijnhemel:
This further processes application extensions as defined in section 26 of the GIF89a specification. Application extensions include things like animation, XMP data, and so on.

### DIFF
--- a/image/gif.ksy
+++ b/image/gif.ksy
@@ -88,6 +88,7 @@ types:
       - id: blue
         type: u1
   block:
+    -webide-representation: '{block_type}'
     seq:
       - id: block_type
         type: u1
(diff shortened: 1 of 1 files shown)


armijnhemel:
The XMP application extension abuses the format and are constructed in such a way that `len_bytes` of a subblock should also be read to get the full XML out of the file.

It would be nice to add some sort of identifier that allows me to access all bytes in a subblock (so `len_bytes` and `bytes` concatenated), but only if the application identifier is "XMP Data" and the application authentication code is "XMP".

GreyCat:
Thanks, looks good to me!

PR #477, opened 2021-04-17
add Mozilla ARchive file format

armijnhemel:
A fairly simple description of the Mozilla ARchive format. This is not in widespread use. I have tested it with several files from the Mozilla FTP site as well as https://github.com/mozilla/build-mar/tree/master/tests

### DIFF
--- a/archive/mozilla_mar.ksy
+++ b/archive/mozilla_mar.ksy
@@ -1,0 +1,116 @@
+meta:
+  id: mozilla_mar
+  title: Mozilla ARchive
+  file-extension: mar
+  license: CC0-1.0
+  endian: be
+doc: |
+  Mozilla ARchive file is Mozilla's own archive format to distribute software updates.
+  Test files can be found on Mozilla's FTP site, for example:
+
+  <http://ftp.mozilla.org/pub/firefox/nightly/partials/>
+doc-ref: https://wiki.mozilla.org/Software_Update:MAR
+seq:
+  - id: magic
+    contents: "MAR1"
+  - id: ofs_index
+    -orig-id: OffsetToIndex
+    type: u4
+  - id: file_size
+    -orig-id: FileSize
+    type: u8
+  - id: len_signatures
+    -orig-id: NumSignatures
+    type: u4
+  - id: signatures
+    type: signature
+    repeat: expr
+    repeat-expr: len_signatures
+  - id: len_additional_sections
+    -orig-id: NumAdditionalSections
+    type: u4
+  - id: additional_sections
+    type: additional_section
+    repeat: expr
+    repeat-expr: len_additional_sections
+instances:
+  index:
+    pos: ofs_index
+    type: mar_index
+types:
+  signature:
+    seq:
+      - id: algorithm
+        -orig-id: SignatureAlgorithmID
+        type: u4
+        enum: signature_algorithms
+      - id: len_signature
+        -orig-id: SignatureSize
+        type: u4
+      - id: signature
+        size: len_signature
+  additional_section:
+    seq:
+      - id: len_block
+        type: u4
+      - id: block_identifier
+        -orig-id: BlockIdentifier
+        type: u4
+        enum: block_identifiers
+      - id: bytes
+        size: len_block - len_block._sizeof - block_identifier._sizeof
+        type:
+          switch-on: block_identifier
+          cases:
+            block_identifiers::product_information_block: product_information_block
+  mar_index:
+    seq:
+      - id: len_index
+        -orig-id: IndexSize
+        type: u4
+      - id: index_entries
+        type: index_entries
+        size: len_index
+  index_entries:
+    seq:
+      - id: index_entry
+        type: index_entry
+        repeat: eos
+  index_entry:
+    seq:
+      - id: ofs_content
+        -orig-id: OffsetToContent
+        type: u4
+      - id: len_content
+        -orig-id: ContentSize
+        type: u4
+      - id: flags
+        type: u4
+        doc: File permission bits (in standard unix-style format).
+      - id: file_name
+        -orig-id: FileName
+        type: strz
+        encoding: UTF-8
+    instances:
+      body:
+        pos: ofs_content
+        io: _root._io
+        size: len_content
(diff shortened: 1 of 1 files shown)


archive/mozilla_mar.ksy
```diff
@@ -0,0 +1,116 @@
...
+  id: mozilla_mar
+  title: Mozilla ARchive
+  file-extension: mar
+  license: CC0-1.0
+  endian: be
+doc: |
+  Mozilla ARchive file is Mozilla's own archive format to distribute software updates.
+  Test files can be found on Mozilla's FTP site, for example:
+
+  <http://ftp.mozilla.org/pub/firefox/nightly/partials/>
```
KOLANICH, line 11: https. Please install https://www.eff.org/https-everywhere

GreyCat:
Looks good to me! Thanks for this contribution!

archive/mozilla_mar.ksy
```diff
@@ -0,0 +1,116 @@
...
+      - id: len_content
+        -orig-id: ContentSize
+        type: u4
+      - id: flags
+        type: u4
+        doc: File permission bits (in standard unix-style format).
+      - id: file_name
+        -orig-id: FileName
+        type: strz
+        encoding: UTF-8
```
generalmimon, line 93: There is no string in other than `UTF-8` encoding in the spec, so why not declare it only in one place in the top-level `/meta/encoding` key, and remove all field-level overrides?

```diff
 meta:
@@ ... @@
   license: CC0-1.0
+  encoding: UTF-8
   endian: be
```

[`archive/mozilla_mar.ksy:90-93`](https://github.com/armijnhemel/kaitai_struct_formats/blob/3b64f2577049b5f941da6cfe347c635778e1745c/archive/mozilla_mar.ksy#L90-L93)

```diff
       - id: file_name
         -orig-id: FileName
         type: strz
-        encoding: UTF-8
```

[`archive/mozilla_mar.ksy:101-110`](https://github.com/armijnhemel/kaitai_struct_formats/blob/3b64f2577049b5f941da6cfe347c635778e1745c/archive/mozilla_mar.ksy#L101-L110)

```diff
       - id: mar_channel_name
         -orig-id: MARChannelName
         size: 64
         type: strz
-        encoding: UTF-8
       - id: product_version
         -orig-id: ProductVersion
         size: 32
         type: strz
-        encoding: UTF-8
```

archive/mozilla_mar.ksy
```diff
@@ -0,0 +1,116 @@
+meta:
+  id: mozilla_mar
+  title: Mozilla ARchive
```
generalmimon, line 3: Probably Firefox can be filled in as a `/meta/application` connected with this format?

archive/mozilla_mar.ksy
```diff
@@ -0,0 +1,116 @@
...
+        -orig-id: IndexSize
+        type: u4
+      - id: index_entries
+        type: index_entries
+        size: len_index
+  index_entries:
+    seq:
+      - id: index_entry
+        type: index_entry
+        repeat: eos
```
generalmimon, line 78: This field should be named `index_entries`. It stores an array of objects, so the name should be in a plural form (https://doc.kaitai.io/ksy_style_guide.html#attr-id).

archive/mozilla_mar.ksy
```diff
@@ -0,0 +1,116 @@
...
+      - id: block_identifier
+        -orig-id: BlockIdentifier
+        type: u4
+        enum: block_identifiers
+      - id: bytes
+        size: len_block - len_block._sizeof - block_identifier._sizeof
+        type:
+          switch-on: block_identifier
+          cases:
+            block_identifiers::product_information: product_information_block
```
generalmimon, line 65: This should not be called `bytes` since it isn't always a byte array (if `block_identifier == block_identifiers::product_information`, you then access the `product_information_block` structure under property `bytes` which doesn't make any sense). It's common to call fields like that a `body`.

archive/mozilla_mar.ksy
```diff
@@ -0,0 +1,116 @@
...
+  mar_index:
+    seq:
+      - id: len_index
+        -orig-id: IndexSize
+        type: u4
+      - id: index_entries
+        type: index_entries
+        size: len_index
```
generalmimon, line 73: It's weird to have a field `len_index` inside a type `mar_index`. Just from the naming, I would assume that the `len_index` apparently means the size of the `mar_index` structure (so the byte size of `index_entries` field in this type would be `size: len_index - len_index._sizeof`). However, this is not the case and `len_index` is used as the size of `index_entries` substream, so why is `len_index` not called `len_index_entries`? If you strictly followed the naming convention of `len_` meaning **size** and `num_` for **number / count**, there would be no ambiguity.

```suggestion
  mar_index:
    seq:
      - id: len_index_entries
        -orig-id: IndexSize
        type: u4
      - id: index_entries
        type: index_entries
        size: len_index_entries
```

generalmimon (commented):
There are value constraints specified in the reference document which might be worth checking as well (https://wiki.mozilla.org/Software_Update:MAR#Constraints):

> ## Constraints
> 
> To protect against invalid inputs the following constraints are in place:
> 
> - There are at most 8 signatures.
> - The file size of the MAR file is at most 500MB.
> - No signature is more than 2048 bytes long.

At least the 1st and 3rd restriction can be checked easily with `valid/max` on the fields `num_signatures`, `len_signature` respectively.

archive/mozilla_mar.ksy
```diff
@@ -0,0 +1,116 @@
...
+        type: index_entry
+        repeat: eos
+  index_entry:
+    seq:
+      - id: ofs_content
+        -orig-id: OffsetToContent
+        type: u4
+      - id: len_content
+        -orig-id: ContentSize
+        type: u4
```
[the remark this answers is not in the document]
armijnhemel: probably better to rename "body" to "content", as that is more in line with the specification?

archive/mozilla_mar.ksy
armijnhemel: It is any Mozilla product really.

PR #478, opened 2021-04-19
Added ATA over Ethernet.

KOLANICH:


network/ata_over_ethernet.ksy
```diff
@@ -0,0 +1,515 @@
...
+        -orig-id: aoe_cfghdr
+        seq:
+          - id: buffer_size_in_messages
+            -orig-id:
+              - Buffer Count
+              - bufcnt
+            type: u2
+            doc: Count of messages that can be buffered. Messages exceeding buffer are dropped.
+            valid:
+              expr: _ == 0 or 1
```
generalmimon, line 313: This is guaranteed **not** to work. I don't know if such ambiguous things work in Python (if the syntax `2 < 4 <= 5` works there, then anything could I guess), but not in Kaitai Struct.

What you expect to happen - `(_ == 0) or 1`, or `_ == (0 or 1)`? The `or` operator only works properly on booleans in KS.

---

I just checked that the compiler allows pretty much _any_ (i.e. anyhow invalid) expression in `valid/expr` for some reason. And the problem is not only in the underscore: when I substitute `_` for the actual attribute name (so that there should be no problem in deriving the type), it makes no difference:

```ksy
  - id: a
    type: u1
    valid:
      expr: a == 0 or 2
```

The above compiles to

```javascript
    if (!( ((this.a == 0) || (2)) )) {
      throw new KaitaiStream.ValidationExprError(this.a, this._io, "/seq/0");
    }
```

without a word. AFAIR probably there is some "precompile" class `TypeDetector`/`TypeValidator` in the compiler that handles this and apparently only looks for the expressions in certain YAML keys, so it needs to be adapted for the new key supporting expressions (`valid`).

But actually, the missing type checks of `valid` expressions is the only thing why https://github.com/kaitai-io/kaitai_struct/issues/435 is still open. So it's nothing too surprising that this is silently allowed.

KOLANICH: `1` was a stub (and I meant it to be considered `true`) to remind me to fill it in. I planned to insert there a condition and have forgotten. In some queries that field (and some other ones, such as firmware version ) must be 0.

PR #479, opened 2021-04-20
Update to use doc-refs

Icekhaos:
Replace comments with doc-refs, since that exists. Did the author even know about this?

### DIFF
--- a/filesystem/ext2.ksy
+++ b/filesystem/ext2.ksy
@@ -11,7 +11,7 @@ meta:
   license: CC0-1.0
   endian: le
 instances:
-  # http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/ext2.txt#n106
+  doc-ref: http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/ext2.txt#n106
   bg1:
     pos: 1024
     type: block_group
@@ -23,14 +23,15 @@ types:
       - id: super_block
         type: super_block_struct
         size: 1024
-      # http://www.nongnu.org/ext2-doc/ext2.html#BLOCK-GROUP-DESCRIPTOR-TABLE
+        doc-ref: http://www.nongnu.org/ext2-doc/ext2.html#BLOCK-GROUP-DESCRIPTOR-TABLE
       - id: block_groups
         type: bgd
         repeat: expr
         repeat-expr: super_block.block_group_count
-  # http://www.nongnu.org/ext2-doc/ext2.html#super_block
-  # http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/fs/ext2/ext2.h#n416
-  # https://ext4.wiki.kernel.org/index.php/Ext4_Disk_Layout#The_Super_Block
+        doc-ref:
+        - http://www.nongnu.org/ext2-doc/ext2.html#super_block
+        - http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/fs/ext2/ext2.h#n416
+        - https://ext4.wiki.kernel.org/index.php/Ext4_Disk_Layout#The_Super_Block
   super_block_struct:
     seq:
       - id: inodes_count
@@ -142,8 +143,9 @@ types:
         1: act_continue
         2: act_ro
         3: act_panic
-  # http://www.nongnu.org/ext2-doc/ext2.html#BLOCK-GROUP-DESCRIPTOR-STRUCTURE
-  # http://web.archive.org/web/20160804172310/http://virtualblueness.net/Ext2fs-overview/Ext2fs-overview-0.1-7.html
+      doc-ref:
+      - http://www.nongnu.org/ext2-doc/ext2.html#BLOCK-GROUP-DESCRIPTOR-STRUCTURE
+      - http://web.archive.org/web/20160804172310/http://virtualblueness.net/Ext2fs-overview/Ext2fs-overview-0.1-7.html
   bgd:
     seq:
       - id: block_bitmap_block
@@ -167,8 +169,9 @@ types:
       inode_bitmap:
         pos: inode_bitmap_block * _root.bg1.super_block.block_size
         size: 1024
-      # http://www.nongnu.org/ext2-doc/ext2.html#INODE-TABLE
-      # http://web.archive.org/web/20161114202411/http://www.virtualblueness.net/Ext2fs-overview/Ext2fs-overview-0.1-10.html
+        doc-ref:
+        - http://www.nongnu.org/ext2-doc/ext2.html#INODE-TABLE
+        - http://web.archive.org/web/20161114202411/http://www.virtualblueness.net/Ext2fs-overview/Ext2fs-overview-0.1-10.html
       inodes:
         pos: inode_table_block * _root.bg1.super_block.block_size
         type: inode
@@ -232,7 +235,7 @@ types:
     seq:
       - id: body
         size: _root.bg1.super_block.block_size
-  # http://www.nongnu.org/ext2-doc/ext2.html#LINKED-DIRECTORY-ENTRY-STRUCTURE
+        doc-ref: http://www.nongnu.org/ext2-doc/ext2.html#LINKED-DIRECTORY-ENTRY-STRUCTURE
   dir:
     seq:
       - id: entries
@@ -259,7 +262,6 @@ types:
       inode:
         value: '_root.bg1.block_groups[(inode_ptr - 1) / _root.bg1.super_block.inodes_per_group].inodes[(inode_ptr - 1) % _root.bg1.super_block.inodes_per_group]'
     enums:
-      # http://www.nongnu.org/ext2-doc/ext2.html#IFDIR-FILE-TYPE
       file_type_enum:
         0: unknown
         1: reg_file
@@ -269,3 +271,4 @@ types:
         5: fifo
         6: sock
         7: symlink
+      doc-ref: http://www.nongnu.org/ext2-doc/ext2.html#IFDIR-FILE-TYPE


filesystem/ext2.ksy
```diff
@@ -23,14 +23,15 @@ types:
...
-      # http://www.nongnu.org/ext2-doc/ext2.html#BLOCK-GROUP-DESCRIPTOR-TABLE
+        doc-ref: http://www.nongnu.org/ext2-doc/ext2.html#BLOCK-GROUP-DESCRIPTOR-TABLE
       - id: block_groups
         type: bgd
         repeat: expr
         repeat-expr: super_block.block_group_count
-  # http://www.nongnu.org/ext2-doc/ext2.html#super_block
-  # http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/fs/ext2/ext2.h#n416
-  # https://ext4.wiki.kernel.org/index.php/Ext4_Disk_Layout#The_Super_Block
+        doc-ref:
```
KOLANICH, line 31: wrong indent

filesystem/ext2.ksy
```diff
@@ -142,8 +143,9 @@ types:
         1: act_continue
         2: act_ro
         3: act_panic
-  # http://www.nongnu.org/ext2-doc/ext2.html#BLOCK-GROUP-DESCRIPTOR-STRUCTURE
-  # http://web.archive.org/web/20160804172310/http://virtualblueness.net/Ext2fs-overview/Ext2fs-overview-0.1-7.html
+      doc-ref:
+      - http://www.nongnu.org/ext2-doc/ext2.html#BLOCK-GROUP-DESCRIPTOR-STRUCTURE
```
KOLANICH, line 147: wrong indent

filesystem/ext2.ksy
```diff
@@ -269,3 +271,4 @@ types:
         5: fifo
         6: sock
         7: symlink
+      doc-ref: http://www.nongnu.org/ext2-doc/ext2.html#IFDIR-FILE-TYPE
```
KOLANICH, line 274: It is invalid.

filesystem/ext2.ksy
```diff
@@ -167,8 +169,9 @@ types:
       inode_bitmap:
         pos: inode_bitmap_block * _root.bg1.super_block.block_size
         size: 1024
-      # http://www.nongnu.org/ext2-doc/ext2.html#INODE-TABLE
-      # http://web.archive.org/web/20161114202411/http://www.virtualblueness.net/Ext2fs-overview/Ext2fs-overview-0.1-10.html
+        doc-ref:
+        - http://www.nongnu.org/ext2-doc/ext2.html#INODE-TABLE
```
KOLANICH, line 173: wrong indent.

KOLANICH (changes requested):
Please fix the issues.

filesystem/ext2.ksy
Icekhaos: Invalid how?

filesystem/ext2.ksy
Icekhaos: Wouldn't a doc-ref a level down cause it to count as an enum? OR should I go a level higher?

generalmimon: @Icekhaos  See [KSY Style Guide > 1. General formatting](https://doc.kaitai.io/ksy_style_guide.html#general):

>
> <table>
> <thead>
> <tr>
> <th align=left>✔️ Good</th>
> <th align=left>❌ Bad</th>
> </tr>
> </thead>
> <tbody>
> <tr>
> <td>
> 
> ```yaml
> imports:
>   - foo
>   - bar
> ```
>
> </td>
> <td>
> 
> ```yaml
> imports:
> - foo
> - bar
> ```
>
> </td>
> </tr>
> </tbody>
> </table>

filesystem/ext2.ksy
KOLANICH: It is within `enums` section. In it keys are names of enums and values are dicts defining enums.

filesystem/ext2.ksy
```diff
@@ -23,14 +23,15 @@ types:
...
         repeat: expr
         repeat-expr: super_block.block_group_count
-  # http://www.nongnu.org/ext2-doc/ext2.html#super_block
-  # http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/fs/ext2/ext2.h#n416
-  # https://ext4.wiki.kernel.org/index.php/Ext4_Disk_Layout#The_Super_Block
+        doc-ref:
+        - http://www.nongnu.org/ext2-doc/ext2.html#super_block
+        - http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/fs/ext2/ext2.h#n416
+        - https://ext4.wiki.kernel.org/index.php/Ext4_Disk_Layout#The_Super_Block
   super_block_struct:
```
generalmimon, line 35: I strongly suspect that these links are not meant to be associated with the attribute `block_groups` but rather with the following type `super_block_struct`. See the original version of this spec:

https://github.com/kaitai-io/kaitai_struct_formats/blob/2eec205abad07f9151fa5097083e4e549d09673f/filesystem/ext2.ksy#L26-L37

filesystem/ext2.ksy
```diff
@@ -11,7 +11,7 @@ meta:
   license: CC0-1.0
   endian: le
 instances:
-  # http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/ext2.txt#n106
+  doc-ref: http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/ext2.txt#n106
```
generalmimon, line 14: Also invalid (the same problem as [`filesystem/ext2.ksy:138-148`](https://github.com/kaitai-io/kaitai_struct_formats/blob/57f08e40704deddd3d854d61ef3d31310875fc2d/filesystem/ext2.ksy#L138-L148) and [`filesystem/ext2.ksy:264-274`](https://github.com/kaitai-io/kaitai_struct_formats/blob/57f08e40704deddd3d854d61ef3d31310875fc2d/filesystem/ext2.ksy#L264-L274)). You can't put `doc-ref` directly under `instances`.

This won't even compile - try this snippet in the [Web IDE](https://ide.kaitai.io/):

```ksy
meta:
  id: test
instances:
  doc-ref: https://kaitai.io/
  a:
    pos: 0
    type: u1
```

You'll get

```
io.kaitai.struct.format.YAMLParseException: /instances/doc-ref: invalid instance ID: 'doc-ref', expected /^[a-z][a-z0-9_]*$/
```

...since `instances` is merely a `identifier->declaration` dictionary (or _map_). There are no identifiers with special meaning here.

filesystem/ext2.ksy
```diff
@@ -23,14 +23,15 @@ types:
       - id: super_block
         type: super_block_struct
         size: 1024
-      # http://www.nongnu.org/ext2-doc/ext2.html#BLOCK-GROUP-DESCRIPTOR-TABLE
+        doc-ref: http://www.nongnu.org/ext2-doc/ext2.html#BLOCK-GROUP-DESCRIPTOR-TABLE
       - id: block_groups
         type: bgd
         repeat: expr
         repeat-expr: super_block.block_group_count
```
generalmimon, line 30: The URL http://www.nongnu.org/ext2-doc/ext2.html#BLOCK-GROUP-DESCRIPTOR-TABLE is **not** associated with the field `super_block` but rather with the following field `block_groups`.

Please use your brain a bit - if the URL screams <code>http\://www.nongnu.org/ext2-doc/ext2.html#**BLOCK-GROUP**-DESCRIPTOR-TABLE</code>, don't bind it to the previous attribute.

generalmimon (changes requested):
Also this URL http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/ext2.txt#n106 and all anchors on page http://www.nongnu.org/ext2-doc/ext2.html no longer work, can you please replace them?

filesystem/ext2.ksy
Icekhaos: Oh, okay.

filesystem/ext2.ksy
Icekhaos: I just redid it in-place, so blame the original author.

Icekhaos:
I'll see about replacing invalid links, but be aware I am not the original author of this spec, and just saw weirds regarding comments instead of doc-refs.

Icekhaos:
And I just realized I forgot a couple. I should really sleep.

filesystem/ext2.ksy
generalmimon: > I just redid it in-place, so blame the original author.

Not really. The _original author_ was just consistently putting reference links always **above** the concerned element. This makes quite sense, because it follows the convention common in programming languages, for example see https://google.github.io/styleguide/jsguide.html#jsdoc-method-and-function-comments:

```js
/** A class that does something. */
class SomeClass extends SomeBaseClass {
  /**
   * Operates on an instance of MyClass and returns something.
   * @param {!MyClass} obj An object that for some reason needs detailed
   *     explanation that spans multiple lines.
   * @param {!OtherClass} obviousOtherClass
   * @return {boolean} Whether something occurred.
   */
  someMethod(obj, obviousOtherClass) { ... }

  /** @override */
  overriddenMethod(param) { ... }
}

/**
 * Demonstrates how top-level functions follow the same rules.  This one
 * makes an array.
 * @param {TYPE} arg
 * @return {!Array<TYPE>}
 * @template TYPE
 */
function makeArray(arg) { ... }
```

Notice that the class and function comments also always come **above** the class or function. If we weren't using YAML and didn't have designated keys `doc` and `doc-ref` for these comments, I'd also recommend this approach. And I'd still use it on places where `doc` and `doc-ref` can't be used, i.e. for enum declarations:

```ksy
enums:
  # https://example.com/fruit-enum.html
  fruit:
    1: apple
    2: banana
```

And you can't just thoughtlessly do some find-and-replace `# http` with `doc-ref: http` and expect to get reasonable results. You always need to check what you're committing, especially when you used some automated tools to make them. _Find and replace_ is only a tool to facilitate tedious work, it won't autonomously work in place of you. It's like buying a Tesla Autopilot and climbing into the back seat while driving, leaving the driver seat empty and letting the vehicle to drive on its own - that's something incredibly stupid, because that not only endangers your own life but also the lives of other people.

PR #480, opened 2021-04-20
Couple stray doc-refs

Icekhaos:
Same as last PR, just a different file.

### DIFF
--- a/archive/zip.ksy
+++ b/archive/zip.ksy
@@ -199,7 +199,7 @@ types:
       local_header:
         pos: ofs_local_header
         type: pk_section
-  # https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT - 4.3.16
+        doc-ref: https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT - 4.3.16
   end_of_central_dir:
     seq:
       - id: disk_of_end_of_central_dir
@@ -335,7 +335,6 @@ enums:
     19: ibm_lz77_z
     98: ppmd
   extra_codes:
-    # https://github.com/LuaDist/zip/blob/b710806/proginfo/extrafld.txt
     0x0001: zip64
     0x0007: av_info
 #    0x0008: reserved for extended language encoding data (PFS) (see APPENDIX D)
@@ -357,3 +356,4 @@ enums:
     0x5455: extended_timestamp
     0x7855: infozip_unix
     0x7875: infozip_unix_var_size
+  doc-ref: https://github.com/LuaDist/zip/blob/b710806/proginfo/extrafld.txt


KOLANICH (changes requested):
The same issues as in the previous PR.

archive/zip.ksy
```diff
@@ -357,3 +356,4 @@ enums:
     0x5455: extended_timestamp
     0x7855: infozip_unix
     0x7875: infozip_unix_var_size
+doc-ref: https://github.com/LuaDist/zip/blob/b710806/proginfo/extrafld.txt
```
generalmimon, line 359: Wrong indentation. This way (with no indent), it will be parsed as if it was at the top level, so it will actually overwrite this `doc-ref` at the beginning of the file, which is really meant to apply to the entire format:

https://github.com/kaitai-io/kaitai_struct_formats/blob/bc8098d6e43b574c98c2d83fc894f2ec0680a670/archive/zip.ksy#L33-L35

Check for yourself at [https://yaml-online-parser.appspot.com/](https://yaml-online-parser.appspot.com/?type=json&url=https%3A%2F%2Fraw.githubusercontent.com%2Fkaitai-io%2Fkaitai_struct_formats%2Fbc8098d6e43b574c98c2d83fc894f2ec0680a670%2Farchive%2Fzip.ksy). The _Output_ looks like this:

```js
{
  "meta": {
    // ...
  },
  "seq": [
    {
      "repeat": "eos", 
      "type": "pk_section", 
      "id": "sections"
    }
  ],
  "doc": "ZIP is a popular archive file format ...",
  "enums": {
    // ...
  },
  "doc-ref": "https://github.com/LuaDist/zip/blob/b710806/proginfo/extrafld.txt",
  // ...
}
```

Note that

```json
  "doc-ref": [
    "https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT", 
    "https://users.cs.jmu.edu/buchhofp/forensics/formats/pkzip.html"
  ],
```

is nowhere to be found - it has been silently overwritten by the later `doc-ref` key on the top level (i.e. by the `doc-ref: https://github.com/LuaDist/zip/blob/b710806/proginfo/extrafld.txt`). Most YAML parsers allow duplicate keys and they keep only one of them (either the last or first one). However, as of 0.9, the JVM version of the `kaitai-struct-compiler` (which uses SnakeYAML as the YAML parser under the hood) even refuses any KSY spec that contains duplicate keys, if I'm not mistaken. So this spec `zip.ksy` would no longer compile.

---

Anyway, `doc` and `doc-ref` are not supported currently for an enum as a whole (see https://github.com/kaitai-io/kaitai_struct/issues/358), so I suggest that you put it above the enum declaration (as it seems common in the programming world, see https://github.com/kaitai-io/kaitai_struct_formats/pull/479#discussion_r616522477). Like this:

```diff
+  # https://github.com/LuaDist/zip/blob/b710806/proginfo/extrafld.txt
   extra_codes:
-    # https://github.com/LuaDist/zip/blob/b710806/proginfo/extrafld.txt
     0x0001: zip64
     0x0007: av_info
```

IMHO it makes slightly more sense than the previous version.

generalmimon (changes requested):

archive/zip.ksy
```diff
@@ -334,6 +334,8 @@ enums:
     18: ibm_terse
     19: ibm_lz77_z
     98: ppmd
+  # doc-ref: https://github.com/LuaDist/zip/blob/b710806/proginfo/extrafld.txt
+  # doc-refs in enums are not currently supported. Note to self: Update when it is supported.
```
KOLANICH, line 338: Create/find an issue about it and use `-affected-by`

generalmimon: @KOLANICH https://github.com/kaitai-io/kaitai_struct/issues/358, see https://github.com/kaitai-io/kaitai_struct_formats/pull/480#discussion_r616613528

Icekhaos: And what do I do to use this ``-affected-by``?

PR #481, opened 2021-04-22
Anti-Virus Quarantine File Formats

fbausch:
Hi,
I recently published a research work about file formats of anti-virus software quarantine.

The research is described here: https://static.ernw.de/whitepaper/ERNW-Whitepaper-71_AV_Quarantine_signed.pdf
Documentation and repo are here: https://github.com/ernw/quarantine-formats

My merge request includes the parsers for:

* Avira
* Windows Defender
* Malwarebytes
* Symantec Endpoint Protection
* G Data
* Sophos Antivirus
* Kaspersky for Windows Server

Please note: some parsers require an external RC4 module and will not work out of the box.

Additionally, I created a parser for Binary Security Descriptors as used by Windows to describe file permissions.
Furthermore, I improved the Registry Hive parser. I found that the current version did not work for my registry hive samples.

PR #484, opened 2021-04-23
Add Avira Quarantine file parser

fbausch:
A parser for the Avira Antivirus quarantine files.

### DIFF
--- a/security/avira.ksy
+++ b/security/avira.ksy
@@ -1,0 +1,50 @@
+meta:
+  id: qua
+  file-extension: qua
+  endian: le
+  title: Avira Antivirus quarantine file parser
+  license: CC-BY-SA-4.0
+  ks-version: 0.9
+doc: |
+  Creator: Florian Bausch, ERNW Research GmbH, https://ernw-research.de
+  License: CC-BY-SA-4.0 https://creativecommons.org/licenses/by-sa/4.0/
+seq:
+  - id: magic
+    size: 16
+    contents: ["AntiVir Qua", 0x00, 0x00, 0x00, 0x00, 0x00]
+  - id: malicious_offset
+    type: u4
+  - id: len_filename
+    type: u4
+  - id: len_addl_info
+    type: u4
+  - id: unknown1
+    size: 0x20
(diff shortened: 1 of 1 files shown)


fbausch:
See also #481

security/avira.ksy
```diff
@@ -0,0 +1,50 @@
...
+  license: CC-BY-SA-4.0
+  ks-version: 0.9
+doc: |
+  Creator: Florian Bausch, ERNW Research GmbH, https://ernw-research.de
+  License: CC-BY-SA-4.0 https://creativecommons.org/licenses/by-sa/4.0/
+seq:
+  - id: magic
+    size: 16
+    contents: ["AntiVir Qua", 0x00, 0x00, 0x00, 0x00, 0x00]
+  - id: malicious_offset
```
KOLANICH, line 15: Could you add  `doc`s, please?

security/avira.ksy
```diff
@@ -0,0 +1,59 @@
...
+  License: MIT
+  https://github.com/ernw/quarantine-formats/blob/master/docs/Avira_Antivirus.md
+  Avira Antivirus quarantine files are usually stored under /ProgramData/Avira/Antivirus/INFECTED.
+  The file extension usually is .qua.
+seq:
+  - id: magic
+    size: 16
+    contents: ["AntiVir Qua", 0x00, 0x00, 0x00, 0x00, 0x00]
+  - id: malicious_offset
+    type: u4
```
KOLANICH, line 19: Why isn't it used?

fbausch: Because the parser also works without using it. I assume the offset is stored there for a quicker lookup of the encrypted malware file, e.g. when restoring the file.

KOLANICH: or maybe it is done for compatibility and some other fields are possible.

fbausch:
I added an instance that also gets the malware sample form the file but uses the malicious_offset field to determine the position. So, afterwards, one can programmatically check whether mal_file and mal_file_offset hold the same data. If not, the structure of the file is different than expected by the parser.

PR #485, opened 2021-04-23
Add G Data quarantine file parser

fbausch:
A parser for G Data quarantine files.

This parser requires an external, custom RC4 module.

### DIFF
--- a/security/gdata.ksy
+++ b/security/gdata.ksy
@@ -1,0 +1,96 @@
+meta:
+  id: gdata
+  file-extension: q
+  endian: le
+  title: G Data quarantine file parser
+  license: CC-BY-SA-4.0
+  ks-version: 0.9
+doc: |
+  Creator: Florian Bausch, ERNW Research GmbH, https://ernw-research.de
+  License: CC-BY-SA-4.0 https://creativecommons.org/licenses/by-sa/4.0/
+seq:
+  - id: magic1
+    size: 4
+    contents: [0xca, 0xfe, 0xba, 0xbe]
+  - id: len_data1
+    type: u4
+  - id: data1
+    size: len_data1
+    type: encrypted_data1
+    process: util.custom_arc4.custom_arc4([0xA7, 0xBF, 0x73, 0xA0, 0x9F, 0x03, 0xD3, 0x11,
+                                           0x85, 0x6F, 0x00, 0x80, 0xAD, 0xA9, 0x6E, 0x9B])
+  - id: magic2
+    size: 4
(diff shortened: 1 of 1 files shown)


fbausch:
See also #481

security/gdata.ksy
```diff
@@ -0,0 +1,96 @@
...
+        type: winfiletime
+      - id: unknown6
+        type: u4
+      - id: filesize2
+        type: u4
+      - id: path
+        type: utf16le
+  utf16le:
+    seq:
+      - id: bom
```
KOLANICH, line 79: Is `utf-16be` possible here? (Otherwise I see no point to have a BOM here)

security/gdata.ksy
```diff
@@ -0,0 +1,96 @@
...
+types:
+  encrypted_data1:
+    seq:
+      - id: unknown1
+        type: u4
+      - id: unknown2
+        type: u4
+      - id: unknown3
+        type: u4
+      - id: quatime
```
KOLANICH, line 45: please, separate words in ids with underscores

security/gdata.ksy
```diff
@@ -0,0 +1,96 @@
...
+        type: u4
+      - id: unknown5
+        type: u4
+      - id: malwaretype
+        type: utf16le
```
KOLANICH, line 50: is it an enum ?

security/gdata.ksy
```diff
@@ -0,0 +1,96 @@
...
+        type: u4
+      - id: filesize
+        type: u4
+      - id: unknownstring1
+        type: utf16le
+      - id: unknown4
+        type: u4
+      - id: unkown5
+        type: u4
+      - id: time1
```
KOLANICH, line 65: can it be created/modified/accessed?

security/gdata.ksy
fbausch: I don't know, if there is any chance that there will ever be UTF-16be at that point. However, fact is that there is a BOM at that location.

security/gdata.ksy
fbausch: No, it's a string saying what malware family/type was detected by the AV.

security/gdata.ksy
KOLANICH: Could you test it by crafting a file with BE and checking if the software would correctly parse it?

fbausch: No, I didn't do this.

fbausch:
I addressed the comments in my last commit.

PR #488, opened 2021-04-23
Add Windows Defender quarantine file parsers

fbausch:
Three parsers to parse the quarantine files of Windows Defender.

These parsers require an external, custom RC4 module.

See also #481


security/windef_entries.ksy
```diff
@@ -0,0 +1,242 @@
...
+                                           0xDB, 0x2F, 0x35, 0xD3, 0xC1, 0xB9, 0xCE, 0xD5, 0x26, 0x36,
+                                           0xF2, 0x76, 0x5E, 0x1A, 0x95, 0xCB, 0x7C, 0xA4, 0xC3, 0xDD,
+                                           0xAB, 0xDD, 0xBF, 0xF3, 0x82, 0x53
+                                          ])
+types:
+  rc4encrypted_header:
+    seq:
+      - id: magic
+        size: 0x10
+        contents: [0xdb, 0xe8, 0xc5, 0x01, 0x01, 0, 0x01, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```
KOLANICH, line 106: `u4 (signature), u1, u1, u1, u1 (bools), u8`?

KOLANICH: Also, have you tried drltrace, if the fields are read individually, it can give some insight kn how they are read?

security/windef_entries.ksy
```diff
@@ -0,0 +1,242 @@
...
+      - id: character
+        type: u2
+        repeat: until
+        repeat-until: character[_index] == 0x0000
+  listelement:
+    seq:
+      - id: length
+        type: u2
+      - id: elementsubtype
+        type: u1
```
KOLANICH, line 193: are these enums?

security/windef_resource_data.ksy
```diff
@@ -0,0 +1,145 @@
...
+        size: len_malfile
+  binarysd:
+    seq:
+    - id: revision
+      type: u1
+    - id: reserved
+      size: 1
+      contents: [0]
+    - id: control_flags
+      type: u2
```
KOLANICH, line 70: should it be a type of bit fields?

security/windef_entries.ksy
fbausch: I did not try to find out how the header is read. But I never saw another header, so I assume it is static.

security/windef_entries.ksy
fbausch: I don't know what this field is doing, but it seems to describe somehow what this element is. However, I only saw that only "elementtype" seems to be relevant for parsing.

fbausch:
I added several improvements with the last commit

PR #489, opened 2021-04-23
Add parser for Windows Defender scan history files

fbausch:
Add a parser for Windows Defender scan history files.

See also #481

### DIFF
--- a/security/windef_scan_history.ksy
+++ b/security/windef_scan_history.ksy
@@ -1,0 +1,115 @@
+meta:
+  id: windef_scan_history
+  endian: le
+  title: Windows Defender scan history file parser
+  license: CC-BY-SA-4.0
+  ks-version: 0.9
+doc: |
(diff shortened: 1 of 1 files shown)


security/windef_scan_history.ksy
```diff
@@ -0,0 +1,115 @@
...
+        size: 2
+        if: 'length > 2'
+      - id: trackingtype
+        type: u4
+        enum: trackingtypes
+      - id: content
+        type:
+          switch-on: trackingtype
+          cases:
+            trackingtypes::flags: u4
```
KOLANICH, line 95: Should it be a type of bit-flags (`b1`)?

fbausch: Maybe, but I don't know how this works because I just see from a string in the file that it's supposed to be flags, but I have no indication how they work.

PR #499, opened 2021-06-14
Fix edid.ksy

N-Nagorny:
I explored https://en.wikipedia.org/wiki/Extended_Display_Identification_Data, fixed some issues and tested the result against a single EDID file.

`refresh_rate_mod` became `b6` because `u1` (`horiz_active_pixels_mod`) + `b2` (`aspect_ratio`) + `b6` (`refresh_rate_mod`) = `u2` as expected. Another evidence is that 123 - 60 = 63, and `b6` gives 2^6 = [0..63] in contrast to `b5` which gives 2^5 = [0..31].

If a standard timing equals to `01 01`, then it's marked as unused. It would be better to mark a whole `std_timing` as `null` in this case but I didn't find how to do that in http://doc.kaitai.io/ksy_diagram.html.

### DIFF
--- a/hardware/edid.ksy
+++ b/hardware/edid.ksy
@@ -54,6 +54,7 @@ seq:
       DMT (Discrete Monitor Timings) modes.
     doc-ref: Standard, section 3.8
   - id: std_timings
+    size: 2
     type: std_timing
     doc: |
       Array of descriptions of so called "standard timings", which are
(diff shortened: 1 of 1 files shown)


generalmimon (changes requested):

generalmimon (approved):
LGTM, thanks for contributing!

PR #500, opened 2021-06-24
Add InstallShield 3 Z archive and self-extracting installer formats

### LINKED ISSUE #328: InstallShield archive format version 3
```yaml
meta:
  id: install_shield_archive_v3
  title: InstallShield archive format version 3
  application:
    - InstallShield (very old versions)
  file-extension:
    - Z
  xref:
    justsolve:
      - InstallShield_Z 
    wikidata: Q1193296

doc: |
  A very old format used by InstallShield installers usually from 199x.
  This version uses the compression implemented in https://github.com/madler/zlib/tree/master/contrib/blast

doc-ref:
  - https://docs.revenera.com/eol/installshield.htm
  - https://www.sac.sk/download/pack/icomp95.zip # i3comp
  - https://github.com/DeclanHoare/stix
  - https://github.com/putara/isdecomp # !!!
  - https://github.com/lephilousophe/idecomp
  - https://github.com/wfr/unshieldv3
  - https://github.com/OmniBlade/isextract
  - https://github.com/agrif/unshield # rust
```

dgelessus:
Closes #328.

Some parts of install_shield_3_z are not tested very well. I only have a single test file in the "extended" format, and no test files that use multiple parts or a password, so those parts of the spec are almost completely untested.

Similarly, I tested install_shield_3_sfx_tail only with a few installer files that I was working with anyway. There are probably other variants of the self-extracting installer data format that aren't handled by this spec.

The dos_datetime_backwards helper spec currently doesn't compile correctly to Python, because of kaitai-io/kaitai_struct#876.

### DIFF
--- a/archive/install_shield_3_sfx_tail.ksy
+++ b/archive/install_shield_3_sfx_tail.ksy
@@ -1,0 +1,51 @@
+meta:
+  id: install_shield_3_sfx_tail
+  title: InstallShield 3 self-extracting installer tail data
+  endian: le
+  imports:
+    - /common/dos_datetime_backwards
+doc: |
+  The data format used in InstallShield 3 self-extracting installers.
+  These installers start with a normal Microsoft PE executable,
+  which is directly followed by the data in this format.
+
+  This is only a very thin wrapper around ZIP files.
+  These ZIP files can also be carved from the installer
+  using a program that can search for ZIP data embedded in a file,
+  such as 7-Zip in "parser mode" (type `#`) or Binwalk.
+doc-ref: http://kannegieser.net/veit/quelle/stix_src.arj # STIX.PAS, STSFX.PAS
+seq:
+  - id: files
+    type: file
+    repeat: eos
+instances:
+  path_encryption_key:
+    value: '[0xb3, 0xf2, 0xea, 0x1f, 0xaa, 0x27, 0x66, 0x13]'
+types:
+  file:
+    seq:
+      - id: len_path_encrypted
+        type: u4
+        doc: Byte length of the encrypted path name.
+      - id: path_encrypted
+        size: len_path_encrypted
+        doc: |
+          Path name for this file,
+          encrypted using a relatively simple algorithm.
+          The path name can be decrypted bytewise using the formula
+          `byte_rot_right((path_encrypted[i] ^ path_encryption_key[7-(i%8)]), 7-(i%8)) ^ path_encryption_key[i%8]`,
+          where `i` is the (0-based) index of the byte in question.
+      - id: modified
+        type: dos_datetime_backwards
+        doc: Modification date/time of the file.
+      - id: len_data
+        type: u4
+        doc: Byte length of the zipped file.
+      - id: data
+        size: len_data
+        doc: |
+          The zipped file.
+          The data should be a ZIP archive containing exactly one file
+          which should be located at top level
+          and whose name should match the one stored in path_encrypted
+          (without any leading directories).
--- a/archive/install_shield_3_z.ksy
+++ b/archive/install_shield_3_z.ksy
@@ -23,6 +23,9 @@ doc-ref:
   - https://github.com/wfr/unshieldv3/blob/0037ff581a4d862757ccc2c584aefdc0fe58a38d/installshieldarchivev3.cpp
   - https://github.com/adrium/unshieldv3/blob/9f47d1c501abb778eb40beb7fdbb24ae556b7c46/installshieldarchivev3.cpp
   - https://github.com/lephilousophe/idecomp/blob/354961c0cfe86632a39e4fea01e0ca521ffac3e0/idecomp.py
+  - https://github.com/putara/isdecomp
+  - https://github.com/agrif/unshield
+  - http://kannegieser.net/veit/quelle/stix_src.arj # STIX.PAS
 seq:
   - id: magic
     contents: [0x13, 0x5d, 0x65, 0x8c]
--- a/archive/install_shield_3_z.ksy
+++ b/archive/install_shield_3_z.ksy
@@ -1,0 +1,481 @@
+meta:
+  id: install_shield_3_z
+  title: InstallShield 3 Z archive
+  application: InstallShield 3
+  # Not to be confused with the Unix compress or pack compression formats,
+  # which use the uppercase Z and lowercase z extensions respectively.
+  file-extension: z
+  xref:
+    justsolve: InstallShield_Z
+  license: MIT
+  ks-version: 0.9
+  imports:
+    - /common/dos_datetime_backwards
+  endian: le
+  bit-endian: le
+doc: |
+  Archive format used by InstallShield 3 installers.
+  File data can optionally be compressed using the PKWARE DCL Implode algorithm.
+  This is not to be confused with the old ZIP Implode compression algorithm of the same name.
+doc-ref:
+  - https://github.com/OpenRA/OpenRA/blob/fe146cb77a25091ed342e38104292ac3cb19019b/OpenRA.Mods.Common/FileSystem/InstallShieldPackage.cs
+  - https://github.com/OmniBlade/isextract/blob/5adb0af87fb0aaad0b436db0b7e7356947af3ee2/src/isextract.cpp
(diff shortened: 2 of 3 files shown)


archive/install_shield_3_z.ksy
```diff
@@ -0,0 +1,484 @@
...
+  - https://github.com/lephilousophe/idecomp/blob/354961c0cfe86632a39e4fea01e0ca521ffac3e0/idecomp.py
+  - https://github.com/putara/isdecomp
+  - https://github.com/agrif/unshield
+  - http://kannegieser.net/veit/quelle/stix_src.arj # STIX.PAS
+seq:
+  - id: magic
+    contents: [0x13, 0x5d, 0x65, 0x8c]
+  - id: len_header
+    type: u1
+    valid: sizeof<header>
```
KOLANICH, line 34: Why not to assumme that someone can reuse this format and extend it, so instead use `size` in `header` field?

archive/install_shield_3_z.ksy
```diff
@@ -0,0 +1,484 @@
...
+          bits 3 and 4 mark volume labels and directories (respectively),
+          so they should never be set on files in an archive.
+
+          For some reason,
+          this field apparently always has bit 7 set if no other attributes are set.
+          This seems to have no meaning and can probably be ignored.
+      - id: len_entry
+        type: u2
+        doc: Total byte size of this file entry.
+      - id: reserved_flags_1
```
KOLANICH, line 373: It may make sense to pack the flags that are packed into a single integer into a separate type to highlight that fact.

archive/install_shield_3_z.ksy
```diff
@@ -0,0 +1,484 @@
...
+            is_split
+            ? (_ > 0 and _ < end_part)
+            : _ == end_part
+        doc: |
+          Number of the split archive part in which the first part of this file's data is stored.
+          Only used in extended format -
+          otherwise this field is 0
+          and the file's data is not split.
+      - id: len_name
+        type: u1
```
KOLANICH, line 428: `-affected-by: 84`

archive/install_shield_3_z.ksy
```diff
@@ -0,0 +1,484 @@
...
+              + 2
+              + reserved_1._sizeof
+              + start_part._sizeof
+              + len_name._sizeof
+              + name_terminator._sizeof
+              + version._sizeof
+              + reserved_2._sizeof
+            )
+        doc: Byte size of the file name.
+      - id: name
```
KOLANICH, line 450: is it always a string of a particular encoding, or is it really allowed to contain arbitrary sequencies of bytes?

archive/install_shield_3_sfx_tail.ksy
```diff
@@ -0,0 +1,51 @@
...
+      - id: len_path_encrypted
+        type: u4
+        doc: Byte length of the encrypted path name.
+      - id: path_encrypted
+        size: len_path_encrypted
+        doc: |
+          Path name for this file,
+          encrypted using a relatively simple algorithm.
+          The path name can be decrypted bytewise using the formula
+          `byte_rot_right((path_encrypted[i] ^ path_encryption_key[7-(i%8)]), 7-(i%8)) ^ path_encryption_key[i%8]`,
```
KOLANICH, line 36: It is not encryption (which implies that the algo is peer-reviewed and considered to be crypto by cryptographers, which implies that it has ever been considered secure by them, which is not the case here), it is just obfuscation.

archive/install_shield_3_sfx_tail.ksy
```diff
@@ -0,0 +1,51 @@
+meta:
```
KOLANICH, line 1: `license` is missing.

KOLANICH:
Thanks for implementing this!

KOLANICH:
I have extracted into a separate repo and packaged https://github.com/KOLANICH/libblast , one can generate packages for Debian/Ubuntu using CPack, Fedora wasn't tested. Python bindings for the decompressor using ctypes (so, for all the impls with them) are almost finished, the only thing remaining is figuring out what is wrong and fixing it (it currently complains about a wrong flag).

I guess for compression I'd package the lib by Ladislav Zezula, as done by other people creating bindings to other languages, but unlike what is done by them, I guess it may make sense to wrap it into a separate python package. These are different libs, anyway.

Anyway, the package should be already useful, i.e. it may be possible to create a compression module for C++ target.

archive/install_shield_3_z.ksy
dgelessus: 1. That is never going to happen for this format :)
2. Every file I tested with has the same value in the size field, so I'm not even sure if that is the field's actual meaning. [idecomp.py apparently thinks so](https://github.com/lephilousophe/idecomp/blob/354961c0cfe86632a39e4fea01e0ca521ffac3e0/idecomp.py#L266-L269), but I don't know where its author got that information from, or if they just guessed. So if there are any files where this field has a different value, I want the parsing to fail at first, because it's hard to say if the rest of the spec will behave properly in that case.
3. The idecomp.py code also only handles this exact header size, so this check probably won't cause any problems in practice.

archive/install_shield_3_z.ksy
dgelessus: Which part exactly? I'm not sure how any of the features in #84 would help with calculating "size of this type except for one of its fields"... Having `_sizeof` on variable-sized fields won't make a difference, because we already have an explicit `len_entry` field.

archive/install_shield_3_z.ksy
dgelessus: This is a format from the pre-NT Windows era, so it probably doesn't have an encoding properly specified. In practice it's probably going to be code page 437, 850, or 1252, assuming a Western system. All of the test files I have are in English with pure ASCII file names, so it's difficult to say.

In cases like these I prefer to use raw byte arrays instead of hardcoding a generic encoding like ASCII or Latin-1, so that in case a file does contain unexpected non-ASCII characters, it can still be parsed using the KSY, and the application code can decide how to deal with the encoding issues (if at all). In my hacky Python script, I have a command-line option for selecting the encoding (defaults to ASCII), and if possible I avoid decoding the name/path fields at all.

archive/install_shield_3_sfx_tail.ksy
dgelessus: Yeah, I know that this is completely insecure and not a proper encryption algorithm :) But this sort of thing is still commonly called "encryption", at least outside of the context of modern cryptography. There's no risk of misunderstanding here I think (nobody is going to think that this is secure encryption), so IMO it would be more confusing to only say "obfuscation" and completely avoid the word "encryption".

dgelessus:
re. decompression: the idecomp repo also contains [a pure Python decompressor for DCL Implode compression](https://github.com/lephilousophe/idecomp/blob/354961c0cfe86632a39e4fea01e0ca521ffac3e0/pwexplode.py). I just didn't spend any time integrating it into my script, because there's already a working Python-based decompressor that uses it, and because all of the files I was working with didn't use compression.

archive/install_shield_3_z.ksy
KOLANICH: >`sizeof(property_name1, property_name2)` is equivalent to
`lea(property_name2) - lea(property_name1) + sizeof(property_name2)`, which is meant to spare us from summing the fields lengths explicitly ourselves

archive/install_shield_3_sfx_tail.ksy
KOLANICH: maybe "scrambling" then?

KOLANICH:
>decompression: the idecomp repo also contains a pure Python decompressor for DCL Implode compression.

Thanks for the info.

>I just didn't spend any time integrating it into my script, because there's already a working Python-based decompressor that uses it

... which is GPLed.

KOLANICH:
Python bindings to libblast: https://github.com/KOLANICH-libs/pkblast.py
kaitai.compress compressor: https://github.com/kaitaiStructCompile/kaitai_compress/blob/python_fixes/python/kaitai/compress/algorithms/implode.py
libimplode: https://github.com/KOLANICH-libs/libimplode (not tested and no bindings have been created currently)

Issue #502, opened 2021-06-30
A packet's size can be either incl_len OR snaplen attribute from the pcap header


2RedSquares:

GreyCat:
Should be fixed now with PR #503.

Issue #501, opened 2021-06-30
Protocol Body - option hop by hop size must be at least 1


2RedSquares:

GreyCat:
Should be fixed now with PR #503.

PR #503, opened 2021-06-30
Pcap reading fixes for #501 and #502 

2RedSquares:


network/protocol_body.ksy
```diff
@@ -52,7 +52,7 @@ types:
       - id: hdr_ext_len
         type: u1
       - id: body
-        size: hdr_ext_len - 1
+        size: 'hdr_ext_len > 0 ? hdr_ext_len - 1 : 1'
```
KOLANICH, line 55: `hdr_ext_len` is `u1`, so it is always `>=0`. So, your condition brings any difference only for `hdr_ext_len=0`. In this case it sets the size to `1`. Even if for `hdr_ext_len=1` it is set to 0. Is this the effect you are trying to achieve and why do you think that it should be achieved?

2RedSquares:
I ran into a test case where I had a PCAP file containing an IPv6 packet with a [Hop by Hop options header](https://datatracker.ietf.org/doc/html/rfc8200#section-4.3).  In this packet, the length value was zero.  See attached image.  

This may be a case where the packet sender is sending an incorrect value.  But the size value was resulting in a value of -1 which was causing failures.  So I'm suggesting setting it to a default value of 1.  

image

GreyCat (approved):

GreyCat:
@2RedSquares Thanks a lot for this contribution and sorry it took 2+ years to merge. Hope it will get better now :)

PR #512, opened 2021-08-03
Add Amlogic eMMC partition format, as used by Meson SoCs

tchebb:
This is an unnamed and undocumented partition table format implemented by the bootloader and kernel that Amlogic provides for their Linux SoCs (Meson series at least, and probably others). They appear to use this rather than GPT, the industry standard, because their BootROM loads and executes the next stage loader from offset 512 (`0x200`) on the eMMC, which is exactly where the GPT header would have to start. So instead of changing their BootROM, Amlogic devised this partition table, which lives at an offset of 36MiB (`0x240_0000`) on the eMMC and so doesn't conflict. This parser expects as input just the partition table from that offset. The maximum number of partitions in a table is 32, which corresponds to a maximum table size of 1304 bytes (`0x518`).

[amlogic-emmc-partitions.zip](https://github.com/kaitai-io/kaitai_struct_formats/files/6926961/amlogic-emmc-partitions.zip) contains a sample partition table (`amlogic-emmc-partitions`), extracted from an Amlogic S905Y2 device I own.

### DIFF
--- a/filesystem/amlogic_emmc_partitions.ksy
+++ b/filesystem/amlogic_emmc_partitions.ksy
@@ -1,0 +1,82 @@
+meta:
+  id: amlogic_emmc_partitions
+  title: Amlogic proprietary eMMC partition table
+  license: CC0-1.0
+  ks-version: 0.9
+  bit-endian: le
+  encoding: UTF-8
+  endian: le
+
+doc: |
+  This is an unnamed and undocumented partition table format implemented by
+  the bootloader and kernel that Amlogic provides for their Linux SoCs (Meson
+  series at least, and probably others). They appear to use this rather than GPT,
+  the industry standard, because their BootROM loads and executes the next stage
+  loader from offset 512 (0x200) on the eMMC, which is exactly where the GPT
+  header would have to start. So instead of changing their BootROM, Amlogic
+  devised this partition table, which lives at an offset of 36MiB (0x240_0000)
+  on the eMMC and so doesn't conflict. This parser expects as input just the
+  partition table from that offset. The maximum number of partitions in a table
+  is 32, which corresponds to a maximum table size of 1304 bytes (0x518).
+
+doc-ref:
+  - http://aml-code.amlogic.com/kernel/common.git/tree/include/linux/mmc/emmc_partitions.h?id=18a4a87072ababf76ea08c8539e939b5b8a440ef
+  - http://aml-code.amlogic.com/kernel/common.git/tree/drivers/amlogic/mmc/emmc_partitions.c?id=18a4a87072ababf76ea08c8539e939b5b8a440ef
+
+seq:
+  - id: magic
+    contents: ["MPT", 0]
+  - id: version
+    size: 12
+    type: strz
+  - id: num_partitions
+    -orig-id: part_num
+    type: s4
+    valid:
+      min: 1
+      max: 32
+  - id: checksum
+    type: u4
+    doc: |
+      To calculate this, treat the first (and only the first) partition
+      descriptor in the table below as an array of unsigned little-endian
+      32-bit integers. Sum all those integers mod 2^32, then multiply the
+      result by the total number of partitions, also mod 2^32. Amlogic
+      likely meant to include all the partition descriptors in the sum,
+      but their code as written instead repeatedly loops over the first
+      one, once for each partition in the table.
+  - id: partitions
+    type: partition
+    repeat: expr
+    repeat-expr: num_partitions
+
+types:
+  partition:
+    seq:
+      - id: name
+        size: 16
+        type: strz
+      - id: size
+        type: u8
+      - id: offset
+        type: u8
+        doc: |
+          The start of the partition relative to the start of the eMMC, in bytes
+      - id: flags
+        type: part_flags
+        size: 4
+      - id: padding
+        size: 4
+
+    types:
+      part_flags:
+        seq:
+          - id: is_code
+            -orig-id: STORE_CODE
+            type: b1
+          - id: is_cache
+            -orig-id: STORE_CACHE
+            type: b1
+          - id: is_data
+            -orig-id: STORE_DATA
+            type: b1


generalmimon:
> [amlogic-emmc-partitions.zip](https://github.com/kaitai-io/kaitai_struct_formats/files/6926961/amlogic-emmc-partitions.zip) contains a sample partition table (`amlogic-emmc-partitions`), extracted from an Amlogic S905Y2 device I own.

Thanks for providing the sample file! Can I add it to the sample repository [kaitai_struct_samples](https://github.com/kaitai-io/kaitai_struct_samples) created recently? If so, would the [CC0-1.0](https://spdx.org/licenses/CC0-1.0.html) license be OK for the sample as well?

tchebb:
I'm fine with adding the sample to the repo, and I'm personally fine with CC0-1.0. However, I don't know if there are any extra licensing restrictions around the sample, since I extracted it from the factory eMMC image of a commercial device. Since it's such a small and trivial amount of data, I doubt anyone will complain though :)

filesystem/amlogic_emmc_partitions.ksy
```diff
@@ -0,0 +1,68 @@
...
+  partition table from that offset. The maximum number of partitions in a table
+  is 32, which corresponds to a maximum table size of 1304 bytes (0x518).
+  
+doc-ref:
+  - https://aml-code.amlogic.com/kernel/common.git/tree/include/linux/mmc/emmc_partitions.h?id=18a4a87072ababf76ea08c8539e939b5b8a440ef
+  - https://aml-code.amlogic.com/kernel/common.git/tree/drivers/amlogic/mmc/emmc_partitions.c?id=18a4a87072ababf76ea08c8539e939b5b8a440ef
+
+seq:
+  - id: magic
+    contents: "MPT\0"
```
generalmimon, line 27: I'd go for the mixed [`contents`](https://doc.kaitai.io/user_guide.html#magic) syntax here:

```suggestion
    contents: ["MPT", 0]
```

While the string escape works too, it's generally a better idea to include bytes as integers in the list.

generalmimon (changes requested):

filesystem/amlogic_emmc_partitions.ksy
```diff
@@ -0,0 +1,68 @@
...
+  - id: version
+    size: 12
+    type: str
+    encoding: UTF-8
+  - id: num_partitions
+    -orig-id: part_num
+    type: s4
```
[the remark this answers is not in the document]
tchebb: Nice! I'd been wondering if there was a way to annotate this, but I didn't find this in the User Guide.

filesystem/amlogic_emmc_partitions.ksy
```diff
@@ -0,0 +1,68 @@
...
+  loader from offset 512 (0x200) on the eMMC, which is exactly where the GPT
+  header would have to start. So instead of changing their BootROM, Amlogic
+  devised this partition table, which lives at an offset of 36MiB (0x240_0000)
+  on the eMMC and so doesn't conflict. This parser expects as input just the
+  partition table from that offset. The maximum number of partitions in a table
+  is 32, which corresponds to a maximum table size of 1304 bytes (0x518).
+  
+doc-ref:
+  - https://aml-code.amlogic.com/kernel/common.git/tree/include/linux/mmc/emmc_partitions.h?id=18a4a87072ababf76ea08c8539e939b5b8a440ef
+  - https://aml-code.amlogic.com/kernel/common.git/tree/drivers/amlogic/mmc/emmc_partitions.c?id=18a4a87072ababf76ea08c8539e939b5b8a440ef
```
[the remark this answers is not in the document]
tchebb: Changed to the HTTP version, since there's no official or semi-official mirror on GitHub. Everything I can find there is someone's personal fork.

tchebb:
Thanks, great feedback! I've pushed an new version with all points addressed.

filesystem/amlogic_emmc_partitions.ksy
```diff
@@ -0,0 +1,72 @@
...
+    seq:
+      - id: name
+        size: 16
+        type: strz
+      - id: size
+        type: u8
+      - id: offset
+        type: u8
+        doc: |
+          The start of the partition relative to the start of the eMMC, in bytes
```
KOLANICH, line 59: Since we know size and position, it may be better to ejcode this info into an instance rather than have it as a free-form description

filesystem/amlogic_emmc_partitions.ksy
```diff
@@ -0,0 +1,72 @@
...
+      - id: is_code
```
KOLANICH, line 60: It may make sense to move the flags into an own subtype.

tchebb: Done.

filesystem/amlogic_emmc_partitions.ksy
tchebb: As currently written, the spec is designed to take just the partition table rather than an entire disk image. As such, offsets into the file don't match offsets into eMMC, meaning any partition that exists before the table location would have to have a negative offset.

The reason I chose to do it this way is because the table lives 36MiB into flash, so requiring an eMMC image as input would require parsing a massive file even if all you care about is the partition metadata.

filesystem/amlogic_emmc_partitions.ksy
```diff
@@ -0,0 +1,77 @@
...
+    -orig-id: part_num
+    type: s4
+    valid:
+      min: 1
+      max: 32
+  - id: checksum
+    type: u4
+    doc: |
+      Unsigned modulo 32 sum of all the bytes in the "partitions" element below,
+      interpreted as a sequence of 32-byte little-endian unsigned integers.
```
generalmimon, line 42: I checked this statement by calculating the checksum in the .ksy based on the instructions like this:

```diff
$ git diff --no-index amlogic_emmc_partitions\@{original,modified}.ksy
diff --git 1/amlogic_emmc_partitions@original.ksy 2/amlogic_emmc_partitions@modified.ksy
index 6d7f63c..19fda28 100644
--- 1/amlogic_emmc_partitions@original.ksy
+++ 2/amlogic_emmc_partitions@modified.ksy
@@ -40,12 +40,44 @@ seq:
     doc: |
       Unsigned modulo 32 sum of all the bytes in the "partitions" element below,
       interpreted as a sequence of 32-byte little-endian unsigned integers.
+  - size: 0
+    if: ofs_partitions < 0
   - id: partitions
     type: partition
     repeat: expr
     repeat-expr: num_partitions
+  - size: 0
+    if: len_partitions < 0
+
+instances:
+  ofs_partitions:
+    value: _io.pos
+  len_partitions:
+    value: _io.pos - ofs_partitions
+  checksum_steps:
+    pos: ofs_partitions
+    size: len_partitions
+    type: checksum_steps_type
+  checksum_expected:
+    value: checksum_steps.steps.last.val

 types:
+  checksum_steps_type:
+    seq:
+      - id: steps
+        type: 'checksum_step(_index != 0 ? steps[_index - 1].val : 0)'
+        repeat: eos
+  checksum_step:
+    -webide-representation: '{cur}'
+    params:
+      - id: prev_val
+        type: u4
+    seq:
+      - id: cur
+        type: u4
+    instances:
+      val:
+        value: '((prev_val + cur) & 0xffff_ffff) >> 0'
   partition:
     seq:
       - id: name
```

It turned out that the checksum calculated this way does not match the stored checksum in the test file [`amlogic-s905y2.bin`](https://github.com/kaitai-io/kaitai_struct_samples/blob/c3829d5/filesystem/amlogic_emmc_partitions/amlogic-s905y2.bin) that you provided:

<pre><code>├─ magic = [77, 80, 84, 0]
├─ version = <b>01.00.00</b>
├─ numPartitions = <b>0x12</b> <i>= 18</i>
├─ checksum = <ins><b>0x4337B596</b></ins> <i>= 1127724438</i>
▷─ partitions
├─ ofsPartitions = <b>0x18</b> <i>= 24</i>
├─ lenPartitions = <b>0x2D0</b> <i>= 720</i>
├─ checksumExpected = <ins><b>0x48EAC954</b></ins> <i>= 1223346516</i>
▷─ checksumSteps</code></pre>

So I looked up the actual code used in the Amlogic's code to generate the checksum ([`emmc_partitions.c:624`](http://aml-code.amlogic.com/kernel/common.git/tree/drivers/amlogic/mmc/emmc_partitions.c?id=18a4a87072ab#n624)) and ran it on the test file [`amlogic-s905y2.bin`](https://github.com/kaitai-io/kaitai_struct_samples/blob/c3829d5/filesystem/amlogic_emmc_partitions/amlogic-s905y2.bin). Here's the console output - for the first 3 partitions, individual checksum addends are printed (see source files in [**amlogic_emmc_partitions-checksum.zip**](https://github.com/kaitai-io/kaitai_struct_formats/files/7925542/amlogic_emmc_partitions-checksum.zip)):

```console
$ gcc -Wall -std=c11 main.c

$ ./a.exe
magic: 'MPT'
part_num: 18
partition[0], initial checksum: 0x0
  [10]: 0x746F6F62
  [9]: 0x64616F6C
  [8]: 0x00007265
  [7]: 0x0
  [6]: 0x00400000
  [5]: 0x0
  [4]: 0x0
  [3]: 0x0
  [2]: 0x0
  [1]: 0x0
partition[1], initial checksum: 0xD9115133
  [10]: 0x746F6F62
  [9]: 0x64616F6C
  [8]: 0x00007265
  [7]: 0x0
  [6]: 0x00400000
  [5]: 0x0
  [4]: 0x0
  [3]: 0x0
  [2]: 0x0
  [1]: 0x0
partition[2], initial checksum: 0xB222A266
  [10]: 0x746F6F62
  [9]: 0x64616F6C
  [8]: 0x00007265
  [7]: 0x0
  [6]: 0x00400000
  [5]: 0x0
  [4]: 0x0
  [3]: 0x0
  [2]: 0x0
  [1]: 0x0
partition[3], initial checksum: 0x8B33F399
partition[4], initial checksum: 0x644544CC
partition[5], initial checksum: 0x3D5695FF
partition[6], initial checksum: 0x1667E732
partition[7], initial checksum: 0xEF793865
partition[8], initial checksum: 0xC88A8998
partition[9], initial checksum: 0xA19BDACB
partition[10], initial checksum: 0x7AAD2BFE
partition[11], initial checksum: 0x53BE7D31
partition[12], initial checksum: 0x2CCFCE64
partition[13], initial checksum: 0x5E11F97
partition[14], initial checksum: 0xDEF270CA
partition[15], initial checksum: 0xB803C1FD
partition[16], initial checksum: 0x91151330
partition[17], initial checksum: 0x6A266463
         checksum: 0x4337B596
checksum_expected: 0x4337B596
```

As you can see, the checksum calculated by the routine matches the stored checksum. At the same time, it's visible what happens here - the checksum routine loops over the first partition over and over again, repeating for the number of partitions. Therefore, the resulting checksum is identical to `((0xD9115133 * 18) & 0xffff_ffff) = 0x4337B596` (run `'0x{:X}'.format((0xD9115133 * 18) & 0xffff_ffff)` in Python) - `0xD9115133` is the checksum of `partition[0]`, and this is repeated 18 times because there are 18 partitions.

Since the official implementation behaves this way (and all partition tables in this format must follow the same approach, otherwise the reference impl will reject them as invalid), I think it should also be documented in the `.ksy` so that the format spec reflects reality.

generalmimon: Also your description contained some flaws, which would have to be corrected anyway:

> Unsigned ~~modulo 32~~ sum of all the bytes in the "partitions" element below,
> interpreted as a sequence of 32-~~byte~~ little-endian unsigned integers.

1. _Modulo 32_ would mean `% 32` (or `& 0b0001_1111`, alternatively `& 0x1f`), which is not the case - it is actually `& 0xffff_ffff` or `% (1 << 32)`, which is equivalent to `% (2 ** 32)`.
2. 32-<ins>**bit**</ins> integers, not 32-~~byte~~.

tchebb: Wow, great catch! Thanks for digging into that! I've updated the documentation to match what you describe and fix the typos. Should I also try to add checksum validation to the format like in your patch? Some of the features you use there are new to me, but I can read up on them.

filesystem/amlogic_emmc_partitions.ksy
```diff
@@ -0,0 +1,82 @@
...
+        size: 4
+
+    types:
+      part_flags:
+        seq:
+          - id: is_code
+            -orig-id: STORE_CODE
+            type: b1
+          - id: is_cache
+            -orig-is: STORE_CACHE
```
tchebb, line 78:
```suggestion
            -orig-id: STORE_CACHE
```

tchebb:
This has been several rounds of review now and I believe I've addressed all comments. Can it be merged?

generalmimon:
@tchebb:

> Can it be merged?

Yes, thanks for the reminder! And sorry for the delay.

PR #515, opened 2021-08-06
Add mach_o_fat format for multiarch Mach-O objects

tchebb:
Multiarch Mach-Os use a special "fat" container format. Currently, we recognize the magic for this in `mach_o.ksy` but then just try to parse it like a normal Mach-O, which fails. Since fat Mach-Os can't be nested inside themselves, I believe that a separate `ksy` to parse them is the correct approach. This PR adds such a `ksy`, which imports the existing Mach-O parser to parse each architecture.

This contribution is on behalf of my company.

### DIFF
--- a/executable/mach_o.ksy
+++ b/executable/mach_o.ksy
@@ -26,12 +26,13 @@ seq:
     repeat-expr: header.ncmds
 enums:
   magic_type:
+    # Note that for multiarch (a.k.a. fat) Mach-O files, which are the primary
+    # kind you find on macOS today, you should instead use mach_o_fat.ksy, which
+    # parses the fat header and embeds mach_o.ksy to parse each arch.
     0xFEEDFACE: macho_be_x86 # MH_MAGIC:    mach-o, big-endian,    x86
     0xCEFAEDFE: macho_le_x86 # MH_CIGAM:    mach-o, little-endian, x86
     0xFEEDFACF: macho_be_x64 # MH_MAGIC_64: mach-o, big-endian,    x64
     0xCFFAEDFE: macho_le_x64 # MH_CIGAM_64: mach-o, little-endian, x64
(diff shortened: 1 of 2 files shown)


executable/mach_o_fat.ksy
```diff
@@ -0,0 +1,45 @@
...
+      - id: size
+        type: u4
+      - id: align
+        type: u4
+
+    instances:
+      object:
+        pos: offset
+        size: size
+        type: mach_o
```
generalmimon, line 45:
```suggestion
      - id: ofs_object
        type: u4
      - id: len_object
        type: u4
      - id: align
        type: u4

    instances:
      object:
        pos: ofs_object
        size: len_object
        type: mach_o
```

executable/mach_o.ksy
```diff
@@ -24,12 +24,13 @@ seq:
     repeat-expr: header.ncmds
 enums:
   magic_type:
+    # Note that for multiarch (a.k.a. fat) Mach-O files, which are the primary
+    # kind you find on macOS today, you should instead use mach_o_fat.ksy, which
+    # parses the fat header and embeds us to parse each arch.
```
generalmimon, line 29: > and embeds **_us_** to parse each arch.

I wouldn't use "embeds *us*" here - depending on how you read it, it may not be immediately clear that "*us*" refers to "this `mach_o.ksy` specification".

generalmimon (approved):
LGTM, I would only suggest a few minor improvements:

tchebb:
Added all suggested plus `-orig-id` keys where names changed from what Apple's code uses. Thanks!

generalmimon:
> Thanks!

Thanks as well for the contribution :blush:

I've added a test file for this format [to kaitai_struct_samples](https://github.com/kaitai-io/kaitai_struct_samples/tree/8a328c39036d3dd3d483fc11f9d55728ed2d3454/executable/mach_o_fat).

PR #516, opened 2021-08-06
Mach-O: support DER-encoded entitlements

tchebb:
This is a code signature field new to iOS binaries as of early 2019 and new to macOS 12. It contains entitlements encoded as ASN.1 DER instead of XML. Also fix up an existing field name to better match Apple's XNU source code.

This contribution is on behalf of my company.


GreyCat:
Makes perfect sense, thanks for this contribution!

PR #522, opened 2021-08-30
Added Joda-Time

KOLANICH:
It's a spec for JodaTime timezone format. Not quite final (I gonna refactor `millis` type (but not right now), and probably we should do something with `pas_str` (Java docs says the UTF is modified there, and we should decide what we do with that (a processor function, I guess) and if for Joda Time we can assumme that modified UTF is considered the same as unmodified). Also some more reversing is needed for better understanding what the pieces do.

@jodastephen


database/joda_time.ksy
```diff
@@ -0,0 +1,177 @@
...
+        value: v >> 6
+
+    types:
+      form_0:
+        params:
+          - id: v
+            type: u1
+        instances:
+          value:
+            value: ((v << (32 - 6)) >> (32 - 6)) * (30 * 60000)
```
generalmimon, line 106: This method of sign extension is poorly portable and generally isn't guaranteed to work in target languages. There is also no guarantee that the default integer type will remain `s4` in the future.  I recommend using the approach I used in `vlq_base128_le` instead:

https://github.com/kaitai-io/kaitai_struct_formats/blob/4c4bfc0709a8a2f6d771a6dc3094ccfa17def3b3/common/vlq_base128_le.ksy#L61-L65

However, you need to take only the lowest 6 bits from `v`. If the highest 2 bits of the `v` byte would be set, it would not work correctly anymore (there are no longer any _back and forth_ shifts which would trim it from both sides).

KOLANICH: Yeah, I know; it was just verbatim copy from the source of Joda Time. I gonna refactor it.

PR #531, opened 2021-09-01
Added AllWinner/sunxi toc0.

KOLANICH:


KOLANICH:
I guess (the search in the Internet have given it as an example of a device using this format), https://www.linksprite.com/image-for-pcduino8-uno/ , though I haven't checked that.

armijnhemel:
> I guess (the search in the Internet have given it as an example of a device using this format), https://www.linksprite.com/image-for-pcduino8-uno/ , though I haven't checked that.

The Android image doesn't use it for sure.

KOLANICH:
One can find source code for firmwares of devices using this format on GitHub by searching by the signatures you can find within the spec. For example https://github.com/pcduino/pcduino8-uno-kernel is found, which I suspect should match some of the files by the link

armijnhemel:
> One can find source code for firmwares of devices using this format on GitHub by searching by the constants you can find within the spec. For example https://github.com/pcduino/pcduino8-uno-kernel is found, which I suspect should match some of the files by the link

I checked, none of the firmware files linked seem to use that particular format.

KOLANICH:
>I checked, none of the firmware files linked seem to use that particular format.

Anyway, ideally we need to make an own set of files not tied anyhow to any real firmware, which only purpose is to be the test files for the specs.

KOLANICH:
>I checked, none of the firmware files linked seem to use that particular format.

Anyway, ideally we need to make an own set of files not tied anyhow to any real firmware, which only purpose is to be the test files for the specs.

PR #535, opened 2021-10-27
add Wii U/3DS-format Mii data

HEYimHeroic:
hi. i believe last year i tried something similar, but the Kaitai files i submitted back then (and that other people submitted of other Mii data formats) were not very good, this Kaitai has been tested and works very well, so i'd like to submit it.

there were a couple of things that i still didn't know how to do, but none that completely break it. there are also a number of unused and unknown bits and bytes. some of the unused bits/bytes are meant to act as a buffer, but others, not so sure. hopefully the unknown bits become known later down the line :) but the majority of the format is documented, even moreso than before, so i'm happy.

anyways, yeah! i'll probably get to rewriting the other Mii data formats later on, but i'm very happy with how this one came out.

### DIFF
--- a/game/mii_data_wiiu_3ds.ksy
+++ b/game/mii_data_wiiu_3ds.ksy
@@ -125,9 +125,9 @@ types:
       - id: unknown_1
         type: b1be
         doc: This bit is currently unknown. It should be set to 0, in most common use cases. Some documentation says only "is DSi Mii?" which doesn't seem right.
-      - id: is_nintendo_mii
+      - id: is_developer_mii
         type: b1be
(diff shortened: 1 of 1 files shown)


HEYimHeroic:
okay, i only have one more question then: both `reserved_12` and `checksum` should be optional, as they only apply to a specific filetype (that is, CFSD and FFSD, whereas CFCD and FFCD do not have them). is there a way to mark them as optional?

KOLANICH:
I guess you need `if: ` and `params`

HEYimHeroic:
there doesn't seem to be any documentation on how i can use `if:` with the file extension?

KOLANICH:
You can use `if` with `params`, but one would have to set the param externally depending on file extension. KS-generated parsers parse arbitrary blobs, not files. `file-extension` is mainly to make the spec more googleable.

PR #536, opened 2021-10-28
Updated DEX format

tins2831:
- Added `code_item` support: https://source.android.com/devices/tech/dalvik/dex-format#code-item
- Added `doc-ref` links wherever possible.
- ~~Improved documentation. (word wrapped, capitalization, references to documents)~~

### DIFF
--- a/executable/dex.ksy
+++ b/executable/dex.ksy
@@ -631,7 +631,7 @@ types:
   code_item:
     doc-ref: 'https://source.android.com/devices/tech/dalvik/dex-format#code-item'
     seq:
-      - id: registers_count
+      - id: register_count
         -orig-id: registers_size
         type: u2
         doc: The number of registers used by this code.
@@ -663,11 +663,13 @@ types:
 
           The format of the data is specified by "debug_info_item".
         doc-ref: 'https://source.android.com/devices/tech/dalvik/dex-format#debug-info-item'
-      - id: insns_size
+      - id: ops_size
+        -orig-id: insns_size
         type: u4
         doc: Size of the instructions list, in 16-bit code units.
-      - id: insns
-        size: insns_size * 2
+      - id: ops
+        -orig-id: insns
+        size: ops_size * 2
         doc: |
           Actual array of bytecode.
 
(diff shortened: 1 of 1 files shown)


armijnhemel:
You couldalso just use https://github.com/armijnhemel/binaryanalysis-ng/blob/master/src/parsers/executable/dex/dex.ksy as a base, as I have added/fixed quite a few things in there (including style fixes).

executable/dex.ksy
```diff
@@ -631,7 +631,7 @@ types:
   code_item:
     doc-ref: 'https://source.android.com/devices/tech/dalvik/dex-format#code-item'
     seq:
-      - id: register_count
+      - id: registers_count
```
KOLANICH, line 634: It seems like the former variant was grammatically more correct. 

https://english.stackexchange.com/questions/11350/use-of-the-plural-with-list-and-when-item-count-is-0

executable/dex.ksy
```diff
@@ -666,8 +666,7 @@ types:
       - id: insns_size
         type: u4
         doc: Size of the instructions list, in 16-bit code units.
-      - id: instructions
-        -orig-id: insns
+      - id: insns
```
KOLANICH, line 669: I like the former variant better

KOLANICH (approved):
Except the mentioned issues, LGTM.

executable/dex.ksy
tins2831: `instructions` is a difficult word to shorten and `insns` looks like a confusing tongue twister. How about `body` and `body_size` instead?

KOLANICH: maybe `opcodes`, but it is not much shorter.

>I could also change the name of the code_item instance in encoded_method from method_body to just code 

Will it be confusing?

tins2831: > maybe opcodes, but it is not much shorter.

That'd be incorrect, as opcodes aren't the same as instructions. An opcode(operation code) is a piece of what makes up an instruction; it defines the operation of the instruction. With that said, it wouldn't make much sense now to name the pool of instructions to `opcodes`.

I vote for `body` and `body_size` or `inst_pool` and `inst_pool_size`.

KOLANICH: >That'd be incorrect, as opcodes aren't the same as instructions. An opcode(operation code) is a piece of what makes up an instruction; it defines the operation of the instruction. 

You are right. But `body` is IMHO too generic, and truncations are non-obvious.

tins2831: True, so `inst_pool` then or do you have another suggestion?

KOLANICH: maybe just `ops`, but it is a truncation.

tins2831: Fair enough. Thanks

PR #537, opened 2021-10-29
ELF: parse ARM attributes

generalmimon:
Based on https://github.com/armijnhemel/binaryanalysis-ng/commit/b728426808cb20f5a27ab6dee878068bf43725e9#diff-3a730cd59a811844bc0277c82828361ffb0b452f5739135300947f8948ae1f10

Cc @armijnhemel 

Some sample files with the `.ARM.attributes` section that I collected:

* [`elf-Linux-ARMv7-ls`](https://github.com/kaitai-io/kaitai_struct_samples/blob/b41101d8cec3824b86234a1ca9f3f48b7644c6e2/executable/elf/elf-Linux-ARMv7-ls) - public ("aeabi") subsection with 16 attributes,
* ```sh
  curl -LO https://busybox.net/downloads/binaries/1.16.1/busybox-armv6l
  ```

  public subsection with 0 attributes,

* ```sh
  curl -LO https://busybox.net/downloads/binaries/1.21.1/busybox-armv7l
  ```

  public subsection with 3 attributes.

If you have any other ELF samples with private subsections or other public subsection types than `FILE`, please share them.

### DIFF
--- a/executable/elf.ksy
+++ b/executable/elf.ksy
@@ -20,6 +20,8 @@ meta:
     - executable
     - linux
   license: CC0-1.0
+  imports:
+    - /common/vlq_base128_le
   ks-version: 0.9
 doc-ref:
   - https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/elf.h;hb=0f62fe0532
@@ -453,6 +455,7 @@ types:
                 'sh_type::note': note_section
                 'sh_type::rel': relocation_section(false)
                 'sh_type::rela': relocation_section(true)
+                'sh_type::arm_attributes': arm_attributes_section
             if: type != sh_type::nobits
           linked_section:
             value: _root.header.section_headers[linked_section_idx]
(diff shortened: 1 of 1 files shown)


armijnhemel:
A good source of ARM binaries are Debian and Fedora, which both have ARM builds,. for example http://ftp.nluug.nl/pub/os/Linux/distr/debian/pool/main/b/bash/bash_5.1-2+b3_arm64.deb and http://ftp.nluug.nl/pub/os/Linux/distr/debian/pool/main/a/aalib/libaa-bin_1.4p5-43_armhf.deb

executable/elf.ksy
```diff
@@ -700,6 +703,258 @@ types:
...
+              is_value_str:
+                value: |
+                  tag.to_i < 32 ? (tag == tags::cpu_raw_name or tag == tags::cpu_name)
+                  : tag.to_i % 2 != 0
+                doc-ref: https://github.com/llvm/llvm-project/blob/014c6b07362c/lldb/source/Plugins/ObjectFile/ELF/ObjectFileELF.cpp#L1260-L1276
+              is_value_int:
+                value: not is_value_str
+            enums:
+              # https://github.com/llvm/llvm-project/blob/014c6b07362c/llvm/include/llvm/Support/ARMBuildAttributes.h#L34-L85
+              tags:
```
KOLANICH, line 814: How about putting large enums into own ksy files?

generalmimon: > How about putting large enums into own ksy files?

This would be nice sometime in the future, but I'm afraid our infrastructure isn't ready for that yet. There are problems in several languages with trying to access an enum in an imported spec, see test `EnumImport` in the [CI dashboard](https://ci.kaitai.io/). And even when this is fixed, we should figure out what to do with these import-only specs at https://formats.kaitai.io/ - they should be at least marked with some special attribute that will exclude them from the normal view. There is no point in displaying such specifications on homepage, that would be confusing.

So for now, let's maintain the status quo.

armijnhemel:
> A good source of ARM binaries are Debian and Fedora, which both have ARM builds,. for example http://ftp.nluug.nl/pub/os/Linux/distr/debian/pool/main/b/bash/bash_5.1-2+b3_arm64.deb and http://ftp.nluug.nl/pub/os/Linux/distr/debian/pool/main/a/aalib/libaa-bin_1.4p5-43_armhf.deb

I checked a few files, also inside Android firmwares, but couldn't find anything other than `FILE`.

generalmimon:
> but couldn't find anything other than `FILE`

@armijnhemel This probably makes sense, because _section_ and _symbol_ attributes are deprecated from the 2.09 release of the Arm ABI:

https://github.com/kaitai-io/kaitai_struct_formats/blob/aa5bf60698aca7034c323f11d5b152171b30c4cd/executable/elf.ksy#L818-L825

As written in the https://developer.arm.com/documentation/ihi0045/h/Addenda-to--and-Errata-in--the-ABI-for-the-Arm-Architecture---ABI-2019Q4-documentation?lang=en document:

> From the 2.09 release, section and symbol attributes are deprecated and optional. Primary object producers are discouraged from generating them and consumers are permitted to ignore them.

So it's OK to ignore them, but I've just realized that my type would probably fail on encountering them, because the _section_ and _symbol_ subsections have a slightly different structure than the _file_ subsection. I'll fix this.

executable/elf.ksy
KOLANICH: >they should be at least marked with some special attribute that will exclude them from the normal view.

I guess it is trivial to detect them automatically.

PR #538, opened 2021-11-01
add specification for ubifs

armijnhemel:
This is a specification for `ubifs`. I found that it works well, but likely this is only true for file systems that have been generated cleanly (so not flash dumps for example).

Note, this is only for `ubifs` not for `ubi`, which is different.

### DIFF
--- a/filesystem/ubifs.ksy
+++ b/filesystem/ubifs.ksy
@@ -1,0 +1,856 @@
+meta:
+  id: ubifs
+  title: UBIFS
+  license: GPL-2.0-only
+  endian: le
+  encoding: UTF-8
+doc: |
+  The UBIFS file system is a file system for flash file systems. It works on
+  top of UBI (Unsorted Block Images), which works as an abstraction layer so
+  UBIFS does not need to care about low level details. This specification only
+  covers UBIFS, not UBI.
+
+  A UBIFS image is divided in several Logical Erase Blocks (LEB). The first
+  LEB (LEB0) contains the superblock node, which includes information about LEB
+  size, amongst others. The next two LEBs (LEB1 and LEB2) comprise the so called
+  "master area" and both contain a copy of the master node, which includes
+  information about where to find the index node, which is then used to access
+  all the files on the file system.
+doc-ref:
+  - https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/ubifs/ubifs-media.h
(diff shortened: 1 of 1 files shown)


filesystem/ubifs.ksy
```diff
@@ -0,0 +1,856 @@
...
+        type: u4
+        doc: number of hard links
+      - id: uid
+        type: u4
+        doc: owner ID
+      - id: gid
+        type: u4
+        doc: group ID
+      - id: mode
+        type: u4
```
KOLANICH, line 259: Should it be a type of bit fields?

filesystem/ubifs.ksy
```diff
@@ -0,0 +1,856 @@
...
+        doc: access flags
+      - id: flags
+        type: u4
```
KOLANICH, line 262: Should it be a type of bit fields?

filesystem/ubifs.ksy
```diff
@@ -0,0 +1,856 @@
...
+        size: len_target
+        io: _root._io
+        type: block
+
+  # Key types
+  key:
+    seq:
+      - id: inode_number
+        type: u4
+      - id: key_value
```
KOLANICH, line 684: Should one use bit-sized types with `bit-endian: le` here?

filesystem/ubifs.ksy
```diff
@@ -0,0 +1,856 @@
...
+        type: u4
+        doc: LEB number of LPT's own lprops table
+      - id: ofs_ltab
+        -orig-id: ltab_offs
+        type: u4
+        doc: offset of LPT's own lprops table
+      - id: leb_save_table
+        -orig-id: lsave_lnum
+        type: u4
+        doc: LEB number of LPT's save table (big model only)
```
KOLANICH, line 418: They are present unconditionally, aren't they? Maybe `makes sense only for big model` then?

filesystem/ubifs.ksy
```diff
@@ -0,0 +1,856 @@
+meta:
+  id: ubifs
+  title: UBIFS
+  license: GPL-2.0-only
```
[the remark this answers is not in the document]
armijnhemel: That was indeed the case. Usually when I work from documents I just pick CC0 (like in #587 ) and when working from source code I pick whatever the source code says. Probably that is a bit too cautious. I will check with a lawyer, just to be sure.

PR #539, opened 2021-11-01
add specification for zchunk

### LINKED ISSUE #197: zchunk
```yaml
meta:
  id: zchunk
  title: zchunk
  application:
    - unzck
    - zck
    - zckdl
    - zck_read_header
  file-extension: zck
doc: |
  A format for incremental distribution of software updates.

doc-ref: https://github.com/zchunk/zchunk/blob/master/zchunk_format.txt
```

armijnhemel:
This is a specification for the `zchunk` format. This is used by for example Fedora for some of the metadata in their repositories: https://fedoraproject.org/wiki/Changes/Zchunk_Metadata

Resolves #197

### DIFF
--- a/archive/zchunk.ksy
+++ b/archive/zchunk.ksy
@@ -1,0 +1,188 @@
+meta:
+  id: zchunk
+  title: Zchunk
+  file-extension: zck
+  license: CC0-1.0
+  endian: le
+doc-ref: https://github.com/zchunk/zchunk/blob/main/zchunk_format.txt
+seq:
+  - id: lead
+    type: lead
+  - id: header
+    type: header
+    size: lead.len_header.value
+  - id: chunks
+    size: chunk_metadata[_index].len_chunk.value
+    repeat: expr
+    repeat-expr: num_chunks
+instances:
+  num_chunks:
+    value: header.index.num_chunks.value - 1
+    doc: the number of chunks includes the header, so -1
+  chunk_metadata:
+    value: header.index.chunks_metadata
+types:
+  lead:
+    seq:
+      - id: magic
+        contents: [0, 'ZCK1']
+      - id: checksum
+        type: compressed_integer
+      - id: len_header
+        type: compressed_integer
+      - id: lead_checksum
+        size: len_checksum
+    instances:
+      checksum_type:
+        value: checksum.value
+        enum: checksum_types
+      len_checksum:
+        value: |
+            checksum_type == checksum_types::sha1 ? 20 :
+            checksum_type == checksum_types::sha256 ? 32 :
+            checksum_type == checksum_types::sha512 ? 64 :
+            checksum_type == checksum_types::sha512_128 ? 16 :
+            0
+  header:
+    seq:
+      - id: preface
+        type: preface
+      - id: len_index
+        type: compressed_integer
+      - id: index
+        type: index
+        size: len_index.value
+      - id: signatures
+        type: signatures
+  preface:
+    seq:
+      - id: checksum
+        size: _root.lead.len_checksum
+      - id: flags
+        type: compressed_integer
+      - id: compression_type
+        type: compressed_integer
+      - id: num_optional_element
+        type: compressed_integer
+        if: has_optional_elements
+      - id: optional_elements
+        type: optional_element
+        repeat: expr
+        repeat-expr: num_optional_element.value
+        if: has_optional_elements
+    instances:
+      has_data_streams:
+        value: flags.value & 0b1 == 0b1
+      has_optional_elements:
+        value: flags.value & 0b10 == 0b10
+  optional_element:
+    seq:
+      - id: element_id
+        type: compressed_integer
+      - id: len_data
+        type: compressed_integer
+      - id: data
+        size: len_data.value
+  index:
+    seq:
+      - id: checksum
+        type: compressed_integer
+      - id: num_chunks
+        type: compressed_integer
+      - id: dict_stream
+        type: compressed_integer
+        if: _parent.preface.has_data_streams
+      - id: dict_checksum
+        size: len_checksum
+      - id: len_dict
+        type: compressed_integer
+      - id: len_uncompressed_dict
+        type: compressed_integer
+      - id: chunks_metadata
+        type: chunk(len_checksum, _parent.preface.has_data_streams)
+        repeat: expr
+        repeat-expr: num_chunks.value - 1
+        doc: the number of chunks includes the header, so -1
+    instances:
+      checksum_type:
+        value: checksum.value
+        enum: checksum_types
+      len_checksum:
+        value: |
+            checksum_type == checksum_types::sha1 ? 20 :
+            checksum_type == checksum_types::sha256 ? 32 :
+            checksum_type == checksum_types::sha512 ? 64 :
+            checksum_type == checksum_types::sha512_128 ? 16 :
+            0
+  chunk:
+    params:
+      - id: len_checksum
+        type: u4
+      - id: has_data_streams
+        type: bool
+    seq:
+      - id: chunk_stream
+        type: compressed_integer
+        if: has_data_streams
+      - id: chunk_checksum
+        size: len_checksum
+      - id: len_chunk
+        type: compressed_integer
+      - id: len_uncompressed_chunk
+        type: compressed_integer
+  signatures:
+    seq:
+      - id: num_signatures
+        type: compressed_integer
+      - id: signatures
+        type: signature
+        repeat: expr
+        repeat-expr: num_signatures.value
+  signature:
+    seq:
+      - id: signature_type
+        type: compressed_integer
+      - id: len_signature
+        type: compressed_integer
+      - id: signature_data
+        size: len_signature.value
+  compressed_integer:
+    seq:
+      - id: groups
+        type: group
+        repeat: until
+        repeat-until: not _.has_next
+    types:
+      group:
+        doc: |
+          One byte group, clearly divided into 7-bit "value" chunk and 1-bit "continuation" flag.
+        seq:
+          - id: b
+            type: u1
+        instances:
+          has_next:
+            value: (b & 0b1000_0000) == 0
+            doc: If true, then we have more bytes to read
+          value:
+            value: b & 0b0111_1111
+            doc: The 7-bit (base128) numeric value chunk of this group
+    instances:
+      len:
+        value: groups.size
+      value:
+        value: >-
+          groups[0].value
+          + (len >= 2 ? (groups[1].value << 7) : 0)
+          + (len >= 3 ? (groups[2].value << 14) : 0)
+          + (len >= 4 ? (groups[3].value << 21) : 0)
+          + (len >= 5 ? (groups[4].value << 28) : 0)
+          + (len >= 6 ? (groups[5].value << 35) : 0)
+          + (len >= 7 ? (groups[6].value << 42) : 0)
+          + (len >= 8 ? (groups[7].value << 49) : 0)
+        doc: Resulting value as normal integer
+enums:
+  checksum_types:
+    0: sha1
+    1: sha256
+    2: sha512
+    3: sha512_128 # first 128 bits of sha512 checksum
--- a/archive/zchunk.ksy
+++ b/archive/zchunk.ksy
@@ -75,6 +75,9 @@ types:
         value: flags.value & 0b1 == 0b1
       has_optional_elements:
         value: flags.value & 0b10 == 0b10
+      compression:
+        value: compression_type.value
+        enum: compression
   optional_element:
     seq:
       - id: element_id
@@ -186,3 +189,6 @@ enums:
     1: sha256
     2: sha512
     3: sha512_128 # first 128 bits of sha512 checksum
+  compression:
+    0: none
+    2: zstd
--- a/archive/zchunk.ksy
+++ b/archive/zchunk.ksy
@@ -127,6 +127,16 @@ types:
         type: compressed_integer
       - id: compression_type_int
         type: compressed_integer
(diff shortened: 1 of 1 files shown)


archive/zchunk.ksy
```diff
@@ -0,0 +1,188 @@
...
+        size: len_index.value
+      - id: signatures
+        type: signatures
+  preface:
+    seq:
+      - id: checksum
+        size: _root.lead.len_checksum
+      - id: flags
+        type: compressed_integer
+      - id: compression_type
```
KOLANICH, line 63: I think it may make sense to create an own type for each type over a compressed integer (also, `-affected-by: 135`). I.e. this one should read a compressed integer in a `seq` and create a `value` instance with `enum`. Also may make sense to utilise `kaitai.compress`.

archive/zchunk.ksy
```diff
@@ -0,0 +1,188 @@
...
+    seq:
+      - id: element_id
+        type: compressed_integer
+      - id: len_data
+        type: compressed_integer
+      - id: data
+        size: len_data.value
+  index:
+    seq:
+      - id: checksum
```
KOLANICH, line 88: We can probably implement types calculating hashes purely in KS. Should we do it?

archive/zchunk.ksy
```diff
@@ -0,0 +1,188 @@
...
+    seq:
+      - id: num_signatures
+        type: compressed_integer
+      - id: signatures
+        type: signature
+        repeat: expr
+        repeat-expr: num_signatures.value
+  signature:
+    seq:
+      - id: signature_type
```
KOLANICH, line 143: is it an enum?

archive/zchunk.ksy
```diff
@@ -0,0 +1,188 @@
...
+        type: compressed_integer
+      - id: len_signature
+        type: compressed_integer
+      - id: signature_data
+        size: len_signature.value
+  compressed_integer:
```
KOLANICH, line 149: It'd be nice to highlight the differences to existing variable length integers.

archive/zchunk.ksy
```diff
@@ -0,0 +1,188 @@
...
+        if: has_optional_elements
+    instances:
+      has_data_streams:
+        value: flags.value & 0b1 == 0b1
+      has_optional_elements:
+        value: flags.value & 0b10 == 0b10
+  optional_element:
+    seq:
+      - id: element_id
+        type: compressed_integer
```
KOLANICH, line 81: It'd be nice to add some docs.

archive/zchunk.ksy
armijnhemel: Yes but the documentation says: " This is an integer containing the type of signature.  Currently there are no
 recognized signature types."

archive/zchunk.ksy
generalmimon: Resolved in https://github.com/armijnhemel/kaitai_struct_formats/commit/e43f79e80c2e63d7ed7e9058736e7ad3bfbed3b1

archive/zchunk.ksy
generalmimon: Signatures are described in the text specification, but the code in https://github.com/zchunk/zchunk doesn't actually implement them:

[zchunk / **zchunk**](https://github.com/zchunk/zchunk) / [`src/lib/header.c:259-264`](https://github.com/zchunk/zchunk/blob/99e51afa38c723e7c25834c2c3b305d20ef55d04/src/lib/header.c#L259-L264)

```c
    /* We don't actually support signatures yet, so bail if there is one */
    zck_log(ZCK_LOG_DEBUG, "Signature count: %i", zck->sigs.count);
    if(zck->sigs.count > 0) {
        set_fatal_error(zck, "Signatures aren't supported yet");
        return false;
    }
```

[zchunk / **zchunk**](https://github.com/zchunk/zchunk) / [`src/lib/header.c:336-338`](https://github.com/zchunk/zchunk/blob/99e51afa38c723e7c25834c2c3b305d20ef55d04/src/lib/header.c#L336-L338)

```c
    for(int i=0; i<zck->sigs.count; i++) {
        // TODO: Add signatures
    }
```


So I actually suggest we do the same: validate that the `num_signatures` is zero, and if it isn't, raise a fatal error. Then we don't have to implement signatures at all, which seems pointless, since it doesn't seem like there are any publicly available implementations of the zchunk format that actually parse (interprets) or generate the signatures. I asked Claude Code if it can find such implementations, and it couldn't:

<details>
<summary>Answer to the question "Are zchunk signatures implemented anywhere?" by Claude Code</summary>

> #### Are zchunk signatures implemented anywhere?
> 
> ##### Short answer
> 
> **No.** There is no publicly known implementation that generates or interprets zchunk signatures. It's a spec-only stub that has been untouched for 8 years, and there are literally no signature types allocated, so there is nothing to interoperate *with*.
> 
> ##### The reference implementation
> 
> It's worse than the two `TODO`s in `src/lib/header.c` suggest — the whole feature is dead code. All references below are permalinked to `99e51af` (v1.5.4, the current `main` at time of writing):
> 
> - [`read_sig()` bails outright](https://github.com/zchunk/zchunk/blob/99e51afa38c723e7c25834c2c3b305d20ef55d04/src/lib/header.c#L259-L264) — `set_fatal_error(zck, "Signatures aren't supported yet")` on any count > 0.
> - [`sig_create()` has an empty loop](https://github.com/zchunk/zchunk/blob/99e51afa38c723e7c25834c2c3b305d20ef55d04/src/lib/header.c#L336-L338) — `// TODO: Add signatures`.
> - **`zck->sigs.count` is never assigned on the write path.** Grepping all of `src/`, the only assignment anywhere is [the read in `read_sig()`](https://github.com/zchunk/zchunk/blob/99e51afa38c723e7c25834c2c3b305d20ef55d04/src/lib/header.c#L256); `sig_create()` just serialises whatever the zeroed-`calloc` left there. The [`zckSig` / `zckSigCollection` structs](https://github.com/zchunk/zchunk/blob/99e51afa38c723e7c25834c2c3b305d20ef55d04/src/lib/zck_private.h#L234-L244) (`hash_type`, `length`, `signature`, `ctx`) have no member written anywhere in the tree.
> - No public API — `zck.h.in` mentions signatures in [exactly one comment](https://github.com/zchunk/zchunk/blob/99e51afa38c723e7c25834c2c3b305d20ef55d04/include/zck.h.in#L223), about header length. No `zck_sign`, no `zck_add_signature`, no verify.
> - No CLI flag in `zck.c` / `unzck.c` / `zck_read_header.c`.
> - The spec itself: [*"Currently there are no recognized signature types."*](https://github.com/zchunk/zchunk/blob/99e51afa38c723e7c25834c2c3b305d20ef55d04/zchunk_format.txt#L231-L233)
> 
> History confirms it was never started: the section landed in `79d1d5a` (2018-04-20, *"Change file format to support streams and signatures"*), last substantively touched in `7bb2219` (2018-11-19). Zero issues or PRs about signatures in the upstream tracker.
> 
> ##### Other implementations
> 
> | Implementation | Signature handling |
> | --- | --- |
> | [zchunk/zchunk-java](https://github.com/zchunk/zchunk-java) (official org, Java, 2020) | **Doesn't even parse it.** `ZChunkHeaderFactory.readSignatureIndex()` ignores its arguments and hardcodes `signatureCount(valueOf(0L))`. `ZChunkHeaderSignatures` has only `getSignatureCount()` — no type/blob fields. (`macaslinbeijay/zchunk-java` is a fork of the same code.) |
> | [yybit/zchunk-rs](https://github.com/yybit/zchunk-rs) (Rust) | **The only one with structural parsing** — `Signature::from_reader` reads type/size/blob and can write it back. But no crypto at all: no signing, no verification, and `Signature::new` is *commented out*, so it can only re-emit signatures it read. Generation always calls `Signatures::new(Vec::new())` → count 0. |
> | [define42/createrepo_go](https://github.com/define42/createrepo_go) (Go) | `skipZchunkSignatures()` — tolerant skip: reads count, then type/size, seeks past the blob with bounds checks. Parse-and-ignore, no interpretation. |
> | [go-deltasync/zchunk](https://github.com/go-deltasync/zchunk) (Go, self-described scaffolding) | Count only; explicitly rejects non-zero, with a comment saying it's matching the reference's *"Signatures aren't supported yet"*. |
> | [tomberek/zchunk-go](https://github.com/tomberek/zchunk-go) (Go) | Nothing. The only `sig` hits in the repo are POSIX signal handlers. |
> | dnf/dnf5, librepo, createrepo_c, mamba/micromamba (via libpowerloader) | All link libzck rather than reimplementing → inherit the fatal error. |
> 
> A GitHub-wide code search for zchunk signature handling turns up only `createrepo_go` outside these repos.
> 
> **The decisive practical point:** because libzck *fatal-errors* on count > 0, a zchunk file carrying a signature is unreadable by the entire RPM ecosystem. So such files effectively cannot exist in the wild — there is no corpus to parse.
> 
> > [!WARNING]
> > One misleading source: the ArchiveTeam file-format wiki [claims](http://fileformats.archiveteam.org/wiki/Zchunk) *"Zchunk also supports an embedded PGP signature."* That statement is uncited and simply wrong — it looks inferred from the spec text. Don't let it (or an LLM that ingested it) convince you otherwise.
> 
> ##### Recommendation for an alternative implementation
> 
> **Don't implement crypto.** There's no type registry, no test vector, no producer.
> 
> **Do implement the structural skip** — it's ~15 lines and it's the difference between a parser that's merely bug-compatible and one that's correct:
> 
> ```
> count = read_compint()
> repeat count times:
>     type = read_compint()
>     size = read_compint()
>     bounds-check size against the remaining header, then skip size bytes
> ```
> 
> This is the `createrepo_go` approach and it's strictly better than the reference's `set_fatal_error`. You can't verify a signature either way, so skipping an unverifiable one is not a security regression relative to libzck — the header checksum and the full-data checksum still hold, and those are what actually protect the file today.
> 
> **On generation**, always emit a single `0x80` (compint 0). Anything else produces a file dnf can't read.
> 
> Two gotchas worth noting, since the reader has to get the offsets right regardless:
> 
> - The signature section *is* inside the header-checksum region — [the spec](https://github.com/zchunk/zchunk/blob/99e51afa38c723e7c25834c2c3b305d20ef55d04/zchunk_format.txt#L49-L53) says the header checksum covers everything "until the end of the signatures".
> - `header_size` includes the signature section, and libzck [warns `"There are unused bytes in the header"`](https://github.com/zchunk/zchunk/blob/99e51afa38c723e7c25834c2c3b305d20ef55d04/src/lib/header.c#L269-L271) if the sections don't exactly fill it, so the count byte must account exactly.
> 
> ##### Sources
> 
> - <https://github.com/zchunk/zchunk> — reference implementation; code links above pinned to [`99e51af`](https://github.com/zchunk/zchunk/commit/99e51afa38c723e7c25834c2c3b305d20ef55d04)
> - <https://github.com/zchunk/zchunk/blob/99e51afa38c723e7c25834c2c3b305d20ef55d04/zchunk_format.txt#L219-L252> — the full signature section of the format spec
> - <https://github.com/zchunk/zchunk-java>
> - <https://github.com/yybit/zchunk-rs>
> - <https://github.com/define42/createrepo_go>
> - <https://github.com/tomberek/zchunk-go>
> - <https://github.com/go-deltasync/zchunk>
> - <http://fileformats.archiveteam.org/wiki/Zchunk>
> - <https://www.jdieter.net/posts/2018/04/30/introducing-zchunk/>

</details>

It suggested not to raise a fatal error if there are any signatures and instead still structurally parse them (even if there is no defined interpretation of them), but I prefer to fail. My thinking is that right now, there should be no `.zck` files in the real world with signatures, because if there were, the reference implementation would reject them. If at some point the reference zchunk implementation starts accepting and generating them (which I don't think is very likely), then I would rather know about that. And the easiest way to learn about it is to have the .ksy spec reject these files - then hopefully someone opens an issue and/or a PR with the fix.

Also, we don't need to be "more correct" (or, as Claude put it, "strictly better") than the reference implementation - I don't see the point in that. There is literally no practical difference between requiring that there are 0 signatures and pretending to parse them in any way, since there are almost certainly no zchunk files with >0 signatures in existence. However, I'd rather not have dead code that can't be tested.

armijnhemel: > So I actually suggest we do the same: validate that the `num_signatures` is zero, and if it isn't, raise a fatal error. Then we don't have to implement signatures at all, which seems pointless, since it doesn't seem like there are any publicly available implementations of the zchunk format that actually parse (interprets) or generate the signatures.

This makes sense, as long as the reason is documented.

generalmimon: Resolved in https://github.com/armijnhemel/kaitai_struct_formats/commit/f80cbb5a1051a99f45b9947d235cd9b8248d5731

archive/zchunk.ksy
generalmimon: Since `checksum_type` appears twice in the format, I extracted it to a reusable type in d3bbc7791864d35eecdae8d46459554e03f66d3e. This also allowed defining the derived `len_checksum` value instance once, rather than having to duplicate it for each of the two `checksum_type` occurrences:

https://github.com/kaitai-io/kaitai_struct_formats/blob/478475f65edbcc6061e98c34c21db55c1b349c7a/archive/zchunk.ksy#L285-L291

`compression_type` appears only here, so I didn't see a compelling reason to create a dedicated type for it.

archive/zchunk.ksy
generalmimon: Calculating SHA-256 purely in KS? No, thanks. That sounds quite insane. Even if it were possible (which I doubt), it would be terribly inefficient.

There is a proposal to add checksum validation as a feature to KS: https://github.com/kaitai-io/kaitai_struct/issues/81

If this were to be implemented at some point in the future, it could include support for hash functions like SHA-256.

PR #540, opened 2021-11-01
add odex specification

armijnhemel:
Add a specification for Android Odex.

### DIFF
--- a/executable/odex.ksy
+++ b/executable/odex.ksy
@@ -1,0 +1,54 @@
+meta:
+  id: odex
+  title: Odex
+  file-extension: odex
+  license: Apache-2.0
+  ks-version: 0.9
+  imports:
+    - /executable/dex
+  endian: le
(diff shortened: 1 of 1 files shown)


executable/odex.ksy
```diff
@@ -0,0 +1,54 @@
...
+    type: u4
+    doc: offset of optimized DEX dependency table
+  - id: len_deps
+    type: u4
+  - id: ofs_opt
+    type: u4
+    doc: file offset of optimized data tables
+  - id: len_opt
+    type: u4
+  - id: flags
```
KOLANICH, line 33: Should it be a type of bit fields?

armijnhemel: I could not find any information in the header file about what this should be:

http://web.archive.org/web/20180816094438/https://android.googlesource.com/platform/dalvik.git/+/master/libdex/DexFile.h#554

PR #542, opened 2021-11-01
add specification for UF2 (USB Flashing Format)

armijnhemel:


firmware/uf2.ksy
```diff
@@ -0,0 +1,169 @@
...
+doc: |
+  UF2 is a file format, developed by Microsoft for PXT (also known as
+  Microsoft MakeCode), that is particularly suitable for flashing
+  microcontrollers over MSC (Mass Storage Class; aka removable flash drive).
+
+  The list of family-ids is stored in a separate JSON file:
+
+  <https://github.com/microsoft/uf2/blob/master/utils/uf2families.json>
+
+  This file is regularly updated and the family_id enum should be kept in
```
KOLANICH, line 16: Probably enums should go into a separate file that is generated by a script.

firmware/uf2.ksy
```diff
@@ -0,0 +1,169 @@
...
+types:
+  uf2_block:
+    seq:
+      - id: magic
+        -orig-id: magicStart0
+        contents: "UF2\n"
+      - id: second_magic
+        -orig-id: magicStart1
+        contents: [0x57, 0x51, 0x5d, 0x9e]
+      - id: flags
```
KOLANICH, line 40: should it be a type of bit-sized fields?

firmware/uf2.ksy
```diff
@@ -0,0 +1,169 @@
+meta:
+  id: uf2
```
KOLANICH, line 2: maybe `usb_flashing_format`. `uf2` is too short, there may be a collision.

firmware/uf2.ksy
armijnhemel: Maybe, but it seems that that list gets very few updates, so perhaps we could leave it as it is.

firmware/uf2.ksy
armijnhemel: UF2 is the most well known name for this format. I did a few queries with several search engines and all the first hits are for this particular format, so I am not too worried about collisions.

firmware/uf2.ksy
```diff
@@ -0,0 +1,315 @@
...
+  <https://micropython.org/download/rp2-pico/>
+
+doc-ref: https://github.com/microsoft/uf2/blob/90e9741f217f5a40c98ba74d663e408041037578/README.md
+seq:
+  - id: uf2_block_start
+    type: uf2_block
+  - id: uf2_blocks
+    type: uf2_block
+    repeat: expr
+    repeat-expr: uf2_block_start.num_blocks - 1
```
generalmimon: I think I know why this isn't just `repeat: eos` (because that wouldn't work well for carving in [BANG](https://github.com/armijnhemel/binaryanalysis-ng)), and I'm generally fine with keeping the .ksy specs self-delimiting whenever possible.

However, I found that when parsing the test file [`RPI_PICO2-20260824-v1.29.0.uf2`](https://micropython.org/resources/firmware/RPI_PICO2-20260824-v1.29.0.uf2) from the first download link **v1.29.0 (2026-08-24) .uf2** on the page https://micropython.org/download/RPI_PICO2/ (which is not the link you provided in `/doc`, which now actually redirects from <code><del>https://micropython.org/download/rp2-pico/</del></code> to <code>https://micropython.org/download/RPI_PICO/</code> - which is a different page), this self-delimits after two blocks, but the file actually has 1292 blocks in total.

I see that from the official spec ([microsoft / **uf2** / `README.md:44`](https://github.com/microsoft/uf2/blob/90e9741f217f5a40c98ba74d663e408041037578/README.md?plain=1#L44)) that `num_blocks` should be indeed "Total number of blocks in file". Therefore, I figured that `RPI_PICO2-20260824-v1.29.0.uf2` is essentially a concatenation of multiple UF2 files, so I made a tiny helper spec `uf2_files.ksy` to test it:

```ksy
meta:
  id: uf2_files
  imports:
    - uf2
seq:
  - id: files
    type: uf2
    repeat: eos
```

To my horror, I found that the second UF2 file fails to parse with `EOFError: requested 4 bytes, but only 0 bytes available` (in the [Web IDE](https://ide.kaitai.io/devel/)). It seems that the cause of the problem is that the very first block declares 2 in `num_blocks`, when in reality there should've been 1. As a result, the second UF2 file attempts to parse one more block after the end of file.

I don't know if whatever tool generated this `RPI_PICO2-20260824-v1.29.0.uf2` file simply has a bug (to be honest, I think that's most likely the case), or if this is somehow valid/expected...

The UF2 spec does say that several files can be concatenated together, see <https://github.com/microsoft/uf2/blob/90e9741f217f5a40c98ba74d663e408041037578/README.md#family-id>:

> If the `familyID` doesn't match, the bootloader should disregard the entire block, including `blockNo` and `numBlocks` fields. In particular, writing a full UF2 file with non-matching `familyID` should not reset the board. **This also allows for several files with different `familyID` to be simply concatenated together**, and the whole resulting file to be copied to the device with only one actually being written to flash.

It says "several files with different `familyID`", which implies that all blocks within one file should have the same family ID. That further indicates that `RPI_PICO2-20260824-v1.29.0.uf2` is invalid, because the first block has family ID `RP2XXX_ABSOLUTE` and the second has `RP2350_ARM_S` (whereas for `"numBlocks": 0x2` to be valid, they would have to be the same):

```json
{
  "uf2BlockStart": {
    "magic": [85, 70, 50, 10],
    "secondMagic": [87, 81, 93, 158],
    "flags": 0xa000,
    "targetAddress": 0x10ffff00,
    "lenPayload": 0x100,
    "blockNumber": 0x0,
    "numBlocks": 0x2,
    "familyId": { "name": "RP2XXX_ABSOLUTE", "value": 0xe48bff57 },
    "data": { "$start": 32, "$end": 507 },
    "endMagic": [48, 111, 177, 10]
  },
  "uf2Blocks": [
    {
      "magic": [85, 70, 50, 10],
      "secondMagic": [87, 81, 93, 158],
      "flags": 0x2000,
      "targetAddress": 0x10000000,
      "lenPayload": 0x100,
      "blockNumber": 0x0,
      "numBlocks": 0x502,
      "familyId": { "name": "RP2350_ARM_S", "value": 0xe48bff59 },
      "data": { "$start": 544, "$end": 1019 },
      "endMagic": [48, 111, 177, 10]
    }
  ]
}
```

armijnhemel: I think that this is because of the flags. The flag of the first blocks indicates that there is extension data. According to the documentation ( https://github.com/microsoft/uf2#extension-tags ):

```
When the 0x8000 flag is set, additional information can be appended right after payload data (i.e., it starts at 32 + payloadSize).

...


Extension tags can, but don't have to, be repeated in all blocks.
```

So what I think what happens is that the very first and very last block in the file are the `num_blocks` listed in the first file, and everything in between is actually "extension data".

armijnhemel: So my comment above apparently isn't entirely correct. The file was created using picotool: https://github.com/raspberrypi/picotool

A related discussion is at https://forums.raspberrypi.com/viewtopic.php?t=375030 where it says:

"See errata E10 in the datasheet for why this block is necessary, and why it has the wrong num_blocks."

The relevant commit in picotool should be: 78c9bd121b09399823b67ee7ea89003ca0d3315f

https://github.com/raspberrypi/picotool/commit/78c9bd121b09399823b67ee7ea89003ca0d3315f

So yes, it seems that the file is indeed not valid according to the specification but to work around a hardware issue.

firmware/uf2.ksy
generalmimon: The `family_id` enum is now generated using the commands listed in `/doc` - see e32fc63347fb25049f6376d6ac908f3fbef41892.

firmware/uf2.ksy
generalmimon: Fixed in 7c4c88802a7c701c5aa1ad564d31f12d062c07d2

firmware/uf2.ksy
generalmimon: See 1b47d195ed27fb1402dbbb336eb7a60bf422ef5b (I decided not to use [bit-sized integers](https://doc.kaitai.io/user_guide.html#_bit_sized_integers), because using value instances and old-school `&`/`!=` unpacking is arguably easier)

PR #543, opened 2021-11-01
add specification for device firmware upgrade files (.dfu)

armijnhemel:


firmware/dfu.ksy
```diff
@@ -0,0 +1,77 @@
...
+    type: target
+    repeat: expr
+    repeat-expr: num_targets
+  - id: device
+    doc: Firmware version
+    type: u2
+  - id: product
+    type: u2
+  - id: vendor
+    type: u2
```
KOLANICH, line 34: should it be an enum?

firmware/dfu.ksy
```diff
@@ -0,0 +1,77 @@
...
+    valid: 0x11a
+  - id: ufd
+    size: 3
+    contents: "UFD"
+  - id: len_suffix
+    type: u1
+    valid: 16
+  - id: checksum
+    type: u4
```
KOLANICH, line 45: which one?

firmware/dfu.ksy
armijnhemel: This refers to USB ids, there are quite a few of those (many thousands) and I am not so sure that would help here.

KOLANICH: We should generate a file with them automatically. I don't remember if I have written a script for that, if not, probably I should.

firmware/dfu.ksy
armijnhemel: It is a crc32. I will clarify this.

PR #544, opened 2021-11-01
add specification for snappy framing format

armijnhemel:
This specification adds support for the snappy framing format. The use of `repeat: until` means that sometimes (if there is trailing data) an extra byte is read. A program using this code could detect this by looking at the `is_valid` attribute.

### DIFF
--- a/archive/snappy.ksy
+++ b/archive/snappy.ksy
@@ -1,0 +1,57 @@
+meta:
+  id: snappy
+  title: snappy framing format
+  license: CC0-1.0
+  ks-version: 0.9
+  encoding: utf-8
+  endian: le
+doc-ref: https://github.com/google/snappy/blob/master/framing_format.txt
+doc: |
+  snappy's framing2 format
+  
+  Test files can be created with snzip: <https://github.com/kubo/snzip>
+seq:
+  - id: chunks
+    type: chunk
+    repeat: until
+    repeat-until: _io.eof or not _.is_valid
+    # This is ugly, as it eats one extra byte, so an external
+    # program processing this could should take this into account
+types:
+  chunk:
+    seq:
+      - id: identifier
+        type: u1
+        enum: chunk_types
+      - id: body
+        type: chunk_body
+        if: is_valid
+    instances:
+        # more values are actually allowed but in practice
+        # these aren't encountered
+        is_valid:
+          value: identifier == chunk_types::compressed or
+            identifier == chunk_types::uncompressed or
+            identifier == chunk_types::padding or
+            identifier == chunk_types::stream_identifier
+  chunk_body:
+    seq:
+      - id: len_chunk
+        type: int3
+      - id: data
+        size: len_chunk.value
+  int3:
+    seq:
+      - id: lower
+        type: u2
+      - id: higher
+        type: u1
+    instances:
+      value:
+        value: higher * 65536 + lower
+enums:
+  chunk_types:
+    0x00: compressed
+    0x01: uncompressed
+    0xfe: padding
+    0xff: stream_identifier


archive/snappy.ksy
```diff
@@ -0,0 +1,57 @@
...
+            identifier == chunk_types::uncompressed or
+            identifier == chunk_types::padding or
+            identifier == chunk_types::stream_identifier
+  chunk_body:
+    seq:
+      - id: len_chunk
+        type: int3
+      - id: data
+        size: len_chunk.value
+  int3:
```
KOLANICH, line 43: `s3` maybe

archive/snappy.ksy
```diff
@@ -0,0 +1,57 @@
...
+    seq:
+      - id: lower
```
KOLANICH, line 45: `hi`, `lo`?

archive/snappy.ksy
KOLANICH: also, have recreated it multiple times in multiple specs. How about modularizing it?

archive/snappy.ksy
```diff
@@ -0,0 +1,57 @@
...
+        type: u2
+      - id: higher
+        type: u1
+    instances:
+      value:
+        value: higher * 65536 + lower
```
KOLANICH, line 51: I usually use bit ops here.

archive/snappy.ksy
armijnhemel: Personally I would be fine with that. I think that this has come up a few times already in the past. Adding @generalmimon

dgelessus: `b24` should also work, right? That wouldn't require an extra type.

armijnhemel: I hadn't thought of that. I will try!

KOLANICH: Only for be.

generalmimon: > Only for be.

If you mean that `b24` would only work for _big-endian_ (`be`) order, well, you can use `type: b24le` (or `meta/bit-endian: le` + `type: b24`) and it will work for little-endian order just as you'd expect.

This is exactly why I discourage not using an explicit `meta/bit-endian` setting and leaving it set to the default value (`be` because of backward compatibility) in the [User Guide](https://doc.kaitai.io/user_guide.html#bit-endian). If you do that, it works, but it essentially suggests that the big-endian order is the only parsing direction supported.

archive/snappy.ksy
generalmimon: > I usually use bit ops here.

Yes, that would definitely make more sense here. It's about expressing the intent: here `higher * 65536` is actually meant to be `higher << 16` (or even `higher << lower._bitsizeof` which is supported [since 0.9](https://kaitai.io/news/2020/10/16/kaitai-struct-v0.9-released.html#release-highlights)) and `+ lower` should be `| lower`. This is quite literally what you want to do - first shift bits of `higher` (`00 00 HH`)  to the left (`HH 00 00`) so that it doesn't collide with `lower` (`00 LL LL`) and then compose them together to get `HH LL LL` - the bitwise OR is enough for that, there is no logical reason why you would want addition `+` with all that carry-in/carry-out madness.

_(This review comment is probably made obsolete by https://github.com/kaitai-io/kaitai_struct_formats/pull/544/files#r744576387 in the sense that the type `int3` would be dropped, but I can't miss the opportunity to share my view on this :-) )_

generalmimon: > (or even `higher << lower._bitsizeof` which would work [since 0.9](https://kaitai.io/news/2020/10/16/kaitai-struct-v0.9-released.html#release-highlights))

OK, it is broken:

```ksy
meta:
  id: test
  endian: le
seq:
  - id: i
    type: int3
types:
  int3:
    seq:
      - id: lower
        type: u2
      - id: higher
        type: u1
    instances:
      value:
        # value: higher * 65536 + lower
        value: higher << lower._bitsizeof | lower
```

```
/types/int3/instances/value/value: error: don't know how to call method '_bitsizeof' of object type 'IntMultiType(false,Width2,Some(LittleEndian))'
```

and that is just another reminder of how much work there is to be done.

But at least `_sizeof` works:

```ksy
    instances:
      value:
        # value: higher * 65536 + lower
        value: higher << (8 * lower._sizeof) | lower
```

though it's slightly less elegant.

archive/snappy.ksy
KOLANICH: Thanks for the info, I should give it a try!

Issue #138, opened 2019-04-27
Add Adobe Photoshop (PSD) file format


davidhicks:
Official specification of the Adobe Photoshop (.psd) file format: https://www.adobe.com/devnet-apps/photoshop/fileformatashtml/

Work in progress branch for implementing a Kaitai Struct format specification for .psd: https://github.com/davidhicks/kaitai_struct_formats/tree/psd

GreyCat:
Alternative PSD spec implementation from @ertaquo:

https://gist.github.com/ertaquo/b1d12c37a21268e3d095d39e196f5863

KOLANICH:
Also https://gist.github.com/vladdancer/832aec8790995a5676dae0839993807a

PR #545, opened 2021-11-01
add specification for Photoshop PSD

armijnhemel:
This specification adds support for a few variants of Photoshop PSD. I have tested with a few of my personal files (which I cannot share) and it seems to work well.

### DIFF
--- a/image/psd.ksy
+++ b/image/psd.ksy
@@ -1,0 +1,574 @@
+meta:
+  id: psd
+  title: Photoshop PSD
+  file-extension: psd
+  license: CC0-1.0
+  ks-version: 0.9
+  endian: be
+doc-ref: https://www.adobe.com/devnet-apps/photoshop/fileformatashtml/
+seq:
+  - id: header
+    type: header
+  - id: color_mode_data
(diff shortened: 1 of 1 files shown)


KOLANICH:
Fixes: #138

image/psd.ksy
```diff
@@ -0,0 +1,574 @@
...
+      - id: magic
+        contents: "8BIM"
+      - id: resource_id
+        type: u2
+        enum: resource_ids
+      - id: name
+        type: pascal_string
+      - id: padding1
+        size: 1
+        if: name.len_string % 2 == 0
```
KOLANICH, line 92: Is padding should be a function of `len_string`, not `_io.pos`?

KOLANICH:
Here are the test files: https://github.com/chinedufn/psd/tree/master/tests/fixtures

image/psd.ksy
armijnhemel: No, not for `_io.pos`. The specification says: "Name: Pascal string, padded to make the size even (a null name consists of two bytes of 0)".

PR #548, opened 2021-11-01
update uimage.ksy

armijnhemel:
* many sanity checks
* magic values used by specific vendors (Cisco, ZyXEL, Genius)
* add support for ASUS specific modifications

The ASUS modifications are for the following product families:

* BRT-*
* GS-*
* GT-*
* RP-*
* RT-*

### DIFF
--- a/firmware/uimage.ksy
+++ b/firmware/uimage.ksy
@@ -17,7 +17,14 @@ types:
   uheader:
     seq:
       - id: magic
-        contents: [0x27, 0x05, 0x19, 0x56]
+        type: u4
+        enum: magic_types
+        valid:
+          any-of:
+            - magic_types::uimage
+            - magic_types::bix
+            - magic_types::bix2
+            - magic_types::bix3
       - id: header_crc
         type: u4
       - id: timestamp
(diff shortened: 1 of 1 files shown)


firmware/uimage.ksy
```diff
@@ -221,12 +306,24 @@ enums:
...
       doc: No Compression Used
-    1: gzip
-    2: bzip2
-    3: lzma
-    4: lzo
-    5: lz4
-    6: zstd
+    1:
+      id: gzip
+      doc: gzip Compression Used
```
KOLANICH, line 311: IMHO it is just redundant and distracting.

firmware/uimage.ksy
```diff
@@ -42,10 +80,57 @@ types:
       - id: compression_type
         type: u1
         enum: uimage_comp
-      - id: name
+        valid:
```
KOLANICH, line 83: Should it (and in other places too) really be validated?

armijnhemel: YMMV of course, but when carving (which I use it for) it allows me to quickly reject false positives, so I would recommend adding these checks.

firmware/uimage.ksy
armijnhemel: Maybe! I am not too attached to this.

generalmimon: > IMHO it is just redundant and distracting.

@KOLANICH I think the same way, I had once removed these previously (https://github.com/kaitai-io/kaitai_struct_formats/commit/3d0a2a5d31004be37131b441ceb81ff2ce04a427), so this is a regression.

armijnhemel: > > IMHO it is just redundant and distracting.
> 
> @KOLANICH I think the same way, I had once removed these previously ([3d0a2a5](https://github.com/kaitai-io/kaitai_struct_formats/commit/3d0a2a5d31004be37131b441ceb81ff2ce04a427)), so this is a regression.

Oops. I thought I had synced properly. I will push a new commit, apologies!

firmware/uimage.ksy
armijnhemel: > YMMV of course, but when carving (which I use it for) it allows me to quickly reject false positives, so I would recommend adding these checks.

So to add a little bit more context: this mostly happens when scanning files that have the alternative magic signatures.

KOLANICH: `-affected-by: 435` (https://github.com/kaitai-io/kaitai_struct/issues/435#issuecomment-389597122)

PR #549, opened 2021-11-01
add specification for libminikin text layout files

armijnhemel:
This specification adds support for text format files for libminikin. As far as I know the format is no longer used on Android and replaced by something else. I will search for some public test files.

### DIFF
--- a/font/minikin_hyb.ksy
+++ b/font/minikin_hyb.ksy
@@ -1,0 +1,122 @@
+meta:
+  id: minikin_hyb
+  title: Libminikin text layout format
+  file-extension: hyb
+  tags:
+    - android
+  license: CC0-1.0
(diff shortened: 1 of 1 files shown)


font/minikin_hyb.ksy
```diff
@@ -0,0 +1,122 @@
+meta:
+  id: minikin_hyb
+  title: Libminikin text layout format
+  file-extension: hyb
+  tags:
+    - android
+  license: CC0-1.0
+  endian: le
+doc: |
+  Text layout format as used on Android
```
KOLANICH, line 10: Not just `text layout`, but `hyphenation`

KOLANICH:
Also you can consider implementing lookup routine as a KS type.

PR #551, opened 2021-11-01
pcap: add more link types, extra sanity check for the version number

armijnhemel:


network/pcap.ksy
```diff
@@ -186,3 +192,26 @@ enums:
...
+    266: usb_darwin
+    268: sdlc
+    270: loratap
+    271: vsock
+    272: nordic_ble
+    273: docsis31_xra31
+    274: ethernet_mpacket
+    275: displayport_aux
+    276: linux_sll2
+    278: openvisla
```
generalmimon, line 205:
```suggestion
    278: openvizsla
```

See https://www.tcpdump.org/linktypes/LINKTYPE_OPENVIZSLA.html

network/pcap.ksy
```diff
@@ -34,6 +34,8 @@ types:
         contents: [0xd4, 0xc3, 0xb2, 0xa1]
       - id: version_major
         type: u2
+        valid:
+          max: 2
```
generalmimon, line 38: @armijnhemel Does the specification support major versions `0` and `1`?

generalmimon: > I would expect very few version 0 and 1 files to exist. I wouldn't object to only making `2` valid and ignore `0` and `1`.

Yep, I'd go for that - we declare that we're only confident that the spec works with major version 2. If there still is some PCAP file version 0 or 1 (which I doubt, given that you write "_Version 2.4 (so major version 2) has been used since at least 1998._"), at least we'll be able to quickly identify it and adapt the spec accordingly, if it makes sense.

```suggestion
        valid:
          eq: 2
```

generalmimon (approved):
@armijnhemel LGTM, thanks for the contribution!

PR #552, opened 2021-11-02
png: extra sanity checks, add Adobe Fireworks, Evernote and "pngattach" chunks

armijnhemel:
This PR adds some sanity checks to the format (width/height cannot be 0) and adds support for Adobe Fireworks, Evernote/Skitch and "pngattach" chunks.

Evernote/Skitch info: https://web.archive.org/web/20210302212148/https://discussion.evernote.com/forums/topic/88532-how-to-extract-annotation-information-from-annotated-evernoteskitch-images/

Evernote/Skitch test file: https://github.com/tleguern/lgpng/blob/396f61e11c7773a3c2bc4a7c19bb26ff757581b5/samples/skRf.png

### DIFF
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -102,6 +102,9 @@ types:
 
             # evernote/skitch chunks
             '"skRf"': evernote_skrf_chunk
+
+            # pngattach: https://nullprogram.com/blog/2021/12/31/
+            '"atCh"': atch_chunk
       - id: crc
         size: 4
   ihdr_chunk:
@@ -411,6 +414,24 @@ types:
         size: 16
       - id: data
         size-eos: true
+  atch_chunk:
+    seq:
+      - id: name
+        type: strz
+        encoding: utf-8
+      - id: compression
+        type: u1
+        enum: compression_attach_methods
+        valid:
+          any-of:
+            - compression_attach_methods::no_compression
+            - compression_attach_methods::zlib
+      - id: data
+        size-eos: true
+    enums:
+      compression_attach_methods:
+        0: no_compression
+        1: zlib
 enums:
   color_type:
     0: greyscale
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -10,6 +10,7 @@ meta:
     justsolve:
       - PNG
       - APNG
+      - Fireworks_PNG
     loc: fdd000153
     mime:
       - image/png
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -61,6 +61,8 @@ types:
         type: str
         size: 4
         encoding: UTF-8
+        valid:
+          expr: type != "\0\0\0\0"
       - id: body
         size: len
         type:
@@ -99,8 +101,12 @@ types:
     seq:
       - id: width
         type: u4
+        valid:
+          min: 1
       - id: height
         type: u4
+        valid:
+          min: 1
       - id: bit_depth
         type: u1
       - id: color_type
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -100,7 +100,8 @@ types:
             '"mkTS"': adobe_fireworks_chunk
             '"prVW"': adobe_fireworks_chunk
 
-            # evernote/skitch chunks
+            # Evernote/Skitch chunks
+            '"skMf"': evernote_skmf_chunk
             '"skRf"': evernote_skrf_chunk
 
             # pngattach
@@ -408,12 +409,38 @@ types:
       - id: preview_data
         process: zlib
         size-eos: true
+  evernote_skmf_chunk:
+    doc-ref: https://web.archive.org/web/20210302212148/https://discussion.evernote.com/forums/topic/88532-how-to-extract-annotation-information-from-annotated-evernoteskitch-images/#comment-451501
+    seq:
+      - id: json
+        type: str
+        encoding: UTF-8
+        size-eos: true
+        doc: |
+          JSON document with information about editable annotations (text,
+          lines, paths, etc.) in Evernote/Skitch.
+
+          It refers to the original image stored in the `skRf` chunk (which
+          usually follows immediately after `skMf`) via the
+          `.children[0].children[0].uri` JSON property. This has the format
+          `"skitch+uuid:///$UUID"`, where `$UUID` is a random UUIDv4 value that
+          matches the `uuid` field in `evernote_skrf_chunk` (i.e. in the first
+          16 bytes of the `skRf` chunk).
   evernote_skrf_chunk:
+    doc-ref: https://web.archive.org/web/20210302212148/https://discussion.evernote.com/forums/topic/88532-how-to-extract-annotation-information-from-annotated-evernoteskitch-images/#comment-451501
+    -webide-representation: '{uuid:uuid}'
     seq:
       - id: uuid
         size: 16
-      - id: data
+        doc: |
+          Random UUIDv4 value used to identify the image. It is referenced by
+          the `skMf` chunk - see the documentation for the `json` field in
+          `evernote_skmf_chunk`.
+      - id: orig_img
         size-eos: true
+        doc: |
+          The original source image without annotations. It's usually a PNG
+          image as well, but it can also be a JPEG or possibly other formats.
   atch_chunk:
     doc-ref:
       - https://github.com/skeeto/scratch/tree/58470254f4a95cdf7a53888e405c851c21eb2cae/pngattach
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -103,7 +103,7 @@ types:
             # evernote/skitch chunks
             '"skRf"': evernote_skrf_chunk
 
-            # pngattach: https://nullprogram.com/blog/2021/12/31/
+            # pngattach
             '"atCh"': atch_chunk
       - id: crc
         size: 4
@@ -415,6 +415,9 @@ types:
       - id: data
         size-eos: true
   atch_chunk:
+    doc-ref:
+      - https://github.com/skeeto/scratch/tree/58470254f4a95cdf7a53888e405c851c21eb2cae/pngattach
+      - https://nullprogram.com/blog/2021/12/31/ A new protocol and tool for PNG file attachments
     seq:
       - id: name
         type: strz
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -94,6 +94,14 @@ types:
             '"acTL"': animation_control_chunk
             '"fcTL"': frame_control_chunk
             '"fdAT"': frame_data_chunk
+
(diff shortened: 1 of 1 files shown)


generalmimon:
@armijnhemel:

> Evernote/Skitch test file: https://content.invisioncic.com/Mevernote/post-269465-0-70688200-1442655592.png

Would you believe that this URL now returns a different file without the `skRf` chunk? 🤦‍♂️

Thankfully, Wayback Machine has archived the [original](https://web.archive.org/web/20210819105020id_/https://content.invisioncic.com/Mevernote/post-269465-0-70688200-1442655592.png) and there is a [copy on GitHub](https://github.com/tleguern/lgpng/blob/396f61e11c7773a3c2bc4a7c19bb26ff757581b5/samples/skRf.png) as well. But it confused me at first.

```console
$ curl -w '%{url_effective}\n' -fsSL -o from-invisioncic.png https://content.invisioncic.com/Mevernote/post-269465-0-70688200-1442655592.png
curl: (22) The requested URL returned error: 403
https://media.invisioncic.com/Mevernote/post-269465-0-70688200-1442655592.png
$ curl -w '%{url_effective}\n' -H 'User-Agent: definitely not curl :)' -fsSL -o from-invisioncic.png https://content.invisioncic.com/Mevernote/post-269465-0-70688200-1442655592.png
https://media.invisioncic.com/Mevernote/post-269465-0-70688200-1442655592.png
$ curl -w '%{url_effective}\n' -fsSL -o from-wayback-machine.png https://web.archive.org/web/20210819105020id_/https://content.invisioncic.com/Mevernote/post-269465-0-70688200-1442655592.png
https://web.archive.org/web/20210819105020id_/https://content.invisioncic.com/Mevernote/post-269465-0-70688200-1442655592.png
$ ll
total 340
drwxrwxr-x 2 ubuntu ubuntu   4096 Jun  2 12:03 ./
drwxr-x--- 1 ubuntu ubuntu   4096 Jun  2 12:01 ../
-rw-rw-r-- 1 ubuntu ubuntu  84499 Jun  2 12:03 from-invisioncic.png
-rw-rw-r-- 1 ubuntu ubuntu 251722 Jun  2 12:03 from-wayback-machine.png
$ ksdump -f json from-invisioncic.png /c/temp/kaitai_struct/formats/image/png.ksy > from-invisioncic.json
$ jq '.chunks | map(.type)' from-invisioncic.json
[
  "gAMA",
  "cHRM",
  "IDAT",
  "IEND"
]
$ ksdump -f json from-wayback-machine.png /c/temp/kaitai_struct/formats/image/png.ksy > from-wayback-machine.json
$ jq '.chunks | map(.type)' from-wayback-machine.json
[
  "gAMA",
  "cHRM",
  "pHYs",
  "IDAT",
  "IDAT",
  "IDAT",
  "IDAT",
  "IDAT",
  "IDAT",
  "IDAT",
  "IDAT",
  "skMf",
  "skRf",
  "IEND"
]
```

This is just another example of how important it is to archive files that are floating around on the internet, rather than just linking to them. That was also the idea of https://github.com/kaitai-io/kaitai_struct_samples - collect sample files from various sources, but ideally always make sure we have a copy in case the original disappears. Sometimes it's unclear what the "license" for a potential sample file is and whether it permits distribution, though...

armijnhemel:
> This is just another example of how important it is to archive files that are floating around on the internet, rather than just linking to them. That was also the idea of https://github.com/kaitai-io/kaitai_struct_samples - collect sample files from various sources, but ideally always make sure we have a copy in case the original disappears. Sometimes it's unclear what the "license" for a potential sample file is and whether it permits distribution, though...

Yeah, i did check beforehand if the file was in the Internet archive.

image/png.ksy
```diff
@@ -385,6 +402,66 @@ types:
...
+    doc-ref:
+      - https://github.com/skeeto/scratch/tree/58470254f4a95cdf7a53888e405c851c21eb2cae/pngattach
+      - https://nullprogram.com/blog/2021/12/31/ A new protocol and tool for PNG file attachments
+    seq:
+      - id: file_name
+        type: strz
+        encoding: utf-8
+        valid:
+          # See https://github.com/skeeto/scratch/blob/58470254f4a95cdf7a53888e405c851c21eb2cae/pngattach/pngattach.c#L466-L468
+          expr: _.length != 0 and _.substring(0, 1) != "."
```
generalmimon, line 455: @armijnhemel Note that this is a bit more stricter handling than BANG currently has, which fall backs to `unpacked_from_png` if the file name is invalid, instead of raising a fatal error (as `valid/expr` does):

https://github.com/armijnhemel/binaryanalysis-ng/blob/cf65db2d9e91961b0fabc4454ce1a22b04c2dc18/src/bang/parsers/image/png/UnpackParser.py#L126-L132

```py
                if i.type == 'atCh':
                    if i.body.name == '':
                        file_path = pathlib.Path('unpacked_from_png')
                    elif i.body.name in ['.', '..', '/']:
                        file_path = pathlib.Path('unpacked_from_png')
                    else:
                        file_path = pathlib.Path(i.body.name)
```

But I don't think it matters in practice, because `pngattach` also rejects such files:

https://github.com/skeeto/scratch/blob/58470254f4a95cdf7a53888e405c851c21eb2cae/pngattach/pngattach.c#L459-L468

```c
// Validate and extract the components of a slurped chunk.
static _Bool
atch_parse(struct atch *a, unsigned char *buf, size_t len)
{
    if (len < 3) {
        return 0;
    }
    if (!buf[0] || buf[0] == '.') {
        return 0;
    }
```

... and it will also not produce them:

https://github.com/skeeto/scratch/blob/58470254f4a95cdf7a53888e405c851c21eb2cae/pngattach/pngattach.c#L310-L324

```c
// Write an atCh chunk to a stream.
static const char *
png_atch(FILE *fo, const char *path, int flags)
{
    struct offlen name = basename(path);
    if (name.len > (uint32_t)-1 - 2) {
        return errwrap("attachment name too large", path);
    }
    if (name.len < 1) {
        return errwrap("attachment name too short", path);
    }

    if (path[name.off] == '.') {
        return "invalid attachment name (begins with .)";
    }
```

image/png.ksy
```diff
@@ -385,6 +402,66 @@ types:
...
+          any-of:
+            - compression_attach_methods::none
+            - compression_attach_methods::zlib
+      - id: data_plain
+        size-eos: true
+        if: compression == compression_attach_methods::none
+      - id: data_zlib
+        size-eos: true
+        process: zlib
+        if: compression == compression_attach_methods::zlib
```
generalmimon, line 485: Essentially the same note as above - if the zlib payload was invalid/corrupted, `process: zlib` would raise a fatal error, instead of silently ignoring it, as BANG does:

https://github.com/armijnhemel/binaryanalysis-ng/blob/cf65db2d9e91961b0fabc4454ce1a22b04c2dc18/src/bang/parsers/image/png/UnpackParser.py#L134-L140

```py
                    if i.body.compression == png.Png.AtchChunk.CompressionAttachMethods.zlib:
                        try:
                            with meta_directory.unpack_regular_file(file_path) as (unpacked_md, outfile):
                                outfile.write(zlib.decompress(i.body.data))
                                yield unpacked_md
                        except:
                            pass
```

I don't know if it matters, which is why I'm pointing it out.

armijnhemel: Oh, good catch. I will add a few extra checks in my code. Although I am sure that pngattach doesn't create such files it could always be the case that things get mangled somehow.

image/png.ksy
armijnhemel: Would it allow names starting with a `/`?

generalmimon: @armijnhemel:

> Would it allow names starting with a `/`?

Yes - the current `valid/expr` only checks that the file name is not empty and does not start with a dot. I only implemented these checks because it's easy to implement them fully in Kaitai Struct.

As for `pngattach`, it also rejects all file names that contain control characters (less than `0x20`) or slashes:

https://github.com/skeeto/scratch/blob/58470254f4a95cdf7a53888e405c851c21eb2cae/pngattach/pngattach.c#L470-L475

```c
    unsigned char *end = buf;
    for (; end < buf+len-1 && *end; end++) {
        if ((*end&0xff) < ' ' || *end == '/' || *end == '\\') {
            return 0;
        }
    }
```

I _could_ add a check that the file name does not start with `/`, but I'm not sure if it's a good idea, because that would check the "does not contain `/`" condition only partially. Perhaps it's better to have a clear separation here - currently, the .ksy spec just checks the requirements put on the file name (from the [official specification](https://github.com/skeeto/scratch/tree/58470254f4a95cdf7a53888e405c851c21eb2cae/pngattach#atch-chunk-specification)) highlighted here **in bold** and doesn't even attempt to check the rest:

> **The name must be at least one byte long, not counting the null terminator. It cannot begin with a period (`0x2e`)**, nor contain control bytes (anything less than `0x20`), nor slash (`0x2f`), nor backslash (`0x5c`), i.e. no directory hierarchies.

The idea is that since the other checks cannot be (easily) implemented in Kaitai Struct, it's better to leave them entirely to the applications using the .ksy spec.

I also tried to document this in the .ksy - let me know if it should be improved:

https://github.com/kaitai-io/kaitai_struct_formats/blob/9108716e58a59de3d4ab2732d0829c86e0cc9b03/image/png.ksy#L464-L471

generalmimon (approved):
@armijnhemel For me, it's good enough - please take a look?

image/png.ksy
```diff
@@ -92,15 +95,31 @@ types:
...
       - id: width
         type: u4
+        valid:
+          min: 1
       - id: height
         type: u4
+        valid:
+          min: 1
       - id: bit_depth
         type: u1
```
armijnhemel, line 124: valid:
          any-of: [1, 2, 4, 8, 16]

armijnhemel:
> @armijnhemel For me, it's good enough - please take a look?

One small nitpick, but looking good!

image/png.ksy
generalmimon: @armijnhemel I'll leave this for the next PR in which I'll incorporate the changes from [`png.ksy` in BANG](https://github.com/armijnhemel/binaryanalysis-ng/blob/1c90fa5447651451d6a70006a26c44844102144a/src/bang/parsers/image/png/png.ksy#L129-L132), because I can see that this is already there ([`png.ksy:129-132`](https://github.com/armijnhemel/binaryanalysis-ng/blob/1c90fa5447651451d6a70006a26c44844102144a/src/bang/parsers/image/png/png.ksy#L129-L132)):

```ksy
      - id: bit_depth
        type: u1
        valid:
          any-of: [1, 2, 4, 8, 16]
```

PR #553, opened 2021-11-03
add two variants of the terminfo format: regular and extended storage format

armijnhemel:
This PR adds two variants of the terminfo format: regular and the extended storage format. It is impossible to see at the start of the file whether or not a file uses the regular format or the extended storage format so when parsing it is recommended to first try the extended storage format and if that fails to use the regular format.

Please note: this PR does *NOT* cover the extended number format, which uses a different magic and 32 bit integers for storing some data instead of 16 bit integers. It should not be too difficult to support this.

### DIFF
--- a/system/terminfo.ksy
+++ b/system/terminfo.ksy
@@ -1,0 +1,80 @@
+meta:
+  id: terminfo
+  title: terminfo
+  license: CC0-1.0
+  encoding: ASCII
+  endian: le
+doc: |
+  terminfo files, format described in the Linux man page for terminfo files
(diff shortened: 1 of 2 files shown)


KOLANICH:
Could these 2 specs be merged into one and the variant be controlled by a param?

armijnhemel:
> Could these 2 specs be merged into one and the variant be controlled by a param?

I think that will be a bit hard (but happily proven incorrect): you cannot see beforehand which of the two formats to choose.

PR #555, opened 2021-11-05
add specification for lz4 (single frame)

armijnhemel:
This specification adds support for parsing a single LZ4 frame, not multi frame LZ4 streams.

### DIFF
--- a/archive/lz4.ksy
+++ b/archive/lz4.ksy
@@ -1,0 +1,93 @@
+meta:
+  id: lz4
+  title: LZ4
+  license: CC0-1.0
+  endian: le
+  encoding: ASCII
+doc-ref: https://github.com/lz4/lz4/blob/dev/doc/lz4_Frame_format.md
+doc: |
+  This specification describes a single LZ4 frame.
+
+  The original LZ4 frame format specification has the following notice:
+
+  Copyright (c) 2013-2020 Yann Collet
+
+  Permission is granted to copy and distribute this document for any purpose and
+  without charge, including translations into other languages and incorporation
+  into compilations, provided that the copyright notice and this notice are preserved,
+  and that any substantive changes or deletions from the original are clearly
+  marked. Distribution of this document is unlimited.
+seq:
+  - id: magic
+    type: u4
+    valid: 0x184d2204
+  - id: frame_descriptor
+    type: frame_descriptor
+  - id: blocks
+    type: block
+    repeat: until
+    repeat-until: _.is_endmark
+  - id: content_checksum
+    type: u4
+    if: frame_descriptor.flag.content_checksum
+types:
+  frame_descriptor:
+    seq:
+      - id: flag
+        type: flag
+      - id: bd
+        type: bd
+      - id: content_size
+        type: u8
+        if: flag.content_size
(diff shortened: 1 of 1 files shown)


archive/lz4.ksy
```diff
@@ -0,0 +1,93 @@
...
+      - id: content_size
+        type: u8
+        if: flag.content_size
+      - id: dictionary_id
+        type: u4
+        if: flag.dictionary_id
+      - id: header_checksum
+        type: u1
+    types:
+      bd:
```
KOLANICH, line 49: Could you, please, make the names more meaningful and add `doc`s? I mean I am not familiar with lz4 internals, but I think KS specs should be adequate replacements of the original specs.

armijnhemel: Which names do you mean specifically? Some of these are not very well explained in the original docs either.

KOLANICH: `bd`. OK then.

archive/lz4.ksy
```diff
@@ -0,0 +1,93 @@
...
+    seq:
+      - id: len_block
+        type: b31
+      - id: uncompressed
+        type: b1
+      - id: data
+        size: len_block
+      - id: checksum
+        type: u4
+        if: _root.frame_descriptor.flag.block_checksum and not is_endmark
```
KOLANICH, line 90: Feel free to reuse

```yaml
seq:
  - id: checksum_calculation
    type: checksum(_index)
    repeat: eos
instances:
  calculated_checksum:
    value: checksum_calculation[checksum_calculation.size - 1].checksum
types:
  checksum:
    params:
      - id: idx
        type: u1
    seq:
      - id: ch
        type: u1
    instances:
      is_first:
        value: idx == 0
      checksum:
        value: "(((is_first ? _parent._io.size : _parent.checksum_calculation[idx - 1].checksum) + ch) & 0xFF).as<u1>"
```

archive/lz4.ksy
armijnhemel: The original docs are not clear about what `bd` actually means: https://github.com/lz4/lz4/blob/dev/doc/lz4_Frame_format.md#frame-descriptor

so I have no idea what it stands for. In the source code it is just referenced as "BD byte".

archive/lz4.ksy
armijnhemel: > Feel free to reuse
> 
> ```yaml
> seq:
>   - id: checksum_calculation
>     type: checksum(_index)
>     repeat: eos
> instances:
>   calculated_checksum:
>     value: checksum_calculation[checksum_calculation.size - 1].checksum
> types:
>   checksum:
>     params:
>       - id: idx
>         type: u1
>     seq:
>       - id: ch
>         type: u1
>     instances:
>       is_first:
>         value: idx == 0
>       checksum:
>         value: "(((is_first ? _parent._io.size : _parent.checksum_calculation[idx - 1].checksum) + ch) & 0xFF).as<u1>"
> ```

Is that an implementation of the xxhash32 checksum?

KOLANICH: No, just a sum modulo 256 from `eink_wbf`

Issue #458, opened 2021-03-27
android_img: doesn't work with version 3


armijnhemel:
Version 3 of the Android boot image header is a significantly simplified version of the boot header:

https://source.android.com/devices/bootloader/boot-image-header

The current specification does not cover this. What would be the best way to add support for it? Create a separate header for v3 and non-v3 and use a switch?

armijnhemel:
There are also other versions where header_version doesn't exist:

https://github.com/LineageOS/android_hardware_samsung/blob/lineage-15.1/dtbhtool/bootimg.h

PR #558, opened 2021-11-10
android_img: support v3 and v4

armijnhemel:
This PR adds support for the v3 and v4 format of the file which are significantly different from v0, v1 and v2. Test files for v3 and v4 can be found in:

* Pixel 5 (firmware: redfin-rd1a.200810.020-factory-c3ea1715.zip ) - v3
* Pixel 6 Pro (firmware: raven-sd1a.210817.015.a4-factory-bd6cb030.zip ) - v4

Firmware files can be grabbed from: https://developers.google.com/android/images

### DIFF
--- a/archive/android_img.ksy
+++ b/archive/android_img.ksy
@@ -7,87 +7,109 @@ meta:
     - android
   license: CC0-1.0
   endian: le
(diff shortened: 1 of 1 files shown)


armijnhemel:
This one should fix #458

PR #560, opened 2021-11-11
java_class: add sanity check for the version number

armijnhemel:
Mach-O files have the same magic bytes. This PR adds a sanity check to avoid false positives.

### DIFF
--- a/executable/java_class.ksy
+++ b/executable/java_class.ksy
@@ -16,6 +16,7 @@ seq:
   - id: version_major
     type: u2
     valid:
+      # https://github.com/file/file/blob/905ca555/magic/Magdir/cafebabe#L11-L12
       min: 43
   - id: constant_pool_count
     type: u2
--- a/executable/java_class.ksy
+++ b/executable/java_class.ksy
@@ -15,6 +15,8 @@ seq:
     type: u2
   - id: version_major
     type: u2
+    valid:
+      min: 43
   - id: constant_pool_count
     type: u2
   - id: constant_pool


executable/java_class.ksy
```diff
@@ -15,6 +15,8 @@ seq:
     type: u2
   - id: version_major
     type: u2
+    valid:
+      min: 43
```
generalmimon, line 19: Any reference/rationale for this particular value?

dgelessus: fwiw, 43 (or 0x2d) is also the lowest class file version recognized by the `file` tool. [The source code](https://github.com/file/file/blob/master/magic/Magdir/cafebabe) says that "the first released Java class format was version 43.0" - not sure where that information comes from. The oldest version that I can find properly documented is version 45 for Java 1.0.2 - see [the JVM 6 spec](https://docs.oracle.com/javase/specs/jvms/se6/html/ClassFile.doc.html#75883).

In any case, the exact value doesn't matter. This comment in the `file` patterns explains it quite well:

> Since Java bytecode and Mach-O universal binaries have the same magic number,  the test must be performed in the same "magic" sequence to get both right. The long at offset 4 in a Mach-O universal binary tells the number of  architectures; the short at offset 4 in a Java bytecode file is the JVM minor  version and the short at offset 6 is the JVM major version. Since there are only only 18 labeled Mach-O architectures at current, and the first released Java class format was version 43.0, we can safely choose any number between 18 and 39 to test the number of architectures against (and use as a hack). Let's not use 18, because the Mach-O people might add another one or two as time goes by...

To be precise, there are 18 "major" architectures - some of them also have sub-architectures (e. g. different PowerPC models or ARM instruction set versions), so the actual total number of architectures is much higher than 18. But in practice, most universal binaries only have 2-4 architectures, e.g. x86_64/arm64 or ppc/ppc64/i386/x86_64, and sub-architectures aren't used much. You'll probably never find a Mach-O binary with more than 10 architectures.

armijnhemel: > Any reference/rationale for this particular value?

I got it from `/usr/share/magic` on my system. This is basically the same information @dgelessus quoted.

generalmimon: @dgelessus @armijnhemel Thanks for clarifying.

generalmimon:
@armijnhemel Thanks!

PR #563, opened 2021-11-13
Resolve name collision

kahlkevin:
- Rename enum certificate_entry.certificate_type to certificate_entry.certificate_type_enum to avoid name collision with sequence field certificat_entry.certificate_type
 - Resolve compilation error in parser generated for csharp target


GreyCat:
In theory, it's better to implement such a feature in KS-to-C# compiler, but probably that won't happen soon, and it's better to have a working module rather than an erroneous one, so the workaround you've proposed seems to hit that :)

Thanks for this contribution!

PR #573, opened 2021-12-16
Add BMFont binary format

vladimirgamalyan:
This PR adds [BMFont binary](https://www.angelcode.com/products/bmfont/doc/file_format.html#bin) format. [This file](https://github.com/kaitai-io/kaitai_struct_formats/files/7727527/test.zip) can be used for testing (zipped, due github upload restriction).

### DIFF
--- a/font/bmfont.ksy
+++ b/font/bmfont.ksy
@@ -1,0 +1,155 @@
+meta:
+  id: bmfont
+  title: BMFont binary file
+  file-extension: fnt
+  license: CC0-1.0
+  endian: le
+  bit-endian: le
+doc-ref: https://www.angelcode.com/products/bmfont/doc/file_format.html  
+seq:
+  - id: magic
+    contents: ['BMF', 3]
+  - id: info
+    type: info
+  - id: common
+    type: common 
+  - id: pages
+    type: pages
+  - id: chars
+    type: chars
+  - id: kernings
+    type: kernings
+    if: not _io.eof
+types:
+  info:
+    seq:
+      - id: magic
+        contents: [1]
+      - id: block_size
+        type: s4
+      - id: size
+        type: s2        
+      - id: smooth
+        type: b1
+      - id: unicode
+        type: b1
+      - id: italic
+        type: b1
+      - id: bold
+        type: b1
+      - id: reserved
+        type: b4        
+      - id: charset
+        type: u1
+      - id: stretch_h
+        type: u2
+      - id: aa
+        type: s1
+      - id: padding_up
+        type: u1
+      - id: padding_right
+        type: u1
+      - id: padding_down
+        type: u1
+      - id: padding_left
+        type: u1        
+      - id: spacing_horiz
+        type: u1
+      - id: spacing_vert
+        type: u1
+      - id: outline
+        type: u1        
+      - id: face
+        type: str
+        size: block_size - 14
+        terminator: 0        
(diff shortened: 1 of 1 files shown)


font/bmfont.ksy
```diff
@@ -0,0 +1,155 @@
...
+types:
+  info:
+    seq:
+      - id: magic
+        contents: [1]
+      - id: block_size
+        type: s4
+      - id: size
+        type: s2        
+      - id: smooth
```
KOLANICH, line 32: Should the flags be packed into a dedicated subtype?

font/bmfont.ksy
```diff
@@ -0,0 +1,155 @@
...
+    type: pages
+  - id: chars
+    type: chars
+  - id: kernings
+    type: kernings
+    if: not _io.eof
+types:
+  info:
+    seq:
+      - id: magic
```
KOLANICH, line 26: Looks like a TLV pattern.

font/bmfont.ksy
```diff
@@ -0,0 +1,155 @@
+meta:
+  id: bmfont
+  title: BMFont binary file
+  file-extension: fnt
```
KOLANICH, line 4: It'd be nice to fill in more info, like the apps using the format.

KOLANICH:
>This file can be used for testing (zipped, due github upload restriction).

BTW, we have https://github.com/kaitai-io/kaitai_struct_samples

font/bmfont.ksy
```diff
@@ -0,0 +1,155 @@
...
+  license: CC0-1.0
+  endian: le
+  bit-endian: le
```
GreyCat, line 7: Usage of `bit-endian` implies that this requires KS 0.9+. Please add `ks-version: 0.9`.

font/bmfont.ksy
```diff
@@ -0,0 +1,155 @@
...
+doc-ref: https://www.angelcode.com/products/bmfont/doc/file_format.html
```
GreyCat, line 8: Adding format specification without even 1-2 paragraphs of text of where it is being used, which applications use that, what era does it refer to, what information does it convey, etc, will be almost meaningless — others who might be looking into this format won't be able to find it with search engines.

Please add top-level `doc` element with at least 1-2 paragraphs of general info about this format.

font/bmfont.ksy
vladimirgamalyan: if I'm not mistaken this will require an intermediate id to access the fields instead of directly accessing, maybe I'm missing some pros?

font/bmfont.ksy
vladimirgamalyan: This is definitely a TLV pattern, but I couldn't figure out how to make first 4 blocks required and in the correct order, except to explicitly list the blocks.

font/bmfont.ksy
vladimirgamalyan: Added a bit

font/bmfont.ksy
vladimirgamalyan: Added a bit

font/bmfont.ksy
vladimirgamalyan: Done

font/bmfont.ksy
KOLANICH: Just insert them and use `valid`

PR #575, opened 2021-12-28
Add support for Chromium file formats

mblsha:
Adds support for Tab restore and Session file types, which Chromium writes using [command_storage_backend.cc](https://source.chromium.org/chromium/chromium/src/+/main:components/sessions/core/command_storage_backend.cc?q=command_storage_backend.cc&ss=chromium).

Adds support for [data_pack.cc](https://source.chromium.org/chromium/chromium/src/+/main:ui/base/resource/data_pack.cc) resource bundles.

### DIFF
--- a/apps/chromium_resource_bundle.ksy
+++ b/apps/chromium_resource_bundle.ksy
@@ -1,7 +1,7 @@
 meta:
   id: chromium_resource_bundle
-  title: Chromium 81 .pak File
-  application: Chromium 81
+  title: Chromium 96 .pak File
+  application: Chromium 96
   file-extension: pak
   endian: le
 
@@ -49,16 +49,9 @@ types:
       resource_size:
         value: 'i < _parent.header.resource_count - 1 ? _parent.entries[i+1].file_offset - _parent.entries[i].file_offset : _io.size - _parent.entries[i].file_offset'
       body:
-        # io: _root._io
         pos: file_offset
         size: resource_size
-        # type: resource
     -webide-representation: '{resource_id}: {file_offset}'
-
-  # resource_png:
-  #   seq:
-  #     - id: magic
-  #       contents: [137, 80, 78, 71, 13, 10, 26, 10]
 
   alias: # DataPack::Alias
     seq:
@@ -69,7 +62,7 @@ types:
     -webide-representation: '{entry_index} -> {resource_id}'
 
 enums:
-  text_encoding_type:
+  text_encoding_type: # ResourceHandle::TextEncodingType
     0: binary
     1: utf8
     2: utf16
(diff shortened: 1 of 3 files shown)


GreyCat:
Thanks for this contribution!

Several high-level observations:

* It looks like this PR actually includes 3 unrelated format specifications — i.e. none of them uses/imports the other. If that's the case, can I ask you to split these into 3 different PRs, so chances of at least some of these being reviewed and merged will be greater?
* It looks like `apps/chromium_resource_bundle.ksy` is duplicating https://formats.kaitai.io/chrome_pak/, which is already present in the repo and peer-reviewed. Can you clarify if your version is somehow different/better?
* We don't have "apps" top-level category, and, in certain sense, that's intentional. Directories are supposed to be used for categorization, and 95-99% of the format specs here could be categorized as having something to do with "an app". Let's think of some better categories for these? E.g. chromium_session.ksy and chromium_tab_restore.ksy might be categorized either as "log" of events which happened in the browser / in the tabs or "database" of such events (that is if the format is actually designed to support efficient inserts or querying). Alternatively, we might want to create top-level group like "chrome", "chromium" or "google" to store everything related to that topic? What do you think?

KOLANICH:
>Alternatively, we might want to create top-level group like "chrome", "chromium" or "google" to store everything related to that topic? What do you think?

I am not he ;) but I think it is an extremily bad idea. Well, we have some software-related dirs like `windows`, but it may make sense to move their contents to other dirs, i.e. `
regf` belongs to `database`, `windows_minidump` to `log`, and `
windows_lnk_file` - to `desktop`.

mblsha:
> It looks like this PR actually includes 3 unrelated format specifications — i.e. none of them uses/imports the other. If that's the case, can I ask you to split these into 3 different PRs, so chances of at least some of these being reviewed and merged will be greater?

Thanks, will do this!

> It looks like apps/chromium_resource_bundle.ksy is duplicating https://formats.kaitai.io/chrome_pak/, which is already present in the repo and peer-reviewed. Can you clarify if your version is somehow different/better?

My version is actually inferior, didn't notice that there's already a chrome_pak decoder (but I created mine last year).

> We don't have "apps" top-level category, and, in certain sense, that's intentional. Directories are supposed to be used for categorization, and 95-99% of the format specs here could be categorized as having something to do with "an app". Let's think of some better categories for these? E.g. chromium_session.ksy and chromium_tab_restore.ksy might be categorized either as "log" of events which happened in the browser / in the tabs or "database" of such events (that is if the format is actually designed to support efficient inserts or querying). Alternatively, we might want to create top-level group like "chrome", "chromium" or "google" to store everything related to that topic? What do you think?

I guess I'll file both remaining formats under `database`, since they're both used for restoring browser state.

PR #576, opened 2022-01-19
Added security/kerberos_pac.ksy - Microsoft Kerberos PAC

kvv81:
The Privilege Attribute Certificate (PAC) Data Structure is used by authentication protocols (protocols
that verify identities) to transport authorization information, which controls access to resources.

Signed-off-by: Volodymyr Khomenko <Khomenko.Volodymyr@gmail.com>

### DIFF
--- a/security/kerberos_pac.ksy
+++ b/security/kerberos_pac.ksy
@@ -1,0 +1,542 @@
+meta:
+  id: kerberos_pac
+  title: Microsoft Kerberos PAC
+  file-extension:
+    - pac
+    - bin
+  application: Active Directory KDC
+  license: MIT
+  endian: le
+doc: |
+  The Privilege Attribute Certificate (PAC) Data Structure is used by authentication protocols (protocols
+  that verify identities) to transport authorization information, which controls access to resources.
+  To fetch PAC file from KDC easily, you may use samba net utility:
+  sudo net ads kerberos pac save -U test_user filename=test_user.pac
+  doc-ref: "https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-pac/c38cc307-f3e6-4ed4-8c81-dc550d96223c"
+seq:
+  - id: pac
+    -orig-id: PACTYPE
+    type: pactype
+    doc: |
+      The PACTYPE structure is the topmost structure of the PAC and specifies the number of elements in
+      the PAC_INFO_BUFFER array. The PACTYPE structure serves as the header for the
+      complete PAC data.
+types:
+  pactype:
+    doc: "See [MS-PAC] 2.3 PACTYPE"
+    seq:
+      - id: c_buffers
+        -orig-id: cBuffers
+        type: u4
+        doc: A 32-bit unsigned integer in little-endian format that defines the number of entries in the buffers array
+      - id: version
+        -orig-id: Version
+        type: u4
+        if: c_buffers != logon_info_magic_number
+        doc: A 32-bit unsigned integer in little-endian format that defines the PAC version; MUST be 0x00000000
+      - id: buffers
+        -orig-id: Buffers
+        type: pac_info_buffer
+        if: c_buffers != logon_info_magic_number
+        repeat: expr
+        repeat-expr: c_buffers
+        doc: |
+          Following the PACTYPE structure is an array of PAC_INFO_BUFFER structures each of
+          which defines the type and byte offset to a buffer of the PAC
+    instances:
+      logon_info_magic_number:
+        value: 0x81001
+        doc: |
+          A magic number formed by pre-defined fields of common_type_header
+          for KERB_VALIDATION_INFO structure. Used to detect partial format of the file
+          See [MS-RPCE] 2.2.6.1 Common Type Header for the Serialization Stream
+      direct_logon_info:
+        pos: 0
+        type: kerb_validation_info
+        if: c_buffers == logon_info_magic_number
+        doc: "To auto-detect file content in case of reduced data (logon_info only) we use version+endianness fields as the magic"
+  pac_info_buffer:
+    doc: "See [MS-PAC] 2.4 PAC_INFO_BUFFER"
+    seq:
+      - id: ul_type
+        -orig-id: ulType
+        type: u4
+        enum: pac_type
+      - id: cb_buffer_size
+        -orig-id: cbBufferSize
+        type: u4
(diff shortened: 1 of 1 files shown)


security/kerberos_pac.ksy
```diff
@@ -0,0 +1,542 @@
...
+    doc: "See [MS-PAC] 2.7 PAC_CLIENT_INFO"
+    seq:
+      - id: client_id
+        -orig-id: ClientId
+        type: filetime
+        doc: A FILETIME structure in little-endian format that contains the Kerberos initial ticket-granting ticket (TGT) authentication time
+      - id: name_length
+        -orig-id: NameLength
+        type: u2
+        doc: An unsigned 16-bit integer in little-endian format that specifies the length, in bytes, of the Name field
```
generalmimon, line 475: 1. should be called `len_name` instead per [style guide](https://doc.kaitai.io/ksy_style_guide.html#attr-id);
2. such a `doc` is useless and actually counter-productive - only comment what isn't described by the spec.

   Omit any type information - that is already obvious from the declarative description (`/meta/endian: le`, `type: u2`):

   > ~~An unsigned 16-bit integer in little-endian format that~~ specifies the length, in bytes, of the Name field

   Avoid extra wording as _specifies the_ (...), or worse _this is a field that specifies_ (...) - :

   > ~~specifies the~~ length, in bytes, of the Name field

   Now we're left with the only potential valuable info:

   > length, in bytes, of the Name field

   However, see what's written in the [style guide](https://doc.kaitai.io/ksy_style_guide.html#seq-attr):
   
   > `doc` SHOULD NOT be included if it’s trivial (i.e. if it is a copy of `id`, and there is really nothing more to say about that attribute).

   The attribute will be called `len_name`, which already describes "_length, in bytes, of the Name field_" well enough, so we have nothing left to say about it and thus `doc` should be omitted.

Please apply these remarks to the entire file.

security/kerberos_pac.ksy
```diff
@@ -0,0 +1,542 @@
...
+      - id: upn
+        type: str
+        size: upn_length
+        encoding: UTF-16LE
+      - id: padding2
+        size: dns_domain_name_offset - upn_offset - upn_length
+      - id: dns_domain_name
+        type: str
+        size: dns_domain_name_length
+        encoding: UTF-16LE
```
generalmimon, line 512: I don't see a reason why not use `instances` - it's easier, less hacky and more descriptive:

```suggestion
    instances:
      upn:
        pos: upn_offset
        type: str
        size: upn_length
        encoding: UTF-16LE
      dns_domain_name:
        pos: dns_domain_name_offset
        type: str
        size: dns_domain_name_length
        encoding: UTF-16LE
```

security/kerberos_pac.ksy
```diff
@@ -0,0 +1,542 @@
...
+        type: u4
+      - id: user_flags
+        -orig-id: UserFlags
+        type: u4
+        doc: A 32-bit unsigned integer that contains a set of bit flags that describe the user's logon information
+      - id: user_session_key
+        -orig-id: UserSessionKey
+        type: u1
+        repeat: expr
+        repeat-expr: 16
```
generalmimon, line 376: This pattern

```ksy
  type: u1
  repeat: expr
  repeat-expr: {N}
```

should be almost always replaced with

```ksy
  size: {N}
```

It will be a byte array, no longer a list of `u1` integers, so it'll be probably easier to work with in the generated parser, too.

security/kerberos_pac.ksy
```diff
@@ -0,0 +1,542 @@
...
+        encoding: UTF-16LE
+  pac_signature_data:
+    doc: "See [MS-PAC] 2.8 PAC_SIGNATURE_DATA"
+    seq:
+      - id: signature_type
+        -orig-id: SignatureType
+        type: u4
+        enum: signature_type_enum
+        doc: A 32-bit unsigned integer value in little-endian format that defines the cryptographic system used to calculate the checksum
+      - id: signature_16
```
generalmimon, line 521: Replace `signature_16` + `signature_12` with a single `signature`:

```ksy
seq:
  # ...
  - id: signature
    size: len_signature
    if: len_signature != -1

instances:
  len_signature:
    value: |
      signature_type == signature_type_enum::kerb_checksum_hmac_md5 ? 16 :
      (
        signature_type == signature_type_enum::hmac_sha1_96_aes128 or
        signature_type == signature_type_enum::hmac_sha1_96_aes256
      ) ? 12 :
      -1
    doc: `-1` if the signature is not present
```

security/kerberos_pac.ksy
```diff
@@ -0,0 +1,542 @@
...
+        -orig-id: Signature
+        type: u1
+        repeat: expr
+        repeat-expr: 12
+        if: '(signature_type == signature_type_enum::hmac_sha1_96_aes128) or (signature_type == signature_type_enum::hmac_sha1_96_aes256)'
+      - id: ro_dc_identifier
+        -orig-id: RODCIdentifier
+        type: u2
+        repeat: eos
+        doc: A 16-bit unsigned integer value in little-endian format that contains the first 16 bits of the key version number ([MS-KILE] section 3.1.5.8) when the KDC is an RODC
```
generalmimon, line 537: According to `doc`, didn't you mean this instead?

```suggestion
      - id: ro_dc_identifier
        -orig-id: RODCIdentifier
        type: u2
        if: not _io.eof
        doc: A 16-bit unsigned integer value in little-endian format that contains the first 16 bits of the key version number ([MS-KILE] section 3.1.5.8) when the KDC is an RODC
```

It should not be an array, if I understand correctly.

generalmimon (changes requested):
In general, please try to follow these naming conventions from the [style guide](https://doc.kaitai.io/ksy_style_guide.html#attr-id):

> * For fields that designate *number / count* of something (in particular, number of repetitions of some other structure), use `num_` prefix and plural form — i.e. `num_questions`, `num_blocks`, `num_nodes`
> * For fields that designate *offset* to some particular data structure, use `ofs_` prefix and name of that data structure (as it would appear in the file) — i.e. `ofs_block`, `ofs_queries`, `ofs_path`
> * For fields that designate *size* of some particular data structure (in bytes or some other fixed units), use `len_` prefix and name of that data structure — i.e. `len_block` (length of a single `block` entry), `len_blocks` (total length of whole `blocks` array, made of `block` entries).

PR #578, opened 2022-01-25
add .ksy for Adobe Flash video

armijnhemel:
A fairly simplistic .ksy for Adobe flash video. This only checks the container structure, but doesn't parse the invididual video audio and script data elements.

### DIFF
--- a/media/flv.ksy
+++ b/media/flv.ksy
@@ -1,0 +1,63 @@
+meta:
+  id: flv
+  title: Adobe Flash Video File Format
+  license: CC0-1.0
+  ks-version: 0.9
+  encoding: utf-8
+  endian: be
+doc-ref: https://wwwimages2.adobe.com/content/dam/acom/en/devnet/flv/video_file_format_spec_v10_1.pdf
+seq:
+  - id: preheader
+    type: preheader
+  - id: rest_of_header
+    size: preheader.len_header - preheader._sizeof
+  - id: previous_tag_size0
+    type: u4
+    valid: 0
+  - id: tags
+    type: tag_and_size
+    repeat: eos
+types:
+  preheader:
+    seq:
+      - id: magic
+        contents: 'FLV'
+      - id: version
+        type: u1
+        valid: 1
+      - id: flags
+        type: u1
+      - id: len_header
+        type: u4
+  tag_and_size:
+    seq:
+      - id: tag
+        type: tag
+      - id: previous_tag_size
+        type: u4
+        valid: tag.len_data + 11
+  tag:
+    seq:
+      - id: reserved
+        type: b2
+      - id: filter
+        type: b1
+      - id: type
+        type: b5
+        enum: data_type
+      - id: len_data
+        type: b24
+      - id: timestamp
+        type: b24
+      - id: timestamp_extended
+        type: u1
+      - id: stream_id
+        type: b24
+        valid: 0
+      - id: data
+        size: len_data
+enums:
+  data_type:
+    8: audio
+    9: video
+    18: script


media/flv.ksy
```diff
@@ -0,0 +1,63 @@
...
+        type: u1
+      - id: len_header
+        type: u4
+  tag_and_size:
+    seq:
+      - id: tag
+        type: tag
+      - id: previous_tag_size
+        type: u4
+        valid: tag.len_data + 11
```
KOLANICH, line 38: Make a subtype and use `sizeof`

media/flv.ksy
```diff
@@ -0,0 +1,63 @@
...
+types:
+  preheader:
+    seq:
+      - id: magic
+        contents: 'FLV'
+      - id: version
+        type: u1
+        valid: 1
+      - id: flags
+        type: u1
```
KOLANICH, line 29: May worth a subtype of bit fields.

media/flv.ksy
```diff
@@ -0,0 +1,63 @@
...
+      - id: reserved
+        type: b2
+      - id: filter
+        type: b1
+      - id: type
+        type: b5
+        enum: data_type
+      - id: len_data
+        type: b24
+      - id: timestamp
```
KOLANICH, line 50: May need some words about conversion of them into objects.

media/flv.ksy
armijnhemel: Why? The specification says that this is always + 11. It has never been anything else.

media/flv.ksy
```diff
@@ -0,0 +1,63 @@
+meta:
+  id: flv
+  title: Adobe Flash Video File Format
+  license: CC0-1.0
+  ks-version: 0.9
+  encoding: utf-8
+  endian: be
+doc-ref: https://wwwimages2.adobe.com/content/dam/acom/en/devnet/flv/video_file_format_spec_v10_1.pdf
```
KOLANICH, line 8: `doc` and samples are missing.

media/flv.ksy
armijnhemel: No. This is just a timestamp according to the specifications.

```
Time in milliseconds at which the data in this tag applies.
```

media/flv.ksy
```diff
@@ -0,0 +1,63 @@
+meta:
+  id: flv
+  title: Adobe Flash Video File Format
+  license: CC0-1.0
+  ks-version: 0.9
+  encoding: utf-8
+  endian: be
```
KOLANICH, line 7:
```yaml
xref:
  wikidata: Q1135858
```

media/flv.ksy
armijnhemel: I will add this.

media/flv.ksy
KOLANICH: What is the epoch? What is `timestamp_extended`?

media/flv.ksy
KOLANICH: Because contents of `tag` except of `data` is exactly 11 bytes, and `11 + len_data` is the len of `tag`. Also it may make sense to use `_io.pos`

armijnhemel: when reading the specifications I think that this is clearer (as it follows the specifications closer). Yes, it is hardcoded, but I don't see it as a problem.

media/flv.ksy
armijnhemel: The epoch is the start of the video. See the documentation, appendix E.

KOLANICH: The one of the goals of having a KS specification is that I won't have to read the original one.

PR #580, opened 2022-02-06
Ttf updates

armijnhemel:
This commit adds/fixes:

* support for `dsig` table
* support for `OS/2` table version 0, 2, 3, 4, 5 (previously only version 1 was supported, which led to parsing errors for version 0 tables)
* rework `cmap` table: there are other formats of TTF files that were not supported, leading to parse errors. Support for some formats was (partially) added

### DIFF
--- a/font/ttf.ksy
+++ b/font/ttf.ksy
@@ -318,7 +318,19 @@ types:
           signature:
             pos: ofs_signature
             size: len_signature
-            io: _root._io
+            type: signature_block
+      signature_block:
+        seq:
+          - id: reserved1
+            type: u2
+            valid: 0
+          - id: reserved2
+            type: u2
+            valid: 0
+          - id: len_signature
+            type: u4
+          - id: signature
+            size: len_signature
   glyf:
     # https://github.com/fonttools/fonttools/blob/678876325ef26ac33e8c6d13f4fb70c3bef5da8e/Lib/fontTools/ttLib/tables/_g_l_y_f.py
     # TODO: sadly, Kaitai currently cannot parse this structure
--- a/font/ttf.ksy
+++ b/font/ttf.ksy
@@ -11,7 +11,9 @@ meta:
 doc: |
   A TrueType font file contains data, in table format, that comprises
   an outline font.
-doc-ref: https://www.microsoft.com/typography/tt/ttf_spec/ttch02.doc
+doc-ref:
+  - https://www.microsoft.com/typography/tt/ttf_spec/ttch02.doc
+  - https://docs.microsoft.com/en-us/typography/opentype/spec/
 seq:
   - id: offset_table
     type: offset_table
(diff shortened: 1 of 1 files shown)


armijnhemel:
Lots of test files: https://github.com/google/fonts.git

font/ttf.ksy
```diff
@@ -11,7 +11,9 @@ meta:
 doc: |
   A TrueType font file contains data, in table format, that comprises
   an outline font.
-doc-ref: https://www.microsoft.com/typography/tt/ttf_spec/ttch02.doc
+doc-ref:
+  - https://www.microsoft.com/typography/tt/ttf_spec/ttch02.doc
```
generalmimon, line 15: This link no longer works (it is redirected to https://docs.microsoft.com/typography), so it would nice to delete or replace it with something else (when you're already updating the links).

The original `ttch02.doc` is still available via the Wayback Machine (https://web.archive.org/web/20160410081432/https://www.microsoft.com/typography/tt/ttf_spec/ttch02.doc), but the question is whether this resource brings any additional value.

font/ttf.ksy
```diff
@@ -750,7 +915,29 @@ types:
         doc: 'The descender metric for Windows.'
       - id: code_page_range
         type: code_page_range
+        if: version > 0
         doc: 'This field is used to specify the code pages encompassed by the font file in the `cmap` subtable for platform 3, encoding ID 1 (Microsoft platform).'
+      - id: x_height
+        type: u2
+        if: version > 1
```
generalmimon, line 922: Wouldn't it make more sense to use the "greater than _or equal to_" (e.g. <code>if: version ><ins><b>=</b></ins> 5</code>) operator rather than the non-inclusive "greater than" (<code>if: version > 4</code>)?

I think that format docs usually introduce some feature _since_ some format version (**including** the version when it was introduced), which suggests the `>=` operator.

armijnhemel: This makes sense, I will change it.

generalmimon: Especially `if: version > 0` looks weird to me - that's equivalent to <code>if: version <ins><b>!=</b></ins> 0</code> (because `version` is unsigned), which suggests that the documentation literally says that the `version == 0` is a special case where some structure is **not present**, otherwise it is (in any other version).

font/ttf.ksy
armijnhemel: The extra `doc-ref` I added has a more complete description of the current format, which might, or might not, be pure TrueType. Most "TrueType" fonts these days are actually OpenType fonts (but converted to TrueType format, with TrueType magic, and so on). It has become a bit of a mix it seems.

armijnhemel: I also added a reference to Apple's TrueType documentation. Many of the fonts that I see that are TrueType are a combination of what I see in Apple's documentation and Microsoft's documentation.

Because TrueType and OpenType are nearly identical (except the first four bytes) I think that it might be worth making this a more generic font description.

PR #581, opened 2022-02-11
Add EDF parser

burakcan:
Spec: https://www.edfplus.info/

### DIFF
--- a/scientific/edf.ksy
+++ b/scientific/edf.ksy
@@ -1,0 +1,182 @@
+meta:
+  id: edf
+  file-extension: edf
+  endian: le
+  license: MIT
+
+seq:
+  - id: header
+    type: header
+
+  - id: body
+    type: body
+
+types:
+  header:
+    instances:
+      ns:
+        value: num_signals.to_i
+      ndr:
+        value: num_data_records.to_i
+
+    seq:
+      - id: "version"
+        size: 8
+        type: str
+        pad-right: 0x20
+        encoding: ASCII
+
+      - id: "patient_id"
+        size: 80
+        type: str
+        encoding: ASCII
+
+      - id: "recording_id"
+        size: 80
+        type: str
+        encoding: ASCII
+
+      - id: "start_date"
+        size: 8
+        type: str
+        encoding: ASCII
+
+      - id: "start_time"
+        size: 8
+        type: str
+        encoding: ASCII
+
+      - id: "header_num_bytes"
+        size: 8
+        type: str
+        encoding: ASCII
+        pad-right: 0x20
+
+      - id: "reserved"
+        size: 44
+        type: str
+        encoding: ASCII
+
+      - id: "num_data_records"
+        size: 8
+        type: str
+        pad-right: 0x20
+        encoding: ASCII
+
+      - id: "data_duration"
+        size: 8
+        type: str
+        pad-right: 0x20
+        encoding: ASCII
+
+      - id: "num_signals"
+        size: 4
+        type: str
+        pad-right: 0x20
+        encoding: ASCII
+
+      - id: "labels"
+        size: 16
+        repeat: expr
+        repeat-expr: ns
+        type: str
+        pad-right: 0x20
+        encoding: ASCII
+
+      - id: "transducers"
+        size: 80
+        repeat: expr
+        repeat-expr: ns
+        type: str
+        pad-right: 0x20
+        encoding: ASCII
+
+      - id: "phys_dimensions"
+        size: 8
+        repeat: expr
+        repeat-expr: ns
+        type: str
+        pad-right: 0x20
+        encoding: ASCII
+
+      - id: "phys_mins"
+        size: 8
+        repeat: expr
+        repeat-expr: ns
+        type: str
+        pad-right: 0x20
+        encoding: ASCII
+
+      - id: "phys_maxes"
+        size: 8
+        repeat: expr
+        repeat-expr: ns
+        type: str
+        pad-right: 0x20
+        encoding: ASCII
+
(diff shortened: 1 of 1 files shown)


scientific/edf.ksy
```diff
@@ -0,0 +1,182 @@
+meta:
+  id: edf
+  file-extension: edf
+  endian: le
+  license: MIT
+
```
KOLANICH, line 6: `doc` and `doc-ref` are mandatory. Also a link to some small samples is desireable.

scientific/edf.ksy
```diff
@@ -0,0 +1,182 @@
...
+types:
+  header:
+    instances:
+      ns:
+        value: num_signals.to_i
+      ndr:
+        value: num_data_records.to_i
+
+    seq:
+      - id: "version"
```
KOLANICH, line 23: No need to quote `id`s here

scientific/edf.ksy
```diff
@@ -0,0 +1,182 @@
...
+      data_recording:
+        seq:
+          - id: "data"
+            type: u2
+            repeat: expr
+            repeat-expr: _root.header.num_samples[_parent.signal_index].to_i
+
+      signal:
+        params:
+          - id: "signal_index"
```
KOLANICH, line 169: 1. Why not just `idx`?
2. Should it really be signed?

scientific/edf.ksy
```diff
@@ -0,0 +1,182 @@
+meta:
+  id: edf
```
KOLANICH, line 2: The id is too short, conflicts are possible. Please mind a more specific `id`.

scientific/edf.ksy
```diff
@@ -0,0 +1,182 @@
...
+  file-extension: edf
+  endian: le
```
KOLANICH, line 4: Here is a template for a metadata block you may find useful:

```yaml
meta:
  id: 
  title: 
  application: 
  file-extension: # todo: https://extension.nirsoft.net/
  xref:
    ansi: 
    din: 
    forensicswiki: # https://forensicswiki.xyz/
    fileformat: # https://www.fileformat.info
    incits: 
    iso: 
    justsolve: # http://fileformats.archiveteam.org/wiki/
    loc: # https://www.loc.gov/preservation/digital/formats/
    mime: 
    pronom: # https://www.nationalarchives.gov.uk/pronom/fmt/
     - x-fmt/
    rfc: 
    uti: 
    ffdg: # https://docs.fileformat.com
    wikidata: Q

doc: | # MANDATORY
  

doc-ref: # MANDATORY.
  - 
  - 
  - 
```

KOLANICH (changes requested):
Thanks for the contribution!

scientific/edf.ksy
KOLANICH: Thanks for the info on how it is going on in JVM-based langs.

generalmimon: > It is semantically incorrect to have negative indexes in an array.

Yes, but unsigned types are not a good solution to this.

KOLANICH: Unsigned types are only good solution for it and only they must be allowed there, any signed type used as an array index must be a syntax error. In the case of sunbtraction - if a compiler can prove that the result is non-negative, it should be cast to unsigned automatically. If the compiler cannot - it must be an error and the programmer must code it so that the compiler can prove it.

It is JVM and Java that are flawed, not unsigned numbers. Wasting 1 bit for a sign that must always be 0 used is waste.

scientific/edf.ksy
```diff
@@ -0,0 +1,182 @@
...
+        value: num_signals.to_i
+      ndr:
+        value: num_data_records.to_i
+
+    seq:
+      - id: "version"
+        size: 8
+        type: str
+        pad-right: 0x20
+        encoding: ASCII
```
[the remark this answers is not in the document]
dgelessus: > Also should be lowercased.

Why? [Searching through the repo](https://github.com/kaitai-io/kaitai_struct_formats/search?q=ascii), both uppercase and lowercase are common (and uppercase is a bit more frequent).

scientific/edf.ksy
```diff
@@ -0,0 +1,182 @@
...
+          - id: "recordings"
+            repeat: expr
+            repeat-expr: _root.header.ndr
+            type: data_recording
+
+    seq:
+      - id: "signals"
+        repeat: expr
+        repeat-expr: _root.header.ns
+        type: signal(_index)
```
generalmimon, line 182: `seq` must come before `types` in the type specification - see https://doc.kaitai.io/ksy_style_guide.html#type:

> ## 2. Order of sections in a type spec
> 
> Use the following order of sections:
> 
> - `meta`, if present, MUST go first
> - `doc`
> - `doc-ref`
> - <ins>**`seq`**</ins>
> - `instances`, <ins>**`types`**</ins>, `enums` — use one’s best judgement to order these 3 to maximize readability

This is what searching for a name of the type in which a `seq` structure is located looks like in practice if this rule is not followed ([`font/ttf.ksy:338-650`](https://github.com/kaitai-io/kaitai_struct_formats/blob/ff19ec06d5fcba3ea382ed7acdcc462b866ed154/font/ttf.ksy#L338-L650)):

ksf-ttf-type-spec-bad-order

scientific/edf.ksy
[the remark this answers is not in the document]
KOLANICH: https://github.com/kaitai-io/kaitai_struct/issues/391#issuecomment-376003487

burakcan:
Thanks for the comments. I will update the pr when I have time.

PR #584, opened 2022-02-16
[File System] Secure Transacted File System

halotroop2288:
STFS is the filesystem used by content packaged on the Microsoft Xbox 360.
STFS filesystems are stored in files, and sometimes are found inside Xbox Content Packages ([XCP](https://free60.org/XCP/ "Free60")) or Game Profile Data packages ([GPD](https://free60.org/GPD/ "Free60")).

Here are a few cross-references for this format:
- https://free60.org/STFS/
- https://github.com/xenia-project/xenia/tree/master/src/xenia/vfs/devices
- https://github.com/hetelek/Velocity/tree/master/XboxInternals/Stfs

Here is a program that can generate these files for samples:
- [Velocity 0.1.0.0](http://www.mediafire.com/?mbr944kv9dp8c "Hetelek's Mediafire")
- [Velocity-XEX+ISO](https://github.com/Gualdimar/Velocity/releases/download/xex%2Biso-branch/Velocity-XEXISO.rar "GitHub Releases")

I'm looking for some pointers on completing this format. So far I have only completed the header and metadata, but the metadata appears to be much larger than I'm expecting (`metadata._io.size > header_size`).
`metadata_size` should be equal to the size of the metadata section (after metadata_size and before the first hash table). (Free60 erroneously calls this value "header_size" but it refers to the size of data after it, not before)


filesystem/stfs.ksy
```diff
@@ -0,0 +1,537 @@
...
+        encoding: 'UTF-8'
+        size: 0x80
+        repeat: expr
+        repeat-expr: 6
+        if: 'metadata_version == metadata_version::two'
+      - id: title_thumbnail_image
+        size: title_thumbnail_size
+      - id: descriptions_ex
+        type: str
+        encoding: 'UTF-8'
```
KOLANICH, line 353: 6 utf-8 vs 3 ascii. I guess utf-8 should be put into `meta`.

filesystem/stfs.ksy
```diff
@@ -0,0 +1,537 @@
...
+  title: 'Secure Transacted Filesystem'
+  application: 'Microsoft Xbox 360'
+  license: 'CC0-1.0'
+  endian: be
+seq:
+  - id: header
+    type: x_content_header
+  - id: licenses
+    type: x_content_license
+    size: 0x10
```
KOLANICH, line 12: Not needed.

PR #585, opened 2022-02-17
add (vastly incomplete) specification of XARA

armijnhemel:
This is a very incomplete specification of the XARA graphics format: https://en.wikipedia.org/wiki/Xara_Flare

Although the format itself is very easy to parse parts of the file can be zlib compressed, which makes things complicated. Unfortunately it isn't known in advance how long the compressed data is (you basically need to scan until you find an "end of compression" tag).

There aren't many test files out there (I have one, but not sure I can share it).

### DIFF
--- a/image/xara.ksy
+++ b/image/xara.ksy
@@ -1,0 +1,387 @@
+meta:
+  id: xara
+  title: XARA
+  license: CC0-1.0
+  ks-version: 0.9
+  encoding: ascii
+  endian: le
+doc-ref:
+ - https://en.wikipedia.org/wiki/Xar_(graphics)
+ - http://site.xara.com/support/docs/webformat/spec/XARFormatDocument.pdf
+seq:
+  - id: header
+    type: header
+  - id: records
+    type: record
+    repeat: until
+    repeat-until: _.tag == tags::end_of_file or _.tag == tags::start_compression
+types:
+  header:
+    seq:
+      - id: magic
+        contents: "XARA"
+      - id: magic2
+        contents: [0xa3, 0xa3, 0x0d, 0x0a]
+  record:
+    seq:
+      - id: tag
+        type: u4
+        enum: tags
+      - id: len_data
+        type: u4
+      - id: data
+        size: len_data
+        type:
+          switch-on: tag
+          cases:
+            tags::file_header: file_header
+            tags::document_nudge: document_nudge
+            tags::document_bitmap_smoothing: document_bitmap_smoothing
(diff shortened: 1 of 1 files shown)


image/xara.ksy
```diff
@@ -0,0 +1,387 @@
...
+        type: strz
+      - id: producer_build
+        type: strz
+  document_nudge:
+    seq:
+      - id: millipoint
+        type: u4
+  document_bitmap_smoothing:
+    seq:
+      - id: flags
```
KOLANICH, line 64: I guess it should be a type of bit fields.

image/xara.ksy
```diff
@@ -0,0 +1,387 @@
...
+      enable_bitmap_smoothing:
+        value: flags & 0x1
+  duplication_offset:
+    seq:
+      - id: coords
+        size: 8
+  start_compression:
+    seq:
+      - id: version
+        size: 3
```
KOLANICH, line 78: Is it a semver with components encoded into separate bytes?

image/xara.ksy
```diff
@@ -0,0 +1,387 @@
+meta:
+  id: xara
+  title: XARA
+  license: CC0-1.0
+  ks-version: 0.9
+  encoding: ascii
+  endian: le
+doc-ref:
+ - https://en.wikipedia.org/wiki/Xar_(graphics)
```
KOLANICH, line 9: Should go into `meta:xref:wikidata` instead.

image/xara.ksy
```diff
@@ -0,0 +1,387 @@
...
+  file_header:
+    seq:
+      - id: filetype
+        size: 3
+      - id: len_file
+        type: u4
+      - id: weblink
+        type: u4
+      - id: refinement_flags
+        type: u4
```
KOLANICH, line 51: Probably should be a type of bit flags.

armijnhemel: Nah:

```
This 32-bit word is held in the File Header record and at the time of writing 
it is defined to always be 0.
```

image/xara.ksy
```diff
@@ -0,0 +1,387 @@
+meta:
+  id: xara
+  title: XARA
+  license: CC0-1.0
+  ks-version: 0.9
+  encoding: ascii
+  endian: le
+doc-ref:
```
generalmimon, line 8:
```console
$ ksylint image/xara.ksy
image/xara.ksy
9:2: wrong indentation: expected 2 but found 1 (indentation)
```

I couldn't have said it better :)

PR #587, opened 2022-02-19
add .ksy for the Git index format

armijnhemel:
This PR adds support for the Git index format (note: not Git pack index, that's a different format).

A few notes:

1. the length of the hash cannot be determined without reading and parsing data using some sort of look ahead. The most used hash right now is SHA1, so the length is set to 20. I am expecting that SHA256 (length 32) might become more common in the future.
2. there are two elements called `object_name`. These are hashes. In the Git documentation they are referred to as "objects":

```
  In a repository using the traditional SHA-1, checksums and object IDs
  (object names) mentioned below are all computed using SHA-1.
```

I am not attached to this name, so if it should be changed to `object_id` I am totally fine with that.

### DIFF
--- a/database/git_index.ksy
+++ b/database/git_index.ksy
@@ -1,0 +1,150 @@
+meta:
+  id: git_index
+  title: Git index
+  license: CC0-1.0
+  ks-version: 0.9
+  endian: be
+  encoding: UTF-8
+doc-ref: https://github.com/git/git/blob/main/Documentation/technical/index-format.txt
+seq:
+  - id: header
+    type: header
+  - id: entries
+    type: entry
+    repeat: expr
+    repeat-expr: header.num_entries
+  - id: extensions
+    type: extension
+    repeat: until
+    repeat-until: _io.pos >= _io.size - len_hash
+  - id: checksum
+    size: len_hash
+instances:
+  len_hash:
+    value: 20   # SHA1, change to 32 for SHA256
+types:
+  header:
+    seq:
+      - id: magic
+        contents: "DIRC"
+      - id: version
+        type: u4
+        valid:
+          any-of: [2]    # only support version 2 now
+          #any-of: [2, 3, 4]
+      - id: num_entries
+        type: u4
+  entry:
+    seq:
+      - id: entry
+        type: entry_body
+      - id: padding
+        #size: (-(_io.pos - _root.header._sizeof) % 8)
+        size: padding_needed
+    instances:
+      padding_needed:
+        value: '(_io.pos - _root.header._sizeof) % 8 == 0 ? 8 : (-(_io.pos - _root.header._sizeof) % 8)'
+  entry_body:
+    seq:
+      - id: ctime_seconds
+        type: u4
+      - id: ctime_nanoseconds
+        type: u4
+      - id: mtime_seconds
+        type: u4
+      - id: mtime_nanoseconds
+        type: u4
+      - id: dev
+        type: u4
+      - id: inode
+        type: u4
+      - id: mode
+        type: mode
+        size: 4
+      - id: uid
+        type: u4
+      - id: gid
+        type: u4
+      - id: file_size
+        type: u4
+        doc: file size, truncated to 32 bit
+      - id: object_name
+        size: _root.len_hash
+      - id: name_flag
+        type: name_flag
+      - id: name
+        type: str
+        size: name_flag.len_name
+  mode:
+    seq:
+      - id: unused
+        type: b16
+      - id: object_type
+        type: b4
+        valid:
+          any-of: [8, 10, 14]
+      - id: unused1
+        type: b3
+        valid: 0
+      - id: permissions
+        type: b9
+  name_flag:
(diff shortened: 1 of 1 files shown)


database/git_index.ksy
```diff
@@ -0,0 +1,150 @@
...
+  - id: header
+    type: header
+  - id: entries
+    type: entry
+    repeat: expr
+    repeat-expr: header.num_entries
+  - id: extensions
+    type: extension
+    repeat: until
+    repeat-until: _io.pos >= _io.size - len_hash
```
generalmimon, line 19: This seems more natural to me and I'm more certain that it is correct:

```suggestion
    repeat-until: _io.pos + len_hash >= _io.size
```

It follows the idea "if I parse the `checksum` from the current position, I'll reach the EOF condition `_io.pos == _io.size`, or `_io.pos > _io.size` (hypothetically) if there are not even enough bytes for `checksum`".

database/git_index.ksy
```diff
@@ -0,0 +1,150 @@
...
+  entry:
+    seq:
+      - id: entry
+        type: entry_body
+      - id: padding
+        #size: (-(_io.pos - _root.header._sizeof) % 8)
+        size: padding_needed
+    instances:
+      padding_needed:
+        value: '(_io.pos - _root.header._sizeof) % 8 == 0 ? 8 : (-(_io.pos - _root.header._sizeof) % 8)'
```
generalmimon, line 46: This could be simplified to this (borrowing from https://github.com/kaitai-io/kaitai_struct_formats/pull/409#discussion_r566748681):

```suggestion
        value: '8 - (_io.pos - _root.header._sizeof) % 8'
```

Although this would probably deserve a comment that inserting the 8-byte padding when it's already "aligned" is intentional, because it would normally look like a mistake.

database/git_index.ksy
```diff
@@ -0,0 +1,150 @@
+meta:
+  id: git_index
+  title: Git index
+  license: CC0-1.0
```
KOLANICH, line 4: Git is GPLed, so are its docs. Git maintainer is pro-GPL, since he has called applying the techniques used to reverse engineer proprietary software in the most licensially-clean way to git as "necessary effort" and has refused to initiate relicensing of the docs into more permissive license.

armijnhemel: No, this is not how it works. I read the e-mail thread and you misunderstood the developers: https://www.spinics.net/lists/git/msg371734.html

```
GPL copyright protects the expression of the document, but the
copyright protects only the expression, and does not protect the
underlying format itself and the idea behind it.  So I do not see a
need to relicense the documentation text at all.
```
So no, I am not going to relicense this file either.

database/git_index.ksy
```diff
@@ -0,0 +1,150 @@
...
+  ks-version: 0.9
+  endian: be
+  encoding: UTF-8
+doc-ref: https://github.com/git/git/blob/main/Documentation/technical/index-format.txt
```
generalmimon, line 8: Not `main` branch, please:

```suggestion
doc-ref: https://github.com/git/git/blob/e6ebfd0e8c/Documentation/technical/index-format.txt
```

database/git_index.ksy
```diff
@@ -0,0 +1,150 @@
...
+    repeat-expr: header.num_entries
+  - id: extensions
+    type: extension
+    repeat: until
+    repeat-until: _io.pos >= _io.size - len_hash
+  - id: checksum
+    size: len_hash
+instances:
+  len_hash:
+    value: 20   # SHA1, change to 32 for SHA256
```
generalmimon, line 24:
```suggestion
    value: 20   # SHA-1, change to 32 for SHA-256
```

database/git_index.ksy
```diff
@@ -0,0 +1,150 @@
...
+        type: str
+        size: name_flag.len_name
+  mode:
+    seq:
+      - id: unused
+        type: b16
+      - id: object_type
+        type: b4
+        valid:
+          any-of: [8, 10, 14]
```
generalmimon, line 85: Maybe an enum?

database/git_index.ksy
generalmimon: Graphs from Wolfram Alpha ([`(-x) mod 8`](https://www.wolframalpha.com/input?i2d=true&i=discontinuities+of+-x+mod+8), [`8 - (x mod 8)`](https://www.wolframalpha.com/input?i2d=true&i=discontinuities+of+8+-+%5C%2840%29x+mod+8%5C%2841%29)):

-x_mod_8
8-x_mod_8

database/git_index.ksy
```diff
@@ -0,0 +1,150 @@
...
+      - id: unused1
+        type: b3
+        valid: 0
+      - id: permissions
+        type: b9
+  name_flag:
+    seq:
+      - id: assume_valid
+        type: b1
```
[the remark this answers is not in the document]
armijnhemel: The style guide doesn't mention where it should go exactly. Feel free to move it around.

PR #590, opened 2022-03-10
rpm: add RPMTAG_ARCHSUFFIX tag

armijnhemel:
add new RPM header tag

### DIFF
--- a/archive/rpm.ksy
+++ b/archive/rpm.ksy
@@ -454,7 +454,6 @@ enums:
       -orig-id: RPMSIGTAG_RESERVEDSPACE
       doc: Space reserved for signatures
   header_tags:
-    # Tags from [lib/rpmtag.h](https://github.com/rpm-software-management/rpm/blob/911448f2/lib/rpmtag.h).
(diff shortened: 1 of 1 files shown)


archive/rpm.ksy
```diff
@@ -455,6 +455,7 @@ enums:
       doc: Space reserved for signatures
   header_tags:
     # Tags from [lib/rpmtag.h](https://github.com/rpm-software-management/rpm/blob/911448f2/lib/rpmtag.h).
+    # Tags from [lib/rpmtag.h](https://github.com/rpm-software-management/rpm/blob/650ba79f/include/rpm/rpmtag.h).
```
generalmimon, line 458: The old revision should be removed:

```suggestion
    # Tags from [lib/rpmtag.h](https://github.com/rpm-software-management/rpm/blob/650ba79f/include/rpm/rpmtag.h).
```

generalmimon (changes requested):

archive/rpm.ksy
armijnhemel: URGH. so stupid.

generalmimon (approved):
Thanks!

PR #592, opened 2022-03-18
Add log/mcap.ksy

jtbandes:
Hi! This PR imports the KSY definition for MCAP from: https://github.com/foxglove/mcap/pull/308 

This is my first time using KSY; I tried to conform to the style guide but please let me know if you have any feedback on the definition.

More info on the MCAP format: https://github.com/foxglove/mcap#readme

Example data:
- Simple examples: https://github.com/foxglove/mcap/tree/main/tests/conformance/data
- To generate more complex example data, use the [`mcap convert` tool](https://github.com/foxglove/mcap/tree/main/go/cli/mcap) on a ROS Bag file, e.g. the files from https://google-cartographer-ros.readthedocs.io/en/latest/data.html

### DIFF
--- a/log/mcap.ksy
+++ b/log/mcap.ksy
@@ -78,7 +78,7 @@ types:
     seq:
       - id: len
         type: u4
-      - id: entry
+      - id: entries
         size: len
         type: entries
     types:
--- a/log/mcap.ksy
+++ b/log/mcap.ksy
@@ -131,35 +131,37 @@ types:
 
   footer:
     seq:
-      - id: summary_start
+      - id: ofs_summary_section
+        -orig-id: summary_start
         type: u8
-      - id: summary_offset_start
+      - id: ofs_summary_offset_section
+        -orig-id: summary_offset_start
         type: u8
       - id: summary_crc32
         -orig-id: summary_crc
         type: u4
-    instances:
-      summary_section:
-        io: _root._io
-        pos: summary_start
-        size: "(summary_offset_start != 0 ? summary_offset_start : _root.ofs_footer) - summary_start"
-        type: records
-        if: summary_start != 0
-      summary_offset_section:
-        io: _root._io
-        pos: summary_offset_start
-        size: "_root.ofs_footer - summary_offset_start"
-        type: records
-        if: summary_offset_start != 0
-      summary_crc32_input:
-        io: _root._io
-        pos: summary_start
-        size: "_root._io.size - summary_start - sizeof<magic> - summary_crc32._sizeof"
-        if: summary_start != 0
         doc: |
           A CRC32 of all bytes from the start of the Summary section up through and
           including the end of the previous field (summary_offset_start) in the footer
           record. A value of 0 indicates the CRC32 is not available.
+    instances:
+      summary_section:
+        io: _root._io
+        pos: ofs_summary_section
+        size: "(ofs_summary_offset_section != 0 ? ofs_summary_offset_section : _root.ofs_footer) - ofs_summary_section"
+        type: records
+        if: ofs_summary_section != 0
+      summary_offset_section:
+        io: _root._io
+        pos: ofs_summary_offset_section
+        size: "_root.ofs_footer - ofs_summary_offset_section"
+        type: records
+        if: ofs_summary_offset_section != 0
+      summary_crc32_input:
+        io: _root._io
+        pos: ofs_summary_section
+        size: "_root._io.size - ofs_summary_section - sizeof<magic> - summary_crc32._sizeof"
+        if: ofs_summary_section != 0
 
   schema:
     seq:
@@ -253,9 +255,11 @@ types:
         type: u8
       - id: message_end_time
         type: u8
-      - id: chunk_start_offset
+      - id: ofs_chunk
+        -orig-id: chunk_start_offset
         type: u8
-      - id: chunk_length
+      - id: len_chunk
+        -orig-id: chunk_length
         type: u8
       - id: len_message_index_offsets
         type: u4
@@ -273,8 +277,8 @@ types:
     instances:
       chunk:
         io: _root._io
-        pos: chunk_start_offset
-        size: chunk_length
+        pos: ofs_chunk
+        size: len_chunk
         type: record
     types:
       message_index_offset:
@@ -299,10 +303,10 @@ types:
         type: prefixed_str
       - id: content_type
         type: prefixed_str
-      - id: data_size
+      - id: len_data
         type: u8
       - id: data
-        size: data_size
+        size: len_data
       # Trigger _io.pos computation: https://github.com/kaitai-io/kaitai_struct/issues/721#issuecomment-623011059
       - id: invoke_crc32_input_end
         size: 0
@@ -362,10 +366,10 @@ types:
         type: u8
       - id: message_end_time
         type: u8
-      - id: channel_message_counts_size
+      - id: len_channel_message_counts
         type: u4
       - id: channel_message_counts
-        size: channel_message_counts_size
+        size: len_channel_message_counts
         type: channel_message_counts
     types:
       channel_message_counts:
@@ -389,17 +393,19 @@ types:
 
   metadata_index:
     seq:
-      - id: offset
+      - id: ofs_metadata
+        -orig-id: offset
         type: u8
-      - id: length
+      - id: len_metadata
+        -orig-id: length
         type: u8
       - id: name
         type: prefixed_str
     instances:
       metadata:
         io: _root._io
-        pos: offset
-        size: length
+        pos: ofs_metadata
+        size: len_metadata
         type: record
 
   summary_offset:
@@ -407,15 +413,17 @@ types:
       - id: group_opcode
         type: u1
         enum: opcode
-      - id: group_start
+      - id: ofs_group
+        -orig-id: group_start
         type: u8
-      - id: group_length
+      - id: len_group
+        -orig-id: group_length
         type: u8
     instances:
       group:
         io: _root._io
-        pos: group_start
-        size: group_length
+        pos: ofs_group
+        size: len_group
         type: records
 
   data_end:
--- a/log/mcap.ksy
+++ b/log/mcap.ksy
@@ -24,7 +24,7 @@ seq:
 
 instances:
   footer:
-    pos: _io.size - 8 - 20 - 9
+    pos: "_io.size - sizeof<magic> - sizeof<footer> - sizeof<record_prefix>"
     size-eos: true
     type: record
 
@@ -86,8 +86,7 @@ types:
     seq:
       - id: records
         type: record
-        repeat: until
-        repeat-until: "_io.pos + 8 >= _io.size"
+        repeat: eos
 
   record_prefix:
     seq:
@@ -194,12 +193,6 @@ types:
         size-eos: true
 
   chunk:
-    types:
-      uncompressed_chunk:
-        seq:
-          - id: records
-            type: record
-            repeat: eos
     seq:
       - id: message_start_time
         type: u8
@@ -218,7 +211,7 @@ types:
         type:
           switch-on: compression.str
           cases:
-            '""': uncompressed_chunk
+            '""': records
 
   message_index:
     types:
--- a/log/mcap.ksy
+++ b/log/mcap.ksy
@@ -11,7 +11,7 @@ doc: |
   durability requirements.
 
   Time values (`log_time`, `publish_time`, `create_time`) are represented in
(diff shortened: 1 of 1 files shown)


generalmimon (changes requested):

log/mcap.ksy
```diff
@@ -0,0 +1,293 @@
...
+    0x0c: metadata
+    0x0d: metadata_index
+    0x0e: summary_offset
+    0x0f: data_end
+
+types:
+  prefixed_str:
+    seq:
+      - { id: len, type: u4 }
+      - { id: str, type: str, size: len, encoding: UTF-8 }
```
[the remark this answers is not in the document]
jtbandes: Ok, I've updated to use the block style! (Maybe I'm missing something but I definitely think it's _harder_ to read the type definitions now — when the field names were visually aligned in a column, it was easier to cross-reference with other code or specs that list out the fields of each record type.)

jtbandes:
Hi! Thanks for your review – is there anything else I can do to help get this PR ready at this point?

KOLANICH (changes requested):
1. `header_magic`, `footer_magic` are not enough DRY. Can be deduped either with YAML references or with a subtype.
2. `8` is a magic number, can it be a `sizeof<>`?
3. should `footer` be renamed into `summary` and prefixes in fields names dropped?
4. Quite a lot of types start with `offset` `length`.

jtbandes:
1, 2: thanks for the suggestions! I didn't know about `sizeof<>` since it doesn't seem to be mentioned in the [user guide](https://doc.kaitai.io/user_guide.html). Implemented in https://github.com/kaitai-io/kaitai_struct_formats/pull/592/commits/28815f3b354ea6db732150858fbd9f22ce86c74f
3, 4: these naming conventions directly follow the field names from the mcap spec: https://github.com/foxglove/mcap/tree/main/docs/specification#footer-op0x02

log/mcap.ksy
```diff
@@ -0,0 +1,397 @@
...
+  - id: records
+    type: record
+    repeat: until
+    repeat-until: _.op == opcode::footer
+
+  - id: footer_magic
+    type: magic
+
+instances:
+  footer:
```
generalmimon, line 26: Why does this instance exist if it's not used from within the spec? Isn't the footer record already available in `records`?

log/mcap.ksy
```diff
@@ -0,0 +1,397 @@
...
+        size: len
+        type: entries
+
+  records:
+    seq:
+      - id: records
+        type: record
+        repeat: eos
+
+  record_prefix:
```
generalmimon, line 91: This is a unused type only referenced in `sizeof<>`, which is disturbing. And when I look at individual pos calculations which involve `sizeof<record_prefix>`, I doubt that they're really necessary.

log/mcap.ksy
```diff
@@ -0,0 +1,397 @@
...
+        type: u8
+      - id: records
+        size: len_records
+        type:
+          switch-on: compression.str
+          cases:
+            '""': records
+
+  message_index:
+    types:
```
generalmimon, line 217: `seq` should come before `types`, see our [style guide](https://doc.kaitai.io/ksy_style_guide.html).

log/mcap.ksy
jtbandes: Does this mean you would recommend replacing them with hard-coded numbers like `8 + 1`? Or is there some way to reference the actually used fields, e.g. `sizeof<record.op> + sizeof<record.len>`? I tried this (and `::`) but it didn't seem to work.

log/mcap.ksy
jtbandes: It is accessible from `records`, however this instance allows direct access via the end of a file (which is a common access pattern for readers that want to use index data — they would not linearly scan all the records in the file from the beginning).

If there were some corruption earlier in the file, this might aid in debugging if the file suffix is intact. At least that was my thinking for including it.

Given that background info, do you think it's better to leave this in or just remove it?

log/mcap.ksy
```diff
@@ -0,0 +1,397 @@
+meta:
+  title: MCAP
+  id: mcap
```
generalmimon, line 3: A minor issue - `id` should come before `title`:

```suggestion
meta:
  id: mcap
  title: MCAP
```

jtbandes:
Hi @generalmimon @KOLANICH, thanks again for looking at this. Are there any more changes that you would like me to make?

KOLANICH:
Is `0` in the signature a format version number? Should it be available as a number and then dispatched, when the new version of the format is introduced?

jtbandes:
Yes, it is a version number — however, the format is currently quite extensible and many changes are possible that would be fully backward-compatible and not require incrementing the version number, and we expect changing the version to be quite rare. Do you think it's important to set up the dispatch on version number now, while only one version of the format exists?

KOLANICH:
>Do you think it's important to set up the dispatch on version number now, while only one version of the format exists?

No, but I think that every format should be designed to allow this kind of dispatch.

I see an issue that the format version is a part of a string `‰MCAP0` with a CRLF line ending.
1. It is not so trivial to scan for a CRLF than for a single byte
2. Is 1 digit enough for a version number? Of course one can use other chars other than CRLF, but it will dusrupt the style of the signature, I guess it was designed to be human-recognizeable in the cases the file is opened with a plain text editor.

If it was me who had designed the format, I'd have used just `MCAP
\n` as a signature folowed by a version number as a binary struct of 3 bytes (the format is binary, one still needs a hex editor to inspect it). But some properties of format signature are already kinda fixed in stone.

So it is proposed to just replace `\r` with `0`. It'll
* keep the signature within 8 bytes, so no scanning for now, just reading these 8 bytes;
* will give you 100 version numbers instead of 10, pretty enough if you are not going to change the format too often (and I guess one shouldn't be afraid to introduce breaking changes, a simple way to do it without supporting every format version within the lib is supporting only 2, the current one and the previous one and having a bunch of lib versions available as separate shared libs, so the converter tool just reads the version number from the header (`n-1`), loads the latest version of the shared lib available in runtime that can write `n`, parses the `n-1` then serializes back into `n`, then repeats until the max version of the format is not reached).
* make scanning for `\n` easier in future if 100 version numbers become not enough and one will have to go beyond 8 bytes.

jtbandes:
I appreciate the detailed analysis and feedback. However, I am also confused, because my intent with this PR was to provide a canonical KSY definition for an existing format, and I expected feedback on the KSY definition, but not the format itself 😅

The MCAP format is officially defined in [this spec document](https://github.com/foxglove/mcap/tree/main/docs/specification), and although it may change in the future (that's why we have a version in the header), the current version has been through months of development and [production](https://foxglove.dev/docs/studio/connection/mcap) [use](https://github.com/ros-tooling/rosbag2_storage_mcap) and there is no room for breaking changes within the current "0" version. It seems like this PR is the wrong place to discuss changes to the MCAP format itself. Would you agree?

By the way, the 8-byte signature chosen for MCAP was inspired by the [HDF5 signature](https://support.hdfgroup.org/HDF5/doc/H5.format.html#Superblock):

> This signature both identifies the file as an HDF5 file and provides for immediate detection of common file-transfer problems. The first two bytes distinguish HDF5 files on systems that expect the first two bytes to identify the file type uniquely. The first byte is chosen as a non-ASCII value to reduce the probability that a text file may be misrecognized as an HDF5 file; also, it catches bad file transfers that clear bit 7. Bytes two through four name the format. The CR-LF sequence catches bad file transfers that alter newline sequences. The control-Z character stops file display under MS-DOS. The final line feed checks for the inverse of the CR-LF translation problem. (This is a direct descendent of the [PNG](http://www.libpng.org/pub/png/spec/iso/index-object.html#5PNG-file-signature) file signature.)

KOLANICH:
>It seems like this PR is the wrong place to discuss changes to the MCAP format itself. Would you agree?

Completely agree. It was just a small offtop remark about the version number. Thank you for the quote from HDF spec.

log/mcap.ksy
```diff
@@ -0,0 +1,397 @@
...
+        type: map_str_str
+
+  message:
+    seq:
+      - id: channel_id
+        type: u2
+      - id: sequence
+        type: u4
+      - id: log_time
+        type: u8
```
generalmimon, line 189: The meaning of `_time` integers is nowhere to be found in the `.ksy` spec. The original specification provides a description of these: https://github.com/foxglove/mcap/tree/ad4e4434e441298e680ce6e6273af74aed71a94b/docs/specification#timestamp.

There are quite a few `_time` fields in the spec, so you can describe them all in the top-level `/doc`.

generalmimon (changes requested):
The naming of `len_*`/`*_size`/`*_length` and `*_start_offset`/`*_start`/`offset` fields is inconsistent and unpredictable (or at least **I** don't personally see a consistent pattern in there).

In general, please try to follow these naming conventions from the [style guide](https://doc.kaitai.io/ksy_style_guide.html#attr-id):

> * For fields that designate **offset** to some particular data structure, use `ofs_` prefix and name of that data structure (as it would appear in the file) — i.e. `ofs_block`, `ofs_queries`, `ofs_path`
> * For fields that designate **size** of some particular data structure (in bytes or some other fixed units), use `len_` prefix and name of that data structure — i.e. `len_block` (length of a single `block` entry), `len_blocks` (total length of whole `blocks` array, made of `block` entries).

If you would like to keep original spelling for reference, there is a `-orig-id` key for that.

log/mcap.ksy
```diff
@@ -0,0 +1,397 @@
...
+      - id: offset
+        type: u8
+      - id: length
+        type: u8
+      - id: log_time
+        type: u8
+      - id: create_time
+        type: u8
+      - id: data_size
+        type: u8
```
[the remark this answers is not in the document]
jtbandes: Both `attachment_index.data_size` and `chunk_index.message_index_length` are referring to data elsewhere in the file (within the Attachment record referenced by the AttachmentIndex, and _after_ the Chunk record referenced by the ChunkIndex, respectively) — not within this stream. They can optionally be used by readers seeking around the file or creating summary output, but are not used directly in the ksy definition. The spellings `data_size` and `message_index_length` are original names from the mcap spec. Would you suggest renaming them and adding `-orig-id`, or leaving them as-is?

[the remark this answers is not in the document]
generalmimon: > Both `attachment_index.data_size` and `chunk_index.message_index_length` are referring to data elsewhere in the file (within the Attachment record referenced by the AttachmentIndex, and _after_ the Chunk record referenced by the ChunkIndex, respectively) — not within this stream. They can optionally be used by readers seeking around the file or creating summary output, but are not used directly in the ksy definition.

OK, thanks for clarification.

> The spellings `data_size` and `message_index_length` are original names from the mcap spec.

Well, in my opinion, names of size fields that are not directly used as `size` of another field do not need to be standardized. The point is especially to make the usual pair `len_{data}`, `{data}` predictable and consistent in naming - in the same spec, but also ideally across the format gallery. However, there's not much pressure to apply this if it the situation isn't that clear and simple.

log/mcap.ksy
```diff
@@ -0,0 +1,436 @@
...
+        io: _root._io
+        pos: ofs_summary_offset_section
+        size: "_root.ofs_footer - ofs_summary_offset_section"
+        type: records
+        if: ofs_summary_offset_section != 0
+      summary_crc32_input:
+        io: _root._io
+        pos: ofs_summary_section
+        size: "_root._io.size - ofs_summary_section - sizeof<magic> - summary_crc32._sizeof"
+        if: ofs_summary_section != 0
```
generalmimon, line 164: According to your commit https://github.com/foxglove/mcap/commit/44355df51a2f77b764b42bb9b3b0985200a168ad, `summary_crc32` is calculated from _something_ (not sure from what exactly) even if the summary section is empty (`ofs_summary_section == 0` or `summary_start == 0` in original identifiers). Can the implementation of `summary_crc32_input` be extended to this case?

jtbandes: Thanks for the reminder, fixed!

generalmimon (approved):
@jtbandes Thanks for the contribution! You did a great job ;)

jtbandes:
@generalmimon Thank you!

Any idea why the SVG diagram is not showing up at https://formats.kaitai.io/mcap/?

The web IDE didn't show any visible error when generating graphviz output from mcap.ksy, but with some console hackery I found this error:

image

jtbandes:
This seems to be the code that is throwing the error (in `$c_Lio_kaitai_struct_GraphvizClassCompiler.prototype.affectedVars__Lio_kaitai_struct_exprlang_Ast$expr__sci_List`):

image

PR #593, opened 2022-03-26
DOS MZ updates

armijnhemel:
This updates fixes the magic and reworks the header which also corrects the offsets of the body.

The previous version did not work for various executables used in FreeDOS 1.2 like `PING.EXE`. These are now correctly (or: "more correctly") parsed.

The biggest change is reworking the header and to look at the actual length of the header (for which a separate `len_header` instance was created). According to <http://www.delorie.com/djgpp/doc/exe/>:

```
 The header includes the relocation entries.
```

### DIFF
--- a/executable/dos_mz.ksy
+++ b/executable/dos_mz.ksy
@@ -11,6 +11,7 @@ meta:
   tags:
     - dos
   license: CC0-1.0
+  ks-version: 0.9
   endian: le
 doc: |
   DOS MZ file format is a traditional format for executables in MS-DOS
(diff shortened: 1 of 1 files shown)


armijnhemel:
I wanted to include data like `num_pages` to calculate the actual size of the body, but this doesn't work in case the binary has been packed or compressed, or if some extender has been used. For example:

```
$ file PING.EXE 
PING.EXE: MS-DOS executable, COFF for MS-DOS, DJGPP go32 DOS extender
```

The data inside the file says that the file should be 2048 bytes, but the actual file is much larger (1818189 bytes). The first 2048 bytes is the stub that is executed to find the rest of the data.

executable/dos_mz.ksy
```diff
@@ -21,22 +22,32 @@ doc: |
   segment of raw CPU instructions), DOS MZ .exe file format allowed
   more flexible memory management, loading of larger programs and
   added support for relocations.
+doc-ref:
+  - http://www.delorie.com/djgpp/doc/exe/
```
dgelessus, line 26:
```suggestion
doc-ref: http://www.delorie.com/djgpp/doc/exe/
```

executable/dos_mz.ksy
```diff
@@ -21,22 +22,32 @@ doc: |
...
+    seq:
+      - id: mz
+        type: mz_header
+      - id: rest_of_header
+        size: mz.len_header - mz._sizeof
   mz_header:
     seq:
       - id: magic
-        size: 2
+        contents: 'MZ'
```
dgelessus, line 50: I think the magic number can also be `ZM` instead of `MZ`, at least in some cases (DOS but not Windows, possibly?). I don't have a good source for this sadly and don't remember where I first read it. But if you search for 'exe magic number "zm"' with your favorite search engine, you can find a number of people mentioning this.

armijnhemel: There are a few sources that say it, but it is apparently very rare. I will add it.

dgelessus (approved):
LGTM, thanks!

PR #594, opened 2022-03-27
add description of COFF

armijnhemel:
A fairly simple grammar for the COFF format as used by some DOS extenders, like DJGPP. An example is the file `PING.EXE` from FreeDOS 1.2:

```
$ file PING.EXE 
PING.EXE: MS-DOS executable, COFF for MS-DOS, DJGPP go32 DOS extender
```

The COFF data follows the MS-DOS executable data.

Other example files can be found in FreeDOS 1.2 inside `FPC.ZIP` and `GOPHERUS.ZIP`

### DIFF
--- a/executable/coff.ksy
+++ b/executable/coff.ksy
@@ -1,0 +1,171 @@
+meta:
+  id: coff
+  title: COFF
+  tags:
+    - dos
+  license: CC0-1.0
+  ks-version: 0.9
+  encoding: ascii
+  endian: le
+doc-ref: https://wiki.osdev.org/COFF
+seq:
+  - id: header
+    type: header
+  - id: optional_header
+    type: optional_header
+    if: header.len_optional_header != 0
+  - id: section_headers
+    type: section_header
+    repeat: expr
+    repeat-expr: header.num_sections
+instances:
+  symbol_table_and_string_table:
+    pos: header.ofs_symbol_table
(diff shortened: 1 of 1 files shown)


executable/coff.ksy
```diff
@@ -0,0 +1,171 @@
...
+  header:
+    seq:
+      - id: magic
+        size: 2
+      - id: num_sections
+        type: u2
+      - id: time_and_date
+        type: u4
+      - id: ofs_symbol_table
+        type: u4
```
generalmimon, line 35: BTW, is this the same structure as

https://github.com/kaitai-io/kaitai_struct_formats/blob/e4f724eaf938f32177b2a3c92f4a983cf5ecb9e0/executable/microsoft_pe.ksy#L61-L72

? If so, the [`executable/microsoft_pe.ksy`](https://github.com/kaitai-io/kaitai_struct_formats/blob/e4f724eaf938f32177b2a3c92f4a983cf5ecb9e0/executable/microsoft_pe.ksy) spec should probably be refactored to import this type from here. Also maybe you can take something from the existing implementation and improve your spec with it?

armijnhemel: It is related, but not completely identical. COFF is a format that was adopted by PE and extended. One of the differences is that the `magic` field in COFF has become the `machine` field in PE.

PR #596, opened 2022-04-03
Squashfs

tisoft:
Added support for the SquashFS filesystem.

Based on the SqashFS Superblock from @abitrolly from #308.

Sample application using this structure: https://github.com/tisoft/jsquashfs

Replaces #308

### DIFF
--- a/filesystem/squashfs.ksy
+++ b/filesystem/squashfs.ksy
@@ -6,6 +6,7 @@ meta:
     - snap
     - sqfs
   endian: le
+  bit-endian: be
   license: CC0-1.0
   xref:
     wikidata: Q389314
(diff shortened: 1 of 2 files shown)


tisoft:
Not all inode types are implemented yet, but the most used basic_file, basic_directory and extended_fil) are implemented. The other types shouldn't be a big problem.

Export and xattrs tables are not yet implemented.

SquashFS uses compressed metadata blocks. To fully support the format, it is needed to uncompress those blocks and parse the resulting bytes. Currently a custom decompress function is used for this. This could probably be replaced with kaitai.compress functions.

Unfortunately it is possible for inodes in the inode table to span 2 metablocks. That is why currently a custom concat function is needed, that concatenates the uncompressed bytes from a list of metadata blocks.

This format is currently very memory intensive. The real SquashFS implementations use caching to only have those metablocks in memory, that are currently needed. I couldn't find a good solution for that in kaitai.

The custom compression function also allows to automatically pad the resulting byte array with 0x00. This is used for sparse files in SquashFS. It is only needed in data blocks, so it is not needed for parsing the metadata. It could be removed from the format and could be handled any the calling application. For now I included it, to have a "complete" format.

filesystem/squashfs.ksy
```diff
@@ -0,0 +1,459 @@
...
+    12: extended_character_device
+    13: extended_named_pipe
+    14: extended_socket
+instances:
+  id_table:
+    io: _root._io
+    pos: superblock.id_table_start
+    type: metablock_reference_list((superblock.bytes_used-superblock.id_table_start)/8)
+  xattr_id_table:
+    if: superblock.xattr_id_table_start != 18446744073709551615
```
generalmimon, line 51:
```suggestion
    if: superblock.xattr_id_table_start != 0xffff_ffff_ffff_ffff
```

PR #601, opened 2022-04-14
zip: update compression and extra_codes enums, sync with upstream APPNOTE.txt

armijnhemel:
This PR updates the `compression` and `extra_codes` enums and syncs it with the latest `APPNOTE.txt`, at least for `compression`. There are still some values in `extra_codes` that already appeared in older versions of `APPNOTE.txt` but were not added to the `.ksy`. I am not sure if this was on purpose or not.


generalmimon (approved):
@armijnhemel Thanks for the contribution!

PR #602, opened 2022-04-15
Zip updates 202204

armijnhemel:
This PR expands parsing of ZIP files. Changed:

* moved section_type into a separate enum, expanded the enum
* process several more sections, including some ZIP64 related data structures
* fixes for `extended_timestamp` which has a different format in the central directory than in a local file
* process several more extra data fields

It should be noted that this will not mean that all files are successfully parsed: there are many variants of ZIP that are slightly different and in some cases it is not possible to successfully parse a ZIP file when parsing from the beginning, as in some cases (possibly related to files created in a streaming fashion) the sizes of the individual elements are only declared *after* the actual data. This is not a problem for regular problems as they rely on the central directory to look up the files in the archive instead.

The ZIP64 extra field has not been enabled yet, because with test files I am seeing different values. This needs some extra research.

Extra not for the `alzip_code_page` extra data element: this field I have so far only seen in files generated in Korea (ALZip is made by a Korean company). The only observed code page is 949, which is the Windows code page for "unified Hangul": <https://en.wikipedia.org/wiki/Unified_Hangul_Code>

### DIFF
--- a/archive/zip.ksy
+++ b/archive/zip.ksy
@@ -56,6 +56,8 @@ types:
             section_types::data_descriptor: data_descriptor
             section_types::archive_extra_data: archive_extra_data
             section_types::digital_signature: digital_signature
+            section_types::zip64_end_of_central_dir: zip64_end_of_central_directory
+            section_types::zip64_end_of_central_dir_locator: zip64_end_of_central_directory_locator
   archive_extra_data:
     seq:
       - id: len_extra_field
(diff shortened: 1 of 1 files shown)


armijnhemel:
[zip64_end_of_central_directory_test.zip](https://github.com/kaitai-io/kaitai_struct_formats/files/8496440/zip64_end_of_central_directory_test.zip)

Test file extracted from an actual ZIP file with several ZIP64 related sections. References to the actual existing files will not work, so you probably need to comment the following in `central_dir_entry` when testing with for example the web IDE:

```
    instances:
      local_header:
        pos: ofs_local_header
        type: pk_section
```

armijnhemel:
[aex_encryption_test.zip](https://github.com/kaitai-io/kaitai_struct_formats/files/8496462/aex_encryption_test.zip)

Test file extracted from an actual ZIP file with AE-x encryption. Like above References to the actual existing files will not work.

armijnhemel:
[kaitai.zip](https://github.com/kaitai-io/kaitai_struct_formats/files/8496467/kaitai.zip)

Full ZIP file (with Kaitai's own contributions so can be fully shared) for testing with the `extended_timestamp` in the central directory and local files.

archive/zip.ksy
```diff
@@ -36,22 +36,34 @@ doc-ref:
...
-            0x0605: end_of_central_dir
-            0x0807: data_descriptor
+            section_types::central_dir_entry: central_dir_entry
+            section_types::local_file: local_file
+            section_types::end_of_central_dir: end_of_central_dir
+            section_types::data_descriptor: data_descriptor
+            section_types::archive_extra_data: archive_extra_data
+            section_types::digital_signature: digital_signature
+            section_types::zip64_end_of_central_dir: zip64_end_of_central_directory
+            section_types::zip64_end_of_central_dir_locator: zip64_end_of_central_directory_locator
```
generalmimon, line 60: I would use `central_dir` everywhere to make the spec consistent, rather than interspersing it with expanded ~~`central_directory`~~.

```diff
-            section_types::zip64_end_of_central_dir: zip64_end_of_central_directory
-            section_types::zip64_end_of_central_dir_locator: zip64_end_of_central_directory_locator
+            section_types::zip64_end_of_central_dir: zip64_end_of_central_dir
+            section_types::zip64_end_of_central_dir_locator: zip64_end_of_central_dir_locator
```

Please rename all relevant fields and types (and their usages) accordingly.

PR #603, opened 2022-05-04
zisofs: add specification

armijnhemel:
This PR adds a specification for zisofs compressed files. I have put it in `archive` as I think that this might be the best place, as it processes a single file that has been compressed with `mkzftree`. That being said, I could imagine it being conceptually belonging to `filesystem`, so if you want to move it, go ahead.

<http://fileformats.archiveteam.org/wiki/Zisofs>

### DIFF
--- a/archive/zisofs.ksy
+++ b/archive/zisofs.ksy
@@ -61,13 +61,16 @@ types:
         value: '(uncompressed_size / block_size) + (uncompressed_size % block_size != 0 ? 1 : 0)'
         doc: ceil(uncompressed_size / block_size)
   block:
+    -webide-representation: '[{ofs_start}, {ofs_end}): {len_data:dec} bytes'
     params:
       - id: ofs_start
         type: u4
       - id: ofs_end
         type: u4
     instances:
+      len_data:
+        value: ofs_end - ofs_start
       data:
         io: _root._io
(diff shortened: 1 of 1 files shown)


armijnhemel:
[ls-zisofs.zip](https://github.com/kaitai-io/kaitai_struct_formats/files/8619717/ls-zisofs.zip)

This file is an ELF binary from Fedora that has been preprocessed with `mkzftree`.

armijnhemel:
> > I have put it in `archive` as I think that this might be the best place, as it processes a single file that has been compressed with `mkzftree`.
> 
> I agree, `archive` seems best. I added 'filesystem' to `/meta/tags`.
> 
> > [ls-zisofs.zip](https://github.com/kaitai-io/kaitai_struct_formats/files/8619717/ls-zisofs.zip)
> > This file is an ELF binary from Fedora that has been preprocessed with `mkzftree`.
> 
> Can you provide more details, URLs or file paths where the file came from? I would like to add it to https://github.com/kaitai-io/kaitai_struct_samples.

Oh, I copied it from a Fedora 33 system (from `coreutils` 21.fc33) and then created the file myself using `mkzftree`

generalmimon (approved):

PR #606, opened 2022-05-15
elf: sync with upstream binutils code

armijnhemel:
elf: sync with upstream binutils code, particularly:

https://gitlab.com/gnutools/binutils-gdb/-/commit/dd207c1302d28d2d2b33cff058bb3fbf5138b480
https://gitlab.com/gnutools/binutils-gdb/-/commit/a7fd11862703e45d2774981a4888bc127d473b06

### DIFF
--- a/executable/elf.ksy
+++ b/executable/elf.ksy
@@ -1784,18 +1784,17 @@ enums:
     28: fini_arraysz     # Size in bytes of DT_FINI_ARRAY
     29: runpath          # Library search path
     30: flags            # Flags for the object being loaded
-    # 32: encoding  # special value (marker):
-                    # Values `v >= ::encoding and v < ::lo_os` follow the rules
-                    # for the interpretation of the d_un union as follows:
-                    # even number == 'd_ptr', odd number == 'd_val' or none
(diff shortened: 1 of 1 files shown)


executable/elf.ksy
```diff
@@ -1791,6 +1792,10 @@ enums:
     32: preinit_array    # Array with addresses of preinit fct
     33: preinit_arraysz  # Size in bytes of DT_PREINIT_ARRAY
     34: symtab_shndx     # Address of SYMTAB_SHNDX section
+    35: relrsz
+    36: relr
+    37: relrent
+    38: encoding
```
generalmimon, line 1798: `DT_ENCODING` is not a real tag value, but a special marker:

https://github.com/kaitai-io/kaitai_struct_formats/blob/06ecf174ef2801750c4480d5986553622e6f512d/executable/elf.ksy#L1787-L1792

Therefore, it must stay commented out as it was (it's not an actual tag, and it may even "collide" with an actual tag on the same numerical value, as seen in the existing state where the `encoding` value is 32). The old `encoding` explanatory comment (shown in the above code snippet) needs to be moved to this new position.

armijnhemel: Whoops, my bad.

generalmimon (approved):
@armijnhemel Thanks!

PR #614, opened 2022-06-19
uimage: add new uimage_type

armijnhemel:
Add new uimage_type (Allwinner TOC0 Boot image)


armijnhemel:
It might be wise to check PR #548 first

Issue #617, opened 2022-07-12
CRC checksum value of ethernet_frame


H4rryK4ne:
Hi,

we want to validate the CRC value of the ethernet frame, and we became aware, that this is not in the ethernet_frame at all.

My first approach would be to use an instance to get the last 4 bytes of the body to get the CRC value. 

Maybe someone has a smarter way to do it.

BR
Martin

H4rryK4ne:
Thanks to @generalmimon there is indeed a smarter solution than I thought, but in the end we found, that all the pcap files we have, don't include the frame check sum.

PR #618, opened 2022-07-14
ethernet_frame checksum

H4rryK4ne:
My solution for issue #617

### DIFF
--- a/network/ethernet_frame.ksy
+++ b/network/ethernet_frame.ksy
@@ -32,13 +32,9 @@ seq:
     type: u2be
     enum: ether_type_enum
     if: ether_type_1 == ether_type_enum::ieee_802_1q_tpid
-  - id: body
+  - id: body_and_checksum
     size-eos: true
-    type:
-      switch-on: ether_type
-      cases:
-        'ether_type_enum::ipv4': ipv4_packet
-        'ether_type_enum::ipv6': ipv6_packet
+
 instances:
   ether_type:
     value: |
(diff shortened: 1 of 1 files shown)


network/ethernet_frame.ksy
```diff
@@ -48,6 +44,20 @@ instances:
       first location bears special marker (0x8100), then it is not the
       real ether frame yet, an additional payload (`tci`) is expected
       and real ether type is upcoming next.
+  body:
+    io: _root._io
+    pos: _root._io.size - body_and_checksum.length
+    size: body_and_checksum.length - 4
```
generalmimon, line 50: I think this is an unnecessarily clumsy and complex implementation. Why not just

```diff
   - id: body
-    size-eos: true
+    size: (_io.size - _io.pos) - checksum._sizeof
     type:
       switch-on: ether_type
       cases:
         'ether_type_enum::ipv4': ipv4_packet
         'ether_type_enum::ipv6': ipv6_packet
+  - id: checksum
+    size: 4
```

...?

network/ethernet_frame.ksy
```diff
@@ -48,6 +44,20 @@ instances:
...
+    type:
+      switch-on: ether_type
+      cases:
+        'ether_type_enum::ipv4': ipv4_packet
+        'ether_type_enum::ipv6': ipv6_packet
+  checksum:
```
generalmimon, line 56: What is the checksum algorithm? If it's CRC-32, I'd prefer calling the field `crc32` (instead of ~~`checksum`~~).

Also, it would be great if you can provide a parse instance for the input byte array of the checksum, preferably called "name of the checksum field + `_input`" (e.g. `crc32_input`). This will make it much easier to check if the checksum is valid (in the application code, or even manually in a hex viewer).

generalmimon (changes requested):

network/ethernet_frame.ksy
H4rryK4ne: > What is the checksum algorithm? If it's CRC-32, I'd prefer calling the field crc32 (instead of checksum).

I wanted to use the same naming as used by TCP or UDP. But I am also fine with `crc32` instead of `checksum`.

> Also, it would be great if you can provide a parse instance for the input byte array of the checksum, preferably called "name of the checksum field + _input" (e.g. crc32_input). This will make it much easier to check if the checksum is valid (in the application code, or even manually in a hex viewer).

I am not sure what you mean with "parse instance". An example bytearray to test the code?

network/ethernet_frame.ksy
H4rryK4ne: Oh, this is indeed a much better solution.

H4rryK4ne:
Oh, what a cluster fuck. We applied this solution on our real data, just to find, that the pcap files don't have any checksum information (facepalm). I am sorry for the hassle :(

PR #619, opened 2022-07-19
Add Quake II MD2 model format

jdjakub:
I've been working with this format recently and saw it was missing, so I plugged the gap! I've followed the syle guide as far as I can tell, though being new to Kaitai I don't know what I should do with the overfull conditional expression lines.

You can find examples as `tris.md2` under `http://tastyspleen.net/~quake2/baseq2/players/`. A nice, simple example is `crate/tris.md2` which is just standard box geometry repeated across all frames.


game/md2.ksy
```diff
@@ -0,0 +1,194 @@
+meta:
+  id: md2
```
generalmimon, line 2: We already have [`game/quake_mdl.ksy`](https://github.com/kaitai-io/kaitai_struct_formats/blob/8b22d829/game/quake_mdl.ksy) and [`game/quake_pak.ksy`](https://github.com/kaitai-io/kaitai_struct_formats/blob/8b22d829/game/quake_pak.ksy) in the format gallery, so this should probably start with `quake_` (i.e. `quake_md2`) as well.

game/md2.ksy
```diff
@@ -0,0 +1,194 @@
...
+  with some `vertices`. Each GL vertex contains `tex_coords_normalized` from 0
+  to 1, and a `vertex_index` into frame vertices.
+
+  A model may also contain `skins`, which are just file paths to PCX images.
+  However, this is empty for many models, in which case it is up to the client
+  (e.g. Q2PRO) to offer skins some other way (e.g. by similar filename in the
+  current directory.)
+doc-ref: |
+  https://icculus.org/~phaethon/q3a/formats/md2-schoenblum.html
+  http://tfc.duke.free.fr/coding/md2-specs-en.html
```
generalmimon, line 34: You should use a list in `doc-ref` (supported since 0.9) instead:

```suggestion
doc-ref:
  - https://icculus.org/~phaethon/q3a/formats/md2-schoenblum.html
  - http://tfc.duke.free.fr/coding/md2-specs-en.html
```

game/md2.ksy
```diff
@@ -0,0 +1,194 @@
...
+  - id: ofs_gl_cmds
+    type: u4
+  - id: ofs_eof
+    type: u4
+instances:
+  skins:
+    pos: ofs_skins
+    size: 64
+    type: str
+    terminator: 0
```
generalmimon, line 76: There is a shortcut for this:

```suggestion
    type: strz
```

which would be I guess _idiomatic_ to use, but I wouldn't really mind keeping the expanded version either (note: there is another `type: str` + `terminator: 0` later in the file, so if you decide to switch to `type: strz` here, please don't forget to do it over there as well).

game/md2.ksy
```diff
@@ -0,0 +1,194 @@
...
+    encoding: ascii
+    repeat: expr
+    repeat-expr: num_skins
+  tex_coords:
+    pos: ofs_tex_coords
+    type: tex_point
+    repeat: expr
+    repeat-expr: num_tex_coords
```
generalmimon, line 84: Maybe `tex_points` would be better?

generalmimon (commented):
Overall, this PR is of pretty good quality.

game/md2.ksy
jdjakub: Renamed to quake2_md2, hope that's fine

game/md2.ksy
jdjakub: Done

game/md2.ksy
jdjakub: Changed in both locations. As long as this still reserves 16 chars and extracts the null-terminated portion, it's ok. Seems to still work in the IDE.

game/md2.ksy
jdjakub: Yeah, I understand but I'm going to keep it like this. I wanted the common phrase "texture coords" to be primary, but then we've got several of them which strictly speaking should be something like texture coords*es*. I chose this instead to still have a singular name for the element type while still being able to tell a programmer looking for "texture coords" that they live there.

game/quake2_md2.ksy
```diff
@@ -0,0 +1,227 @@
...
+  A model may also contain `skins`, which are just file paths to PCX images.
+  However, this is empty for many models, in which case it is up to the client
+  (e.g. Q2PRO) to offer skins some other way (e.g. by similar filename in the
+  current directory.)
+
+  There are 198 `frames` in total, partitioned into a fixed set of ranges used
+  for different animations. Each frame has a standard `name` for humans, but the
+  client just uses their index and the name can be arbitrary. The frames are:
+
+   INDEX     NAME SUFFIX NOTES
```
jdjakub, line 37: I added this section, formatted the table like this from the POV of someone reading a text file. However, now I think about it, will the doc section be rendered as Markdown in a web page? In which case, I should make this a Markdown table?

jdjakub: I've made this a Markdown table so it should display correctly. I think this PR is ready to merge.

game/quake2_md2.ksy
```diff
@@ -0,0 +1,249 @@
...
+        type: f4
+        repeat: expr
+        repeat-expr: 2
+      - id: vertex_index
+        type: u4
+enums:
+  gl_primitive:
+    0: triangle_strip
+    1: triangle_fan
+  anim_start_index:
```
jdjakub, line 229: I have added this machine-readable version of part of the animation table. I would also like to include machine-readable data for how many frames each anim has, but I can't see how to do this in Kaitai. For example, the three pain anims all have 4 frames, which would clash if I tried to use another enum for this purpose. Is there any way to include constant key-value definitions where several keys might share the same value?

game/quake2_md2.ksy
generalmimon: @jdjakub:

> will the doc section be rendered as Markdown in a web page?

Yes. The top-level `doc` key is rendered at https://formats.kaitai.io/ as the format description.

> I've made this a Markdown table so it should display correctly.

Unfortunately, the Markdown dialect we use is [CommonMark](https://commonmark.org/), which doesn't provide markup for tables. So you could use either use HTML syntax (`<table>` etc. - it would probably lose the readability a bit), or you may insert a plaintext code block `` ``` `` with an ASCII-formatted table.

ASCII table is fine, if you ask me.

jdjakub: I've now expanded the table as a preformatted block. Do you have any thoughts on my Sep 21 comment below beginning "I have added this machine-readable version of part of the animation table..."?

game/quake2_md2.ksy
generalmimon: @jdjakub:

> For example, the three pain anims all have 4 frames, which would clash if I tried to use another enum for this purpose.

Yes, an enum in Kaitai Struct is a strictly bidirectional mapping ([one-to-one correspondence](https://en.wikipedia.org/wiki/Bijection)).

> Is there any way to include constant key-value definitions where several keys might share the same value?

You can use [value instances](https://doc.kaitai.io/user_guide.html#_value_instances) to define constants:

```ksy
instances:
  my_const:
    value: 4
```

Technically you can also declare arrays there (though I'm not sure whether it makes sense here):

```ksy
instances:
  my_array:
    value: |
      [2, 3, 5, 7, 11, 13]
```

Another option (but it depends on what you mean by machine-readable) may be to add a custom YAML key somewhere. KS compiler ignores any properties starting with a dash `-` (see https://doc.kaitai.io/user_guide.html#orig-id), so this can be used to add arbitrary metadata. But these keys won't be provided in generated parsers.

jdjakub: Thanks. I have replaced the enum with three value instances and updated the doc to refer to them. If it's OK then please go ahead and merge, I want to be able to look it up on the website :)

game/quake2_md2.ksy
```diff
@@ -0,0 +1,245 @@
...
+        value: z_compressed * _parent._parent.scale.z + _parent._parent.translate.z
+  vertex:
+    seq:
+      - id: position
+        type: compressed_vec
+      - id: normal_index
+        type: u1
+        doc: `normal = bytedirs[normal_index]`
+        doc-ref: |
+          https://github.com/skullernet/q2pro/blob/master/src/common/math.c#L80
```
generalmimon, line 205: Referring to ever-changing references (branches) like `master` is not future-proof. Even more so if you link to a specific line number (which is a good thing, and it should be done whenever possible) - in general, there's a very good chance that soon there will be at least something quite different at line 80 than it is now. In the long run (several years) it's also not uncommon that code gets restructured/moved and the referenced source file will no longer exist on the `master` branch.

Therefore, you should use something permanent like a version tag (if available) or a commit hash (I like it shortened, because that's usually enough to identify the commit and doesn't disturb much), for example:

```suggestion
          https://github.com/skullernet/q2pro/blob/f4faabd/src/common/math.c#L80
```

game/quake2_md2.ksy
```diff
@@ -0,0 +1,245 @@
...
+  The MD2 format is used for 3D animated models in id Sofware's Quake II.
+
+  A model consists of named `frames`, each with the same number of `vertices`
+  (`vertices_per_frame`). Each such vertex has a `position` in model space,
+  and a `normal_index` which you must look up to get its normal. Each vertex
+  has the same topological "meaning" across frames, in terms of triangle and
+  texture info; it just varies in position and normal for animation purposes.
+
+  How the vertices form triangles is defined via disjoint `triangles` or via
+  `gl_cmds` (which allows strip and fan topology.) Each triangle contains three
```
generalmimon, line 18:
```suggestion
  `gl_cmds` (which allows strip and fan topology). Each triangle contains three
```

Please do the same for the rest of the occurrences.

game/quake2_md2.ksy
```diff
@@ -0,0 +1,245 @@
...
+  - https://icculus.org/~phaethon/q3a/formats/md2-schoenblum.html
+  - http://tfc.duke.free.fr/coding/md2-specs-en.html
+  - http://tastyspleen.net/~panjoo/downloads/quake2_model_frames.html
+  - http://wiki.polycount.com/wiki/OldSiteResourcesQuake2FramesList
+seq:
+  - id: magic
+    contents: IDP2
+  - id: version
+    type: u4
+    doc: Always 8, apparently.
```
generalmimon, line 81: It may be a good idea to enforce the expected version using the `valid` key (see https://github.com/kaitai-io/kaitai_struct/issues/435; it's supported since 0.9, but still has no proper docs):

```suggestion
  - id: version
    type: u4
    valid: 8
```

If the version wouldn't be 8, it provides feedback (a validation error) by rejecting the file immediately.

generalmimon (changes requested):

game/quake2_md2.ksy
```diff
@@ -0,0 +1,245 @@
...
+        value: y_compressed * _parent._parent.scale.y + _parent._parent.translate.y
+      z:
+        value: z_compressed * _parent._parent.scale.z + _parent._parent.translate.z
+  vertex:
+    seq:
+      - id: position
+        type: compressed_vec
+      - id: normal_index
+        type: u1
+        doc: `normal = bytedirs[normal_index]`
```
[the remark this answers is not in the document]
jdjakub: I'm surprised; throughout, I've been using the Kaitai online IDE to make sure it compiles and that doesn't give me any errors. Just a heads up in case it's using a different version.

game/quake2_md2.ksy
jdjakub: Aha, very good point. I will do this in future.

jdjakub:
I've incorporated your suggested changes.

generalmimon (changes requested):

game/quake2_md2.ksy
```diff
@@ -0,0 +1,249 @@
...
+        type: strz
+        encoding: ascii
+      - id: vertices
+        type: vertex
+        repeat: expr
+        repeat-expr: _root.vertices_per_frame
+  gl_cmd:
+    seq:
+      - id: cmd_num_vertices
+        type: s4
```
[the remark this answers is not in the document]
jdjakub: Oh no... that's my custom model breaking it 😨 Thanks for all the work in coming up with those suggestions, I wouldn't have thought of it because I'm not experienced with Kaitai.

game/quake2_md2.ksy
```diff
@@ -0,0 +1,249 @@
...
+      z:
+        value: z_compressed * _parent._parent.scale.z + _parent._parent.translate.z
+  vertex:
+    seq:
+      - id: position
+        type: compressed_vec
+      - id: normal_index
+        type: u1
+        doc: |
+          `normal = bytedirs[normal_index]`
```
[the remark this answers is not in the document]
jdjakub: Ideally, I'd like to include the entire table as a value instance like I did with `anim_num_frames`. Would Kaitai permit the nested array literal? Like so:

```
anorms_table:
  value: |
    [
      [-0.525731, 0.000000, 0.850651], 
      [-0.442863, 0.238856, 0.864188], 
      [-0.295242, 0.000000, 0.955423],
      ...
    ]
```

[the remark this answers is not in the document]
jdjakub: Alright, I've gone one further and included `anorms_table` as a giant value instance. I was glad to verify in the IDE that Kaitai accepts the nested array literal. If the user sees it in a doc block then the poor sap would have to copy and paste it somewhere, change the brace syntax ... better to just have the constants there.

jdjakub:
Thanks for your detailed dive into all the issues. I've incorporated your changes. Back to you :)

game/quake2_md2.ksy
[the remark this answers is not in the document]
generalmimon: > If the user sees it in a doc block then the poor sap would have to copy and paste it somewhere, change the brace syntax ... better to just have the constants there.

Yeah, why not, actually 😉

game/quake2_md2.ksy
```diff
@@ -0,0 +1,429 @@
...
+    # in http://tfc.duke.free.fr/coding/src/md2.c initially read as `int *`
+    # (a pool of `int`s) => hence `s4`
+    size: 4 * num_gl_cmds # FIXME: replace `4` with `sizeof<s4>` once `sizeof` is implemented for GraphViz
+    type: gl_cmds_list
+  anim_names:
+    value: |
+      ['stand', 'run', 'attack', 'pain1', 'pain2', 'pain3', 'jump', 'flip',
+      'salute', 'taunt', 'wave', 'point', 'crstnd', 'crwalk', 'crattak',
+      'crpain', 'crdeath', 'death1', 'death2', 'death3']
+  anim_start_indexes:
```
generalmimon, line 151:
```suggestion
  anim_start_indices:
```

Please don't forget to change the occurrence in `/doc` too:

https://github.com/kaitai-io/kaitai_struct_formats/blob/2f1b418810ff416aaab92efef57d9a6328d4f2d0/game/quake2_md2.ksy#L41-L42

game/quake2_md2.ksy
```diff
@@ -0,0 +1,429 @@
...
+  triangles:
+    pos: ofs_triangles
+    type: triangle
+    repeat: expr
+    repeat-expr: num_triangles
+  frames:
+    pos: ofs_frames
+    type: frame
+    repeat: expr
+    repeat-expr: num_frames
```
generalmimon, line 139: The use of `bytes_per_frame` is missing here:

```suggestion
  frames:
    pos: ofs_frames
    size: bytes_per_frame
    type: frame
    repeat: expr
    repeat-expr: num_frames
```

game/quake2_md2.ksy
```diff
@@ -0,0 +1,429 @@
...
+      z:
+        value: z_compressed * _parent._parent.scale.z + _parent._parent.translate.z
+  vertex:
+    seq:
+      - id: position
+        type: compressed_vec
+      - id: normal_index
+        type: u1
+        doc: |
+          `normal = anorms_table[normal_index]`
```
generalmimon, line 382: Given that you included the `anorms_table` definition, you can actually resolve the `normal` vector now:

```suggestion
      - id: normal_index
        type: u1
    instances:
      normal:
        value: _root.anorms_table[normal_index]
```

jdjakub: Excellent! I was hoping this was possible.

game/quake2_md2.ksy
```diff
@@ -0,0 +1,429 @@
...
+        [-0.425325, -0.688191, -0.587785], 
+        [-0.587785, -0.425325, -0.688191], 
+        [-0.688191, -0.587785, -0.425325] ]
+types:
+  tex_point:
+    seq:
+      - id: u_px
+        type: u2
+      - id: v_px
+        type: u2
```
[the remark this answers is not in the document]
jdjakub: Fair enough.

game/quake2_md2.ksy
```diff
@@ -0,0 +1,433 @@
...
+      4, 5, 6, 6, 8]
+  anorms_table:
+    doc-ref: |
+      https://github.com/skullernet/q2pro/blob/f4faabd/src/common/math.c#L80
+      from Quake anorms.h
+    value: |
+      [
+        [-0.525731, 0.000000, 0.850651], 
+        [-0.442863, 0.238856, 0.864188], 
+        [-0.295242, 0.000000, 0.955423],
```
generalmimon, line 170: There should be no trailing whitespace (on _any_ line; I've only highlighted some of them). It's recommended to set up your editor to honor the [project `.editorconfig`](https://github.com/kaitai-io/kaitai_struct_formats/blob/1eae5145325bb992e20dae838dd68d04814dda46/.editorconfig#L8) (see https://editorconfig.org/ for explanation).

If you use VS Code, it doesn't support EditorConfig natively, but there is an official [_EditorConfig_](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig) extension; for other editors, see https://editorconfig.org/#download.

jdjakub: Cool, wasn't aware of that. I'm actually using the GitHub web editor for these small changes.

jdjakub:
Hello! I think this is mergeable now.

PR #621, opened 2022-08-22
Fix dbf.ksy

alexander982:
More flexible and detailed dbf file parser/analyser. Do not tested on ksc - only in Web IDE

### DIFF
--- a/database/dbf.ksy
+++ b/database/dbf.ksy
@@ -28,8 +28,10 @@ seq:
   - id: header1
     type: header1
   - id: header2
-    size: header1.len_header - 12
+    size: header1.len_header - 13
     type: header2
+  - id: header_terminator
+    contents: [0x0D]
   - id: records
     size: header1.len_record
     repeat: expr
@@ -65,8 +67,7 @@ types:
         type: header_dbase_7
       - id: fields
         type: field
-        repeat: expr
-        repeat-expr: 11
+        repeat: eos
   header_dbase_3:
     seq:
       - id: reserved1
--- a/database/dbf.ksy
+++ b/database/dbf.ksy
@@ -130,7 +130,8 @@ types:
         type: str
         encoding: ASCII
         size: _root.header2.fields[_index].length
-        repeat: eos
+        repeat: expr
+        repeat-expr: _root.header2.fields.size
 enums:
   delete_state:
     0x2a: true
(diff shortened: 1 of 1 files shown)


generalmimon (changes requested):

database/dbf.ksy
```diff
@@ -119,3 +121,25 @@ types:
...
+        repeat-expr: len
+  record:
+    seq:
+      - id: deleted
+        type: u1
+        enum: delete_state
+      - id: record_fields
+        type: record_data(_root.header2.fields[_index].length)
+        repeat: expr
+        repeat-expr: _root.header2.fields.size
```
generalmimon, line 141: I meant like this:

```suggestion
      - id: record_fields
        size: _root.header2.fields[_index].length
        repeat: expr
        repeat-expr: _root.header2.fields.size
```

and the `record_data` type is longer needed.

alexander982:
Yes. The fact that "type: u1" and "size" do not work together was a surprise for me.

generalmimon:
@alexander982:

> Yes. The fact that "type: u1" and "size" do not work together was a surprise for me.

There is no need for `type: u1` here, `size` alone is enough to create a byte array - see https://doc.kaitai.io/user_guide.html#fixed-size-struct:

> With that out of the way, we use a `seq` element with an array (ordered sequence of elements) in it to describe which attributes this structure consists of. Every attribute includes several keys, namely:
> 
> - `id` is used to give the attribute a name
> - `type` designates the attribute type:
>   - **no type means that we’re dealing with just a raw byte array; `size` is to be used to designate number of bytes in this array**

There is actually a semantic difference between what you get using

- just `size`: a raw byte array or byte string in some languages, i.e. the same data type that you can use to feed raw data to a Kaitai Struct parser at the beginning; target languages have often built-in methods to convert native types, e.g. numbers and character strings, exactly to/from this byte array/byte string data type (such methods may be called `from_bytes`/`fromBytes` and `to_bytes`/`toBytes`);
- `type: u1` + `repeat: expr`: this is a *list*/*collection* of 0-255 integers, something meaningful in this final form, i.e. each such integer value would be meaningful on its own (every number can be for example an array index, a number of objects of some kind, etc.)

alexander982:
Thanks. It is more clear for me now

generalmimon (approved):
@alexander982 Looks good to me, thanks for the contribution!

PR #625, opened 2022-08-30
Implemented CRC checksums in KS

KOLANICH:


KOLANICH:
@armijnhemel, @generalmimon, @dgelessus

generalmimon:
@KOLANICH I'm amazed at how much work you put in these, and it quite well showcases the capabilities of the Kaitai Struct language. It's also interesting to see how many CRC variants are in use.

However, I personally don't think this is ever going to be merged, sorry. It's just not a good fit for this repo. But you can certainly have your own repository for these.

This may very well qualify for https://github.com/kaitai-io/kaitai_struct/issues/165 - keep these specs in a separate repository, and then (when it's implemented) we may add a link to there from https://formats.kaitai.io/ so people can find it.

KOLANICH:
>I'm amazed at how much work you put in these

Well, the most of the files are auto-generated from the supplied TSV file using the supplied python scripts. What has really took the most of effort is to fill that file from different sources.

>However, I personally don't think this is ever going to be merged, sorry. It's just not a good fit for this repo.

1. I know that the impl of CRC in KS is inefficient and eats memory, but to be honest, the efficiency can be improved by implementing certain KSC features (https://github.com/kaitai-io/kaitai_struct/issues/88, https://github.com/kaitai-io/kaitai_struct/issues/991)

2. There are quite some specs in the repo using CRC checksums. Currently they don't validate them. Having these specs here would enable using them from within those specs to actually do validation.

3. KS as a language is the tool to specify formats. Making the generated code efficient is the job of compiler, so inefficiency of implementing certain things in KS shouldn't be an obstacle to merging them.

4. KS currently has no functionality to use specs from external repos https://github.com/kaitai-io/kaitai_struct/issues/71

PR #627, opened 2022-09-19
rpm: add new tags 5099/5100/5101/5102

armijnhemel:
This commit adds some of the new RPM tags that were added in the past 6 months.

### DIFF
--- a/archive/rpm.ksy
+++ b/archive/rpm.ksy
@@ -454,7 +454,7 @@ enums:
       -orig-id: RPMSIGTAG_RESERVEDSPACE
       doc: Space reserved for signatures
   header_tags:
-    # Tags from [lib/rpmtag.h](https://github.com/rpm-software-management/rpm/blob/650ba79f/include/rpm/rpmtag.h).
+    # Tags from [lib/rpmtag.h](https://github.com/rpm-software-management/rpm/blob/9ed9d3fc/include/rpm/rpmtag.h).
     # This includes (almost) all tags. Some have `_unimplemented`, `_internal`
     # or `_obsolete` suffix (if more than one applies, the first applicable in
     # this order is used).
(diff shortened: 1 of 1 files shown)


archive/rpm.ksy
```diff
@@ -1448,6 +1448,18 @@ enums:
...
+      -orig-id: RPMTAG_SPEC
+    5100:
+      id: translation_url
+      -orig-id: RPMTAG_TRANSLATIONURL
+    5101:
+      id: upstream_releases
+      -orig-id: RPMTAG_UPSTREAMRELEASES
+    5102:
+      id: source_license
+      -orig-id: RPMTAG_SOURCELICENSE
```
generalmimon, line 1462: This tag is marked as *internal* (https://github.com/rpm-software-management/rpm/blob/9ed9d3fc/include/rpm/rpmtag.h#L383):

```c
    RPMTAG_SOURCELICENSE	= 5102, /* internal */
```

I used to declare such state in the tag identifier, according to the following description:

https://github.com/kaitai-io/kaitai_struct_formats/blob/c34d640bb8680b856ee2389f24f93e7dac8f1c28/archive/rpm.ksy#L458-L460

For example:

https://github.com/kaitai-io/kaitai_struct_formats/blob/c34d640bb8680b856ee2389f24f93e7dac8f1c28/archive/rpm.ksy#L545-L547

@armijnhemel Do you think it makes sense? If so, this would be:

```suggestion
    5102:
      id: source_license_internal
      -orig-id: RPMTAG_SOURCELICENSE
```

armijnhemel: Personally I would not use `_internal` (and just leave it out) but consistency is important. So feel free to replace!

generalmimon: > Personally I would not use `_internal` (and just leave it out)

OK, so we may drop these later.

generalmimon:

PR #628, opened 2022-09-29
Added archive format for Game "4x4 Evolution"

wattostudios:


KOLANICH (commented):
1. `encoding: UTF-8` can be in `meta`, and also adviced to be lowercase
2. `filename` -> `file_name` is IMHO better
3. `str`, `terminator: 0`: maybe just `strz`?

wattostudios:
Updated as suggested.

For reference/feedback:
1. `UTF-8` in uppercase is written all through the user guide at [](https://doc.kaitai.io/user_guide.html) so would be great to update that if that's not the recommended format
2. The documentation doesn't indicate that `type: strz` can be used in conjunction with `size: 80` . The section where you describe the combination of `terminator` + `size` only mentions `type:str` , and even though you could argue that it's implied for `strz` as well, I'd argue that if you have an explicit part of the document that deals with `terminator` + `size` combinations, `strz` should be listed as being compatible.

KOLANICH (approved):
LGTM

game/4x4_evolution_pod.ksy
```diff
@@ -0,0 +1,40 @@
+meta:
+  id: game_4x4_evolution_pod
+  file-extension: pod
+  application: 4x4 Evolution
+  license: CC0-1.0
+  endian: le
+  encoding: utf-8
```
generalmimon, line 7: I know that @KOLANICH advised to switch to lowercase, but uppercase `UTF-8` is actually more common in the format gallery.

game/4x4_evolution_pod.ksy
```diff
@@ -0,0 +1,40 @@
...
+doc-ref: 'https://www.watto.org/specs.html?specs=Archive_POD'
+seq:
+  - id: header
+    type: archive_header
+  - id: details_directory
+    type: details_directory
```
generalmimon, line 13: The type should not be called `details_directory`, because that's not what the type represents. It represents a single file, so naming it a "directory" is confusing. Using `details_directory` as the name of the array is somewhat less confusing, but it doesn't comply with the naming convention specified [in the style guide](https://doc.kaitai.io/ksy_style_guide.html#attr-id):

> - For an array of objects (i.e. with `repeat: something`), use plural form — e.g. `files`, `transactions`.

So I'd personally go for `files` and the type `file`.

game/4x4_evolution_pod.ksy
```diff
@@ -0,0 +1,40 @@
...
+        size: 80
+        doc: A description for the archive
+  details_directory:
+    seq:
+      - id: file_name
+        type: strz
+        size: 32
+      - id: length
+        type: u4
+        doc: The length of the file data
```
generalmimon, line 33: These comments are probably obvious and redundant.

generalmimon (changes requested):

KOLANICH:
I can bulk-convert the field in the whole gallery to lowercase 8f you want :)

Ithamar:
Just a side-note, this format is used by a whole bunch of `Terminal Reality, Inc`'s games (and there are several versions of the format too). Maybe it would make sense to use call it `terminalreality_pod.ksy`, as there are totally different POD formats out there from other companies.

PR #630, opened 2022-10-17
luks.ksy: Added partial support for luks2

Sebastian2023:
Most distributions nowadays use luks2. I have implemented support for it. But not completely, since the JSON areas and (thus also the keyslot areas referenced in it) are difficult to parse in Kaitai.

### DIFF
--- a/filesystem/luks.ksy
+++ b/filesystem/luks.ksy
@@ -11,45 +11,120 @@ meta:
   encoding: ASCII
   endian: be
 doc: |
-  Linux Unified Key Setup (LUKS) is a format specification for storing disk
+  Linux Unified Key Setup (LUKS2) is a format specification for storing disk
   encryption parameters and up to 8 user keys (which can unlock the master key).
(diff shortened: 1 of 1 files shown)


filesystem/luks.ksy
```diff
@@ -11,45 +11,120 @@ meta:
...
+          max: 0x02
+      # LUKS1
       - id: cipher_name_specification
         type: str
         size: 32
+        if: version2 == 0x01
       - id: cipher_mode_specification
         type: str
         size: 32
+        if: version2 == 0x01
```
generalmimon, line 58: You should extract this bunch of consecutive fields specific to LUKS1/LUKS2 to new types, which'll allow you to drop `if: version2 == {0x01,0x02}` on the individual fields - you will use an `if` only once when you include these types in the header.

This has all the benefits of better readability, maintainability and easier (and less error-prone) usage of generated parsers in users' applications.

generalmimon (changes requested):

KOLANICH:
>JSON areas and (thus also the keyslot areas referenced in it) are difficult to parse due to Kaitai language constraints.

Look at #636, that format also relies on JSON.

PR #633, opened 2022-11-05
games/renderware_binary_stream: add atomic struct definition

Ithamar:
Atomic is commonly the top level section in DFF files. This makes geometry is parsable in many older RenderWare game files.

### DIFF
--- a/game/renderware_binary_stream.ksy
+++ b/game/renderware_binary_stream.ksy
@@ -25,6 +25,7 @@ seq:
         sections::geometry_list: list_with_header
         sections::texture_dictionary: list_with_header
         sections::texture_native: list_with_header
+        sections::atomic: list_with_header
 instances:
   version:
     value: 'library_id_stamp & 0xFFFF0000 != 0 ? (library_id_stamp >> 14 & 0x3FF00) + 0x30000 | (library_id_stamp >> 16 & 0x3F) : library_id_stamp << 8'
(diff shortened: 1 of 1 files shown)


generalmimon (approved):
@Ithamar LGTM, if you don't want to analyze `flags` (https://gtamods.com/wiki/Atomic_(RW_Section)#Flags). The PR is good even without that, but it would be a nice addition.

I think of value `instances` that test the relevant bits (e.g. `instances/is_something/value: flags & 0x04 != 0`).

KOLANICH:
A brief reminder: IMHO it makes sense to implement flags as a struct of bit fields, not as a struct of instances where values are manually shifted, ANDed, and compared to 1 or 0.

generalmimon:
@KOLANICH:

> IMHO it makes sense to implement flags as a struct of bit fields

I'd generally agree, but it's not a hard-and-fast rule for me. Using [bit-sized integers](https://doc.kaitai.io/user_guide.html#_bit_sized_integers) is obviously the most declarative way (which is a good thing), which allows it to work best for [serialization](https://doc.kaitai.io/serialization.html) as well (you can directly set the fields that will be written, as opposed to having to manually set the bits of `flags` - which also forces you to duplicate the bit shifts/masks in the application code - when the value instances would be used).

But if the flags are too sparse (which is this case, because there are only 2 known bits in 32), I don't mind using the value instances to unpack the 4-byte `flags` integer. The benefit is that all the bits are still accessible by their original bit masks.

Maybe it's still better to use bit integers even for sparse flags, I'm not sure.

Ithamar:
I've added the flags now too, using bit-sized integers. I tend to use the raw ksy files as format documentation, and bit fields make more sense then masks/shifts, though it does require to understand kaitai's bit field endianness magic ;)

game/renderware_binary_stream.ksy
```diff
@@ -54,12 +56,28 @@ types:
...
+      - id: frame_index
+        type: u4
+      - id: geometry_index
+        type: u4
+      - id: flag_render
+        type: b1
+      - type: b1
+      - id: flag_collision_test
+        type: b1
+      - type: b24
```
generalmimon, line 78: These bits must add up to `32`, for example:

```suggestion
      - id: flag_render
        type: b1
      - type: b1
      - id: flag_collision_test
        type: b1
      - type: b29
```

It kind of _works_ even if they don't, but there will be an alignment of the position to a byte boundary just before the following `unused` field, so the 5 bits that you don't parse will be skipped and not accessible in any field.

generalmimon (approved):
@Ithamar Thanks!

PR #640, opened 2023-02-17
Make all SQLite database pages parseable

pepijnve:
This PR improves a couple of things in the SQLite3 format:
- The string serial type is now conditionally parsed
- Database pages beyond page 1 are now declared as instances which makes them parseable

With these tweaks the Kaitai web IDE and visualizer can be used to browse through the entire database contents.


ddevienne:
Just a few remarks. There is at least one other type of page, ptr ones used with incremental vacuuming. There’s also a special page instance (not type) around 1GB with some bytes excluded, because used for locks. Maybe leave TODOs about them?

pepijnve:
@ddevienne I've added the necessary declarations to handle ptrmap pages and the lock-byte page. The only remaining unimplemented bit is that free pages will be interpreted as btree pages even though they're not used. I don't think I can express the necessary logic to check the free list in Kaitai's declarative language.

pepijnve:
My initial PR was probably a bit premature. I've continued working and refining the PR and it's now at the point where you can browse the database either as a flat list of pages (with the caveat that free pages and cell payload overflow pages are misclassified as b-tree pages) or hierarchically by following page number 'pointers'. I've squashed this down to one commit; the separate ones didn't make much sense anymore.

database/sqlite3.ksy
```diff
@@ -32,213 +32,431 @@ doc: |
...
-    doc: The "Application ID" set by PRAGMA application_id.
-  - id: reserved
-    size: 20
-  - id: version_valid_for
-    type: u4
-  - id: sqlite_version_number
-    type: u4
-  - id: root_page
-    type: btree_page
+  - id: header
```
milahu, line 35: header_string

https://www.sqlite.org/fileformat.html

>The header string: "SQLite format 3\000"

database/sqlite3.ksy
```diff
@@ -32,213 +32,431 @@ doc: |
...
+        # TODO: Free pages and cell overflow pages are incorrectly interpreted as btree pages
+        # This is unfortunate, but unavoidable since there's no way to recognize these types at
+        # this point in the parser.
+        2: btree_page(_index + 1)
+    pos: 0
+    size: header.page_size
+    repeat: expr
+    repeat-expr: header.page_count
 types:
+  database_header:
```
milahu, line 53: any ideas where these come from? (python)

```
>>> db.header.j
819
>>> db.header.m
489
```

database/sqlite3.ksy
```diff
@@ -32,213 +32,431 @@ doc: |
   (`root_page`) is always "btree_page".
```
milahu, line 32: `root_page` is still referenced in the docstring

database/sqlite3.ksy
```diff
@@ -32,213 +32,431 @@ doc: |
...
-  cell_table_leaf:
+            btree_page_type::table_leaf: table_leaf_cell
+            btree_page_type::table_interior: table_interior_cell
+            btree_page_type::index_leaf: index_leaf_cell
+            btree_page_type::index_interior: index_interior_cell
+  table_leaf_cell:
     doc-ref: 'https://www.sqlite.org/fileformat.html#b_tree_pages'
     seq:
-      - id: len_payload
+      - id: p
```
milahu, line 261: `p` is cryptic ... why not keep `len_payload`? or `payload_size`?

https://www.sqlite.org/fileformat.html

> Table B-Tree Leaf Cell (header 0x0d):
> 
> - A varint which is the total number of bytes of payload, including any overflow
> - A varint which is the integer key, a.k.a. "[rowid](https://www.sqlite.org/lang_createtable.html#rowid)"
> - The initial portion of the payload that does not spill to overflow pages.
> - A 4-byte big-endian integer page number for the first page of the overflow page list - omitted if all payload fits on the b-tree page.

database/sqlite3.ksy
```diff
@@ -32,213 +32,431 @@ doc: |
...
       - id: code
         type: vlq_base128_be
     instances:
-      is_blob:
-        value: 'code.value >= 12 and (code.value % 2 == 0)'
-      is_string:
-        value: 'code.value >= 13 and (code.value % 2 == 1)'
-      len_content:
+      value_type:
+        value: 'code.value >= 12 ? ((code.value % 2 == 0) ? 12 : 13 + _root.header.text_encoding - 1) : code.value'
```
milahu, line 326: rename `code` to `raw_value`?
rename `value_type` to `type`?

raw_value = type + size

```
>>> value_type.code.value
23
>>> value_type.value_type
13
```

```
>>> serial_type.code.value
23
>>> serial_type.value_type
13
```

milahu: can we have enum for type?

```
0: null
# big-endian twos-complement integers
1: int8
2: int16
3: int24
4: int32
5: int48
6: int64
7: float64 # big-endian IEEE 754-2008 64-bit float
8: int0 # the integer 0
9: int1 # the integer 1
# Reserved for internal use
10: internal10
11: internal11
12: blob
13: string
```

database/sqlite3.ksy
```diff
@@ -32,213 +32,431 @@ doc: |
...
+        type: u4
+        doc: The "user version" as read and set by the user_version pragma.
+      - id: is_incremental_vacuum
+        type: u4
+        doc: True (non-zero) for incremental-vacuum mode. False (zero) otherwise.
+      - id: application_id
+        type: u4
+        doc: The "Application ID" set by PRAGMA application_id.
+      - id: reserved
+        size: 20
```
milahu, line 118:
```diff
-db.header.reserved
+db.header.reserved_header_bytes
 db.header.reserved_space
```

?

database/sqlite3.ksy
```diff
@@ -32,213 +32,431 @@ doc: |
...
+      call_content_area_start:
+        value: 'cell_content_area_start_raw == 0 ? 65536 : cell_content_area_start_raw'
+      cell_content_area:
+        pos: call_content_area_start
+        size: _root.header.usable_size - call_content_area_start
+      reserved_area:
+        pos: _root.header.page_size - _root.header.reserved_space
+        size-eos: true
+        if: _root.header.reserved_space != 0
+  cell_pointer:
```
milahu, line 244: fails to read `cell_pointer` from a large database

```py
    class CellPointer(KaitaiStruct):
        def __init__(self, _io, _parent=None, _root=None):
            self._io = _io
            self._parent = _parent
            self._root = _root if _root else self
            self._read()

        def _read(self):
            self.content_offset = self._io.read_u2be()
```

```
  File "pysqlite3/connection.py", line 228, in _tables
    page = self._db.pages[0] # FIXME this fails on large datbase
  File "pysqlite3/parser/sqlite3.py", line 842, in pages
    Sqlite3.BtreePage((i + 1), _io__raw__m_pages, self, self._root)
  File "pysqlite3/parser/sqlite3.py", line 115, in __init__
    self._read()
  File "pysqlite3/parser/sqlite3.py", line 137, in _read
    self.cells.append(Sqlite3.CellPointer(self._io, self, self._root))
  File "pysqlite3/parser/sqlite3.py", line 697, in __init__
    self._read()
  File "pysqlite3/parser/sqlite3.py", line 700, in _read
    self.content_offset = self._io.read_u2be()
  File "kaitaistruct.py", line 173, in read_u2be
    return KaitaiStream.packer_u2be.unpack(self.read_bytes(2))[0]
  File "kaitaistruct.py", line 326, in read_bytes
    raise EOFError(
EOFError: requested 2 bytes, but only 0 bytes available
```

milahu: one problem is that `db.pages` is not made for random-access
when i read `db.pages[0]` it will loop through ALL pages

```py
class Sqlite3(KaitaiStruct):
    # ...
    @property
    def pages(self):
        if hasattr(self, "_m_pages"):
            return self._m_pages

        _pos = self._io.pos()
        self._io.seek(0)
        self._raw__m_pages = []
        self._m_pages = []
        for i in range(self.header.page_count):
```

the other problem is that `db.root_pages` is missing
when i follow `right_ptr` then i get a list of all `root_pages`
until `right_ptr.page_number == db.header.page_count`

quickfix for python in https://github.com/milahu/pysqlite3/commit/1256408d5479a4902df17728a2ae6f355fcc44e9

database/sqlite3.ksy
pepijnve: paragraph updated

database/sqlite3.ksy
pepijnve: Thos somewhat cryptic single letter variables are all references to the documentation.
`J` is defined at https://www.sqlite.org/fileformat2.html#pointer_map_or_ptrmap_pages

> In a database with ptrmap pages, the first ptrmap page is page 2. A ptrmap page consists of an array of 5-byte entries. Let J be the number of 5-byte entries that will fit in the usable space of a page. (In other words, J=U/5.) The first ptrmap page will contain back pointer information for pages 3 through J+2, inclusive. The second pointer map page will be on page J+3 and that ptrmap page will provide back pointer information for pages J+4 through 2*J+3 inclusive. And so forth for the entire database file.

`M` is described in https://www.sqlite.org/fileformat2.html#b_tree_pages

> The amount of payload that spills onto overflow pages also depends on the page type. For the following computations, let U be the usable size of a database page, the total page size less the reserved space at the end of each page. And let P be the payload size. In the following, symbol X represents the maximum amount of payload that can be stored directly on the b-tree page without spilling onto an overflow page and symbol M represents the minimum amount of payload that must be stored on the btree page before spilling is allowed.

The paragraphs after that describe how to calculate the values.

database/sqlite3.ksy
pepijnve: See my earlier comment. This was also also a reference to the formulas in the file format documentation. After restructuring the format a couple of times, retaining the short name doesn't make much sense anymore. I've changed it to `payload_size`.

database/sqlite3.ksy
pepijnve: Done

database/sqlite3.ksy
pepijnve: Sure. I changed
`reserved_space` -> `reserved_bytes_per_page`
`reserved` -> `reserved_header_bytes`
to make it clearer what these two value are.

database/sqlite3.ksy
pepijnve: I'll have to experiment a bit for this one. Not sure what's going on with the right page pointer. There might simply be an error in the definition.

I currently declared the btree_page_pointer as
```
  btree_page_pointer:
    seq:
      - id: page_number
        type: u4
    instances:
      page:
        io: _root._io
        pos: (page_number - 1) * _root.header.page_size
        size: _root.header.page_size
        type: btree_page(page_number)
        if: page_number != 0
```

What you would actually want to do here is describe that this is a pointer to `db.pages[page_number]`. I'm not sure if there's a way to describe that in Kaitai Struct. I added the `page` instance mainly for my own convenience when browsing through the database file in KSV.

I'm not sure if there's much I can do about the generated python code. The `pages` instances section doesn't really require unmarshalling all database pages up front. I don't know if there's a definition level concept that can inform the Python code generator to produce a lazy collection instead of an eager one.

database/sqlite3.ksy
pepijnve: > Ideally use the same name-format for all sizes/lengths: payload.header_length, page.cell_count, ...

I've reworked the definition in an attempt to keep things consistent. `count` is a number of things, `size` is byte size, `length` is only used for blob and string now. I personally think it's clearer to keep count and size distinct.

database/sqlite3.ksy
pepijnve: Note that `db.root_pages` doesn't really exist at the level of the database format. You need to query the `sqlite_schema` table (the table b-tree with page 1 as root)  to construct that list if you want it.

database/sqlite3.ksy
pepijnve: I ended up with `num_...` instead to silence the Kaitai compiler warnings. Compiles without warnings now.

generalmimon: @pepijnve:

> I ended up with `num_...` instead to silence the Kaitai compiler warnings.

That's correct. It's preferred to use the naming convention as described in https://doc.kaitai.io/ksy_style_guide.html#attr-id. Before your changes, the spec tried to stick to them (`num_cells`, `ofs_cells`, `ofs_body`, `len_payload`, `len_page`), so I'm not sure why you felt the need to change them.

> Compiles without warnings now.

You shouldn't be only fixing the warnings (they are not complete, and don't work all the time) - please at least check out https://doc.kaitai.io/ksy_style_guide.html#attr-id and make sure you follow the recommendations. For example, offsets should be called `ofs_{structure}` - the compiler doesn't generate warnings for this yet.

pepijnve: I tested the produced code with the Java backend on a database with a table b-tree that has both interior and leaf pages. Seems to work correctly. The root page is an interior page and has a non-null right pointer. If I follow that I end up at a leaf page with a null right pointer.

One pitfall to be aware of here is that page numbers are 1-based in the file format, while the pages list is 0-based?

In the Java implementation, the Kaitai compiler also generates an eager `ArrayList` for the `pages` instance. Kind of useless for large files.

generalmimon: > In the Java implementation, the Kaitai compiler also generates an eager `ArrayList` for the `pages` instance. Kind of useless for large files.

Well, it works the way you implemented it. The `pages` instance itself is lazily parsed, but that doesn't mean the individual pages will automatically be lazy too. In this case, it's definitely possible to make them lazy, only there isn't the magic switch you'd like.

You can do for example something like this (each page `body` will be parsed lazily):

```ksy
instances:
  pages:
    type: page(_index + 1, header.page_size * _index)
    repeat: expr
    repeat-expr: header.page_count # note: this should be called `num_pages` per KS conventions

types:
  page:
    params:
      - id: page_number
        type: s4
      - id: ofs_body
        type: s4
    instances:
      body:
        pos: ofs_body
        size: _parent.header.page_size
        # type: ...
```

database/sqlite3.ksy
milahu: > In a database with ptrmap pages, the first ptrmap page is page 2. A ptrmap page consists of an array of 5-byte entries. Let J be the number of 5-byte entries that will fit in the usable space of a page.

J = ptrmap_entry_count_max

> symbol X represents the maximum amount of payload that can be stored directly on the b-tree page without spilling onto an overflow page

> symbol M represents the minimum amount of payload that must be stored on the btree page before spilling is allowed.

X = overflow_payload_size_max
M = overflow_payload_size_min

generalmimon: I agree with @milahu - the single-letter "names" are incomprehensible, and it doesn't matter whether any credible source uses them or not. When using a `.ksy` specification, there should be no need to have the original SQLite specification on another monitor to decrypt the nonsensical names.

@pepijnve You can use the `-orig-id` key to keep a reference to the original identifiers if you want, but the main `id` names should be as clear and consistent as possible. See https://doc.kaitai.io/ksy_style_guide.html#transcribing:

> Use `-orig-id` key to specify original names of fields for purposes of maintaining a reference link to parts of original spec, but, otherwise, feel free to use a more consistent and language-neutral approach in naming attributes and types.

database/sqlite3.ksy
milahu: yes, the parser should be minimal
ie, derive values only when they are needed to continue parsing

ideally kaitai-struct-compiler would have a `debug` option
to generate a more verbose (and more expensive) parse tree

for completeness: 

<details>
<summary>
table_pages vs root_pages
</summary>

> You need to query the `sqlite_schema` table

these are `table_pages`

```py
    def _table_page_numbers(self):
        """
        get the page numbers of all tables,
        excluding the sqlite_schema table in page 1
        """
        page = self._db.root_pages[0]
        assert page.page_type.value == 0x0D  # Table B-Tree Leaf Cell
        for cell in page.cells:
            yield cell.content.payload.values[3].value
```

`root_pages` are more: `table_pages` plus `index_pages` (etc)

```py
    def _root_page_numbers(self):
        """
        get the page numbers of all root pages,
        excluding page 1
        """
        if self._db.header.page_count == 1:
            return []
        i = 1 # page 2
        i_max = self._db.header.page_count - 1
        while i < i_max: # `i == i_max` means "end of file"
            page = self._db.pages[i]
            yield page.page_number
            # go to next root page
            if hasattr(page, "right_ptr"):
                i = page.right_ptr.page_number - 1
            else:
                i = i + 1
```

</details>

database/sqlite3.ksy
pepijnve: > so I'm not sure why you felt the need to change them.

Potayto, potahto. What's legible to one person is gobbledygook to someone else. See for instance Microsoft's Hungarian notation. Personally I find a suffix stlye (`foobar_offset`) easier to read than a prefix style (`ofs_foobar`). The former I can read in the order it's written, the other asks for interpretation and flipping the word order.

I made this more complete definition for a SQLite internals sessions for my colleagues. It was mainly written with that in mind and since it was more complete and correct than what was already I simply made the PR with what I had at that point.
No objections to sticking to the conventions. Should be in line with what you asked for now.

database/sqlite3.ksy
```diff
@@ -27,218 +27,476 @@ doc: |
...
+        doc: The "usable size" of a database page
+      overflow_min_payload_size:
+        value: ((usable_size-12)*32/255)-23
+        doc: The minimum amount of inline b-tree cell payload
+      table_max_overflow_payload_size:
+        value: usable_size - 35
+        doc: The maximum amount of inline table b-tree cell payload
+      index_max_overflow_payload_size:
+        value: ((usable_size-12)*64/255)-23
+        doc: The maximum amount of inline index b-tree cell payload
```
milahu, line 139: docstrings should be more precise regarding overflow

i would just copy the original descriptions

```diff
       overflow_min_payload_size:
         value: ((usable_size-12)*32/255)-23
-        doc: The minimum amount of inline b-tree cell payload
+        doc: The minimum amount of payload that must be stored on the btree page before spilling is allowed
       table_max_overflow_payload_size:
         value: usable_size - 35
-        doc: The maximum amount of inline table b-tree cell payload
+        doc: The maximum amount of payload that can be stored directly on the b-tree page without spilling onto an overflow page. Value for table page
       index_max_overflow_payload_size:
         value: ((usable_size-12)*64/255)-23
-        doc: The maximum amount of inline index b-tree cell payload
+        doc: The maximum amount of payload that can be stored directly on the b-tree page without spilling onto an overflow page. Value for index page
```

database/sqlite3.ksy
```diff
@@ -27,218 +27,476 @@ doc: |
...
+      lock_byte_page_index:
+        value: '1073741824 / page_size'
+      ptrmap_max_num_entries:
+        value: usable_size/5
+        doc: The number of ptrmap entries per ptrmap page
```
milahu, line 144: the maximum number

database/sqlite3.ksy
pepijnve: I'm at the point where I've put way much more time in this than I was willing to. Feel free to take over from here and tinker on the definition.

database/sqlite3.ksy
milahu: > I bumped into a Kaitai limit here where I couldn't source the string encoding from an expression.

can you elaborate?

```
# line 465
  serial:
    # ...
    # Value is a string in the text encoding and (N-13)/2 bytes in length. The nul terminator is
    # not stored.
    13: string_utf8
    14: string_utf16_le
    15: string_utf16_be
```


at least add comment for 14 and 15

ideally avoid this workaround

database/sqlite3.ksy
milahu: thanks for your work : )

database/sqlite3.ksy
pepijnve: I had put the string value of the encoding (`UTF-8`, `UTF16-LE`, or `UTF-16BE`) in an instance attribute in the header. There didn't seem to be a way to reference that from the `encoding:` attribute of an `str` typed value. As a consequence I needed to have three distinct string types and a way to reference them. That's what those three enum constants are for.

pepijnve: Thinking about this a bit more, you could get rid of these `string_...` enum values, replace `text_encoding` value in the header with another enum, and then switch on that in a single string type. I think you'll still need the three string types to actually be able to specify the encoding for Kaitai.

The resulting tree is going to be a bit uglier though. I was kind of optimising things for display in the KSV tree view and cross-referencing with the SQLite technical documentation.

PR #642, opened 2023-03-02
java_class: add tags since Java 9, add version checks

KOLANICH:


armijnhemel:
Why references to openjdk instead of the official spec? https://docs.oracle.com/javase/specs/jvms/se19/html/jvms-4.html

KOLANICH:
Fixed

executable/java_class.ksy
```diff
@@ -230,6 +238,19 @@ types:
         type: u2
       - id: name_and_type_index
         type: u2
+  class_cp_module_package:
+    doc-ref:
+      - https://docs.oracle.com/javase/specs/jvms/se19/html/jvms-4.html#jvms-4.4.11
+      - https://docs.oracle.com/javase/specs/jvms/se19/html/jvms-4.html#jvms-4.4.12
```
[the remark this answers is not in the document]
generalmimon: > Perhaps we can also work out what `version_major` value corresponds to the format version that added the `CONSTANT_Module` and `CONSTANT_Package` tags and document it in the spec?

@KOLANICH Thanks to your reference you added in `doc-ref`, this turns out to be easier than I thought ([`classFileParser.cpp:355-360`](https://github.com/openjdk/jdk/blob/72de24e59a80a38ea4ea6a8a3f966f555987ac86/src/hotspot/share/classfile/classFileParser.cpp#L355-L360)):

```cpp
      case JVM_CONSTANT_Module:
      case JVM_CONSTANT_Package: {
        // Record that an error occurred in these two cases but keep parsing so
        // that ACC_Module can be checked for in the access_flags.  Need to
        // throw NoClassDefFoundError in that case.
        if (_major_version >= JAVA_9_VERSION) {
```

[`classFileParser.cpp:121`](https://github.com/openjdk/jdk/blob/72de24e59a80a38ea4ea6a8a3f966f555987ac86/src/hotspot/share/classfile/classFileParser.cpp#L121):

```cpp
#define JAVA_9_VERSION                    53
```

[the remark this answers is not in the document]
armijnhemel: > Perhaps we can also work out what `version_major` value corresponds to the format version that added the `CONSTANT_Module` and `CONSTANT_Package` tags and document it in the spec?

That would be table 4.4-B in the specification.

executable/java_class.ksy
```diff
@@ -83,28 +87,82 @@ types:
...
+          id: module
+          -orig-id: CONSTANT_Module
+        20:
+          id: package
+          -orig-id: CONSTANT_Package
+
+  version_guard:
+    doc: |
+      1.0.2 is so ancient, that assummed as granted, and further versions add features only on major releases
+      see
```
generalmimon, line 154: Are we supposed to see something?

executable/java_class.ksy
```diff
@@ -83,28 +87,82 @@ types:
...
+        19:
+          id: module
+          -orig-id: CONSTANT_Module
+        20:
+          id: package
+          -orig-id: CONSTANT_Package
+
+  version_guard:
+    doc: |
+      1.0.2 is so ancient, that assummed as granted, and further versions add features only on major releases
```
generalmimon, line 153:
```suggestion
      1.0.2 is so ancient that it's taken for granted and further versions only add features in major releases
```

executable/java_class.ksy
```diff
@@ -83,28 +88,85 @@ types:
...
+          id: package
+          -orig-id: CONSTANT_Package
+
+  version_guard:
+    doc: |
+      `class` file format version 45.3 (appeared in the very first publicly known release of Java SE 1.0.2 and JDK 1.0, released 23th January 1996) is so ancient that it's taken for granted. Earlier formats seem to be undocumented. Changes of `version_minor` don't change `class` format.
+      Earlier `version_major`s likely belong to Oak programming language, the proprietary predecessor of Java.
+    doc-ref:
+      - James Gosling, Bill Joy and Guy Steele. The Java Language Specification. English. Ed. by Lisa Friendly. Addison-Wesley, Aug. 1996, p. 825. ISBN: 0-201-63451-1.
+      - Frank Yellin and Tim Lindholm. The Java Virtual Machine Specification. English. Ed. by Lisa Friendly. Addison-Wesley, Sept. 1996, p. 475. ISBN: 0-201-63452-X.
```
generalmimon, line 158: This is a key-value pair in YAML. I hope it wouldn't slip through the compiler :grimacing:

```suggestion
    doc-ref:
      - 'James Gosling, Bill Joy and Guy Steele. The Java Language Specification. English. Ed. by Lisa Friendly. Addison-Wesley, Aug. 1996, p. 825. ISBN: 0-201-63451-1.'
      - 'Frank Yellin and Tim Lindholm. The Java Virtual Machine Specification. English. Ed. by Lisa Friendly. Addison-Wesley, Sept. 1996, p. 475. ISBN: 0-201-63452-X.'
```

generalmimon: > I hope it wouldn't slip through the compiler 😬

Fortunately, it doesn't:

```console
$ kaitai-struct-compiler-0.10 -- --verbose file -t java java_class.ksy
parsing java_class.ksy...
reading java_class.ksy...
java_class.ksy: /types/version_guard/doc-ref/0:
        error: expected string, got Map(James Gosling, Bill Joy and Guy Steele. The Java Language Specification. English. Ed. by Lisa Friendly. Addison-Wesley, Aug. 1996, p. 825. ISBN -> 0-201-63451-1.) (class scala.collection.immutable.MapLike$$anon$2)
```

generalmimon (changes requested):
> * 77e5472 (HEAD, KOLANICH-specs/java_class_more_tags) Split `dynamic` from `invoke_dynamic` in `java_class` and set the proper version check for it.
> * 584c0a0 Add `Dynamic` tag into `java_class`
> * 019ecf9 Add `Module` and `Package` tags into `java_class`
> * cba8a7f Add version checks for `java_class`
> * ee774c1 Add additional `doc-ref`s int `java_class.ksy`.
> * ec49c52 Add `-orig-id`s to constants types and enums. Sorted constants by their index, as done in `The Java® Virtual Machine Specification Java SE 19 Edition`
> * 6a08446 Remove unneeded quotes in `java_class.ksy`

For commits clearly belonging to a single format spec, I'd prefer to normalize the commit messages to always start with `java_class: ` + a lowercase letter.

KOLANICH:
Fixed

generalmimon:
> Fixed

@KOLANICH Sorry for being ambiguous with respect to formatting - I don't think it's necessary to format the "java_class" in as inline code span with monospace font in commit titles (as <code><del>\`java_class\`</del>: ...</code> - please use <code><ins>**java_class**</ins>: ...</code> instead). I'd like to see something similar [to this](https://github.com/kaitai-io/kaitai_struct_formats/commits/master?after=729989ef249c90be4af72776aac565d9ced004db+34&branch=master&qualified_name=refs%2Fheads%2Fmaster):

```
icc_4: use correct repeat-expr (see section 10.15 of specification)
icc_4: fix type of curve type (u2 instead of u4, see section 10.5 of 
icc: add chromaticity_type which was referenced, but not defined in an 
Fix {warcraft_2_pud,dicom}.ksy to compile with KSC 0.10 
zisofs: add specification (#603) 
napp_header: rename to android_nanoapp_header, improve metadata
napp_header: shorten /doc-ref commit hash
napp_header: remove bitmask duplication in flag extraction
add the Android Nano app header
java_class: add reference for 43.0 as the minimum version
java_class: add sanity check for the version number as there is a clash 
Add log/mcap.ksy (#592) 
elf: sync with upstream binutils code (#606) 
dtb: use hyphen - to join words in /doc (not the UTF-8 en dash)
zip: update compression and extra_codes enums, sync with upstream APP… 
nt_mdt_pal: fix encoding - UTF-16LE, not UTF-16 
java_class: improve is_two_entries to prevent 'ConstantPoolEntry' obj… 
Merge pull request #593 from armijnhemel/dos_mz 
dos_mz: fix for kaitai struct compiler (the web ide doesn't complain 
dos_mz: correctly compute the body length
dos_mz: allow both MZ and ZM magic
```

I've been following this convention for a while in the KSF repo, and I'd suggest to do the same so that we have a consistent style in the commit history.

KOLANICH:
I propose to have all the identifiers be marked with back ticks to make them rendered in markdown as identifiers. Those inline comments were created for things like that - the quotes from source code. An identifier 8s a quote from source code, so IMHO it should be marked as such.

KOLANICH:
Also it'd be nice to use the same convention for the commits introducing formats:

>`uuid`: Create the spec
>`nt_mdt`, `nt_mdt_pal`: Create the specs for NT-MDT formats

generalmimon:
@KOLANICH:

> I propose to have all the identifiers be marked with back ticks to make them rendered in markdown as identifiers.

Sorry, I will not negotiate on this specifically. This convention is used in this repo for a long time, it works well, I'm happy with it and I see absolutely no reason to change it.

> Those inline comments were created for things like that - the quotes from source code. An identifier 8s a quote from source code, so IMHO it should be marked as such.

Feel free to (and I'd even encourage to) use backticks for YAML path keys (e.g. `/doc-ref`, `encoding`, `/meta/xref`, etc.), .ksy type/enum/field identifiers (e.g. `header` type, `len_data` field, `machine` enum, `sh_type::nobits` enum value), expressions in KS expression language (e.g. `header.data_size == 0xffff_ffff`), and so on.

Just the commit title has to start with the .ksy base filename without the .ksy extension, then a colon (`:`) and space ` `.

If you would like a more precise classification, this Python code may help:

```python
import unittest
from pathlib import PurePath


def check_ksf_commit_title(title: str, changed_file_paths: list[str]) -> None:
    if len(changed_file_paths) != 1:
        return
    file_path = PurePath(changed_file_paths[0])
    if file_path.name.endswith('.ksy'):
        expected_prefix = file_path.stem + ": "
        assert title.startswith(expected_prefix), \
            f"expected title \"{title}\" to start with \"{expected_prefix}\", but it did not"


class TestCheckKsfCommitTitle(unittest.TestCase):
    def test_invalid_1(self):
        with self.assertRaisesRegex(AssertionError, "expected title.* to start with \"java_class: \""):
            check_ksf_commit_title("`java_class`: remove unneeded quotes", ['executable/java_class.ksy'])

    def test_invalid_2(self):
        with self.assertRaisesRegex(AssertionError, "expected title.* to start with \"java_class: \""):
            check_ksf_commit_title("Add more tags into `java_class`", ['executable/java_class.ksy'])

    def test_valid(self):
        check_ksf_commit_title("java_class: remove unneeded quotes", ['executable/java_class.ksy'])


if __name__ == '__main__':
    unittest.main()

```

generalmimon:
@KOLANICH:

> Also it'd be nice to use the same convention for the commits introducing formats:
> 
> > `uuid`: Create the spec
> > `nt_mdt`, `nt_mdt_pal`: Create the specs for NT-MDT formats

No, I specifically don't think this would be nice. Commits that add a `.ksy` spec have often the same title as the [pull requests](https://github.com/kaitai-io/kaitai_struct_formats/pulls) and I don't see a single PR that would spontaneously use it. Instead, you often want to include the format _title_ in the commit headline when you're describing a new format (like for example [Add android's DTB/DTBO partitions format](https://github.com/kaitai-io/kaitai_struct_formats/commit/eabe2bce606b0c930cb43dff8f030cba093a3ff7)). 

It's also clear from your examples that after you start the commit message with an identifier of the `.ksy` spec you're adding, you don't have anything meaningful to write. So you had to add some filling ("create the spec"), but this filling could have many variants (all equally meaningless), for example "add", "add the .ksy", "create", "describe the format", ... (which is a great room for inconsistency). And also commits adding vs. modifying a spec are then not so easy to visually identify in the commit history.

KOLANICH:
>Instead, you often want to include the format title in the commit headline when you're describing a new format (like for example Add android's DTB/DTBO partitions format).

Yeah. In fact if we want to search for commits introducing or modifying a certain id, we can use git for that without putting additional data into commits.

>Just the commit title has to start with the .ksy base filename without the .ksy extension

`.ksy` base filename without the `.ksy` extension is the contents of `meta.id`. It generates the python module with the same name, that is an identifier of the module to be imported.

>It's also clear from your examples that after you start the commit message with an identifier of the .ksy spec you're adding, you don't have anything meaningful to write.

That's true. A first commit touching a certain spec is highly likely a commit introducing this spec. The commit message can be completely empty TBH, but it'd just look not very good within the changelog.

>And also commits adding vs. modifying a spec are then not so easy to visually identify in the commit history.

They start with the words describing adding. Though you are right, commits adding a feature also start with these words.

>This convention is used in this repo for a long time

Approximately since c1627d239cd493ed775ed352a29255e4da4c2ef1 it started being actively used.


>Sorry, I will not negotiate on this specifically. 
>it works well, I'm happy with it and I see absolutely no reason to change it.

OK.

executable/java_class.ksy
```diff
@@ -219,19 +293,49 @@ types:
...
+    seq:
+      - type: version_guard(55)
+      - id: bootstrap_method_attr_index
+        type: u2
+      - id: name_and_type_index
+        type: u2
   invoke_dynamic_cp_info:
-    doc-ref: 'https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.4.10'
+    doc-ref: https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.4.10
+    -orig-id: CONSTANT_Dynamic_info
```
generalmimon, line 313: `CONSTANT_InvokeDynamic_info`?

executable/java_class.ksy
```diff
@@ -219,19 +293,49 @@ types:
...
     seq:
+      - type: version_guard(51)
       - id: bootstrap_method_attr_index
         type: u2
       - id: name_and_type_index
         type: u2
+  module_package_cp_info:
```
generalmimon, line 320: This is now the only `*_cp_info` type without `-orig-id`.

I think you can list both:

```suggestion
  module_package_cp_info:
    -orig-id:
      - CONSTANT_Module_info
      - CONSTANT_Package_info
```

KOLANICH:
Thanks for the review.

generalmimon (approved):
@KOLANICH Thanks for contributing!

PR #644, opened 2023-03-11
Add additional structs used within other Windows structs.

KOLANICH:
Changed `windows_minidump` to use them.

Samples: https://github.com/kaitai-io/kaitai_struct_samples/pull/19


windows/windows_minidump.ksy
```diff
@@ -74,6 +74,10 @@ types:
             'stream_types::exception': exception_stream
             # TODO: support more stream types
   system_info:
+    meta:
+      imports:
+        - /windows/windows_suite_mask
+        - /windows/windows_version_info
```
generalmimon, line 80: `/meta/imports` should only be used at the top level and not in nested types. See [style guide](https://doc.kaitai.io/ksy_style_guide.html#meta):

> For intermediate-level meta sections, use the following order of keys:
> 
> 1. Defaults (in alphabetic order):
> 
>    a. `encoding`
>    b. `endian`
> 
> <table>
> <tr>
> <th>NOTE</th>
> <td>
> 
> KS syntax allows usage of some top-level elements deep inside the hierarchy — this can be useful during development, for example, for purpose of grafting one .ksy file into another quickly. However, in production-quality .ksy files, one MUST NOT use keys like `title`, `imports` or `ks-version` (i.e. everything except explicitly listed in a list above) on intermediate levels.
> </td>
> </tr>
> </table>

PR #646, opened 2023-03-17
add Portable Image Format (pif)

armijnhemel:
This PR adds a fairly basic specification of the PIF format, a fairly new image format for embedded systems and microcontrollers: https://hackaday.com/2022/05/08/creating-an-image-format-for-embedded-hardware/

### DIFF
--- a/image/pif.ksy
+++ b/image/pif.ksy
@@ -14,7 +14,7 @@ seq:
   - id: color_table
     size: information_header.len_color_table
   - id: data
-    size: header.len_file - header._sizeof - information_header._sizeof - information_header.len_color_table
+    size: header.len_image
 types:
   header:
     seq:
--- a/image/pif.ksy
+++ b/image/pif.ksy
@@ -2,11 +2,21 @@ meta:
   id: pif
   title: Portable Image Format
   file-extension: pif
+  xref:
+    justsolve: PIF_(Portable_Image_Format)
   license: LGPL-2.1
   ks-version: 0.9
   endian: le
   bit-endian: le
-doc-ref: https://github.com/gfcwfzkm/PIF-Image-Format/blob/cc256d5/Specification/PIF%20Format%20Specification.pdf
+doc: |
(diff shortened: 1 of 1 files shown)


image/pif.ksy
```diff
@@ -0,0 +1,58 @@
+meta:
+  id: pif
+  title: Portable Image Format
+  file-extension: pif
+  license: LGPL-2.1
+  ks-version: 0.10
```
generalmimon, line 6: @armijnhemel It should work in 0.9 too, doesn't it?

image/pif.ksy
```diff
@@ -0,0 +1,58 @@
...
+doc-ref: https://github.com/gfcwfzkm/PIF-Image-Format/blob/cc256d5/Specification/PIF%20Format%20Specification.pdf
+seq:
+  - id: header
+    type: header
+  - id: information_header
+    type: information_header
+  - id: color_table
+    size: information_header.len_color_table
+  - id: data
+    size: header.len_file - header._sizeof - information_header._sizeof - information_header.len_color_table
```
generalmimon, line 17: Is there a reason why you don't use

```suggestion
    size: header.len_image
```

instead? It also seems more correct.

image/pif.ksy
armijnhemel: Yes, it should be. I just default to using the latest version. I will change this to 0.9.

image/pif.ksy
armijnhemel: Yes, that's actually better. I think that the intention of the author of the format is to not have them diverse, so that should be correct.

generalmimon:
@armijnhemel If it still works for you correctly, I'm fine with merging this.

generalmimon:
@armijnhemel Did you have a chance to look at this?

armijnhemel:
> @armijnhemel Did you have a chance to look at this?

I went through all of the test images, and as far as I can see they are all parsed correctly.

generalmimon:
> Thanks for checking!

@armijnhemel And of course thanks for the contribution too 😉

PR #660, opened 2023-04-01
fix string condition

milahu:
the parser was trying to decode non-string fields as string

```
import parser.sqlite3 as parser_sqlite3
db = parser_sqlite3.Sqlite3.from_file("test.db")
cell = db.root_page.cells[0]
print(cell.body)
```

error was

```
Traceback (most recent call last):
  File "parse.py", line 9, in <module>
    cell.body
  File "parser/sqlite3.py", line 337, in body
    self._m_body = Sqlite3.CellTableLeaf(self._io, self, self._root)
  File "parser/sqlite3.py", line 191, in __init__
    self._read()
  File "parser/sqlite3.py", line 198, in _read
    self.payload = Sqlite3.CellPayload(_io__raw_payload, self, self._root)
  File "parser/sqlite3.py", line 210, in __init__
    self._read()
  File "parser/sqlite3.py", line 224, in _read
    Sqlite3.ColumnContent(
  File "parser/sqlite3.py", line 270, in __init__
    self._read()
  File "parser/sqlite3.py", line 297, in _read
    self.as_str = (self._io.read_bytes(self.serial_type.len_content)).decode(
  File "parser/kaitaistruct.py", line 302, in read_bytes
    if n < 0:
TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

### DIFF
--- a/database/sqlite3.ksy
+++ b/database/sqlite3.ksy
@@ -230,6 +230,7 @@ types:
         type: str
         size: serial_type.len_content
         encoding: UTF-8
+        if: serial_type.is_string
 #        if: _root.text_encoding == encodings::utf_8 and serial_type.is_string
     instances:
       serial_type:


database/sqlite3.ksy
```diff
@@ -230,6 +230,7 @@ types:
         type: str
         size: serial_type.len_content
         encoding: UTF-8
+        if: serial_type.is_string
 #        if: _root.text_encoding == encodings::utf_8 and serial_type.is_string
```
generalmimon, line 234: I wonder why the `if` was commented out 🤔 (do you see anything wrong with it?). I think it's more correct because you don't check which encoding the file uses and always parse the string as `UTF-8`, which seems to be incorrect.

generalmimon (changes requested):

generalmimon:
Actually, I should first check if https://github.com/kaitai-io/kaitai_struct_formats/pull/640 hasn't already solved the problems the **sqlite3.ksy** has.

@milahu Could you also try it out and see if it fixes the errors you're getting?

generalmimon:
> Actually, I should first check if #640 hasn't already solved the problems the **sqlite3.ksy** has.

Yeah, it appears that it solves the problem you attempted to fix in this PR (https://github.com/kaitai-io/kaitai_struct_formats/pull/640#issue-1588560917):

> * Fix string serial type parsing by making it conditional

---

So I'm closing this PR in favor of #640.

PR #661, opened 2023-04-02
sqlite3: fix parser

milahu:
continue #640

### DIFF
--- a/database/sqlite3.ksy
+++ b/database/sqlite3.ksy
@@ -27,218 +27,476 @@ doc: |
   versions, size of page, etc). After the header, normal contents of
   the first page follow.
 
-  Each page would be of some type, and generally, they would be
-  reached via the links starting from the first page. First page type
-  (`root_page`) is always "btree_page".
+  Each page would be of some type (btree, ptrmap, lock_byte, or free),
+  and generally, they would be reached via the links starting from the
+  first page. The first page is always a btree page for the implicitly
+  defined `sqlite_schema` table.
 doc-ref: https://www.sqlite.org/fileformat.html
 seq:
-  - id: magic
-    contents: ["SQLite format 3", 0]
-  - id: len_page_mod
-    type: u2
-    doc: |
-      The database page size in bytes. Must be a power of two between
-      512 and 32768 inclusive, or the value 1 representing a page size
-      of 65536.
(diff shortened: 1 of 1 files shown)


database/sqlite3.ksy
```diff
@@ -27,218 +27,487 @@ doc: |
...
+        size: _root.header.page_size
+        type:
+          switch-on: '(page_index == _root.header.idx_lock_byte_page ? 0 : page_index >= _root.header.idx_first_ptrmap_page and page_index <= _root.header.idx_last_ptrmap_page ? 1 : 2)'
+          cases:
+            0: lock_byte_page(page_number)
+            1: ptrmap_page(page_number)
+            # TODO: Free pages and cell overflow pages are incorrectly interpreted as btree pages
+            # This is unfortunate, but unavoidable since there's no way to recognize these types at
+            # this point in the parser.
+            2: btree_page(page_number)
```
milahu, line 64: @generalmimon no luck with lazy `db.pages`

this still loops all pages when i read `db.pages[0]`

```py
    @property
    def pages(self):
        if hasattr(self, "_m_pages"):
            return self._m_pages

        self._m_pages = []
        for i in range(self.header.num_pages):
            self._m_pages.append(
                Sqlite3.Page(
                    (i + 1), (self.header.page_size * i), self._io, self, self._root
                )
            )
```

generalmimon: @milahu:

> this still loops all pages when i read `db.pages[0]`

Yes, it creates the objects but does not parse them.

If you don't want to even create the empty objects (because the total memory usage of too many empty objects would be too high), you can provide an "unused" `page` type and let the users of the parser instantiate it themselves for the page number they want (and even dispose the object afterwards to keep the memory usage low). This approach is essentially explained in https://stackoverflow.com/a/73332294/12940655.

KS unfortunately doesn't support truly unused types very well at the moment (i.e. when you define a type in the `types` but don't use it absolutely anywhere), but this can be easily worked around by the `if: false` trick as explained in the linked SO post.

edzillion:
Yeah, I am trying to parse 'app size' sqlite dbs on limited hardware so lazy parsing isn't an issue. That's an interesting application you have, thanks for the pointers, I will have a look.

PR #663, opened 2023-04-19
rockchip specific logo container format

armijnhemel:
This PR adds a spec for a format used by Rockchip on at least some of their eInk devices. I could not find any specification about what the fields are whatsoever, but it seems to be fairly simple (`height` and `width` are quite obvious when searching for the particular chipset). My guess is that what I am calling `unknown_2` is probably a setting for the image quality and `unknown_3` could perhaps be X/Y coordinates.

The data itself: possibly something raw or BMPs without a header (which is not uncommon on Android).

### DIFF
--- a/image/rockchip_rkel.ksy
+++ b/image/rockchip_rkel.ksy
@@ -54,13 +54,13 @@ types:
         type: u2
       - id: width
         type: u2
-      - id: unknown_3
+      - id: unknown_2
         type: u4
       - id: ofs_image_data
         type: u4
       - id: len_image_data
         type: u4
-      - id: unknown_4
+      - id: unknown_3
         size: 8
     instances:
       image_data:
--- a/image/rockchip_rkel.ksy
+++ b/image/rockchip_rkel.ksy
@@ -1,0 +1,69 @@
+meta:
+  id: rkel
+  title: Rockchip RKEL
+  license: CC0
+  endian: le
+  encoding: UTF-8
+doc: |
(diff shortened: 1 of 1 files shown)


KOLANICH:
One field is called `magic` and another one is `signature`... Feels a bit inconsistent.

KOLANICH:
```yaml
size: 64
....
size-eos: true
```

May it have sense to just replace `size-eos` with the needed size?

armijnhemel:
> One field is called `magic` and another one is `signature`... Feels a bit inconsistent.

Perhaps. One is for the top level identifier, the other for the individual image entries. I don't really care about what it's called.

armijnhemel:
> ```yaml
> size: 64
> ....
> size-eos: true
> ```
> 
> May it have sense to just replace `size-eos` with the needed size?

I don't see a good reason to. This looked like the cleanest solution to me at the moment, because there are no specifications whatsoever so I don't even know what the rest of the data means (and I only have two test files).

KOLANICH:
I guess `image_entry` should be nested into `images`, shouldn't it?

armijnhemel:
> I guess `image_entry` should be nested into `images`, shouldn't it?

As far as I know it is nested (clicking through it in the Web IDE now), so I don't know what you mean exactly.

KOLANICH:
Now it is

```yaml
types:
  images:
    ...
  image_entry:
```

, but I propose it be

```yaml
types:
  images:
    ...
    types:
      image_entry:
```

PR #664, opened 2023-04-19
add specification for lrzip

armijnhemel:
This PR adds a specification for the lrzip format. It *mostly* works, except that if there is a trailing MD5 sum it isn't easy to automatically recognize it because it is necessary to first find out where the data ends and propagate that (and I don't know how to do that cleanly with instances).

Also, this one will not work with very big files, as I was too lazy to implement something like `b40le`, `b48le` and `b56le`, so there is still room for improvement.

### DIFF
--- a/archive/lrzip.ksy
+++ b/archive/lrzip.ksy
@@ -1,0 +1,149 @@
+meta:
+  id: lrzip
+  title: lrzip 0.6
+  file-extension: lrz
+  license: CC0-1.0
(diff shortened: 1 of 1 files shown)


KOLANICH:
I guess

```yaml
switch-on: byte_width
cases:
  1: u1
  2: u2
  3: b24le
  4: u4
  8: u8
```

can be an own type.

PR #665, opened 2023-04-20
add specification for SerpentOS .stone files

armijnhemel:
This PR adds a parser for the `.stone` packaging format used in SerpentOS (a fairly new Linux distribution). I only found one file to test with in their latest ISO (only containing metadata), so there isn't much beyond structural parsing.

### DIFF
--- a/archive/stone.ksy
+++ b/archive/stone.ksy
@@ -1,0 +1,69 @@
+meta:
+  id: stone
+  title: SerpentOS Stone
+  license: Zlib
+  endian: be
+doc-ref: <https://github.com/serpent-os/libmoss/blob/841a6d67/source/moss/format/binary/archive_header.d>
+seq:
+  - id: header
+    type: header
+    size: 32
+  - id: payloads
+    type: payload
+    repeat: expr
+    repeat-expr: header.num_payloads
+types:
+  header:
+    seq:
+      - id: signature
+        contents: [0, 0x6d, 0x6f, 0x73]
+      - id: num_payloads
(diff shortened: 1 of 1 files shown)


armijnhemel:
Well, found some more files to test with, so I will go back to the drawing board first:

https://dev.serpentos.com/pool/b/bash/

KOLANICH (approved):

KOLANICH:
>`be`
>in 2021, when the only sensible endian (`le`) is mainstream

:facepalm:

armijnhemel:
> > `be`
> > in 2021, when the only sensible endianness (`le`) is mainstream in hardware
> 
> facepalm

Yet the format uses `be`

armijnhemel:
> Well, found some more files to test with, so I will go back to the drawing board first:
> 
> https://dev.serpentos.com/pool/b/bash/

It seems that `zstd` is not universally supported by `kaitai-compress` so I will leave this one as it is right now.

archive/stone.ksy
```diff
@@ -0,0 +1,69 @@
+meta:
+  id: stone
+  title: SerpentOS Stone
+  license: Zlib
+  endian: be
+doc-ref: <https://github.com/serpent-os/libmoss/blob/841a6d67/source/moss/format/binary/archive_header.d>
```
generalmimon, line 6:
```suggestion
doc-ref: https://github.com/serpent-os/libmoss/blob/841a6d67/source/moss/format/binary/archive_header.d
```

PR #666, opened 2023-04-23
add specification for Qt translation files

armijnhemel:
This PR adds a fairly minimal specification for Qt translation files. It is not complete and misses a few things:

* processors for sourcetext16 and context16. I could not find any test files that have these tags
* processor for the hashes section

What also doesn't work correctly is correctly processing the translation. Currently I am simply returning a byte array for translations, but it should be a string, almost certainly `utf-16be`.

### DIFF
--- a/database/qt_translation.ksy
+++ b/database/qt_translation.ksy
@@ -1,0 +1,92 @@
+meta:
+  id: qt_translation
+  title: Qt Translation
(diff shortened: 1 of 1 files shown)


KOLANICH (approved):
maybe `qt_qm`?

PR #668, opened 2023-04-29
add specification for netgear .chk files

armijnhemel:
This PR adds a specification for Netgear `.chk` files. It is a fairly simple format: a header specifying some meta information and lengths, followed by data. The checksums in the header are computed using a Netgear specific method.


KOLANICH:
I guess you can try to implement checksumming in KS using my impls of various checksums (#623, #624, #625) as a base.

PR #671, opened 2023-05-02
add specification for Fullhan .FLS files

armijnhemel:
A fairly simple file format for a Chinese brand of IP cameras. Basically it is a header, with a number of entries, each containing a name, offset and length and then data. I don't know what each field means, so there is also a big chunk of "unknown". The header also contains two names, which I *think* refer to a model and submodel, or a chip and a board, or something like that. I could not find reliable information about that.

### DIFF
--- a/firmware/fullhan_fls.ksy
+++ b/firmware/fullhan_fls.ksy
@@ -1,0 +1,53 @@
+meta:
+  id: fullhan_fls
+  title: Fullhan FLS
+  license: CC0-1.0
+  endian: le
+  encoding: UTF-8
+doc: |
+  Test files:
+  - https://www.nvripc.com/new-chinese-ip-camera-firmwares-update/
+  - https://www.herospeed.net/hs/ipc_test/fuhan/
+seq:
+  - id: header
+    type: header
+  - id: entries
+    type: entry
+    repeat: expr
+    repeat-expr: header.num_entries
+  - id: data
+    size: header.len_file - ofs_data
+    type: dummy
+instances:
+  ofs_data:
+    value: header._sizeof + header.num_entries * sizeof<entry>
+types:
+  dummy: {}
+  header:
+    seq:
+      - id: model
+        size: 8
+        type: strz
+      - id: submodel
+        size: 8
+        type: strz
+      - id: len_file
+        type: u4
+      - id: num_entries
+        type: u4
+  entry:
+    seq:
+      - id: name
+        size: 128
+        type: strz
+      - id: len_data
+        type: u4
+      - id: ofs_data
+        type: u4
+      - id: unknown
+        size: 120
+    instances:
+      data:
+        io: _root.data._io
+        pos: ofs_data - _root.ofs_data
+        size: len_data


armijnhemel:
It seems like other (cheap) brands are also using this file format, so I guess that it is not specific to Fullhan. What should we rename it to?

KOLANICH:
https://github.com/kaitai-io/kaitai_struct/issues/1031 should be related.

KOLANICH:
```yaml
meta:
  id: fullhan_fls
  title: Fullhan FLS
  license: CC0-1.0
  endian: le
  encoding: UTF-8
doc: |
  Test files:
  - https://www.nvripc.com/new-chinese-ip-camera-firmwares-update/
  - https://www.herospeed.net/hs/ipc_test/fuhan/
seq:
  - id: header
    type: header
  - id: blob
    size: header.len_file - sizeof<header>
    type: body
types:
  header:
    seq:
      - id: model
        size: 8
        type: strz
      - id: submodel
        size: 8
        type: strz
      - id: len_file
        type: u4
      - id: num_entries
        type: u4
  body:
    seq:
      - id: entries
        type: entry
        repeat: expr
        repeat-expr: _root.header.num_entries
      - id: blob_area
        size-eos: true
        type: dummy
    instances:
      lea_blob_area:
        -affected-by: 84
        #value: entries._sizeof
        value: _root.header.num_entries * sizeof<entry>
      lea_lea_blob_area_global:
        value: sizeof<header> + lea_blob_area
    types:
      dummy: {}
      entry:
        seq:
          - id: name
            size: 128
            type: strz
          - id: len_data
            type: u4
          - id: ofs_data
            type: u4
          - id: unknown
            size: 120
        instances:
          offs_base_blob_area:
            -affected-by: 1031
            value: ofs_data - _parent.lea_lea_blob_area_global
          data:
            -affected-by: 1031
            io: _parent.blob_area._io
            pos: offs_base_blob_area
            size: len_data
```

firmware/fullhan_fls.ksy
```diff
@@ -0,0 +1,53 @@
+meta:
+  id: fullhan_fls
+  title: Fullhan FLS
```
generalmimon, line 3:
```suggestion
meta:
  id: fullhan_fls
  title: Fullhan FLS
  file-extension: fls
```

firmware/fullhan_fls.ksy
```diff
@@ -0,0 +1,54 @@
...
+  file-extension: fls
+  license: CC0-1.0
+  endian: le
+  encoding: UTF-8
+doc: |
+  Test files:
```
generalmimon, line 9: Thanks for declaring where to find test files (I'll check them out), but I think `/doc` should also include some general description of the format. You put the .ksy spec in the `firmware` directory, which suggests that the format is somehow firmware-related, but anything beyond that is left to the reader to guess. I guess it's a firmware update format of IP cameras of a certain brand Fullhan? At least one or two sentences would go a long way for someone not knowing the purpose of the format or perhaps trying to find a .ksy spec for it via certain keywords.

PR #675, opened 2023-05-18
bugfix extended timestamp zip.ksy

proukornew:


armijnhemel:
I can confirm that this is indeed a problem. Attached is a minimal test file.

[test.zip](https://github.com/kaitai-io/kaitai_struct_formats/files/11523156/test.zip)

PR #676, opened 2023-06-06
elf: fix naming warnings

avaunit02:
when I compile `elf.ksy` I get a load of warnings about the naming of `program_header_entry_size`, `section_header_entry_size`, `qty_program_header`, `qty_section_header`

this PR renames those fields to fix the warnings (to match what's in the style guide)

### DIFF
--- a/executable/elf.ksy
+++ b/executable/elf.ksy
@@ -299,16 +299,16 @@ types:
       - id: e_ehsize
         type: u2
       # e_phentsize
-      - id: program_header_entry_size
+      - id: len_program_headers
         type: u2
       # e_phnum
-      - id: qty_program_header
+      - id: num_program_headers
         type: u2
       # e_shentsize
-      - id: section_header_entry_size
+      - id: len_section_headers
         type: u2
       # e_shnum
-      - id: qty_section_header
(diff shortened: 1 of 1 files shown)


GreyCat:
@avaunit02 Thanks a lot for this contribution, looks good to me, merged!

executable/elf.ksy
```diff
@@ -703,23 +703,23 @@ types:
...
         type: program_header
         repeat: expr
-        repeat-expr: qty_program_header
+        repeat-expr: num_program_headers
       section_headers:
         pos: section_header_offset
-        size: section_header_entry_size
+        size: len_section_headers
         type: section_header
         repeat: expr
```
generalmimon, line 714: This renaming is actually misleading - the old `section_header_entry_size` correctly suggested that the field specifies the size of **one** entry, whereas the new `len_section_headers` name incorrectly indicates that it's the total size of all entries, which is not true.

Note that the compiler naming suggestions are pretty crude, taking none of these semantics into account; they can also easily be contradictory, e.g. if you have a `num_*` field that you use in two different `repeat-expr`s. So we always have to think about whether the suggested names make sense.

PR #682, opened 2023-09-28
add specification for TP-Link's minifs

armijnhemel:
This PR adds a specification for a file system found in certain TP-Link routers and seems to be used for their own RTOS. An example can be found in the RE200 V5 firmware: https://www.tp-link.com/nl/support/download/re200/#Firmware

### DIFF
--- a/filesystem/minifs.ksy
+++ b/filesystem/minifs.ksy
@@ -1,0 +1,112 @@
+meta:
+  id: minifs
+  title: MINIFS
+  license: CC0-1.0
+  endian: be
+  encoding: ASCII
+doc: |
+  MINIFS is a file system found in certain TP-Link firmware files, such as
+  RE450(V4)_210531.zip, for TP-Link's own TPOS RTOS.
+
+  It consists of a header, followed by a list of file names, followed by
+  meta information for each file, meta information about the LZMA compressed
+  blobs including length, offset and uncompressed length, and then a number
+  of LZMA compressed blobs.
+
+  The LZMA blobs can contain the concatenated contents of multiple files.
+  The meta information of a file contains an offset to the directory name,
+  a file name, the LZMA blob (numbering starts at 0), the offset of the file
+  in the uncompressed data and the size of the uncompressed file.
+seq:
+  - id: header
+    type: header
+    size: 32
+  - id: filenames
+    type: filenames
+    size: header.len_filenames
+  - id: inodes
+    type: 'inode(_index != 0 ? inodes[_index - 1].cur_max_lzma_blob : 0)'
+    repeat: expr
+    repeat-expr: header.num_files
+  - id: lzma_metas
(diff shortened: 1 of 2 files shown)


filesystem/minifs.ksy
```diff
@@ -0,0 +1,112 @@
+meta:
+  id: minifs
```
generalmimon, line 2: `tplink_minifs`? Maybe it helps someone to include `tplink` in there to find the format they need, if the format is TP-Link specific.

armijnhemel: This makes sense.

filesystem/tplink_minifs.ksy
```diff
@@ -0,0 +1,112 @@
...
+        pos: ofs_directory
+        type: strz
+  lzma_meta:
+    params:
+      - id: prev_max_lzma_blob_len
+        type: u4
+    seq:
+      - id: ofs_blob
+        type: u4
+        valid: prev_max_lzma_blob_len
```
generalmimon, line 105: This seems to enforce that

1. this must be `0` in the first `lzma_meta`, then
2. `lzma_metas[0].ofs_blob + lzma_metas[0].len_blob` in the second `lzma_meta`,
3. `lzma_metas[1].ofs_blob + lzma_metas[1].len_blob` in the third `lzma_meta`

... and so on. In other words, computing the "maximum" seems to be unnecessary here, because this `valid` ensures sequential order without any holes, so we know we can just read the size of `lzma_blobs_area` from the last `lzma_meta` entry.

In that case, I think this would better express what's going on:

```diff
   - id: lzma_metas
-    type: 'lzma_meta(_index != 0 ? lzma_metas[_index - 1].cur_max_lzma_blob_len : 0)'
+    type: 'lzma_meta(_index != 0 ? lzma_metas[_index - 1].ofs_end : 0)'
     repeat: expr
     repeat-expr: inodes.last.cur_max_lzma_blob + 1
     doc: Count starts at 0, so there is one more blob than the highest blob number
   - id: lzma_blobs_area
-    size: lzma_metas.last.cur_max_lzma_blob_len
+    size: lzma_metas.last.ofs_end
     type: dummy
```

```diff
   lzma_meta:
     params:
-      - id: prev_max_lzma_blob_len
+      - id: prev_ofs_end
         type: u4
     seq:
       - id: ofs_blob
         type: u4
         valid: prev_ofs_end
       - id: len_blob
         type: u4
       - id: len_blob_uncompressed
         type: u4
     instances:
-      cur_max_lzma_blob_len:
-        value: 'ofs_blob + len_blob > prev_max_lzma_blob_len ? ofs_blob + len_blob : prev_max_lzma_blob_len'
+      ofs_end:
+        value: ofs_blob + len_blob
```

armijnhemel: This makes sense. In the files I have researched so far the metadata is ordered and there are no holes, but they could possibly happen although I think this is very unlikely given that it is running on RTOS devices with relatively little flash space, so I think that it is a fair assumption to make.

PR #687, opened 2024-01-15
Add Qualcomm QDB (Qshrink Database) format

tchebb:
This is a very simple wrapper around compressed XML used by Qualcomm to store modem debug info. I gather that these files are used by Qualcomm's official QXDM diagnostic software, but I derived the format without reference to any documentation or tooling; it's just that simple.

Not quite sure if firmware/ is the right directory for this. Suggestions welcome!


tchebb:
Here's a sample file from the `modem` partition of a Samsung Galaxy S10e, firmware version G970USQU9IWH2. The file is Qualcomm/Samsung proprietary licensed, as far as I'm aware.

[sample.zip](https://github.com/kaitai-io/kaitai_struct_formats/files/13941125/sample.zip)

PR #703, opened 2024-11-06
microsoft pe debug info parse

aoleg94:


executable/microsoft_pe.ksy
```diff
@@ -58,6 +58,11 @@ types:
         size: optional_hdr.data_dirs.certificate_table.size
         type: certificate_table
         if: optional_hdr.data_dirs.certificate_table.virtual_address != 0
+      debug_table:
+        pos: optional_hdr.data_dirs.debug.virtual_address - (sections[1].virtual_address - sections[1].pointer_to_raw_data)
```
aoleg94, line 62: assumes that second section is .rdata

a better way to convert RVA to raw offset is needed here

PR #711, opened 2025-03-23
Change the type of timeval.sec of log/glibc_utmp.ksy to follow the change of GNU libc

nmantani:
The type of tv_sec in bits/utmp.h of GNU libc was changed from int32_t to __uint32_t on 19 April 2024 to address the year 2038 problem.

The relevant commit of GNU libc:
https://sourceware.org/git/?p=glibc.git;a=commit;h=5361ad3910c257bc327567be76fde532ed238e42

### DIFF
--- a/log/glibc_utmp.ksy
+++ b/log/glibc_utmp.ksy
@@ -60,7 +60,7 @@ types:
   timeval:
     seq:
       - id: sec
-        type: s4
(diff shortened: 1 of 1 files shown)


generalmimon (approved):
@nmantani Thanks!

Issue #158, opened 2019-06-04
WebAssembly


GreyCat:
@sophoslabs have started a WebAssembly parser:

* [webassembly.ksy](https://github.com/sophoslabs/WebAssembly/blob/master/Tools/Kaitai-Wasm/webassembly.ksy)
* [Python disassembler based on that .ksy](https://github.com/sophoslabs/WebAssembly/blob/master/Tools/Kaitai-Wasm/wasm-disassembler.py)

```yaml
meta:
  id: webassembly
  title: Web Assembly parser
  file-extension: wasm
  endian: le
  license: CC0-1.0
  file-extension:
    - wasm
  imports:
    - vlq_base128_le
doc: |
  WebAssembly is a web standard that defines a binary format and a corresponding
  assembly-like text format for executable code in Web pages. It is meant to
  enable executing code nearly as fast as running native machine code.
doc-ref: https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md
```

KOLANICH:
This post still doesn't follow the proposed format.

PR #723, opened 2025-10-07
add webassembly binary format

henrythasler:
A [Kaitai Struct](https://kaitai.io) definition for [WebAssembly](https://webassembly.org/) binary files. 

depends on: `/common/vlq_base128_le`

- Supports Wasm 1.0 according to [WebAssembly Core Specification — W3C Recommendation, 5 December 2019](https://www.w3.org/TR/wasm-core-1)
- Naming as close as possible to the WebAssembly Specification
- As permissive as possible to allow parsing of uncommon files
- Cross-references to the specific chapters of the specification
- Follows the [KSY Style Guide](https://doc.kaitai.io/ksy_style_guide.html)

see also: https://github.com/henrythasler/wasm-kaitai-struct

### DIFF
--- a/executable/webassembly.ksy
+++ b/executable/webassembly.ksy
@@ -1,0 +1,385 @@
+meta:
+  id: webassembly
+  title: Binary format for WebAssembly modules
+  file-extension: wasm
+  tags: 
+    - webassembly
+    - schema
+  license: MIT
+  ks-version: 0.9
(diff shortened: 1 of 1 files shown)


armijnhemel:
Also see #158 

The sophoslabs version linked there seems to be unmaintained.

PR #726, opened 2026-01-10
Fix typos (mostly) found by codespell

DimitriPapadopoulos:
There are other typos in `id` keys, but I fear that fixing them may break backwards compatibility:

https://github.com/kaitai-io/kaitai_struct_formats/blob/5f37c8178a631ae1d50e044be7af58dced0d0b69/media/id3v1_1.ksy#L127

https://github.com/kaitai-io/kaitai_struct_formats/blob/5f37c8178a631ae1d50e044be7af58dced0d0b69/network/dns_packet.ksy#L189

https://github.com/kaitai-io/kaitai_struct_formats/blob/5f37c8178a631ae1d50e044be7af58dced0d0b69/network/icmp_packet.ksy#L46-L47

### DIFF
--- a/archive/phar_without_stub.ksy
+++ b/archive/phar_without_stub.ksy
@@ -219,7 +219,7 @@ types:
         SHA-512 signature types.
       * 1.1.1: Supported since phar extension 2.0.0 (released 2009-07-29 and
         included with PHP 5.3 and later). (PHP_Archive 0.12.0 also supports
-        all features from API verison 1.1.1, but it reports API version 1.1.0.)
+        all features from API version 1.1.1, but it reports API version 1.1.0.)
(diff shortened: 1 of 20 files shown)


generalmimon:
@DimitriPapadopoulos:

> There are other typos in `id` keys, but I fear that fixing them may break backwards compatibility: (...)

I think it would be good to fix these as well. There's probably not much point in worrying about backward compatibility of .ksy specifications in the format gallery - if we took it seriously, we wouldn't be able to change almost anything.

DimitriPapadopoulos:
Fixed all typos I was able to find.

generalmimon (approved):
@DimitriPapadopoulos Thanks!

Issue #728, opened 2026-02-05
NITF struct format issues


SprinterDave:
The [nitf.ksy](https://github.com/kaitai-io/kaitai_struct_formats/blob/master/image/nitf.ksy) format has some issues...

1. The determination of `image_segment.has_mask` is incorrect.  It currently check for `img_compression` of `MM` but the [NITF 2.1 spec](https://web.archive.org/web/20181105050336/https://www.gwg.nga.mil/ntb/baseline/docs/2500c/2500C.pdf) section 5.4.3.2 says that IC values of NM, M1, M3, M4, M5, M6, M7, and M8 indicate masked imagery.

2. If an image does not have a mask, the `image_data_field` is not being read.  It should be read regardless

3. `image_geo_loc` is a conditional field (see NITF 2.1 spec, Table A-3, IGEOLO) and should not be parsed if `image_coordinate_rep` (ICORDS) is `' '`

4. `compression_rate_code` is a conditional field (see NITF 2.1 spec, Table A-3, COMRAT) and should not be parsed if `img_compression` (IC) is `NC` or `NM`

5. The `tre_header` field currently shows `header_data` as a simple byte array.  It's possible to show the contents as an array of `tre` objects (which are defined in the ksy, but never utilized.)

SprinterDave:
I do have proposed fixes for these issues.  I'm currently testing them and will have a PR up when testing is completed.

SprinterDave:
FYI some NITF samples for testing can be found at https://earth.esa.int/eogateway/ftp/missions/sample-data/third-party-missions/rapideye/REScene_Basic_Analytic_nitf.zip

PR #729, opened 2026-02-05
Fixes for https://github.com/kaitai-io/kaitai_struct_formats/issues/728

SprinterDave:
Changes have been tested using https://ide.kaitai.io/ with a few sample NITFs from https://earth.esa.int/eogateway/ftp/missions/sample-data/third-party-missions/rapideye/REScene_Basic_Analytic_nitf.zip as well as some proprietary NITFs.

The new TRE array display looks like this...
image

haitai-struct-compiler -all nitf.ksy gives the following warnings...
```
C:/dev/dwelch/kaitai_struct_formats/image/nitf.ksy: /types/image_data_mask/instances/tpxcd_size/id:
        warning: use `len_tpxcd` instead of `tpxcd_size`, given that it's only used as a byte size of `tpxcd` (see https://doc.kaitai.io/ksy_style_guide.html#attr-id)

C:/dev/dwelch/kaitai_struct_formats/image/nitf.ksy: /types/image_data_mask/instances/bmrtmr_count/id:
        warning: use `num_bmrbnd` instead of `bmrtmr_count`, given that it's only used as repeat count of `bmrbnd` (see https://doc.kaitai.io/ksy_style_guide.html#attr-id)

C:/dev/dwelch/kaitai_struct_formats/image/nitf.ksy: /types/image_data_mask/instances/bmrtmr_count/id:
        warning: use `num_tmrbnd` instead of `bmrtmr_count`, given that it's only used as repeat count of `tmrbnd` (see https://doc.kaitai.io/ksy_style_guide.html#attr-id)
```

No attempt was made to address these warnings as they existed prior to the changes in this PR.

### DIFF
--- a/image/nitf.ksy
+++ b/image/nitf.ksy
@@ -297,11 +297,10 @@ types:
         type: image_data_mask
         if: has_mask
       - id: image_data_field
-        if: has_mask
-        size: _parent.header.linfo[idx].length_image_segment.to_i - image_data_mask.total_size
+        size: 'has_mask ? _parent.header.linfo[idx].length_image_segment.to_i - image_data_mask.total_size : _parent.header.linfo[idx].length_image_segment.to_i'
     instances:
       has_mask:
-        value: image_sub_header.img_compression.substring(0, 2) == 'MM'
+        value: image_sub_header.img_compression.substring(0, 1) == 'M' or image_sub_header.img_compression.substring(1, 2) == 'M'
   image_sub_header:
     seq:
       - id: file_part_type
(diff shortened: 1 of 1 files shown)


generalmimon:
@SprinterDave Thanks for the PR and apologies for the delay.

Any chance you could split the commit into several atomic commits, each fixing one specific issue? Please try following https://cbea.ms/git-commit/#seven-rules while writing the commit messages.

Also, please don't put issue references directly into the _title_ of a PR or a commit (i.e. the first line of the commit), put them into the _body_ of a PR or a commit message. Titles should contain a brief human-readable summary without requiring the reader to click anywhere.

If you look at https://github.com/kaitai-io/kaitai_struct_formats/commits/master/, you'll probably notice a pattern/convention that we use for commit titles in this repo, which is that we use the `<format_id>: ` prefix. So in this case, all commit titles should start with `nitf: `.

PR #739, opened 2026-06-10
png: pull changes from BANG (decode zlib-compressed `iTXt`, add PNG 3e HDR chunks)

generalmimon:
This incorporates the changes made to the downstream `png.ksy` spec in the the BANG project by @armijnhemel: https://github.com/armijnhemel/binaryanalysis-ng/blob/1c90fa5447651451d6a70006a26c44844102144a/src/bang/parsers/image/png/png.ksy

### DIFF
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -152,7 +152,7 @@ types:
       - id: b
         type: u1
   cicp_chunk:
-    doc-ref: https://w3c.github.io/PNG-spec/#cICP-chunk
+    doc-ref: https://www.w3.org/TR/png/#cICP-chunk
     seq:
       - id: colour_primaries
         type: u1
@@ -163,7 +163,7 @@ types:
       - id: video_full_range_flag
         type: u1
   clli_chunk:
-    doc-ref: https://w3c.github.io/PNG-spec/#cLLi-chunk
+    doc-ref: https://www.w3.org/TR/png/#cLLI-chunk
     seq:
       - id: maximum_content_light_level
         type: u4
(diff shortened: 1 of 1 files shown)


image/png.ksy
```diff
@@ -77,8 +78,11 @@ types:
 
             # Ancillary chunks
             '"cHRM"': chrm_chunk
+            '"cICP"': cicp_chunk
+            '"cLLi"': clli_chunk
             '"gAMA"': gama_chunk
             # iCCP
+            '"mDCv"': mdcv_chunk
```
generalmimon: @armijnhemel These chunk types actually changed in the [official W3C standard](https://www.w3.org/TR/2025/REC-png-3-20250624/) compared to previous drafts:

- <code><del>cLLi</del></code> -> <code><ins>**cLLI**</ins></code>
- <code><del>mDCv</del></code> -> <code><ins>**mDCV**</ins></code>

`pngcheck` confirms this: https://github.com/pnggroup/pngcheck/blob/bd33ad6490269df07cac81e5305f4ebf56c2b637/pngcheck.c#L3639-L3657

```c
    /*------*
    | cLLi |
    *------*/
    } else if (strcmp(chunkid, "cLLi") == 0) {
            if (verbose)
        printf("\n    "
          "Old version of CLLI, do not use\n");
        set_err(kMinorError);
        last_is_IDAT = last_is_JDAT = 0;

    /*------*
    | mDCv |
    *------*/
    } else if (strcmp(chunkid, "mDCv") == 0) {
            if (verbose)
        printf("\n    "
          "Old version of mDCV, do not use\n");
        set_err(kMinorError);
        last_is_IDAT = last_is_JDAT = 0;
```

I assume it's fine to only support the new types `cLLI` and `mDCV` (as per the official spec) and leave `cLLi` and `mDCv` undefined?

See

- https://github.com/w3c/png/pull/473
- https://github.com/w3c/png/pull/474

armijnhemel: Yes. I think that these names were temporary so they probably can be ignored. I have never seen these chunks in the wild.

PR #743, opened 2026-06-18
png: miscellaneous improvements (mainly in text chunks)

generalmimon:
This brings various improvements to the PNG format spec:

1. parse CRCs as `u4` (not as byte arrays) - a4b8eb32d2dfb7608b9030f622e054cbc96a7d32
1. fix non-canonical encoding names for which KSC 0.11 issues warnings - 64655707f082ed1a92744bcfdf6b579942155e60 (<code><del>iso8859-1</del></code> -> <code><ins>**ISO-8859-1**</ins></code>, <code><del>utf-8</del></code> -> <code><ins>**UTF-8**</ins></code>)

   See https://doc.kaitai.io/ksy_style_guide.html#encoding-name

1. fix the encoding of the `keyword` field in `iTXt` and `zTXt` chunks - 2692b8f2343b55a4a19b7a9276847b1ac0b1fe4d (i.e. in types `international_text_chunk` and `compressed_text_chunk`)

   It was `UTF-8`, but the [official PNG spec](https://www.w3.org/TR/2025/REC-png-3-20250624/#11textinfo) mandates Latin-1 (`ISO-8859-1`). Perhaps the [old PNG 1.2 spec](https://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html#C.Anc-text) says it even more clearly:

   > Keywords are always interpreted according to the ISO/IEC 8859-1 (Latin-1) character set [[ISO/IEC-8859-1](https://www.libpng.org/pub/png/spec/1.2/PNG-References.html#B.I-8859-1)].

   Note: strictly speaking, ISO/IEC 8859-1 is not the same as `ISO-8859-1`, because in "ISO/IEC 8859-1", code points 0x00-0x1F, 0x7F and 0x80-0x9F are unassigned (rather than being assigned to C0 and C1 control codes as in `ISO-8859-1`), but it's the closest equivalent (superset).

1. actually parse the text string in the `zTXt` chunk - 40f998fcba50e5044ee5b20c7c719a40c8cfa29a (previously, parsing was left at decompressing the zlib bytes and providing as a raw byte array in the `text_datastream` field)
1. improve the docstrings (`doc` keys) in all text chunks - 2692b8f2343b55a4a19b7a9276847b1ac0b1fe4d, 839a3eca693849b51113c3230ac4105319ec9a91 (in particular, correct and clarify the set of allowed characters)

1. add `-webide-representation` to all three types of text chunks (`iTXt`, `tEXt`, and `zTXt`) - a57a4bcbdd7219fc5e6921c387bc4c5fadf9dc88

   Although the `Chunk` object still needs to be expanded first, the `body` field now displays the keyword in the collapsed state without having to expand it:

   webide-png-object-tree-with-webide-repr-in-text-chunks

1. fully validate every chunk type using `valid/expr` - 0e3ab749d8626a5c5bc3946cd2c693e3df96403f

   The [PNG spec](https://www.w3.org/TR/2025/REC-png-3-20250624/#table51) requires that a chunk type consists entirely of uppercase and lowercase ASCII letters (`A-Z` and `a-z`). Now the `png.ksy` spec rejects files that don't meet this condition. For example:

   webide-png-object-tree-chunk-type-valid-expr-error

1. improve the `gAMA` chunk - 1c9455abe73f7eed437e7cbf87206423414c0b92

   Previously, the only value instance provided was `gamma_ratio`, which actually calculated the inverse / reciprocal of the gamma value (for example, it returned `2.2` when the actual gamma value in the chunk was `0.45455`).

   Now there are two value instances available: `gamma` (e.g. `1/2.2 = 0.45455`) and `inv_gamma`, which is the inverse of the gamma (calculated in exactly the same way as the previous `gamma_ratio`).

1. validate that the gamma value in `gAMA` is not 0 - 774bd7d2f0fb4979c99d9bbf377d11d784fa867f

1. decode property bits of the chunk type - f9f15c9f77664c97c9cbff8dd1734d82d9ad9d44 (provide `is_ancillary`, `is_private`, `reserved_bit` and `is_safe_to_copy` boolean value instances)

1. change the `doc-ref` links in the APNG chunks so that they point to https://www.w3.org/TR/png/ instead of https://wiki.mozilla.org/APNG_Specification - 014900e40c498a702180280c01f95d8c16f57b6f

1. add `-orig-id` to enum members associated with the APNG chunk `fcTL` - e7b8ec24538756b28e7d22b868cd7dfe869b795d (the [official spec](https://www.w3.org/TR/png/#fcTL-chunk) always refers to them by these `APNG_DISPOSE_OP_*` and `APNG_BLEND_OP_*` names, so I think makes sense to mention them)

1. validate enum fields in all chunks, mostly using `valid/in-enum` - e277cf2b4c2e5cc624d3f049d87e6a46f1fda2d0, c3849b18ceaa9b40c0431fda719b0948474408b1, 0801979fb8fe4f6608b6ca32cf950462f9d38e4e

1. improve docstrings in APNG chunks - 34d4f3eeb0a275973c49de3c738fcb02841f5ab3 (previously, some sentences were slightly inaccurate; now we use more wording from the official specification, plus some of my own)

1. add value instances `dots_per_inch_{x,y}` to the `pHYs` chunk, which calculate horizontal/vertical resolution (DPI) - 1594eeebf6a97d619bd653d7c33e26d7ca7863d1

### DIFF
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -257,13 +257,23 @@ types:
       y:
         value: y_int / 100000.0
   gama_chunk:
+    -webide-representation: '{gamma:dec} (= 1/{inv_gamma:dec})'
     doc-ref: https://www.w3.org/TR/png/#11gAMA
     seq:
       - id: gamma_int
         type: u4
+        doc: |
+          Image gamma multiplied by 100000 (a gamma value of 1/2.2 is stored as
+          45455)
     instances:
-      gamma_ratio:
+      gamma:
+        value: gamma_int / 100000.0
+        doc: Image gamma, typically 0.45455 = 1/2.2
+      inv_gamma:
         value: 100000.0 / gamma_int
+        doc: |
+          Inverse of the image gamma (1 / gamma), typically 2.2 (not considering
+          rounding)
   mdcv_chunk:
     doc-ref:
       - https://www.w3.org/TR/png/#mDCV-chunk
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -648,10 +648,14 @@ types:
       - id: dispose_op
         type: u1
         enum: dispose_op_values
+        valid:
+          in-enum: true
         doc: Type of frame area disposal to be done after rendering this frame
       - id: blend_op
         type: u1
         enum: blend_op_values
+        valid:
+          in-enum: true
         doc: Type of frame area rendering for this frame
     instances:
       delay:
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -366,16 +366,27 @@ types:
         type: u1
   international_text_chunk:
     doc: |
-      International text chunk effectively allows to store key-value string pairs in
-      PNG container. Both "key" (keyword) and "value" (text) parts are
-      given in pre-defined subset of ISO-8859-1 without control
-      characters.
+      International textual data (`iTXt`) chunk effectively allows you to store
+      key-value string pairs in the PNG container.
+
+      The "key" part (`keyword`) is restricted to printable ISO-8859-1 (Latin-1)
+      characters and spaces. The translated keyword and the "value" part
+      (`text`) are stored in UTF-8 and thus can store text in any language -
+      this language can be indicated via the language tag (`language_tag`).
     doc-ref: https://www.w3.org/TR/png/#11iTXt
     seq:
       - id: keyword
         type: strz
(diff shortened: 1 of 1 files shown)


armijnhemel (approved):
Very thorough! looking good.

PR #747, opened 2026-06-24
png: add notice required by the W3C license to `/doc`

generalmimon:
Since the `png.ksy` spec contains many excerpts from the W3C specification https://www.w3.org/TR/png/ to document types and fields using the `doc` key, I believe this notice is likely required.

https://www.w3.org/TR/2025/REC-png-3-20250624/ has the following notice:

> [Copyright](https://www.w3.org/policies/#copyright) © 1996-2025 [World Wide Web Consortium](https://www.w3.org/). W3C® [liability](https://www.w3.org/policies/#Legal_Disclaimer), [trademark](https://www.w3.org/policies/#W3C_Trademarks) and [permissive document license](https://www.w3.org/copyright/software-license-2023/) rules apply.

From this, I concluded that the W3C PNG spec is under the [Software and Document license - 2023 version](https://www.w3.org/copyright/software-license-2023/). I noticed there is also a [Document license - 2023 version](https://www.w3.org/copyright/document-license-2023/), but it doesn't seem to be used by the PNG spec (it's probably used by some other W3C documents).

### DIFF
--- a/image/png.ksy
+++ b/image/png.ksy
@@ -38,6 +38,54 @@ doc: |
   Copyright © 1996-2025 [World Wide Web Consortium](https://www.w3.org/).
   <https://www.w3.org/copyright/software-license-2023/>
 
+  The full text of the license for the original W3C PNG specification is
+  provided below:
+
+  > ## Software and Document license - 2023 version
+  >
+  > This work is being provided by the copyright holders under the following
+  > license.
+  >
+  > ### License
+  >
+  > By obtaining and/or copying this work, you (the licensee) agree that you
+  > have read, understood, and will comply with the following terms and
+  > conditions.
+  >
+  > Permission to copy, modify, and distribute this work, with or without
+  > modification, for any purpose and without fee or royalty is hereby granted,
+  > provided that you include the following on ALL copies of the work or
+  > portions thereof, including modifications:
+  >
+  > * The full text of this NOTICE in a location viewable to users of the
+  >   redistributed or derivative work.
+  > * Any pre-existing intellectual property disclaimers, notices, or terms and
+  >   conditions. If none exist, the [W3C software and document short
+  >   notice](https://www.w3.org/Consortium/Legal/2023/copyright-software-short-notice.html)
+  >   should be included.
+  > * Notice of any changes or modifications, through a copyright statement on
+  >   the new code or document such as "This software or document includes
+  >   material copied from or derived from [title and URI of the W3C document].
+  >   Copyright © [$year-of-document] [World Wide Web
+  >   Consortium](https://www.w3.org/).
+  >   <https://www.w3.org/copyright/software-license-2023/>"
+  >
+  > ### Disclaimers
+  >
+  > THIS WORK IS PROVIDED "AS IS," AND COPYRIGHT HOLDERS MAKE NO REPRESENTATIONS
+  > OR WARRANTIES, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO, WARRANTIES
+  > OF MERCHANTABILITY OR FITNESS FOR ANY PARTICULAR PURPOSE OR THAT THE USE OF
+  > THE SOFTWARE OR DOCUMENT WILL NOT INFRINGE ANY THIRD PARTY PATENTS,
+  > COPYRIGHTS, TRADEMARKS OR OTHER RIGHTS.
+  >
+  > COPYRIGHT HOLDERS WILL NOT BE LIABLE FOR ANY DIRECT, INDIRECT, SPECIAL OR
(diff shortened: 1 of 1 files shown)


image/png.ksy
```diff
@@ -32,6 +32,14 @@ meta:
     - exif
   endian: be
 doc: |
+  NOTICE: Many of the documentation comments (or docstrings) in this file were
+  copied from or derived from the [Portable Network Graphics (PNG) Specification
+  (Third Edition)](https://www.w3.org/TR/2025/REC-png-3-20250624/).
+  Copyright © 1996-2025 [World Wide Web Consortium](https://www.w3.org/).
+  <https://www.w3.org/copyright/software-license-2023/>
```
generalmimon, line 39: Maybe I should also copy-paste the full license text as per https://www.w3.org/copyright/software-license-2023/ 🤔

armijnhemel: would it make sense to include the short notice and then to add the full license text separately?

generalmimon: @armijnhemel:

> would it make sense to include the short notice and then to add the full license text separately?

This highlights a deeper problem: there is no established practice in this repository of including full license texts. This is despite the fact that the [MIT](https://choosealicense.com/licenses/mit/) license that [23 .ksy files already use](https://github.com/search?q=repo%3Akaitai-io%2Fkaitai_struct_formats+path%3A%2F%5C.ksy%24%2F+%22license%3A+MIT%22&type=code) (see also https://formats.kaitai.io/xref.html) also requires distributing the full license text, yet we don't do so.

Given that we currently use the `/meta/license` key to specify a [SPDX license expression](https://spdx.github.io/spdx-spec/v3.0.1/annexes/spdx-license-expressions/), I think it makes sense to continue doing so. The W3C license actually has a SPDX identifier [`W3C-20150513`](https://spdx.org/licenses/W3C-20150513.html), but I'm not sure if it makes sense to include it in `/meta/license`, because it's not like W3C itself implemented the .ksy spec. The W3C license is quite permissive and reasonable, though, so it probably wouldn't be a problem to relicense the .ksy to `/meta/license: W3C-20150513`. To me, it feels a bit weird, but at least it would simplify the confusing situation of "the .ksy implementation is under `CC0-1.0`, but the documentation comments / docstrings are `W3C-20150513`" 🤔

I see a few ways out of this:

1. Keep `/meta/license: CC0-1.0`, include the W3C license text in `/doc` next to the notice.

   I feel like the advantage of this is that the license text would then be automatically included in the generated code as well.

2. Keep `/meta/license: CC0-1.0`, include the W3C license text as a YAML comment.

   Just like the previous approach, but a YAML comment will not be included in the generated code.

3. Keep `/meta/license: CC0-1.0`, include the W3C license text as a separate file (the [REUSE tool](https://codeberg.org/fsfe/reuse-tool) provided as part of the [REUSE](https://reuse.software/) project would choose the path `LICENSES/W3C-20150513.txt`, just like [in our samples repo](https://github.com/kaitai-io/kaitai_struct_samples/tree/5f2b1a917c8ad8841863bc968a3e552d3d4e7974/LICENSES), where we use the REUSE tool) and link to it from `/doc` next to the notice.

   With this approach, I'm not sure whether `image/png.ksy` should have a relative link like `[W3C license](../LICENSES/W3C-20150513.txt)` or an absolute link like `[W3C license](https://github.com/kaitai-io/kaitai_struct_formats/blob/{'master' or commit SHA}/LICENSES/W3C-20150513.txt)`. The relative link would be broken at https://formats.kaitai.io/ (unless we made changes to the site generator) or if someone copied the `png.ksy` spec to their own project. The absolute link would work everywhere, but then I don't see a difference compared to just linking to https://www.w3.org/copyright/software-license-2023/ as this PR already does.

4. Relicense to `/meta/license: W3C-20150513`.

   The advantage would be a simpler licensing situation - not only for humans, but also for tools. We could (and probably should) then use something like https://codeberg.org/fsfe/reuse-tool to enforce the requirement that all referenced license identifiers must have a corresponding license text in the `LICENSES/` directory. The REUSE tool currently doesn't recognize the `.ksy` file extension and would only pick up license identifiers from `SPDX-License-Identifier` tags, typically in a code comment at the beginning of the file. But we could adapt it to our needs - at least recognizing the `.ksy` suffix would be good 🙂

---

Honestly, I'd rather not deal with this and move on to more fun things, but then I probably shouldn't have brought it up in the first place 😭

armijnhemel: In the end it is about whether or not people are correctly informed. I would choose either 1 or 3.

generalmimon: I went with 1, see fabdd9d5b47c1ee562cf420a08f2b794516c000d

PR #749, opened 2026-06-26
Add Norton Utilities File Info format

dfandrich:
I'm not sure if database/ is the best location, but I didn't see anything better.
There is a VM at https://www.pcjs.org/software/pcx86/util/norton/4.50/advanced/
with Norton Utilities installed. Norton Utilities 4.50 (Adv 2) has the NI.EXE
program that provides usage information, and Adv 3 has the program itself,
along with a populated fileinfo.fi file that can be used for testing.


dfandrich:
I've created a [program](https://github.com/dfandrich/fileinfofi/) that uses this definition and it includes a sample [test file](https://github.com/dfandrich/fileinfofi/blob/master/fileinfo.fi).

database/fileinfo_fi.ksy
```diff
@@ -0,0 +1,59 @@
...
+  tags:
+    - dos
+  license: CC0-1.0
+  endian: le
+doc: |
+  fileinfo.fi files store detailed file description alongside file names
+  that are displayed by the fi.exe program. This format was used by Norton
+  Utilities 4 on MS-DOS systems.
+doc-ref:
+  - https://en.wikipedia.org/wiki/Norton_Utilities
```
generalmimon: `doc-ref` is typically used for resources that serve as a reference for understanding the structure of the format, so a link to the general page about the application associated with the format is not a good fit.

Is there any other resource that specifically covers this format and its structure? If so, it would be better to use that one instead.

database/fileinfo_fi.ksy
```diff
@@ -0,0 +1,59 @@
...
+        contents: ['PNCI', 0]
+      - id: num_records
+        type: u2
+        doc: Number of records in the file.
+      - id: checksum
+        type: u2
+        doc: Simple checksum of remaining header bytes.
+      - id: bitmask
+        size: 128
+        doc: >-
```
generalmimon: We always use `doc: |`, so let's stick to this convention:

```suggestion
        doc: |
```

database/fileinfo_fi.ksy
```diff
@@ -0,0 +1,59 @@
...
+          Bit mask of valid records, starting with the highest bit in the first
+          byte and ending with the lowest bit in the bitmask_len byte. This
+          structure limits the number of records in the file to 1024.
+  record:
+    seq:
+      - id: file_name
+        size: 12
+        type: str
+        terminator: 0
```
generalmimon: It's idiomatic to abbreviate this to `type: strz`:

```suggestion
        type: strz
```

database/fileinfo_fi.ksy
```diff
@@ -0,0 +1,59 @@
...
+        doc: Number of records in the file.
+      - id: checksum
+        type: u2
+        doc: Simple checksum of remaining header bytes.
+      - id: bitmask
+        size: 128
+        doc: >-
+          Bit mask of valid records, starting with the highest bit in the first
+          byte and ending with the lowest bit in the bitmask_len byte. This
+          structure limits the number of records in the file to 1024.
```
generalmimon: Technically, it's possible to parse this in Kaitai Struct down to the individual bits:

```ksy
      - id: bitmask
        type: b1
        repeat-expr: 1024
```

(Perhaps it should then be renamed from `bitmask` to some better name... 🤔)

It would also be recommended to add this (https://doc.kaitai.io/user_guide.html#bit-endian):

```ksy
meta:
  bit-endian: be
```

In the current Kaitai Struct implementation, the resulting type would be a list of booleans (because `b1` is automatically mapped to `bool` by default), which in most languages means 1 byte per element. So it wouldn't be the most efficient thing, but for 1024 elements, the overhead is certainly negligible. https://github.com/kaitai-io/kaitai_struct/issues/112 proposes adding bitsets as a feature to Kaitai Struct, but it apparently hasn't gained traction (yet).

generalmimon (changes requested):

database/fileinfo_fi.ksy
dfandrich: I found next to nothing about this program. The ksy file is now the best documentation on the file format that I know of, and that was created by reverse engineering.

database/fileinfo_fi.ksy
dfandrich: Done.

database/fileinfo_fi.ksy
dfandrich: Done.

database/fileinfo_fi.ksy
dfandrich: This looks a lot cleaner, and the extra overhead is unimportant. Done.

database/fileinfo_fi.ksy
```diff
@@ -0,0 +1,105 @@
+meta:
+  id: fileinfo_fi
```
generalmimon: I propose renaming this format to `norton_fileinfo_fi` (both the value of `/meta/id` and the .ksy file name, which should always be kept in sync). `fileinfo` is a rather generic name, so I think this longer name will make the format less mysterious when skimming through the list of formats at https://formats.kaitai.io/, for example.

database/fileinfo_fi.ksy
```diff
@@ -0,0 +1,105 @@
...
+      - id: checksum
+        type: u2
+        doc: Checksum of remaining header bytes, equal to header_checksum.
+      - id: record_in_use
+        type: b1
+        repeat: expr
+        repeat-expr: 1024
+        doc: |
+          Bit mask of valid records, starting with the highest bit in the first
+          byte and ending with the lowest bit in the bitmask_len byte. This
```
generalmimon: I'm not sure what "bitmask_len" refers to? There is no such field in the .ksy spec.

database/fileinfo_fi.ksy
```diff
@@ -0,0 +1,105 @@
...
+            (_root.header.record_in_use[idx * 8 + 4].as<u2> << 3) +
+            (_root.header.record_in_use[idx * 8 + 5].as<u2> << 2) +
+            (_root.header.record_in_use[idx * 8 + 6].as<u2> << 1) +
+            (_root.header.record_in_use[idx * 8 + 7].as<u2> << 0))
+        doc: |
+          Add one additional checksum byte from 8 record_in_use bits. The
+          calculation is done in groups of 8 to reduce the amount of recursion
+          necessary to calculate the entire record_in_use checksum, which
+          otherwise exceeds the maximum Python recursion limit.
+instances:
```
generalmimon: Please move the `instances` section before `types`. This is a rule documented in https://doc.kaitai.io/ksy_style_guide.html#type. It generally makes .ksy specs more readable, because you don't need to scroll through the entire `types` section (which can be very long) just to get to the `instances` section, which actually applies to the root level again, just like `seq` (so `seq` should come immediately before `instances`).

database/fileinfo_fi.ksy
```diff
@@ -0,0 +1,105 @@
...
+            (_root.header.record_in_use[idx * 8 + 3].as<u2> << 4) +
+            (_root.header.record_in_use[idx * 8 + 4].as<u2> << 3) +
+            (_root.header.record_in_use[idx * 8 + 5].as<u2> << 2) +
+            (_root.header.record_in_use[idx * 8 + 6].as<u2> << 1) +
+            (_root.header.record_in_use[idx * 8 + 7].as<u2> << 0))
+        doc: |
+          Add one additional checksum byte from 8 record_in_use bits. The
+          calculation is done in groups of 8 to reduce the amount of recursion
+          necessary to calculate the entire record_in_use checksum, which
+          otherwise exceeds the maximum Python recursion limit.
```
generalmimon: The recursion can actually be avoided, you just need to pass the previous intermediate result into a parameter as well:

```ksy
instances:
  record_checksum_reduction:
    type: 'record_checksum_iter(_index, _index == 0 ? 0 : record_checksum_reduction[_index - 1].res)'
    repeat: expr
    repeat-expr: _root.header.record_in_use.size / 8
    doc: |
      Structure used in calculating record_checksum.
```

```ksy
  record_checksum_iter:
    params:
      - id: idx
        type: u1
        doc: |
          The index into the record_in_use array in calculating the checksum.
          This is the index into groups of 8 bits in record_in_use, so ranges
          from 0..127.
      - id: prev
        type: u2
```

Note that this also removes the need to use the `.as<record_checksum_iter>` and `.as<u2>` type casts.

database/fileinfo_fi.ksy
```diff
@@ -0,0 +1,105 @@
...
+    type: record_checksum_iter(_index)
+    repeat: expr
+    repeat-expr: _root.header.record_in_use.size / 8
+    doc: |
+      Structure used in calculating record_checksum.
+  record_checksum:
+    value: record_checksum_reduction[_root.header.record_in_use.size / 8 - 1].res
+    doc: |
+      Partial sum used in calculating header_checksum.
+  header_checksum:
```
generalmimon:
```suggestion
  expected_header_checksum:
```

database/fileinfo_fi.ksy
```diff
@@ -0,0 +1,105 @@
...
+    value: (header.magic[0].as<u1> + header.magic[1].as<u1> + header.magic[2].as<u1> + header.magic[3].as<u1> + header.num_records / 256 + header.num_records % 256 + record_checksum) % 65536
```
generalmimon: It's advisable to format such long expressions over multiple lines, for example as follows:

```suggestion
    value: |
      (
        header.magic[0].as<u1>
        + header.magic[1].as<u1>
        + header.magic[2].as<u1>
        + header.magic[3].as<u1>
        + header.num_records / 256
        + header.num_records % 256
        + record_checksum
      ) % 65536
```

database/fileinfo_fi.ksy
```diff
@@ -0,0 +1,105 @@
...
+        value: |
+          prev + (
+            (_root.header.record_in_use[idx * 8 + 0].as<u2> << 7) +
+            (_root.header.record_in_use[idx * 8 + 1].as<u2> << 6) +
+            (_root.header.record_in_use[idx * 8 + 2].as<u2> << 5) +
+            (_root.header.record_in_use[idx * 8 + 3].as<u2> << 4) +
+            (_root.header.record_in_use[idx * 8 + 4].as<u2> << 3) +
+            (_root.header.record_in_use[idx * 8 + 5].as<u2> << 2) +
+            (_root.header.record_in_use[idx * 8 + 6].as<u2> << 1) +
+            (_root.header.record_in_use[idx * 8 + 7].as<u2> << 0))
```
generalmimon: First of all, the correct way to convert a boolean into an integer is to use the [`bool.to_i`](https://doc.kaitai.io/user_guide.html#bool-methods) method. The `.as<u2>` cast will (by design) do nothing in dynamically typed languages, which in this case is not a problem in Python, where `bool` is a subclass of `int`, but for example in Ruby, `true << 5` fails with `undefined method '<<' for true (NoMethodError)`.

This is a relatively common mistake and the compiler should reject this, but unfortunately, it currently doesn't check the validity of type casts at all - see https://github.com/kaitai-io/kaitai_struct/issues/696.

---

After you apply my suggestion above to eliminate recursion - for compatibility with our statically typed target languages, it might be necessary to cast this to `u2` (currently, the derived type of this value instance is `s4`, which won't fit into the `prev` parameter, which is of type `u2`):

```suggestion
      res:
        value: |
          (prev + (
            (_root.header.record_in_use[idx * 8 + 0].to_i << 7) +
            (_root.header.record_in_use[idx * 8 + 1].to_i << 6) +
            (_root.header.record_in_use[idx * 8 + 2].to_i << 5) +
            (_root.header.record_in_use[idx * 8 + 3].to_i << 4) +
            (_root.header.record_in_use[idx * 8 + 4].to_i << 3) +
            (_root.header.record_in_use[idx * 8 + 5].to_i << 2) +
            (_root.header.record_in_use[idx * 8 + 6].to_i << 1) +
            (_root.header.record_in_use[idx * 8 + 7].to_i << 0)
          )).as<u2>
```

database/fileinfo_fi.ksy
```diff
@@ -0,0 +1,105 @@
...
+          is often IBM437 but can be something else depending on the code page
+          in use when the entry was created.
+  record_checksum_iter:
+    params:
+      - id: idx
+        type: u1
+        doc: |
+          The index into the record_in_use array in calculating the checksum.
+          This is the index into groups of 8 bits in record_in_use, so ranges
+          from 0..127.
```
generalmimon: Wrapping identifiers in backticks makes them easier to spot when reading the text:

```suggestion
        doc: |
          The index into the `record_in_use` array in calculating the checksum.
          This is the index into groups of 8 bits in `record_in_use`, so ranges
          from 0..127.
```

I noticed there are more unformatted identifiers in `doc`s throughout the .ksy - please correct them as well.

See also https://doc.kaitai.io/ksy_style_guide.html#doc:

> - use backticks `` ` `` to wrap identifiers and small pieces of code

generalmimon (changes requested):

database/fileinfo_fi.ksy
generalmimon: @dfandrich:

> The ksy file is now the best documentation on the file format that I know of, and that was created by reverse engineering.

That makes it even more valuable. Thank you for your effort!

I'm still convinced that this link doesn't belong to `doc-ref`, since `doc-ref` is reserved for references to format documentation. Maybe you could incorporate this link in `/doc` instead?

The top-level `doc` key supports Markdown (specifically [CommonMark](https://commonmark.org/) - see https://doc.kaitai.io/ksy_style_guide.html#doc) and it will be properly formatted at https://formats.kaitai.io/.

dfandrich: Added as a link in the doc section.

database/fileinfo_fi.ksy
dfandrich: Done.

database/fileinfo_fi.ksy
dfandrich: This was a holdover from an earlier version. Removed.

database/fileinfo_fi.ksy
dfandrich: Done.

database/fileinfo_fi.ksy
dfandrich: Done.

database/fileinfo_fi.ksy
dfandrich: I thought "expected" could be ambiguous (it could mean the expected value of the calculated checksum) so I changed it to `calculated_header_checksum`.

database/fileinfo_fi.ksy
dfandrich: Done.

database/fileinfo_fi.ksy
dfandrich: Done.

database/fileinfo_fi.ksy
dfandrich: Done.

database/fileinfo_fi.ksy
dfandrich: I also removed the 8-bit loop unrolling to simplify the calculation.

PR #752, opened 2026-06-29
elf: pull changes from BANG

generalmimon:
This incorporates the changes made to the downstream `elf.ksy` spec in the the BANG project by @armijnhemel: https://github.com/armijnhemel/binaryanalysis-ng/blob/19570ae9156878076b10d11c4fec44d3beb27785/src/bang/parsers/executable/elf/elf.ksy

### DIFF
--- a/executable/elf.ksy
+++ b/executable/elf.ksy
@@ -835,24 +835,132 @@ enums:
     1: le
     # ELFDATA2MSB
     2: be
+  # https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=include/elf/common.h;h=1ae68221a89723773b4ec5bf17c7455def7b90b8;hb=refs/tags/binutils-2_46_1#l60
+  # https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/elf.h;h=46a01281cb0fb5322d5124f0443c11dea4d5b721;hb=refs/tags/glibc-2.43#l134
+  # https://github.com/llvm/llvm-project/blob/ca7933e47d3a3451d81e72ac174dcb5aa28b59d1/llvm/include/llvm/BinaryFormat/ELF.h#L344 (Git tag "llvmorg-22.1.8")
+  # https://gabi.xinuos.com/v42/elf/b-osabi.html
   os_abi:
-    0: system_v
-    1: hp_ux
-    2: netbsd
-    3: gnu
-    6: solaris
-    7: aix
-    8: irix
-    9: freebsd
-    0xa: tru64 # Compaq TRU64 UNIX
-    0xb: modesto # Novell Modesto
-    0xc: openbsd
-    0xd: openvms
-    0xe: nsk # Hewlett-Packard Non-Stop Kernel
-    0xf: aros # Amiga Research OS
-    0x10: fenixos # The FenixOS highly scalable multi-core OS
-    0x11: cloudabi # Nuxi CloudABI
-    0x12: openvos # Stratus Technologies OpenVOS
+    0:
+      id: system_v
+      -orig-id:
+        - ELFOSABI_SYSV
+        - ELFOSABI_NONE
+      doc: UNIX System V ABI
+    1:
+      id: hp_ux
+      -orig-id: ELFOSABI_HPUX
+      doc: HP-UX
+    2:
+      id: netbsd
+      -orig-id: ELFOSABI_NETBSD
+      doc: NetBSD
+    3:
+      id: gnu
+      -orig-id:
+        - ELFOSABI_GNU
+        - ELFOSABI_LINUX
+      doc: Object uses GNU ELF extensions.
+    6:
+      id: solaris
+      -orig-id: ELFOSABI_SOLARIS
+      doc: Solaris
+    7:
+      id: aix
+      -orig-id: ELFOSABI_AIX
+      doc: IBM AIX
+    8:
+      id: irix
+      -orig-id: ELFOSABI_IRIX
+      doc: IRIX by Silicon Graphics (SGI)
+    9:
+      id: freebsd
+      -orig-id: ELFOSABI_FREEBSD
+      doc: FreeBSD
+    10:
+      id: tru64
+      -orig-id: ELFOSABI_TRU64
+      doc: Compaq TRU64 UNIX
+    11:
+      id: modesto
+      -orig-id: ELFOSABI_MODESTO
+      doc: Novell Modesto
+    12:
+      id: openbsd
+      -orig-id: ELFOSABI_OPENBSD
+      doc: OpenBSD
+    13:
+      id: openvms
+      -orig-id: ELFOSABI_OPENVMS
+      doc: OpenVMS
+    14:
+      id: nsk
+      -orig-id: ELFOSABI_NSK
+      doc: Hewlett-Packard NonStop Kernel
+    15:
+      id: aros
+      -orig-id: ELFOSABI_AROS
+      doc: AROS Research Operating System
+    16:
+      id: fenixos
+      -orig-id: ELFOSABI_FENIXOS
+      doc: FenixOS
+    17:
+      id: cloudabi
+      -orig-id: ELFOSABI_CLOUDABI
+      doc: Nuxi CloudABI
+    18:
+      id: openvos
+      -orig-id: ELFOSABI_OPENVOS
+      doc: Stratus Technologies OpenVOS
+    51:
+      id: cuda
+      -orig-id: ELFOSABI_CUDA
+      doc: NVIDIA CUDA architecture
+      doc-ref:
+        - https://github.com/llvm/llvm-project/blob/ca7933e47d3a3451d81e72ac174dcb5aa28b59d1/llvm/include/llvm/BinaryFormat/ELF.h#L364 Git tag "llvmorg-22.1.8"
+        - https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=include/elf/common.h;h=1ae68221a89723773b4ec5bf17c7455def7b90b8;hb=refs/tags/binutils-2_46_1#l79
+        - 'https://docs.nvidia.com/cuda/cuda-binary-utilities/index.html search for `"ei_osabi": 51,`'
+    # 64-255: Architecture-specific value range
+    64:
+      id: arm_aeabi
+      -orig-id: ELFOSABI_ARM_AEABI
+      doc: ARM EABI (symbol versioning extensions)
+    # 64:
+    #   id: amdgpu_hsa
+    #   -orig-id: ELFOSABI_AMDGPU_HSA
+    #   doc: AMD HSA runtime
+    #   doc-ref: https://github.com/llvm/llvm-project/blob/ca7933e47d3a3451d81e72ac174dcb5aa28b59d1/llvm/include/llvm/BinaryFormat/ELF.h#L367 Git tag "llvmorg-22.1.8"
+    # 64:
+    #   id: c6000_elfabi
+    #   -orig-id: ELFOSABI_C6000_ELFABI
+    #   doc: Bare-metal TMS320C6000
+    65:
+      id: arm_fdpic
+      -orig-id: ELFOSABI_ARM_FDPIC
+      doc: ARM FDPIC
+      doc-ref: https://github.com/llvm/llvm-project/blob/ca7933e47d3a3451d81e72ac174dcb5aa28b59d1/llvm/include/llvm/BinaryFormat/ELF.h#L371 Git tag "llvmorg-22.1.8"
+    # 65:
+    #   id: amdgpu_pal
+    #   -orig-id: ELFOSABI_AMDGPU_PAL
+    #   doc: AMD PAL runtime
+    #   doc-ref: https://github.com/llvm/llvm-project/blob/ca7933e47d3a3451d81e72ac174dcb5aa28b59d1/llvm/include/llvm/BinaryFormat/ELF.h#L368 Git tag "llvmorg-22.1.8"
+    # 65:
+    #   id: c6000_linux
+    #   -orig-id: ELFOSABI_C6000_LINUX
+    #   doc: Linux TMS320C6000
+    66:
+      id: amdgpu_mesa3d
+      -orig-id: ELFOSABI_AMDGPU_MESA3D
+      doc: AMD GCN GPUs (GFX6+) for MESA runtime
+      doc-ref: https://github.com/llvm/llvm-project/blob/ca7933e47d3a3451d81e72ac174dcb5aa28b59d1/llvm/include/llvm/BinaryFormat/ELF.h#L369 Git tag "llvmorg-22.1.8"
+    97:
+      id: arm
+      -orig-id: ELFOSABI_ARM
+      doc: ARM
+    255:
+      id: standalone
+      -orig-id: ELFOSABI_STANDALONE
+      doc: Standalone (embedded) application
   # e_type
   obj_type:
     # ET_NONE
--- a/executable/elf.ksy
+++ b/executable/elf.ksy
@@ -101,52 +101,76 @@ types:
       mask_proc:
         value: value & 0xf0000000 != 0
   section_header_flags:
+    doc-ref: https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/elf.h;h=46a01281cb0fb5322d5124f0443c11dea4d5b721;hb=refs/tags/glibc-2.43#l468
     params:
       - id: value
         type: u4
     instances:
       write:
+        -orig-id: SHF_WRITE
         value: value & 0x01 != 0
-        doc: "writable"
+        doc: Writable
       alloc:
+        -orig-id: SHF_ALLOC
         value: value & 0x02 != 0
-        doc: "occupies memory during execution"
+        doc: Occupies memory during execution
       exec_instr:
+        -orig-id: SHF_EXECINSTR
         value: value & 0x04 != 0
-        doc: "executable"
+        doc: Executable
       merge:
+        -orig-id: SHF_MERGE
         value: value & 0x10 != 0
-        doc: "might be merged"
+        doc: Might be merged
       strings:
+        -orig-id: SHF_STRINGS
         value: value & 0x20 != 0
-        doc: "contains nul-terminated strings"
+        doc: Contains nul-terminated strings
       info_link:
+        -orig-id: SHF_INFO_LINK
         value: value & 0x40 != 0
-        doc: "'sh_info' contains SHT index"
+        doc: |
+          `sh_info` contains SHT index
       link_order:
+        -orig-id: SHF_LINK_ORDER
         value: value & 0x80 != 0
-        doc: "preserve order after combining"
-      os_non_conforming:
+        doc: Preserve order after combining
+      os_nonconforming:
+        -orig-id: SHF_OS_NONCONFORMING
         value: value & 0x100 != 0
-        doc: "non-standard OS specific handling required"
+        doc: Non-standard OS specific handling required
       group:
+        -orig-id: SHF_GROUP
         value: value & 0x200 != 0
-        doc: "section is member of a group"
+        doc: Section is member of a group
       tls:
+        -orig-id: SHF_TLS
         value: value & 0x400 != 0
-        doc: "section hold thread-local data"
-      ordered:
-        value: value & 0x04000000 != 0
-        doc: "special ordering requirement (Solaris)"
-      exclude:
-        value: value & 0x08000000 != 0
-        doc: "section is excluded unless referenced or allocated (Solaris)"
+        doc: Section hold thread-local data
+      compressed:
+        -orig-id: SHF_COMPRESSED
+        value: value & 0x800 != 0
+        doc: Section with compressed data
       mask_os:
-        value: value & 0x0ff00000 != 0
-        doc: "OS-specific"
+        -orig-id: SHF_MASKOS
+        value: value & 0x0ff0_0000 != 0
+        doc: OS-specific
       mask_proc:
-        value: value & 0xf0000000 != 0
-        doc: "Processor-specific"
+        -orig-id: SHF_MASKPROC
+        value: value & 0xf000_0000 != 0
+        doc: Processor-specific
+      retain:
+        -orig-id: SHF_GNU_RETAIN
+        value: value & 0x0020_0000 != 0
+        doc: Not to be GCed by linker
+      ordered:
+        -orig-id: SHF_ORDERED
+        value: value & 0x4000_0000 != 0
+        doc: Special ordering requirement (Solaris)
+      exclude:
+        -orig-id: SHF_EXCLUDE
+        value: value & 0x8000_0000 != 0
+        doc: Section is excluded unless referenced or allocated (Solaris)
   dt_flag_values:
     doc-ref:
       - 'https://refspecs.linuxbase.org/elf/gabi4+/ch5.dynamic.html Figure 5-11: DT_FLAGS values'
--- a/executable/elf.ksy
+++ b/executable/elf.ksy
@@ -102,6 +102,7 @@ types:
         value: value & 0xf0000000 != 0
   section_header_flags:
     doc-ref:
+      - https://docs.oracle.com/en/operating-systems/solaris/oracle-solaris/11.4/linkers-libraries/section-headers.html#GUID-2CBE4879-2E76-426E-BB7F-CF0CB1D87C52__CHAPTER6-10675
       - https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=include/elf/common.h;h=1ae68221a89723773b4ec5bf17c7455def7b90b8;hb=refs/tags/binutils-2_46_1#l614
       - https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/elf.h;h=46a01281cb0fb5322d5124f0443c11dea4d5b721;hb=refs/tags/glibc-2.43#l468
     params:
@@ -110,20 +111,20 @@ types:
     instances:
       write:
         -orig-id: SHF_WRITE
-        value: value & 0x01 != 0
-        doc: Writable data during execution
+        value: value & 0x1 != 0
+        doc: Writable during execution
       alloc:
         -orig-id: SHF_ALLOC
-        value: value & 0x02 != 0
+        value: value & 0x2 != 0
         doc: Occupies memory during execution
       exec_instr:
         -orig-id: SHF_EXECINSTR
-        value: value & 0x04 != 0
+        value: value & 0x4 != 0
         doc: Executable machine instructions
       merge:
         -orig-id: SHF_MERGE
         value: value & 0x10 != 0
-        doc: Data in this section can be merged
+        doc: Data in this section can be merged to eliminate duplication
       strings:
         -orig-id: SHF_STRINGS
         value: value & 0x20 != 0
@@ -132,7 +133,7 @@ types:
         -orig-id: SHF_INFO_LINK
         value: value & 0x40 != 0
         doc: |
-          `sh_info` holds section header table index
+          Section header's `sh_info` field holds a section header table index
       link_order:
         -orig-id: SHF_LINK_ORDER
         value: value & 0x80 != 0
@@ -140,7 +141,7 @@ types:
       os_nonconforming:
         -orig-id: SHF_OS_NONCONFORMING
         value: value & 0x100 != 0
-        doc: Non-standard OS specific handling required
+        doc: Special OS-specific handling required
       group:
         -orig-id: SHF_GROUP
         value: value & 0x200 != 0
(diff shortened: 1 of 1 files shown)


executable/elf.ksy
```diff
@@ -1040,619 +1091,873 @@ enums:
...
       id: picojava
       -orig-id: EM_PJ
       doc: picoJava
     92:
-      id: openrisc
-      doc: OpenRISC 32-bit embedded processor
+      id: or1k
+      -orig-id:
+        - EM_OR1K
+        - EM_OPENRISC # Old constant that might be in use by some software.
```
armijnhemel, line 1981: glibc uses EM_OPENRISC

https://gitlab.com/gnutools/glibc/-/blob/master/elf/elf.h?ref_type=heads#L249

armijnhemel: On the other hand: https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=commit;h=73589c9dbddc7906fa6a150f2a2a0ff6b746e8ba

armijnhemel (approved):

executable/elf.ksy
generalmimon: Yeah, it's a bit complicated:

- https://www.openwall.com/lists/musl/2014/07/17/3
- https://git.musl-libc.org/cgit/musl/commit/include/elf.h?h=v1.1.18&id=308e4bc61883c82aeda70e20367b656cd35899f1

As you pointed out, binutils renamed `EM_OPENRISC` to `EM_OR1K` [in commit `73589c9dbdd`](https://sourceware.org/git/?p=binutils-gdb.git;a=blobdiff;f=include/elf/common.h;h=e8ae3ac2e97c0053255765195b782584dabf75b4;hp=9643a3959ac3b8d50de6d0e46469a957d156326d;hb=73589c9dbddc7906fa6a150f2a2a0ff6b746e8ba;hpb=a75fef0e5bc3bfeb4eac434c1c68831f55ab597d) and also removed the `EM_OPENRISC_OLD` (`0x3426`) and `EM_OR32` (`0x8472`) constants. However, other projects generally still use `EM_OPENRISC`.

It's interesting that https://openrisc.io/revisions/r1.4, which was last updated February 20, 2022 (which wasn't that long ago), still states this:

> The `e_machine` member of the ELF header contains the decimal value 33906 (hexadecimal 0x8472) that is defined as the name EM_OR32.

I don't know if this is intentional or they just forgot to update it. It's frustrating that there is no single source we can trust...

---

To be honest, I don't care whether we use `or1k` or `openrisc`. I renamed it to `or1k` because binutils convinced me that it was the new, more correct name. That might very well be true, but it's a bit weird when most other projects (except the ones that have clearly adopted the entire set of binutils constants verbatim) keep using `EM_OPENRISC`. Which one would you prefer?

armijnhemel: So there are three or four values? Fun! I think that or1k makes sense then.

executable/elf.ksy
```diff
@@ -174,214 +224,303 @@ types:
...
+            io: _root._io
+            pos: ofs_body
+            size: len_body
+            type:
+              switch-on: type
+              cases:
+                'ph_type::interp': ph_interpreter
+                'ph_type::dynamic': ph_dynamic_section
+                'ph_type::note': note_section
+            if: type != ph_type::null_type
```
generalmimon: @armijnhemel I'm somewhat unsure about whether this will work. BANG has the following condition instead (https://github.com/armijnhemel/binaryanalysis-ng/blob/19570ae9156878076b10d11c4fec44d3beb27785/src/bang/parsers/executable/elf/elf.ksy#L400):

```ksy
            if: offset != 0 and filesz != 0
```

However, I see this as something reached empirically, not actually following from the specification, and I'd like it to follow from the specification, or at least perhaps from an existing established implementation...

What makes it clear that this is an empiric condition is this commit - https://github.com/armijnhemel/binaryanalysis-ng/commit/1c83d31888f8bd93eb0037791c8bd74351388767:

> **elf: fix program header parsing for ELF files found in .gnu_debugdata sections**
>
> ```diff
> -            if: offset != 0
> +            if: offset != 0 and filesz != 0
> ```

Perhaps you might remember where the ELF files where this change was needed came from, and how to get them for testing?

armijnhemel: The data in these .gnu_debugdata sections are actually full ELF files, but most of the sections are of type `NOBITS` and for those sections `filesz` doesn't make sense. For example, `vim` in F39:

```
There are 32 section headers, starting at offset 0x3b8a8:

Section Headers:
  [Nr] Name              Type            Address          Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            0000000000000000 000000 000000 00      0   0  0
  [ 1] .interp           NOBITS          0000000000000318 000318 00001c 00   A  0   0  1
  [ 2] .note.gnu.property NOTE            0000000000000338 000334 000050 00   A  0   0  8
  [ 3] .note.gnu.build-id NOTE            0000000000000388 000384 000024 00   A  0   0  4
  [ 4] .note.ABI-tag     NOTE            00000000000003ac 0003a8 000020 00   A  0   0  4
  [ 5] .note.package     NOTE            00000000000003cc 0003c8 000088 00   A  0   0  4
  [ 6] .gnu.hash         NOBITS          0000000000000458 000450 00005c 00   A  7   0  8
  [ 7] .dynsym           NOBITS          00000000000004b8 000450 001848 18   A  8   1  8
  [ 8] .dynstr           NOBITS          0000000000001d00 000450 0009ee 00   A  0   0  1
  [ 9] .gnu.version      NOBITS          00000000000026ee 000450 000206 02   A  7   0  2
  [10] .gnu.version_r    NOBITS          00000000000028f8 000450 000150 00   A  8   4  8
  [11] .rela.dyn         NOBITS          0000000000002a48 000450 03af08 18   A  7   0  8
  [12] .rela.plt         NOBITS          000000000003d950 000450 0016b0 18  AI  7  25  8
  [13] .init             NOBITS          000000000003f000 001000 00001b 00  AX  0   0  4
  [14] .plt              NOBITS          000000000003f020 001000 000f30 10  AX  0   0 16
  [15] .plt.sec          NOBITS          000000000003ff50 001000 000f20 10  AX  0   0 16
  [16] .text             NOBITS          0000000000040e70 001000 322a23 00  AX  0   0 16
  [17] .fini             NOBITS          0000000000363894 001000 00000d 00  AX  0   0  4
  [18] .rodata           NOBITS          0000000000364000 001000 028684 00   A  0   0 32
  [19] .eh_frame_hdr     NOBITS          000000000038c684 001000 00b2ec 00   A  0   0  4
  [20] .eh_frame         NOBITS          0000000000397970 001000 038dd4 00   A  0   0  8
  [21] .init_array       NOBITS          00000000003d1c10 000c10 000008 08  WA  0   0  8
  [22] .fini_array       NOBITS          00000000003d1c18 000c10 000008 08  WA  0   0  8
  [23] .data.rel.ro      NOBITS          00000000003d1c20 000c10 0129c0 00  WA  0   0 32
  [24] .dynamic          NOBITS          00000000003e45e0 000c10 000250 10  WA  8   0  8
  [25] .got              NOBITS          00000000003e4830 000c10 0007d0 08  WA  0   0  8
  [26] .data             NOBITS          00000000003e5000 000c10 03381d 00  WA  0   0 32
  [27] .bss              NOBITS          0000000000418820 000c10 00e5d8 00  WA  0   0 32
  [28] .annobin.notes    STRTAB          0000000000000000 000450 000125 01  MS  0   0  1
  [29] .symtab           SYMTAB          0000000000000000 000578 0219c0 18     30 4394  8
  [30] .strtab           STRTAB          0000000000000000 021f38 019839 00      0   0  1
  [31] .shstrtab         STRTAB          0000000000000000 03b771 000137 00      0   0  1
Key to Flags:
  W (write), A (alloc), X (execute), M (merge), S (strings), I (info),
  L (link order), O (extra OS processing required), G (group), T (TLS),
  C (compressed), x (unknown), o (OS specific), E (exclude),
  D (mbind), l (large), p (processor specific)
```

armijnhemel: [vim.debug.zip](https://github.com/user-attachments/files/29971084/vim.debug.zip)

generalmimon: @armijnhemel Thanks for providing the sample file - it was very helpful.

I concluded that `if: len_body != 0` is enough - the commit message of 5fdb6a7f52bd23238848aacc6c2acddb1a5ebcbe contains details. I inferred from inspecting several ELF files in the Web IDE that `ofs_body == 0` apparently doesn't mean that the program header has no body - on the contrary, executable ELF files tend to have a `LOAD` segment with `ofs_body == 0`, which looks like a normal `LOAD` segment.

For example, https://github.com/kaitai-io/kaitai_struct_samples/blob/6e000c8d792cca51b0f698dc63fd27a8d7cea8a1/executable/elf/hello_world.out has a `LOAD` segment with `execute|read` flags, which has `ofs_body == 0` and includes the ELF header, program headers and the `.text` section. This segment is the only executable one in the ELF file, so it cannot be ignored - clearing the `executable` flag or setting `len_body` to zero in this segment results in a segmentation fault.

armijnhemel: Dumb me, I had section headers in mind, not program headers. Still it has the same root cause.

executable/elf.ksy
```diff
@@ -423,7 +594,7 @@ types:
             type: u4
           - id: info
```
armijnhemel, line 617: According to:

https://docs.oracle.com/en/operating-systems/solaris/oracle-solaris/11.4/linkers-libraries/section-headers.html

"Two members in the section header, sh_link and sh_info, hold special information, depending on section type. "

Is this something that would be possible to parse here (note: I don't have test files), or do you think that it would be better processed at a higher level?

generalmimon: I actually use it in `verdef_section`:

https://github.com/kaitai-io/kaitai_struct_formats/blob/9c0cd9fc173bc58d3a23c60761a1cb0922dffee5/executable/elf.ksy#L1046-L1049

... and `verneed_section`:

https://github.com/kaitai-io/kaitai_struct_formats/blob/9c0cd9fc173bc58d3a23c60761a1cb0922dffee5/executable/elf.ksy#L1217-L1220

This is also why I changed the type to `u4`. The [Solaris table](https://docs.oracle.com/en/operating-systems/solaris/oracle-solaris/11.4/linkers-libraries/section-headers.html#GUID-2CBE4879-2E76-426E-BB7F-CF0CB1D87C52__CHAPTER6-47976) shows that it's either `0`, some section header index or the number of some entries, so the `u4` type seems to be always appropriate.

executable/elf.ksy
```diff
@@ -807,21 +1646,57 @@ enums:
...
+        (it's unclear whether it was ever used in practice at all).
     9:
       id: s370
-      doc: IBM System/370
-    10:
-      id: mips_rs3_le
-      doc: MIPS R3000 little-endian
+      -orig-id: EM_S370
+      doc: IBM System/370 (S/370)
+    # 10:
```
armijnhemel: Would it make sense to keep it regardless? I guess that there are very few binaries that use this value, so I don't care that much about it.

generalmimon:
@armijnhemel As I mentioned privately, I updated the links to https://sourceware.org/git/ so that they point to the mirrors at https://forge.sourceware.org/ instead: https://github.com/kaitai-io/kaitai_struct_formats/compare/9c0cd9fc173bc58d3a23c60761a1cb0922dffee5...30bd672fabb918083526136ad481b711c8ffe5d8

I don't know what the uptime is for https://forge.sourceware.org/, but based on my experience, it's definitely better than https://sourceware.org/git/ (which isn't exactly setting the bar very high). If I understand correctly, both https://forge.sourceware.org/glibc/glibc-mirror and https://forge.sourceware.org/binutils-gdb/binutils-gdb-mirror are automatically synced with the canonical repos at https://sourceware.org/git/.

executable/elf.ksy
generalmimon: Done: 655fd3c8261aea676941be70dea22c4a60d0196f

executable/elf.ksy
```diff
@@ -1537,25 +2679,94 @@ enums:
       id: moxie_old
       -orig-id: EM_MOXIE_OLD
       doc: Old, unofficial value for Moxie
+  # https://forge.sourceware.org/glibc/glibc-mirror/src/tag/glibc-2.43/elf/elf.h#L715
+  # https://forge.sourceware.org/binutils-gdb/binutils-gdb-mirror/src/tag/binutils-2_46_1/include/elf/common.h#L472
+  # https://docs.oracle.com/en/operating-systems/solaris/oracle-solaris/11.4/linkers-libraries/program-header.html#GUID-49F99618-9CDC-4A08-A94C-E2AA264AA01A__CHAPTER6-69880
   ph_type:
```
armijnhemel, line 2708: It might be good to incorporate some of the LLVM specific values:

https://llvm.org/doxygen/BinaryFormat_2ELF_8h_source.html

starting at line 1565

executable/elf.ksy
```diff
@@ -1537,25 +2679,94 @@ enums:
...
+        - https://forge.sourceware.org/binutils-gdb/binutils-gdb-mirror/src/tag/binutils-2_46_1/include/elf/arm.h#L41
+        - https://github.com/ARM-software/abi-aa/blob/daa7a94ca55973736c0e434a67a6e4bbcd35d7fa/aaelf32/aaelf32.rst#61program-header Git tag "2025Q4"
     # 0x7fffffff: hi_proc
-    0x6474e550: gnu_eh_frame
-    0x6474e551: gnu_stack
-    0x6474e552: gnu_relro
-    0x6474e553: gnu_property
   # https://docs.oracle.com/en/operating-systems/solaris/oracle-solaris/11.4/linkers-libraries/section-headers.html#GUID-2CBE4879-2E76-426E-BB7F-CF0CB1D87C52__CHAPTER6-73445
   # https://github.com/illumos/illumos-gate/blob/1d806c5f41/usr/src/boot/sys/sys/elf_common.h#L377-L462
   sh_type:
```
armijnhemel, line 2903: It might be good to sync with values from LLVM:

https://llvm.org/doxygen/BinaryFormat_2ELF_8h_source.html

starting at line 1154

In some ELF binaries that I have from recent Android devices I see:

```
  [21] .llvm_addrsig     LOOS+0xfff4c03  0000000000000000 008c48 00005a 00   E  0   0  1

```

which is `SHT_LLVM_ADDRSIG`

As it seems that they chose a value that isn't overlapping with other section types I guess that it is safe to add these, similar to the ones from GNU.

armijnhemel: Also:

https://android.googlesource.com/platform/bionic/+/6f12bfece5dcc01325e0abba56a46b1bcf991c69/tools/relocation_packer/src/elf_file.cc#37

as I am seeing these in Android files:

```
  [ 9] .rel.dyn          LOOS+0x1        00001948 001948 000045 01   A  4   0  4
  [10] .relr.dyn         LOOS+0xfffff00  00001990 001990 000014 04   A  0   0  4
```

(observed in `android.hardware.audio.parameter_parser.example_service` from `menari_lte-cp2a.260603.001-factory-4e10d715.zip` downloaded from the Android factory images website)

armijnhemel: Not related to this PR: there are overlapping values in `sh_type`, for example:

```ksy
    0x70000001: amd64_unwind
    # 0x70000001: arm_exidx
```

It would be good to find a neat solution for this problem. In this case both AMD64 and ARM are in widespread use.

generalmimon: @armijnhemel:

> It would be good to find a neat solution for this problem. In this case both AMD64 and ARM are in widespread use.

I posted my reply in a new issue: https://github.com/kaitai-io/kaitai_struct_formats/issues/764

generalmimon: @armijnhemel:

> It might be good to sync with values from LLVM:
> 
> https://llvm.org/doxygen/BinaryFormat_2ELF_8h_source.html
> 
> starting at line 1154
> 
> In some ELF binaries that I have from recent Android devices I see:
> 
> ```
>   [21] .llvm_addrsig     LOOS+0xfff4c03  0000000000000000 008c48 00005a 00   E  0   0  1
> ```
> 
> which is `SHT_LLVM_ADDRSIG`
> 
> As it seems that they chose a value that isn't overlapping with other section types I guess that it is safe to add these, similar to the ones from GNU.

---

> Also:
> 
> https://android.googlesource.com/platform/bionic/+/6f12bfece5dcc01325e0abba56a46b1bcf991c69/tools/relocation_packer/src/elf_file.cc#37
> 
> as I am seeing these in Android files:
> 
> ```
>   [ 9] .rel.dyn          LOOS+0x1        00001948 001948 000045 01   A  4   0  4
>   [10] .relr.dyn         LOOS+0xfffff00  00001990 001990 000014 04   A  0   0  4
> ```

Both of these comments were resolved in 1f6c98eee865390d4cadbbe52f2ba5ddf4ff4f40.

executable/elf.ksy
generalmimon: Done: 439936f980eb33feb3c1a04bd474059a78179a46

generalmimon:
@armijnhemel Is this good enough to merge now?

armijnhemel:
> @armijnhemel Is this good enough to merge now?

Yes, please merge!

generalmimon:
I asked Claude Code 2.1.233 (model `claude-opus-5`, high effort) to review this PR. The prompt was the following:

<details>
<summary>Full prompt for Claude Code</summary>

> Thoroughly review and test this PR: focus on correctness (potential bugs, logic errors, edge cases, ...), consistency, etc. You should try to verify against original sources. If you cannot fetch normally, you can use `curl`. Also check the comments/discussion from the PR review. Report even minor issues. You should test whether the generated Python code works, so once you're done with the initial review, create a uv project (`uv init`) in the scratch dir and develop a harness for the testing against input ELF files. Possible ELF sample sources: https://github.com/JonathanSalwan/binary-samples, https://github.com/aboutcode-org/elf-inspector/tree/main/tests/data + others you can find yourself (e.g. download and unpack some Ubuntu packages). You can also craft your own samples. Environment notes: the Git-ignored `_scratch/` directory should be used as a scratch dir (don't use `/tmp`). This is a Git for Windows shell, so prefer Bash commands. Available CLI commands: 7-Zip (`7z`), `jq`, ripgrep (`rg`), `kaitai-struct-compiler`, GitHub CLI (`gh`) set up with a public access token, `readelf`. Python is installed via `uv`, so it must be run either as `python3.14` (if outside of a `uv` project) or as `uv run python` (`python` or `python3` will NOT work). If you want to install any PyPI packages, set up a proper `uv` project (i.e. `uv init` + `uv add ...`) - that way, there'll be a proper lockfile. And venv will be managed automatically if you run everything via `uv run` (e.g. `uv run python ...`). `readelf` can be used for authoritative parsing of ELF, in particular the versinoning sections can be parsed as `readelf -V`.

</details>

It produced the following HTML file with the review comments: [**review.html**](https://github.com/user-attachments/files/31216177/review.html)

The most serious finding is the first one, copied below:

> R1
> 
> ### The new `program_header.body` breaks on separate debug-info files
> 
> Highest impact new in this PR
> 
> The `if: len_body != 0` guard (5fdb6a7f52bd23238848aacc6c2acddb1a5ebcbe) fixes ELFs embedded in `.gnu_debugdata`, where `p_filesz` was zeroed. It does not cover the much more common sibling case: a separate debug file produced by `objcopy --only-keep-debug`, which every distribution ships in its `-dbg`/`-debuginfo` packages. There the program headers keep their *original* `p_offset` and `p_filesz`, but the sections they cover were turned into `SHT_NOBITS`, so those file offsets now hold unrelated bytes. `readelf` shows the same thing — it prints `[Requesting program interpreter: ^D]` for `labrea.debug`.
> 
> Three of the 148 corpus files hit this:
> 
> | File | Segment | Error from the compiled parser |
> |----|----|----|
> | [`analyze.so.debug`](https://github.com/aboutcode-org/elf-inspector/blob/bc6b862a788f1082bc1e87c7b9547eb3e6def011/tests/data/dwarf_and_elf/analyze.so.debug) | PH #&#x200B;3 · PT\_DYNAMIC | EndOfStream: 4 bytes requested, 0 available (segment holds `.dynstr` text, never reaches `DT_NULL`) |
> | [`latex2emf.debug`](https://github.com/aboutcode-org/elf-inspector/blob/bc6b862a788f1082bc1e87c7b9547eb3e6def011/tests/data/dwarf_and_elf/latex2emf.debug) | PH #&#x200B;4 · PT\_DYNAMIC | EndOfStream: 4 bytes requested, 0 available |
> | [`labrea.debug`](https://github.com/aboutcode-org/elf-inspector/blob/bc6b862a788f1082bc1e87c7b9547eb3e6def011/tests/data/dwarf_and_elf/labrea.debug) | PH #&#x200B;6 · PT\_NOTE | EndOfStream: 5 590 599 bytes requested, 32 available (garbage `n_descsz`) |
> 
> > **Suggestion.** Either widen the heuristic — e.g. also skip the body when the segment's file range is not backed by any non-`NOBITS` section — or leave the behaviour and extend the (already excellent) comment above the `if` to say that `body` is expected to fail on `--only-keep-debug` output. Because `body` is a lazy instance the rest of the file still parses, but the Web IDE and any consumer that walks all instances will surface it as an error.

I find the fact that such ELF files exist to be interesting (and unfortunate), which is why I'm mentioning it here. However, I'm afraid that realistically, we can't solve this problem on the .ksy side, if we want to parse `body` in `program_header` at all. The only indication that the plausibly looking values of `ofs_body` and `len_body` in these program headers are actually bogus is that the corresponding section headers have the `NOBITS` type, but we're unable to establish a connection between program headers and section headers within the .ksy spec itself.

So unfortunately, if such ELF files should be parsed successfully by an application that relies on `elf.ksy`, I believe these cases must be handled on the application side.

---

The other Claude Code's findings are largely intentional (and most of them are quite small details, but to be fair, I included "Report even minor issues." in the prompt, so that's fine), so I don't feel the need to address any of them right now.

generalmimon:
I addressed the finding from Claude Code posted in my [previous comment](https://github.com/kaitai-io/kaitai_struct_formats/pull/752#issuecomment-5374057275) by documenting the issue in the .ksy spec: e85bd0debfb8a75b1ff3cc9173b8f816c0d46f41

https://github.com/kaitai-io/kaitai_struct_formats/blob/e85bd0debfb8a75b1ff3cc9173b8f816c0d46f41/executable/elf.ksy#L531-L552

It turned out that, although Claude was largely right and pointed me in the right direction, there were still some inaccuracies. Specifically, I'm quite sure that the `*.debug` files where the EOF errors occurred were **not** produced by `objcopy --only-keep-debug` from GNU Binutils, but by `eu-strip --strip-debug` from elfutils. The clue to this was that the `.comment` sections in all 3 `*.debug` files say something similar to `GCC: (GNU) 4.7.0 20120223 (Red Hat 4.7.0-0.15)`, and Red Hat distros (e.g. Fedora) traditionally use elfutils instead of GNU Binutils (used by Debian/Ubuntu).

This is an important difference, because `objcopy` from GNU Binutils zeroes `len_body` (so `body` will be disabled due to the `if: len_body != 0` condition) in the segments whose contents were omitted from the resulting `*.debug` files, but `eu-strip` from elfutils doesn't. I reproduced this on Ubuntu 26.04 with both the `binutils` and `elfutils` packages installed:

```console
$ cat > hello.c << 'EOF'
#include <stdio.h>
int main(void) {
    printf("hello, world\n");
    return 0;
}
EOF
$ gcc -g -O0 -o hello hello.c
$ objcopy --only-keep-debug hello hello.debug
$ cp -v hello hello-eu
'hello' -> 'hello-eu'
$ eu-strip --strip-debug -f hello-eu.debug hello-eu
$ for f in hello hello.debug hello-eu.debug; do readelf -lW -- "$f" > "$f".readelf-lW.txt; done
readelf: Error: Unable to find program interpreter name
$ git diff --no-index --word-diff hello.readelf-lW.txt hello.debug.readelf-lW.txt
diff --git a/hello.readelf-lW.txt b/hello.debug.readelf-lW.txt
index f0c8149..97d47fb 100644
--- a/hello.readelf-lW.txt
+++ b/hello.debug.readelf-lW.txt
@@ -1,25 +1,24 @@

Elf file type is DYN [-(Position-Independent Executable-]{+(Shared object+} file)
Entry point 0x1060
There are 14 program headers, starting at offset 64

Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  PHDR           0x000040 0x0000000000000040 0x0000000000000040 0x000310 0x000310 R   0x8
  INTERP         0x000374 0x0000000000000374 0x0000000000000374 [-0x00001c-]{+0x000000+} 0x00001c R   0x1[-[Requesting program interpreter: /lib64/ld-linux-x86-64.so.2]-]
  LOAD           0x000000 0x0000000000000000 0x0000000000000000 [-0x000608-]{+0x000374+} 0x000608 R   0x1000
  LOAD           0x001000 0x0000000000001000 0x0000000000001000 [-0x000175-]{+0x000000+} 0x000175 R E 0x1000
  LOAD           [-0x002000-]{+0x001000+} 0x0000000000002000 0x0000000000002000 0x000148 0x000148 R   0x1000
  LOAD           [-0x002db8-]{+0x000db8+} 0x0000000000003db8 0x0000000000003db8 [-0x000258-]{+0x000000+} 0x000260 RW  0x1000
  DYNAMIC        [-0x002dc8-]{+0x001db8+} 0x0000000000003dc8 0x0000000000003dc8 [-0x0001f0-]{+0x000000+} 0x0001f0 RW  0x8
  NOTE           0x000350 0x0000000000000350 0x0000000000000350 0x000024 0x000024 R   0x4
  NOTE           [-0x0020f8-]{+0x0010f8+} 0x00000000000020f8 0x00000000000020f8 0x000030 0x000030 R   0x8
  NOTE           [-0x002128-]{+0x001128+} 0x0000000000002128 0x0000000000002128 0x000020 0x000020 R   0x4
  GNU_PROPERTY   [-0x0020f8-]{+0x0010f8+} 0x00000000000020f8 0x00000000000020f8 0x000030 0x000030 R   0x8
  GNU_EH_FRAME   [-0x002014-]{+0x001000+} 0x0000000000002014 0x0000000000002014 [-0x000034-]{+0x000000+} 0x000034 R   0x4
  GNU_STACK      0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000 RW  0x10
  GNU_RELRO      [-0x002db8-]{+0x001db8+} 0x0000000000003db8 0x0000000000003db8 [-0x000248-]{+0x000000+} 0x000248 R   0x1

 Section to Segment mapping:
  Segment Sections...
$ git diff --no-index --word-diff hello.readelf-lW.txt hello-eu.debug.readelf-lW.txt
diff --git a/hello.readelf-lW.txt b/hello-eu.debug.readelf-lW.txt
index f0c8149..e7f607a 100644
--- a/hello.readelf-lW.txt
+++ b/hello-eu.debug.readelf-lW.txt
@@ -1,5 +1,5 @@

Elf file type is DYN [-(Position-Independent Executable-]{+(Shared object+} file)
Entry point 0x1060
There are 14 program headers, starting at offset 64

@@ -7,7 +7,7 @@ Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  PHDR           0x000040 0x0000000000000040 0x0000000000000040 0x000310 0x000310 R   0x8
  INTERP         0x000374 0x0000000000000374 0x0000000000000374 0x00001c 0x00001c R   0x1
      [Requesting program interpreter: [-/lib64/ld-linux-x86-64.so.2]-]{+]+}
  LOAD           0x000000 0x0000000000000000 0x0000000000000000 0x000608 0x000608 R   0x1000
  LOAD           0x001000 0x0000000000001000 0x0000000000001000 0x000175 0x000175 R E 0x1000
  LOAD           0x002000 0x0000000000002000 0x0000000000002000 0x000148 0x000148 R   0x1000
@@ -24,16 +24,16 @@ Program Headers:
 Section to Segment mapping:
  Segment Sections...
   00
   01     .interp {+.comment .symtab .strtab+}
   02     .note.gnu.build-id .interp .gnu.hash .dynsym .dynstr .gnu.version .gnu.version_r .rela.dyn .rela.plt
   03     .init .plt .plt.got .plt.sec .text .fini
   04     .rodata .eh_frame_hdr .eh_frame[-.note.gnu.property .note.ABI-tag-]
   05     .init_array .fini_array .dynamic .got .data .bss
   06     .dynamic
   07     .note.gnu.build-id {+.comment .symtab .strtab+}
   08     [-.note.gnu.property-]{+.comment .symtab .strtab+}
   09     [-.note.ABI-tag-]{+.comment .symtab .strtab+}
   10     [-.note.gnu.property-]{+.comment .symtab .strtab+}
   11     .eh_frame_hdr
   12
   13     .init_array .fini_array .dynamic .got
```

You can see that `objcopy` zeroes `FileSiz` (shown as `{+0x000000+}` in the diff), while `eu-strip` leaves the program header table unchanged (which apparently also confuses causes confusion in the `Section to Segment mapping:` section of `readelf`'s output). If you try parsing the `hello-eu.debug` file using the current `elf.ksy`, you'll get EOF errors. With `hello.debug`, you won't.

generalmimon:
Now that this issue has been resolved, I believe this PR is ready to be merged.

armijnhemel:
> You can see that `objcopy` zeroes `FileSiz` (shown as `{+0x000000+}` in the diff), while `eu-strip` leaves the program header table unchanged (which apparently also confuses causes confusion in the `Section to Segment mapping:` section of `readelf`'s output). If you try parsing the `hello-eu.debug` file using the current `elf.ksy`, you'll get EOF errors. With `hello.debug`, you won't.

So if I understand correctly this will only happen when parsing via the program headers, not when parsing via the section headers?

generalmimon:
@armijnhemel:

> So if I understand correctly this will only happen when parsing via the program headers, not when parsing via the section headers?

Yes. The section headers whose contents are missing from the file have type `NOBITS` in any case, so the `body` in `section_header` will not be parsed (and therefore no error will occur):

https://github.com/kaitai-io/kaitai_struct_formats/blob/ccad5db7b39174e6857b8cea61c4c39f8ea39af3/executable/elf.ksy#L652

The issue is only with the `body` in `program_header`. It breaks the (generally reasonable) assumption that if you fully parse a valid file with the .ksy spec (including invoking all instances), no errors will occur. Now, I'm actually not sure whether what `eu-strip` does is correct (because the fact that `Offset` and `FileSiz` point past the end of the ELF file feels wrong to me), but it's a popular implementation, so we just have to deal with it.

Of course, if you prefer parsing the ELF file via section headers (and ignore the bodies of program headers when section headers are available), you should be able to easily avoid this problem.

PR #753, opened 2026-06-30
Add ballisticng_ctl.ksy for custom track layout

Theaninova:


game/ballistic_ng_ctl.ksy
```diff
@@ -0,0 +1,234 @@
...
+  file-extension: ctl
+  endian: le
+  license: CC0-1.0
+doc-ref:
+  - https://ballisticng-documentation.readthedocs.io/en/latest/ingame/layout_creator.html#ctl-custom-track-layout-specification
+seq:
+  - id: version
+    type: s4
+  - id: auto_juncion_threshold
+    type: f4
```
generalmimon: 1. Please fix the typo:

   ```suggestion
     - id: auto_junction_threshold
       type: f4
   ```

2. I noticed that the official spec lists this as a boolean:

   ```
   bool autoJunctionThreshold
   ```

   I know that you pointed this out on the https://discord.gg/neognosis-games server, so I believe you that `f4` is correct:

   Image

   But until it is fixed in the official spec (which we have no control over), you should at least document this via a field-level [docstring](https://doc.kaitai.io/user_guide.html#docstrings) (the `doc` key).

game/ballistic_ng_ctl.ksy
```diff
@@ -0,0 +1,234 @@
+meta:
+  id: ballistic_ng_ctl
+  title: BallisticNG Custom Track
+  file-extension: ctl
+  endian: le
+  license: CC0-1.0
+doc-ref:
+  - https://ballisticng-documentation.readthedocs.io/en/latest/ingame/layout_creator.html#ctl-custom-track-layout-specification
```
generalmimon: Nit: if there is only one reference, it is customary not to make it a list:

```suggestion
doc-ref: https://ballisticng-documentation.readthedocs.io/en/latest/ingame/layout_creator.html#ctl-custom-track-layout-specification
```

game/ballistic_ng_ctl.ksy
```diff
@@ -0,0 +1,234 @@
...
+    size: reference_image_path_len
+    if: version > 5
+  - id: reference_is_locked
+    type: u1
+    if: version > 6
+  - id: reference_pos
+    type: f4
+    if: version > 5
+    repeat: expr
+    repeat-expr: 3
```
generalmimon: I don't like that XYZ coordinates are described as `type: f4` + `repeat-expr: 3`. It would probably be better to create a dedicated type with `id: x`, `id: y`, `id: z` fields.

game/ballistic_ng_ctl.ksy
```diff
@@ -0,0 +1,234 @@
...
+  control_point:
+    seq:
+      - id: pos
+        type: f4
+        repeat: expr
+        repeat-expr: 3
+      - id: rot
+        type: f4
+        repeat: expr
+        repeat-expr: 4
```
generalmimon: Similarly as for the XYZ coordinates, the quaternion here representing the rotation should be defined in a dedicated type instead of `type: f4` + `repeat-expr: 4`.

I assume that the semantics of the `[x y z w]` values is the following? https://www.opengl-tutorial.org/intermediate-tutorials/tutorial-17-quaternions/#quaternions

> A quaternion is a set of 4 numbers, [x y z w], which represents rotations the following way:
> 
> ```csharp
> // RotationAngle is in radians
> x = RotationAxis.x * sin(RotationAngle / 2)
> y = RotationAxis.y * sin(RotationAngle / 2)
> z = RotationAxis.z * sin(RotationAngle / 2)
> w = cos(RotationAngle / 2)
> ```

If so, that further underscores why a dedicated type with named fields makes more sense here than just a list of four numbers, which does not convey the meaning as well (e.g. `w` has a qualitatively different meaning than the rest of the numbers).

game/ballistic_ng_ctl.ksy
```diff
@@ -0,0 +1,234 @@
+meta:
+  id: ballistic_ng_ctl
+  title: BallisticNG Custom Track
```
generalmimon:
```suggestion
  title: BallisticNG Custom Track Layout
```

I think this is an important distinction, because "Custom Tracks" seems to refer to something else: https://ballisticng-documentation.readthedocs.io/en/latest/unity_tools/custom_tracks/index.html

game/ballistic_ng_ctl.ksy
```diff
@@ -0,0 +1,234 @@
+meta:
+  id: ballistic_ng_ctl
```
generalmimon: I suggest renaming the spec to `ballisticng_ctl`. The reason is that when I search for [`"ballisticng"`](https://github.com/search?q=%22ballisticng%22&type=code) on GitHub, I get many more results than when I search for [`"ballistic_ng"`](https://github.com/search?q=%22ballistic_ng%22&type=code).

game/ballistic_ng_ctl.ksy
```diff
@@ -0,0 +1,234 @@
...
+    repeat: expr
+    repeat-expr: route_count
+  - id: physics_zone_count
+    type: s4
+    if: version > 7
+  - id: physics_zones
+    type: physics_zone
+    if: version > 7
+    repeat: expr
+    repeat-expr: physics_zone_count
```
generalmimon: According to https://doc.kaitai.io/ksy_style_guide.html#seq-attr, the `if` key should come last:

```suggestion
  - id: physics_zones
    type: physics_zone
    repeat: expr
    repeat-expr: physics_zone_count
    if: version > 7
```

game/ballistic_ng_ctl.ksy
```diff
@@ -0,0 +1,234 @@
...
+    repeat-expr: 3
+  - id: reference_width
+    type: f4
+    if: version > 5
+  - id: route_count
+    type: s4
+  - id: routes
+    type: route
+    repeat: expr
+    repeat-expr: route_count
```
generalmimon: I understand that this name comes from the official spec, but we have a standard naming convention for number/count fields, see https://doc.kaitai.io/ksy_style_guide.html#attr-id:

The desktop `kaitai-struct-compiler` also issues warnings about this:

```
ballistic_ng_ctl.ksy: /seq/13/id:
        warning: use `num_routes` instead of `route_count`, given that it's only used as repeat count of `routes` (see https://doc.kaitai.io/ksy_style_guide.html#attr-id)

ballistic_ng_ctl.ksy: /seq/15/id:
        warning: use `num_physics_zones` instead of `physics_zone_count`, given that it's only used as repeat count of `physics_zones` (see https://doc.kaitai.io/ksy_style_guide.html#attr-id)

ballistic_ng_ctl.ksy: /types/route/seq/9/id:
        warning: use `num_control_points` instead of `control_point_count`, given that it's only used as repeat count of `control_points` (see https://doc.kaitai.io/ksy_style_guide.html#attr-id)
```

So I suggest renaming `route_count` and other `*_count` fields. It is best practice to use `-orig-id` to record the original name in the official specification for reference (see https://doc.kaitai.io/ksy_style_guide.html#transcribing):

```suggestion
  - id: num_routes
    -orig-id: routeCount
    type: s4
  - id: routes
    type: route
    repeat: expr
    repeat-expr: num_routes
```

Please do the same for `physics_zone_count` and `control_point_count`.

game/ballistic_ng_ctl.ksy
```diff
@@ -0,0 +1,234 @@
...
+        repeat: expr
+        repeat-expr: 3
+      - id: right_wall_extent
+        type: f4
+        repeat: expr
+        repeat-expr: 3
+  physics_zone:
+    seq:
+      - id: unk_1
+        type: s4
```
generalmimon: It's recommended to omit `id` for unknown fields (see https://doc.kaitai.io/ksy_style_guide.html#attr-id):

```suggestion
      - type: s4
```

Kaitai Struct compiler will then generate a name automatically based on the index in the `seq` array. In this case, it will be `_unnamed0`. It will be accessible and displayed under this name even in visualizers: both in the Web IDE and in the [console visualizer](https://github.com/kaitai-io/kaitai_struct_visualizer) (`ksv`).

game/ballistic_ng_ctl.ksy
```diff
@@ -0,0 +1,234 @@
+meta:
+  id: ballistic_ng_ctl
+  title: BallisticNG Custom Track
+  file-extension: ctl
+  endian: le
+  license: CC0-1.0
+doc-ref:
```
generalmimon: It's good that you specified `doc-ref` that points to official docs with all the necessary context about what this format is for. I'd appreciate if you also added a top-level `doc` with at least a brief description of the format (conventionally, it should come above/before `doc-ref`).

You can use existing specs in this repo for inspiration: https://github.com/search?q=repo%3Akaitai-io%2Fkaitai_struct_formats%20%2F%5Edoc%3A%2F&type=code

generalmimon (changes requested):

Theaninova:
Thanks for the review, I've updated the branch with all your suggestions!

game/ballisticng_ctl.ksy
```diff
@@ -0,0 +1,235 @@
+meta:
+  id: ballisticng_ctl
```
generalmimon, line 2: Please also rename the file so that the file name matches the new `/meta/id` 🙏

Theaninova: Oops, right

game/ballisticng_ctl.ksy
```diff
@@ -0,0 +1,235 @@
...
+    type: s4
+  - id: auto_junction_threshold
+    type: f4
+    doc: Official spec lists this as a bool, which seems to be a mistake.
+  - id: hyper_speed_enabled
+    type: s4
+    if: _root.version > 5
+  - id: speed_class
+    type: s4
+    if: _root.version > 5
```
generalmimon, line 23: For consistency:

```suggestion
  - id: hyper_speed_enabled
    type: s4
    if: version > 5
  - id: speed_class
    type: s4
    if: version > 5
```

game/ballisticng_ctl.ksy
```diff
@@ -0,0 +1,235 @@
...
+    5: in_quint
+    6: in_circ
+    7: in_quad
+    8: in_quart
+    9: in_expo
+    10: out_sine
+    11: out_cubic
+    12: out_quint
+    13: out_circ
+    14: out_quat
```
generalmimon, line 219:
```suggestion
    14: out_quad
```

game/ballisticng_ctl.ksy
```diff
@@ -0,0 +1,235 @@
...
+      - id: tangent_yaw
+        type: f4
+        if: _root.version > 3
+      - id: tangent_pitch
+        type: f4
+        if: _root.version > 4
+      - id: track_shape_interpolation_type
+        type: u1
+        enum: track_shape_interpolation_type
+        if: _root.version > 10
```
generalmimon, line 135: The official spec lists this as `int32 trackShapeInterpolationType`, but this is a `type: u1` (i.e. `uint8`). As with `auto_junction_threshold`, it's worth documenting that this discrepancy is also intentional.

game/ballisticng_ctl.ksy
```diff
@@ -0,0 +1,235 @@
...
+      - id: gravity_multiplier
+        type: f4
+      - id: track_tracking_multiplier
+        type: f4
+      - id: grip_multiplier
+        type: f4
+      - id: air_resistance_multiplier
+        type: f4
+      - id: zero_gravity
+        type: u1
```
generalmimon, line 162: Same as above - this is `float32 zeroGravity` in the official spec.

game/ballisticng_ctl.ksy
```diff
@@ -0,0 +1,235 @@
...
+      - id: zero_gravity_tracking_multiplier
+        type: f4
+      - id: zero_gravity_pitch_rotation_multiplier
+        type: f4
+      - id: zero_gravity_pitch_force
+        type: f4
+      - id: zero_gravity_pitch_speed
+        type: f4
+      - id: spring_to_track
+        type: u1
```
generalmimon, line 172: And this is `float32 springToTrack`.

game/ballisticng_ctl.ksy
```diff
@@ -0,0 +1,235 @@
+meta:
+  id: ballisticng_ctl
+  title: BallisticNG Custom Track Layout
+  file-extension: ctl
+  endian: le
+  license: CC0-1.0
```
generalmimon, line 6: Should be swapped (https://doc.kaitai.io/ksy_style_guide.html#meta):

```suggestion
  license: CC0-1.0
  endian: le
```

game/ballisticng_ctl.ksy
```diff
@@ -0,0 +1,235 @@
...
+  Format for custom race track layouts in BallisticNG.
+  Only encodes non-decorative information about the track,
+  such as spline control points and physics zones.
+doc-ref: https://ballisticng-documentation.readthedocs.io/en/latest/ingame/layout_creator.html#ctl-custom-track-layout-specification
+seq:
+  - id: version
+    type: s4
+  - id: auto_junction_threshold
+    type: f4
+    doc: Official spec lists this as a bool, which seems to be a mistake.
```
generalmimon, line 17: https://doc.kaitai.io/ksy_style_guide.html#doc:

> - use backticks `` ` `` to wrap identifiers and small pieces of code

```suggestion
    doc: Official spec lists this as a `bool`, which seems to be a mistake.
```

game/ballisticng_ctl.ksy
```diff
@@ -0,0 +1,235 @@
...
+    type: f4
+    if: version > 12
+  - id: reference_image_path_len
+    type: u1
+    if: version > 5
+  - id: reference_image_path
+    type: str
+    encoding: UTF-8
+    size: reference_image_path_len
+    if: version > 5
```
generalmimon, line 43: - https://doc.kaitai.io/ksy_style_guide.html#attr-id
- https://doc.kaitai.io/ksy_style_guide.html#seq-attr

```suggestion
  - id: len_reference_image_path
    type: u1
    if: version > 5
  - id: reference_image_path
    size: len_reference_image_path
    type: str
    encoding: UTF-8
    if: version > 5
```

game/ballisticng_ctl.ksy
```diff
@@ -0,0 +1,235 @@
...
+        type: vec3
+      - id: left_wall_extent
+        type: vec3
+      - id: right_floor_extent
+        type: vec3
+      - id: right_wall_extent
+        type: vec3
+  physics_zone:
+    seq:
+      - type: s4
```
generalmimon, line 146: Thanks for removing the bogus `id` here! Now I'm thinking that it should probably also have a `doc` key that briefly explains that this is a field of unknown purpose which is missing from the official spec, but is actually present in every .ctl file. After all, it's also a discrepancy from the official spec, and it's good to document these.

generalmimon (changes requested):

PR #754, opened 2026-06-30
Add nulib file formats

Theaninova:


generalmimon:
This is actually quite a big PR (thanks for that!), which means that reviewing it will probably take at least a few months or even years.

You can make the review easier by incorporating the feedback from https://github.com/kaitai-io/kaitai_struct_formats/pull/753 (I saw for example several `unknown*` fields here as well after a quick glance) and of course following the [style guide](https://doc.kaitai.io/ksy_style_guide.html) (I guess you could use help of an LLM, if you want).

I'd also appreciate sample files for testing all formats so that I don't have to search for them on the internet (which takes time).

Theaninova:
Thanks for the feedback, I'll have a look at fixing all these things!

PR #767, opened 2026-08-20
exif: fix and improve `ifd_field` parsing

### LINKED ISSUE #765: exif: `ifd_field` should expose the value as byte array regardless of `is_immediate_data`
Cc @armijnhemel

In [`image/exif.ksy`](https://github.com/kaitai-io/kaitai_struct_formats/blob/33a32b5cbba16a1238c54c0ffe8f2b811d1aeb79/image/exif.ksy), the `data` property in the `ifd_field` type is only available if `is_immediate_data` is false, which happens if the value is over 4 bytes in length:

https://github.com/kaitai-io/kaitai_struct_formats/blob/33a32b5cbba16a1238c54c0ffe8f2b811d1aeb79/image/exif.ksy#L69-L75

If the IFD field value fits in 4 bytes, `is_immediate_data` is true, which means that the value is only accessible via the `ofs_or_data` field:

https://github.com/kaitai-io/kaitai_struct_formats/blob/33a32b5cbba16a1238c54c0ffe8f2b811d1aeb79/image/exif.ksy#L62-L63

This is inconvenient to use. The consumers of the .ksy spec shouldn't have to care what property the value is stored in depending on whether `is_immediate_data` is true or false (which is just a serialization detail) - they should be able to read a property which provides the byte array with the value in both cases.

On top of that, `type: u4` is actually wrong - it should be a byte array too. This is clear from the official Exif 2.2 specification - see https://web.archive.org/web/20131018091152id_/http://exif.org/Exif2-2.PDF, page 14 (the PDF viewer says page 20):

> If the value is smaller than 4 bytes, the value is stored in the 4-byte area starting from the left, i.e., from the lower end of the byte offset area. For example, in big endian format, if the type is SHORT and the value is 1, it is recorded as 00010000.H.

In this example from the spec, `ofs_or_data` would be read as the integer `0x00010000`. So in order to unpack the actual value, users of the `exif.ksy` spec would first have to convert it to bytes themselves, before they can read the SHORT value from the first two bytes correctly.

generalmimon:
Fixes #765

On top of that, the `exif.ksy` spec now parses the individual types of IFD field values (rather than just parsing all values as byte arrays, as suggested in #765) and descends into sub-IFDs: "Exif IFD", "Interoperability IFD" and "GPS IFD".


generalmimon:
Tip for reviewing this PR: because of the commit 01ae1e8f57b8bbb0c41f786f9bb671689595e641, which made a pure whitespace change of decreasing the indent of the `enums` section, it's a good idea to enable GitHub's "Hide whitespace" option to make the diff much clearer. This link will do that for you: https://github.com/kaitai-io/kaitai_struct_formats/pull/767/changes?w=1

PR #774, opened 2026-09-15
rpm: support v6 packages and large (4+ GiB) v4 packages

generalmimon:
In September 2025, [RPM 6.0.0](https://rpm.org/releases/6.0.0) was released with quite a few changes, including the new v6 package format. Note that Fedora 44 ships rpm 6.0.x, but applies a [patch](https://src.fedoraproject.org/rpms/rpm/blob/7099d81c3b5ecf1777a43095be429cb198bcc566/f/rpm-6.0-rpmformat.patch) to keep generating v4 packages.

This PR updates `archive/rpm.ksy` so that it fully supports packages written by RPM 6.0+, both v4 and v6, including v4 packages [using 64-bit sizes](https://github.com/rpm-software-management/rpm/blob/ec9ea8c43808c346da4b6cb454cdc58aef8e506a/docs/manual/large_files.md) (rather than the usual 32-bit sizes) because the combined size of the header and compressed payload, or the size of the uncompressed payload, is 4 GiB or more.

1. Enums were synced with [`ec9ea8c4`](https://github.com/rpm-software-management/rpm/tree/ec9ea8c43808c346da4b6cb454cdc58aef8e506a), the latest upstream RPM repo commit when this PR was made. In particular, the `signature_tags` and `header_tags` enums were updated - b27869008ee881f3cc807aa121794c9950bf30fc
2. `architectures::not_set` = 0 and `operating_systems::not_set` = 0 were added. RPM 6.0 no longer populates `lead.architecture` and `lead.os` when writing both v4 and v6 packages, so they are left zeroed - 8b754e4121aff6290fe9cef31cbd51780e7bea78
3. `payload` is now available in v6 packages and in v4 packages with 64-bit sizes - 7efb5e302a7219e0d2a12bb34163e5c0a2fb9840

   Previously, the size of `payload` was always computed from the 32-bit `signature_tags::size` tag, which is only present in v4 packages that don't use the [Large File Support](https://github.com/rpm-software-management/rpm/blob/ec9ea8c43808c346da4b6cb454cdc58aef8e506a/docs/manual/large_files.md) extension (which switches to 64-bit sizes).

   v6 packages have a 64-bit tag `header_tags::payload_size` (`RPMTAG_PAYLOADSIZE`) and v4 packages with 64-bit sizes have `signature_tags::long_size` (`RPMSIGTAG_LONGSIZE`) instead, so we use those to determine the size of `payload` as well.

4. Many `doc`/`doc-ref` updates - cfa195c3c3c6b64f582d18d72576df674c3c6029
5. Validate lengths in the `header_record` type (nowadays called `struct header_intro` - see [The Header Structure](https://github.com/rpm-software-management/rpm/blob/ec9ea8c43808c346da4b6cb454cdc58aef8e506a/docs/manual/format_header.md#intro) in RPM docs) against RPM's limits - 05f301bd43a5604b210901ba9dc8ae27cdacddb7
6. Validate that the `record_type` in the `header_index_record` type (nowadays called `struct index_entry` - see [The Header Structure](https://github.com/rpm-software-management/rpm/blob/ec9ea8c43808c346da4b6cb454cdc58aef8e506a/docs/manual/format_header.md#index) in RPM docs) is known, just as RPM does - 5129be6a3a84faa3c2aa75cadef0b4532a2f7e18

   Also remove `record_types::not_implemented` = 0 (`RPM_NULL_TYPE`), which RPM considers an invalid data type since version 4.17 (released in 2021):
https://github.com/rpm-software-management/rpm/commit/42f53a13c4300c046f4f16d91712bb28763da4c3


armijnhemel:
> 6\. Also remove `record_types::not_implemented` = 0 (`RPM_NULL_TYPE`), which RPM considers an invalid data type since version 4.17 (released in 2021)

Has it ever been valid?

generalmimon:
@armijnhemel:

> > 6. Also remove `record_types::not_implemented` = 0 (`RPM_NULL_TYPE`), which RPM considers an invalid data type since version 4.17 (released in 2021)
> 
> Has it ever been valid?

I don't think so. It seems that it has always been an internal sentinel value. From https://github.com/rpm-software-management/rpm/commit/42f53a13c4300c046f4f16d91712bb28763da4c3 (which first appeared in RPM 4.17.0):

> **Fix RPM_MIN_TYPE definition, RPM_NULL_TYPE is not a legit tag type**
>
> RPM_NULL_TYPE is not a legit tag type, but RPM_MIN_TYPE is defined
> and used as if it was. (...)

Older versions would technically let `RPM_NULL_TYPE` through if it (hypothetically) were to appear in an input .rpm file. However, it would have a size of zero ([`lib/header.c:54`](https://github.com/rpm-software-management/rpm/blob/3659b8a04f5b8bacf6535e0124e7fe23f15286bd/lib/header.c#L54)), so such entries couldn't contain any data. I don't think any version of RPM would ever write such files, though. The [RPM docs](https://github.com/rpm-software-management/rpm/blob/ec9ea8c43808c346da4b6cb454cdc58aef8e506a/docs/manual/format_header.md#index) don't mention the "NULL" type at all (only the ancient LSB spec did - I think that's just because it was a transcript of the original RPM implementation, which was even older).

